PR #201, opened 2015-01-05
fix for SYNPY-161 enhanced column metadata

cbare:


synapseclient/table.py
```diff
@@ -321,75 +321,48 @@ def row_labels_from_id_and_version(rows):
 def row_labels_from_rows(rows):
     return row_labels_from_id_and_version([(row['rowId'], row['versionNumber']) for row in rows])
 
-def cast_row(row, columns, headers):
+def cast_values(values, headers):
```
brucehoff: may want to double check that headers and row values always appear in the corresponding order

PR #203, opened 2015-01-07
dirty hack to check if we have row id and version in CSV queries

cbare:


brucehoff:
I suggest that Chris uses the same approach as does the R client, i.e. rather than checking the csv column count, check for the presence of columns named ROW_ID and ROW_VERSION.  Chris proposed that the change be done as a subsequent pull request.

PR #211, opened 2015-02-10
Fix synapse cat printing extra line breaks

kdaily:
Updated 'synapse cat' to use sys.stdout to write instead of print, so extra line breaks are not added.

Also changed a couple of variable names ('file' to 'filepath', 'input' to 'inputfile') as 'file' and 'input' are reserved Python built-ins or names.


kdaily:
@larssono can you review this with me?

kdaily:
Did code review with @larssono before merging.

PR #212, opened 2015-02-24
Add feature to command line to set download location in 'get'

kdaily:
Main functionality works, EXCEPT when using `synapse get` with a file path instead of a synapse ID; `entity.path` is `None` in this case. Can't figure it out!

Need to update the integration tests - the basic parsing of the `Created foo.txt` should include the path (possibly `./` for current directory, which is the sensible default for the parameter).


kdaily:
Merged develop in and still works, even with @larssono's recursive get. Still needs tests.

cbare:
This PR is associated with: SYNPY-237 Set data download location from command line.

PR #213, opened 2015-03-02
Write progress messages to stderr instead of stdout

kdaily:
I'm working on a feature for the command line client to query Synapse tables. However, the progress messages get in the way of that output. Hence, simply writing progress to stderr allows the user to separate that output from the table output.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2839,11 +2839,11 @@ class Synapse:
         while time.time()-start_time < self.table_query_timeout:
             result = self.restGET(uri+'/get/%s'%async_job_id['token'])
             if result.get('jobState', None) == 'PROCESSING':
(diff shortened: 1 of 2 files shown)


larssono:
I believe that will make the Galaxy client break as Galaxy assumes that anything output to stderr means that the plugin failed.

kdaily:
Ha, Chris just told me the same thing. I'm going to look how to implement a
flag in all commands that progress/status out.

kellrott:
The galaxy 'stderr means error' behavior can be disabled: https://wiki.galaxyproject.org/Admin/Tools/ToolConfigSyntax#A.3Cstdio.3E.2C_.3Cregex.3E.2C_and_.3Cexit_code.3E_tag_sets

PR #217, opened 2015-04-11
add downloadTableFile without caching

cbare:


synapseclient/client.py
```diff
@@ -1771,6 +1771,12 @@ def returnDict(destination):
             elif scheme == 'http' or scheme == 'https':
                 #TODO add support for username/password
                 response = requests.get(url, headers=self._generateSignedHeaders(url, {}), stream=True)
+
+                ## get filename from content-disposition, if we don't have it already
+                if os.path.isdir(destination):
+                    filename = utils.extract_filename(response.headers['content-disposition'])
```
cbare: Test if content-disposition exists

synapseclient/client.py
```diff
@@ -1771,6 +1771,12 @@ def returnDict(destination):
             elif scheme == 'http' or scheme == 'https':
                 #TODO add support for username/password
                 response = requests.get(url, headers=self._generateSignedHeaders(url, {}), stream=True)
+
```
brucehoff: if no 'content-disposition' header then take the 'tail' of the URL (after the "/"), taking care to drop the query string part of the URL (unlike what Bruce did!)

PR #220, opened 2015-04-22
Adding method to query wiki attachments

kellrott:
Sorry for all the auto formatting. My editor hates whitespaces ;-)

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -55,18 +55,18 @@ from synapseclient.entity import Entity, File, Project, Folder, split_entity_nam
 from synapseclient.table import Schema, Column, RowSet, Row, TableQueryResult, CsvFileTable
 from synapseclient.dict_object import DictObject
 from synapseclient.evaluation import Evaluation, Submission, SubmissionStatus
-from synapseclient.wiki import Wiki
+from synapseclient.wiki import Wiki, WikiAttachment
 from synapseclient.retry import _with_retry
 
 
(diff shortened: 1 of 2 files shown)


synapseclient/client.py
```diff
@@ -2818,6 +2818,11 @@ def _downloadWikiAttachment(self, owner, wiki, filename, destination=None):
             destination = os.path.join(destination, filename)
         return self._downloadFile(url, destination)
 
+    def getWikiAttachments(self, wiki):
```
kellrott: This is where the relevant changes start.

synapseclient/wiki.py
```diff
@@ -113,42 +113,52 @@ def __init__(self, **kwargs):
...
         return '/entity/%s/wiki/%s' % (self.ownerId, self.id)
 
     def deleteURI(self):
         """For internal use."""
-        
+
         return '/entity/%s/wiki/%s' % (self.ownerId, self.id)
 
 
+class WikiAttachment(DictObject):
```
kellrott: New Class added here

synapseclient/client.py
larssono: Looks good.  We need integration test for this.  If you don't have the bandwidth we can see if we can do it on our side.

P.S.  I like that we are finally getting all of those PEP8 violating spaces cleaned-up :)

kellrott: I'm busy working on the SMC-Het docs, but need these attachment functions so I can sync the doc images to synapse. If you guys can help getting this into the API, it would be a great help.

PR #243, opened 2015-09-09
Support for deleting entity versions.

kkdang:


synapseclient/entity.py
```diff
@@ -262,8 +262,12 @@ def postURI(self):
     def putURI(self):
         return '/entity/%s' %self.id
 
-    def deleteURI(self):
-        return '/entity/%s' %self.id
+    def deleteURI(self,versionNumber):
```
cbare: versionNumber=None

synapseclient/client.py
```diff
@@ -1027,24 +1027,38 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
 
         return bundle
 
-    def delete(self, obj):
+    def delete(self, obj, version=False):
```
cbare: version=None

PR #247, opened 2015-10-23
Changes so that username and password are prompted for if not provided.

pgm:
How do you feel about receiving pull requests?  

This is a bit of a cosmetic change to make the "synapse" command handle authentication in a slightly more interactive way.  (Although I can imagine one could debate how it should work.)

I have an aversion to putting passwords in scripts, so I didn't want to use "syn.login("username", "secret")"

I then read I could cache credentials with the command line client via "synapse login -u username -p secret --remember-me" but I have an aversion to typing passwords as arguments (it means the password is visible in the output of "ps -o args")   In this case, the process is so short lived, it's not a real threat, but I prefer to be prompted for passwords with no echo.

I introduced changes so that if no credentials are cached, it prompts for username/password.  If the username is provided, it only prompts for the password.  

On a related note, the examples on the synapse site for the command line give the following for downloading a file:

    # Login
    synapse -u synapse_username -p synapse_password

    # Download file
    synapse get syn5007782

However, that gave me an error, saying not enough arguments.  I believe the example should be:

    # Login
    synapse -u synapse_username -p synapse_password login --remember-me

    # Download file
    synapse get syn5007782

or after my change, you can just do:
    synapse login --remember-me

(I also think --remember-me should always be true, because the "login" command doesn't really do anything if it's not caching credentials.  All other operations appear to automatically login before performing the operation.)

### DIFF
--- a/synapseclient/__main__.py
+++ b/synapseclient/__main__.py
@@ -56,6 +56,7 @@ from synapseclient import Activity
 import utils
 import signal
 import json
+import getpass
 from synapseclient.exceptions import *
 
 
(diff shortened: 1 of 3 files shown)


cbare:
Thanks pgm. Looks like your corrections to the example for the 'synapse login' command is spot on. I like the idea of prompting for the password as well!

PR #150, opened 2014-06-04
2to3

adrinjalali:
Hey,

Not sure why your code is not compatible with python3. Anyway I ported it mostly using 2to3, and some manual changes, and tested some basic features like login, and get.

I'll be using python3, and will try to fix issues if I encounter any.

Cheers,
Adrin.

PR #256, opened 2015-11-16
Python3

cbare:
This branch represents the continuation of the Python 3 compatibility work started by @adrinjalali in [PR#150 "2to3"](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/150). We hope to merge Python 3 compatibility into the develop branch after a period of testing.


cbare:
Some items yet to look at:
* some redundant uses of list() introduced by 2to3
* inconsistent translation of py2's map.iteritems() as six.iteritems(map) or list(map.items)
* make sure all uses of str() do from builtins import str
* remove any remaining uses of u(), remove utils.u and utils.to_bytes
* messages removed from exceptions - don't know why this happens
* standardize all .py files with a set of imports from future and utf declaration

adrinjalali:
About messages being removed from exceptions, it I'm not mistaken it's because in python 2 the code raises a string, but now it should raise an Exception class, so they got replaced by an empty exception class. Don't exactly remember.

PR #258, opened 2015-12-02
Py3 4 dreamtools

cokelaer:
Dear all,

I have to make dreamtools python3 compatible and realised that synapseclient is not. I therefore forked and fix the code. This may not be 100% safe but all tests passed on python 3.4 and 2.7 except for :

- tests that uses raw_input/input in nosetests, which seems to be a known issue/feature in nose with python3.
- and a couple of tests that requires specific credentials.

thanks
thomas


cbare:
Merged in PR #266

PR #271, opened 2016-01-08
Add Link Class

thomasyu888:
SYNPY-64 Extract targetId from Links and follow.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -633,6 +633,8 @@ class Synapse:
                                  Defaults to True
         :param downloadLocation: Directory where to download the Synapse File Entity.
                                  Defaults to the local cache.
+        :followLink:             Whether the link returns the target Entity.
+                                 Defaults to True
         :param ifcollision:      Determines how to handle file collisions.
                                  May be "overwrite.local", "keep.local", or "keep.both".
                                  Defaults to "keep.both".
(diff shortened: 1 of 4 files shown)


synapseclient/client.py
```diff
@@ -739,6 +739,12 @@ def _getWithEntityBundle(self, entityBundle, entity=None, **kwargs):
         ifcollision = kwargs.get('ifcollision', 'keep.both')
         submission = kwargs.get('submission', None)
 
+        #If Link, get target ID entity bundle
+        if entityBundle['entity']['concreteType'] == 'org.sagebionetworks.repo.model.Link':
+            targetId = entityBundle['entity']['linksTo']['targetId']
+            targetVersion = entityBundle['entity']['linksTo']['targetVersionNumber']
+            entityBundle = self._getEntityBundle(targetId, targetVersion)
```
larssono: How do you now delete or find meta information about a link?

thomasyu888: If you add a link to a file, shouldn't the annotations should be the same as that of the file its being linked to?  So the meta information about a link is just the meta information of the target file.

thomasyu888: Updated. Added Param followLink which defaults to true, but if false, syn.get returns meta information of Link

cbare:
SYNPY-64 Extract targetId from Links and follow.

PR #282, opened 2016-01-27
Update copy function- Use fileHandleId

thomasyu888:


synapseclient/__main__.py
```diff
@@ -219,12 +220,22 @@ def copy(args,syn):
     for i in search:
         if i['file.name'] == ent.name:
             raise ValueError('Filename exists in directory you would like to copy to, either rename or check if file has already been copied!')
-    new_ent = synapseclient.File(ent['path'],parent=args.parentid)
+    #CHECK: If the user created the file, copy the file by using fileHandleId else hard copy
+    if profile == ent.properties.createdBy:
+        tempFile = utils.make_bogus_data_file()
```
cbare: It shouldn't be necessary to make and throw away a dummy file here. Try something like:

```python
new_ent = File(name=ent.name, description=ent.description, parentId=args.parentId)
new_end.dataFileHandleId = ent.dataFileHandleId
```

synapseclient/__main__.py
```diff
@@ -219,12 +220,22 @@ def copy(args,syn):
     for i in search:
         if i['file.name'] == ent.name:
             raise ValueError('Filename exists in directory you would like to copy to, either rename or check if file has already been copied!')
-    new_ent = synapseclient.File(ent['path'],parent=args.parentid)
+    #CHECK: If the user created the file, copy the file by using fileHandleId else hard copy
+    if profile == ent.properties.createdBy:
```
cbare: Could it ever happen that a File was created by the current user, but it's fileHandle wasn't?

thomasyu888: When you try to copy a file that you don't own, an error will the thrown if you try to set fileHandleId.

larssono: Use: syn.restGET('/entity/%s/filehandles' %f.id) to get the filehandles and check the createdBy

larssono: OR use GET /fileHandle/{handleId} and go over to download/upload of the file if you get a 403 error back

PR #297, opened 2016-03-14
Complete copy function

thomasyu888:


synapseclient/client.py
```diff
@@ -1134,11 +1134,186 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+            copiedId = self._copyTable(ent.id, parentId)
+        else:
+            raise ValueError("Not able to copy this type of file")
+        #have to assign copied id something or else an error may be thrown
+        if copiedId != None:
+            print("Copied %s to %s" % (ent.id,copiedId))
+            mapping[ent.id] = copiedId
+        return(mapping)
+
+    def _copyFolder(self, entity, parentId, mapping=dict(), **kwargs):
```
cbare: parentId might be called destinationId

synapseclient/client.py
```diff
@@ -1134,11 +1134,186 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+        :param mapping:         Takes a mapping {'oldSynId': 'newSynId'} to help with copying syn ids that aren't already part of the entity
+                                Default to dict() and builds the mapping of all the synapse entities copied
+        """
+
+        version = kwargs.get('version', None)
+        setProvenance = kwargs.get('setProvenance', "traceback")
+        recursive = kwargs.get('recursive',True)
+        copiedId = None
+        ent = self.get(entity,downloadFile=False)
+        if isinstance(ent, Project):
```
cbare: Can you create a new copy of a project by not giving a destination synapse ID?

synapseclient/client.py
```diff
@@ -1134,13 +1135,199 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+        :param copyTable:       Determines whether tables are copied
+                                Default is True
+
+        :param kwargs:          Parameters that can be passed to the hidden copy functions called
+        """
+        mapping = kwargs.get('mapping', dict())
+        updateLinks = kwargs.get('updateLinks', True)
+        updateSynIds = kwargs.get('updateSynIds', True)
+        if onlyCopyWiki:
+            self._copyWiki(entity, destinationId, updateLinks=updateLinks, updateSynIds=updateSynIds, entityMap=mapping)
```
cbare: expose syn.copyWiki(...) for when we only want to copy wiki. Then syn.copy(...) can optionally call syn.copyWiki(...).

synapseclient/client.py
```diff
@@ -1134,13 +1135,199 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+        """
+
+        version = kwargs.get('version', None)
+        setProvenance = kwargs.get('setProvenance', "traceback")
+        recursive = kwargs.get('recursive',True)
+        copyTable = kwargs.get('copyTable',True)
+        copiedId = None
+        ent = self.get(entity,downloadFile=False)
+        if isinstance(ent, Project):
+            if destinationId == None:
```
cbare: destinationId is None

synapseclient/client.py
```diff
@@ -1134,13 +1135,199 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+                newProject = self.store(Project("Copied %s %s" %(time.time(),ent.name)))
+                destinationId = newProject.id
+            elif not isinstance(self.get(destinationId),Project):
+                raise ValueError("You must give a destinationId of a new project to copy projects")
+            mapping[ent.id] = destinationId
+            entities = self.chunkedQuery('select id, name from entity where parentId=="%s"' % ent.id)
+            for i in entities:
+                mapping = self._copyRecursive(i['entity.id'], destinationId, mapping=mapping, **kwargs)
+        elif isinstance(ent, Folder):
```
cbare: Friendly error msg for the case where destinationId is None and we're not copying a project

synapseclient/client.py
```diff
@@ -1134,13 +1135,199 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+        elif isinstance(ent, File):
+            copiedId = self._copyFile(ent.id, destinationId, version=version, setProvenance=setProvenance)
+        elif isinstance(ent, Link):
+            copiedId = self._copyLink(ent.id, destinationId)
+        elif isinstance(ent, Schema) and copyTable:
+            copiedId = self._copyTable(ent.id, destinationId)
+        else:
+            raise ValueError("Not able to copy this type of file")
+        #have to assign copied id something or else an error may be thrown
+        if copiedId != None:
```
cbare: if copiedId is not None:

synapseclient/client.py
```diff
@@ -1134,13 +1135,199 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+        if isinstance(ent, Project):
+            if destinationId == None:
+                newProject = self.store(Project("Copied %s %s" %(time.time(),ent.name)))
+                destinationId = newProject.id
+            elif not isinstance(self.get(destinationId),Project):
+                raise ValueError("You must give a destinationId of a new project to copy projects")
+            mapping[ent.id] = destinationId
+            entities = self.chunkedQuery('select id, name from entity where parentId=="%s"' % ent.id)
+            for i in entities:
+                mapping = self._copyRecursive(i['entity.id'], destinationId, mapping=mapping, **kwargs)
```
cbare: copiedId = destinationId

synapseclient/client.py
```diff
@@ -1134,13 +1135,199 @@ def _getEntityBundle(self, entity, version=None, bitFlags=0x800 | 0x400 | 0x2 |
...
+        :param onlyCopyWiki:    Determines whether only the wiki of the entity is copied
+                                Default is False
+
+        :param copyTable:       Determines whether tables are copied
+                                Default is True
+
+        :param kwargs:          Parameters that can be passed to the hidden copy functions called
+        """
+        mapping = kwargs.get('mapping', dict())
+        updateLinks = kwargs.get('updateLinks', True)
```
cbare: updateLInks and updateSynIds in arg list with default values

synapseclient/client.py
```diff
@@ -1337,12 +1337,12 @@ def setAnnotations(self, entity, annotations={}, **kwargs):
 
         :returns: A dictionary
         """
-
-        uri = '/entity/%s/annotations' % id_of(entity)
+        entity = self.get(entity,downloadFile=False)
```
larssono: I would call restGET(/entity/%s/annotations) instead.  It would be less costly as we don't need the entire entity.

synapseclient/client.py
```diff
@@ -1413,12 +1337,12 @@ def setAnnotations(self, entity, annotations={}, **kwargs):
...
+        uri = '/entity/%s/annotations' % entity.id
```
cbare: This looks like it's addressing SYNPY-362, which came about as a consequence of stripping out the redundant synapse properties from annotations (SYNPY-175).

I agree w/ Larsson. We might only look up the etag if we don't already have it (in the entity), possibly something like:

   if 'etag' not in synapseAnnos:
       if 'etag' in entity:
           synapseAnnos['etag'] = entity['etag']
       else:
           old_annos = syn.restGET(uri)
           synapseAnnos['etag'] = old_annos['etag']

Whether it's worth an extra couple lines, I don't know. In the common case, syn.store(entity) already adds the etag of the entity to its annotations before calling setAnnotations(). It's subtleties like this that'll make you crazy.

As an aside, some think setAnnotations should be changed to work only in memory as the R client does. See SYNPY-232.

PR #313, opened 2016-08-12
check MD5 of downloaded files SYNPY-308

cbare:


synapseclient/client.py
```diff
@@ -998,6 +998,10 @@ def store(self, obj, **kwargs):
                                                              md5=local_state.get('md5', None),
                                                              fileSize=local_state.get('fileSize', None))
                 properties['dataFileHandleId'] = fileHandle['id']
+                print(fileHandle)
```
cbare: remove

PR #316, opened 2016-08-23
resume partial downloads

cbare:
SYNPY-309 resume partial downloads using range header
SYNPY-388 fix sporadic test failures related to team creation

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -45,6 +45,7 @@ import collections
 import math, os, sys, stat, re, json, time
 import base64, hashlib, hmac
 import six
+import uuid
 
 try:
     from urllib.parse import urlparse
@@ -106,6 +107,7 @@ ROOT_ENTITY = 'syn4489'
 PUBLIC = 273949  #PrincipalId of public "user"
 AUTHENTICATED_USERS = 273948
 DEBUG_DEFAULT = False
+REDIRECT_LIMIT = 5
 
 
 # Defines the standard retry policy applied to the rest methods
@@ -114,8 +116,9 @@ DEBUG_DEFAULT = False
 STANDARD_RETRY_PARAMS = {"retry_status_codes": [429, 502, 503, 504],
                          "retry_errors"      : ["proxy error", "slow down", "timeout", "timed out",
                                                 "connection reset by peer", "unknown ssl protocol error",
-                                                "couldn't connect to host", "slowdown", "try again"],
-                         "retry_exceptions"  : ["ConnectionError", "Timeout", "timeout"],
+                                                "couldn't connect to host", "slowdown", "try again",
+                                                "connection reset by peer"],
+                         "retry_exceptions"  : ["ConnectionError", "Timeout", "timeout", "ChunkedEncodingError"],
                          "retries"           : 8,
                          "wait"              : 1,
                          "back_off"          : 2}
@@ -1758,7 +1761,7 @@ class Synapse:
         return self._downloadFile(url, destination, expected_md5=entity.get("md5", None))
 
 
-    def _downloadFile(self, url, destination, expected_md5=None):
+    def _downloadFile(self, url, destination, file_handle_id=None, expected_md5=None):
         """
         Download a file from a URL to a the given file path.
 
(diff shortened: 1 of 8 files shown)


synapseclient/client.py
```diff
@@ -1770,82 +1773,146 @@ def returnDict(destination):
...
 
-        if response.status_code in [301,302,303,307,308]:
-            url = response.headers['location']
+        :param url: source of download
+        :param destination: destination on local file system
+        :param file_handle_id: (optional) if given, the file will be given a
+                               temporary name that includes the file handle id
+                               which allows resuming partial downloads of the same
+                               file from previous sessions
+        :param expected_md5:   (optional) if given, check that the MD5 of the
```
larssono: Can't this be handled by the filename instead and an extension?

synapseclient/client.py
```diff
@@ -1770,82 +1773,146 @@ def returnDict(destination):
                      'files': [None] if destination is None else [os.path.basename(destination)],
                      'cacheDir': None if destination is None else os.path.dirname(destination) }
 
-        if expected_md5 == '':
-            expected_md5 = None
+        return returnDict(self._download_with_retries(url, destination, file_handle_id, expected_md5))
```
larssono: Why not make this function _downlod_with_retries?

synapseclient/client.py
cbare: I broke it up for clarity - keeping the retry logic separate from the loop that handles redirects. 

Also, I like that _download_with_retries keeps the original URL in its scope, whereas _downloadFile is free to modify its copy of the URL as it handles redirects. Otherwise, you'd have to stash a copy of the original URL so you could refresh the (time limited) presigned S3 URL when you get a partially completed download.

synapseclient/client.py
cbare: I broke it up into 2 functions for clarity - separating the retry logic from the loop handling redirects.

I also like that __download_with_retries_ keeps a copy of the original URL in its scope leaving __download_ free to modify its copy or URL as it gets redirected. Otherwise, you'd have to stash a copy of the original so you could refresh the presigned and time-limited S3 URL.

synapseclient/client.py
```diff
@@ -1770,82 +1773,146 @@ def returnDict(destination):
...
-        download_via_http = True
 
-        # We expect to be redirected to a signed S3 URL or externalURL
-        #The assumption is wrong - we always try to read either the outer or inner requests.get
-        #but sometimes we don't have something to read.  I.e. when the type is ftp at which point
-        #we still set the cache and filepath based on destination which is wrong because nothing was fetched
-        response = _with_retry(lambda: requests.get(url, headers=self._generateSignedHeaders(url), allow_redirects=False), verbose=self.debug, **STANDARD_RETRY_PARAMS)
+    def _download_with_retries(self, url, destination, file_handle_id=None, expected_md5=None, retries=5):
+        while retries > 0:
+            try:
```
larssono: Reset retry counter if we have made progress since last loop.   Can be checked by keeping track of size of destination.

synapseclient/utils.py
```diff
@@ -62,7 +62,7 @@
 BUFFER_SIZE = 8*KB
 
 
-def md5_for_file(filename, block_size=2*MB):
+def md5_for_file(filename, block_size=2*MB, hash=None):
     """
     Calculates the MD5 of the given file.  See `source <http://stackoverflow.com/questions/1131220/get-md5-hash-of-a-files-without-open-it-in-python>`_.
```
cbare: add doc for 'hash' parameter

tests/integration/test_download.py
```diff
@@ -46,3 +46,36 @@ def test_download_check_md5():
...
+    schedule_for_cleanup(original_file)
+
+    temp_dir = tempfile.gettempdir()
+
+    url = '%s/entity/%s/file' % (syn.repoEndpoint, entity.id)
+    path = syn._download(url, destination=temp_dir, file_handle_id=entity.dataFileHandleId, expected_md5=entity.md5)
+
+    ## simulate an imcomplete download by putting the
+    ## complete file back into its temporary location
+    tmp_path = os.path.join(temp_dir, "synapse_download_" + str(entity.dataFileHandleId))
```
cbare: pull out function to compute the temporary file name and use from test and from client.py

tests/unit/unit_test_download.py
```diff
@@ -0,0 +1,238 @@
...
+    ])
+
+    ## patch requests.get and also the method that generates signed
+    ## headers (to avoid having to be logged in to Synapse)
+    with patch.object(requests, 'get', side_effect=mock_requests_get), \
+         patch.object(synapseclient.client.Synapse, '_generateSignedHeaders', side_effect=mock_generateSignedHeaders):
+        path = syn._download_with_retries(url, destination=temp_dir, file_handle_id=12345, expected_md5=contents_md5)
+
+
+    ## 6. 206 Range header not supported, respond with 200 and full file
```
cbare: fix me

PR #319, opened 2016-08-27
Retry changes

larssono:


synapseclient/retry.py
```diff
@@ -45,51 +46,48 @@ def foo(a, b, c): return [a, b, c]
...
-                ## if the response is not JSON, look for retryable errors in its text content
-                elif any([msg.lower() in response.text.lower() for msg in retry_errors]):
-                    if verbose: sys.stderr.write("retrying"+ response.text+'\n')
+                ## For all other non 200 messages look for retryable errors in the body or reason field
+                response_message = _get_message(response)
+                if any([msg.lower() in response_message.lower() for msg in retry_errors]):
+                    retry = True
+                    log_error('retrying %s' %response_message, verbose)
+                ## special case for message throttling
+                elif 'Please slow down.  You may send a maximum of 10 message' in response:
```
brucehoff: consider getting rid of this case since, (1) generic exp back off may handle it, (2) we shouldn't have logic that switches based on the text of an error message

brucehoff:
Good job, Larsson!

PR #321, opened 2016-09-08
Update client.py

thomasyu888:
https://sagebionetworks.jira.com/browse/SYNPY-391

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -3128,7 +3128,7 @@ class Synapse:
         max_files_per_request = kwargs.get('max_files_per_request', 2500)
         #Rowset tableQuery result not allowed
         if isinstance(table, TableQueryResult):
-            raise ValueError("Cannot specify resultsAs='rowset' in syn.tableQuery()")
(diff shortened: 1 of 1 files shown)


synapseclient/client.py
```diff
@@ -3126,7 +3126,9 @@ def downloadTableColumns(self, table, columns, **kwargs):
         RETRIABLE_FAILURE_CODES = ["EXCEEDS_SIZE_LIMIT"]
         MAX_DOWNLOAD_TRIES = 100
         max_files_per_request = kwargs.get('max_files_per_request', 2500)
-
+        #Rowset tableQuery result not allowed
+        if isinstance(table, TableQueryResult):
+            raise ValueError("Cannot specify resultsAs='rowset' in syn.tableQuery()")
```
larssono: I actually think it should be possible to use a rowset.  It would require a new way of interacting with it though.
But this is a good first pass until we fix it.  It might be worth updating the error message though to indicate that downloadTableColumn doesn't work with rowsets.

Issue #327, opened 2016-10-18
Slow synapse get on large projects


fidlr:
On large projects with many files, the synapse get command is very slow.
One suggestion I can make to speeding it up is using multi-threading / multi-processing for making the calls to 'get' concurrent.

A simple patch that suggests one way to do this is pasted below (sorry, for some reason github wouldn't let me upload it, just save to a txt file and apply) -

From d6ae4c2426afc5e2132842db3e2859410d83759a Mon Sep 17 00:00:00 2001
From: fidlr <me@fidlr.org>
Date: Tue, 18 Oct 2016 10:35:51 +0300
Subject: [PATCH] multi-threaded get

---
 synapseutils/sync.py | 23 ++++++++++++++++++++---
 1 file changed, 20 insertions(+), 3 deletions(-)

diff --git a/synapseutils/sync.py b/synapseutils/sync.py
index dfbfab8..de399a8 100644
--- a/synapseutils/sync.py
+++ b/synapseutils/sync.py
@@ -2,6 +2,13 @@ import errno
 from synapseclient.entity import is_container
 from synapseclient.utils import id_of
 import os
+from concurrent.futures import ThreadPoolExecutor
+
+pool = ThreadPoolExecutor(max_workers=3)  # Synapse allows up to 3 concurrent requests
+
+def getOneEntity(syn, entity_id, downloadLocation, ifcollision, allFilesList):
+    ent = syn.get(entity_id, downloadLocation=downloadLocation, ifcollision=ifcollision)
+    allFilesList.append(ent)  # lists are thread-safe
 
 
 def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFiles = None):
@@ -36,7 +43,11 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
         for f in entities:
             print(f.path)
     """
-    if allFiles is None: allFiles = list()
+    global pool
+    wait_at_finish = False
+    if allFiles is None:  # initial call
+        allFiles = list()
+        wait_at_finish = True
     id = id_of(entity)
     results = syn.chunkedQuery("select id, name, nodeType from entity where entity.parentId=='%s'" %id)
     for result in results:
@@ -53,6 +64,12 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
                 new_path = None
             syncFromSynapse(syn, result['entity.id'], new_path, ifcollision, allFiles)
         else:
[...]

PR #328, opened 2016-10-19
multi-threaded get

fidlr:
@larssono thanks for your comment on issue #327 
I'm attaching it as a pull request for your review.

Best,
Fidlr.


larssono:
We will work this into a future release but will have to rewrite to fit changes made in 1.7 for file downlaods.

PR #336, opened 2016-11-10
Be able to download links recursively as well 

thomasyu888:
This will have issues if the entity that the link points to doesn't exist anymore or don't have permissions to view the entity.  But this will allow for people to create Links to files and for people to download them recursively.


thomasyu888:
It will also fail if the link is dead.  Will have to look into this further when I have time

PR #344, opened 2017-01-18
wiki markdown attachment is now has its own type

thomasyu888:


thomasyu888:
This is still in staging, don't pull in yet

PR #345, opened 2017-01-22
Create Docker class

thomasyu888:


thomasyu888:
I don't know if you want to have the option of passing in the path of where the Dockerfile is and have synapse build and push the image to a synapse project.  This can definitely be done, because there is a python docker client.  However, that is just another optional dependency that we will have to enforce if we allow for this

synapseclient/entity.py
```diff
@@ -464,6 +464,18 @@ def __init__(self, name=None, properties=None, annotations=None, local_state=Non
         super(Project, self).__init__(concreteType=Project._synapse_entity_type, properties=properties,
                                       annotations=annotations, local_state=local_state, **kwargs)
 
+class Docker(Entity):
+    """
+    Represents a Docker in Synapse.
```
kdaily: Represents a Docker container in Synapse.

synapseclient/entity.py
```diff
@@ -464,6 +464,18 @@ def __init__(self, name=None, properties=None, annotations=None, local_state=Non
...
+
+    You cannot create a Synapse Docker image with this function, please follow instructions here (http://docs.synapse.org/articles/docker.html) to create one.
```
kdaily: Confusing about what you are creating then to have this warning...maybe this should be shown when a user creates, and an image doesn't exist (or both here and there)?

thomasyu888:
Figure out how to pass in username and password

kdaily:
Use environment variables for passwords, or a config file.

thomasyu888:
You need to log into the registry to be able to create a docker container

thomasyu888:
There is a newer pull request done by Ziming.

PR #352, opened 2017-03-11
Set external URL on submissions when submissions are links

thomasyu888:


larssono:
This is already handled by downloadFile=False in getSubmission.

PR #362, opened 2017-04-06
SYNPY-443

zimingd:


synapseclient/cache.py
```diff
@@ -139,8 +139,9 @@ def contains(self, file_handle_id, path):
 
             cached_time = cache_map.get(path, None)
             if cached_time:
-                return True if compare_timestamps(_get_modified_time(path), cached_time) else False
-
+                return compare_timestamps(_get_modified_time(path), cached_time)
+
```
larssono: Single space between lines if significant different code blocks.  2 blank lines between functions only (Pep8)

synapseclient/cache.py
```diff
@@ -173,9 +174,27 @@ def get(self, file_handle_id, path=None):
...
+                    removed_entry_from_cache = False # determines if cache_map needs to be rewritten to disk
+
+                    # iterate a copy of cache_map to allow modifying original cache_map
+                    for cached_file_path, cached_time in six.iteritems(dict(cache_map)):
                         if path == os.path.dirname(cached_file_path):
-                            return cached_file_path if compare_timestamps(_get_modified_time(cached_file_path), cached_time) else None
+                            if os.path.exists(cached_file_path):
+                                if compare_timestamps(_get_modified_time(cached_file_path), cached_time):
+                                    matching_unmodified_directory = cached_file_path
+                                    break
```
larssono: Combine repeated if statements into single if with and

synapseclient/cache.py
```diff
@@ -173,9 +174,27 @@ def get(self, file_handle_id, path=None):
...
+                    matching_unmodified_directory = None
+                    removed_entry_from_cache = False # determines if cache_map needs to be rewritten to disk
+
+                    # iterate a copy of cache_map to allow modifying original cache_map
+                    for cached_file_path, cached_time in six.iteritems(dict(cache_map)):
                         if path == os.path.dirname(cached_file_path):
-                            return cached_file_path if compare_timestamps(_get_modified_time(cached_file_path), cached_time) else None
+                            if (os.path.exists(cached_file_path)
+                                and compare_timestamps(_get_modified_time(cached_file_path), cached_time)):
+                                    matching_unmodified_directory = cached_file_path
```
kdaily: Even if you have to do something more complex (shouldn't it fail fast if the path doesn't exist, meaning timestamp comparison isn't needed?) What would `_get_modified_time` do with a path that doesn't exist?

kdaily: The explicit check is redundant but should be faster (`compare_timestamps` returns `False` if either value is `None`)

PR #366, opened 2017-04-12
SYNPY-344

zimingd:


synapseclient/client.py
```diff
@@ -178,6 +178,27 @@ def _test_import_sftp():
              "\n\n\n"))
         raise
 
+def _extract_zip_file_to_cache(zip_file, zip_entry_name, cache_dir):
+    """
```
larssono: Make more generic: extract files into specified directory.  Move to synapseclient/utils.py

larssono: Also, while moving to utils also move the _is_integer function to utils.

PR #368, opened 2017-04-15
Remove chunkedQuery dependency and added getChildren() (SYNPY-435)

thomasyu888:
Add helper function to get children of containers.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2525,6 +2525,7 @@ class Synapse:
         submission_id = id_of(id)
         uri = Submission.getURI(submission_id)
         submission = Submission(**self.restGET(uri))
+        
         # Pre-fetch the Entity tied to the Submission, if there is one
         if 'entityId' in submission and submission['entityId'] is not None:
             related = self._getWithEntityBundle(
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2525,7 +2525,7 @@ class Synapse:
         submission_id = id_of(id)
         uri = Submission.getURI(submission_id)
         submission = Submission(**self.restGET(uri))
-        
+
         # Pre-fetch the Entity tied to the Submission, if there is one
         if 'entityId' in submission and submission['entityId'] is not None:
             related = self._getWithEntityBundle(
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2545,15 +2545,19 @@ class Synapse:
         submission_id = id_of(id)
         uri = Submission.getURI(submission_id)
         submission = Submission(**self.restGET(uri))
-
+        entityBundle = json.loads(submission['entityBundleJSON'])
+        #Take account into URL submissions
+        toDownload = True if entityBundle['fileHandles'][0].get('externalURL') is None else False
         # Pre-fetch the Entity tied to the Submission, if there is one
         if 'entityId' in submission and submission['entityId'] is not None:
             related = self._getWithEntityBundle(
-                                entityBundle=json.loads(submission['entityBundleJSON']),
+                                entityBundle=entityBundle,
                                 entity=submission['entityId'],
-                                submission=submission_id, **kwargs)
+                                submission=submission_id, downloadFile=toDownload, **kwargs)
             submission.entity = related
             submission.filePath = related.get('path', None)
+            if not toDownload:
+                submission.externalURL = related.externalURL
 
         return submission
 
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -67,6 +67,7 @@ import mimetypes
 import tempfile
 import warnings
 import getpass
+import json
 from collections import OrderedDict
 
 import synapseclient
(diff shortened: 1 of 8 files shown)


synapseutils/walk.py
```diff
@@ -35,12 +50,12 @@ def _helpWalk(syn,synId,newpath=None):
         dirpath = (newpath,synId)
     dirs = []
     nondirs = []
-    results = syn.chunkedQuery('select id, name, nodeType from entity where parentId == "%s"'%synId)
+    results = _getChildren(syn, synId)
     for i in results:
-        if is_container(i):
-            dirs.append((i['entity.name'],i['entity.id']))
```
larssono: You should fix is_container instead of working around the function.

synapseutils/walk.py
```diff
@@ -1,6 +1,21 @@
...
+    entityChildrenRequest = {'parentId':parentId,
+                             'includeTypes':includeTypes,
+                             'sortBy':sortBy,
+                             'sortDirection':sortDirection,
+                             'nextPageToken':nextPageToken}
+
+    resultPage = syn.restPOST('/entity/children',body =json.dumps(entityChildrenRequest))
+    for result in resultPage['page']:
+        yield result
+    if resultPage.get('nextPageToken') is not None:
```
larssono: Why make this recursive?  You always run the risk of a 1000 page request which would result in a recursive limit.  Since this a very linear it is easier to write a while loop.

thomasyu888:
Larsson wrote:

```
Why make this recursive? You always run the risk of a 1000 page request which would result in a recursive limit. Since this a very linear it is easier to write a while loop.
```

use syn.restGET("/REST/resources/schema/?resourceId=org.sagebionetworks.repo.model.EntityType") to get allowable entity types (somewhere in client should be defined)

^ does it break the client to define all of them in getChildren?

kdaily:
No but if there's a new entity type you're going to have to remember to add it here, or else it will break.

synapseutils/walk.py
```diff
@@ -1,8 +1,24 @@
...
+                             'sortBy':sortBy,
+                             'sortDirection':sortDirection,
+                             'nextPageToken':None}
+    allChildren = []
+    resultPage = syn.restPOST('/entity/children',body =json.dumps(entityChildrenRequest))
+    allChildren = resultPage['page']
+    while resultPage.get('nextPageToken') is not None:
+        entityChildrenRequest['nextPageToken'] = resultPage['nextPageToken']
+        resultPage = syn.restPOST('/entity/children',body =json.dumps(entityChildrenRequest))
+        allChildren.extend(resultPage['page'])
```
kdaily:
```
for child in resultPage['page']:
    yield child
```

synapseutils/walk.py
```diff
@@ -1,8 +1,24 @@
...
 import os
+import json
 
-def walk(syn, synId):
+def _getChildren(syn, parentId, includeTypes=["folder","file","table","link","entityview","dockerrepo"], sortBy="NAME", sortDirection="ASC"):
+    entityChildrenRequest = {'parentId':parentId,
+                             'includeTypes':includeTypes,
+                             'sortBy':sortBy,
+                             'sortDirection':sortDirection,
+                             'nextPageToken':None}
```
kdaily: `nextPageToken` not required.

synapseutils/walk.py
thomasyu888: But I would still need to call the next page right?  I would still have to use recursion which would run into that 1000 limit?

larssono: Tom:  To make this an iterator (whcih I recomnend) you need to make a for loop within your while loop where you yield the individual results in your resultPage['page']

synapseutils/walk.py
```diff
@@ -11,6 +26,9 @@ def walk(syn, synId):
 
     :param synId:          A synapse ID of a folder or project
 
+    :param includeTypes:   Must be a list of entity types (ie. ["folder","file"]) which can be found here:
+                           http://docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityType.html
+
     Example::
```
larssono: Whoa:  Does this work, that would mean that the return tuple would be different depeneding on how you called this function.  That seems confusing from a user perspective and also incompatible with how os.walk works.

Is there a broader goal here?

thomasyu888: Ya.  Basically people can choose what they want returned.  So if they only want files and nothing else they would just specify ["file"] - although that may break the function as it needs folders to generate the dirpath and dirname.  Maybe Ill have to specify that you must include ['folder','file'] at the minimum?

thomasyu888:
What did you want to call the getChildren function again? And did you want it to be a hidden function _get...?

synapseclient/client.py
```diff
@@ -1378,6 +1379,33 @@ def setAnnotations(self, entity, annotations={}, **kwargs):
     ##                        Querying                        ##
     ############################################################
 
+    def getChildren(self, parentId, includeTypes=["folder","file","table","link","entityview","dockerrepo"], sortBy="NAME", sortDirection="ASC"):
+        """
+        :param syn:            A synapse object: syn = synapseclient.login()- Must be logged into synapse
```
larssono: does not. it is now a function on the object.

synapseclient/entity.py
```diff
@@ -640,6 +641,8 @@ def is_container(entity):
     """Test if an entity is a container (ie, a Project or a Folder)"""
     if 'concreteType' in entity:
         concreteType = entity['concreteType']
+    elif 'type' in entity:
+        concreteType = entity['type']
     elif isinstance(entity, collections.Mapping):
```
larssono: Isn't type 'project' or 'folder'?  but entity.concreteType org.sagebionetworks.repo.model....

larssono:
1) Need unit test at least for getChildren
2) Need to modify test of is_container to go through new path of code.  (Also verify that this works).

synapseclient/entity.py
thomasyu888: The rest call for get children returns the concrete type in entity['type'] . This is why i added that line.

PR #380, opened 2017-05-09
fixed password prompt for windows python 2.7 (SYNPY-461)

zimingd:


synapseclient/__main__.py
```diff
@@ -787,10 +787,16 @@ def login_with_prompt(syn, user, password, rememberMe=False, silent=False, force
...
         # if there were no credentials in the cache nor provided, prompt the user and try again
-        if user is None:
+        while not user:
             user = input("Synapse username: ")
 
-        passwd = getpass.getpass("Password for " + user + ": ")
+        passwd_prompt = "Password for " + user + ": "
+        if six.PY2 and os.name == 'nt': #if using windows convert form unicode literal into str
+            passwd_prompt = passwd_prompt.encode('utf-8')
+
```
larssono: Can't we not just encode always.  None of the characters here are unicode anyway.

PR #394, opened 2017-05-30
Added DockerRepository, fixed ViewSchema and renamed it to EntityViewSchema (SYNPY-415)

zimingd:


synapseclient/table.py
```diff
@@ -538,12 +531,99 @@ def _before_synapse_store(self, syn):
...
+    def __init__(self, name=None, columns=None, parent=None, scopes = None, view_type=None, add_default_columns = True, properties=None, annotations=None, local_state=None, **kwargs):
+        if view_type:
+            kwargs['type'] = view_type
+
+        super(EntityViewSchema, self).__init__(name=name, columns=columns, properties=properties,
+                                               annotations=annotations, local_state=local_state, parent=parent, **kwargs)
+
+        #This is a hacky solution to make sure we don't try to add default columns to schemas that we retrieve from synapse
+        self.add_default_columns = add_default_columns and not (properties or annotations or local_state)
+
```
larssono: What happens if I am creating something new but want to add a dictionary of annotations?

thomasyu888:
I don't know if you want to have the option of passing in the path of where the Dockerfile is and have synapse build and push the image to a synapse project.  This can definitely be done, because there is a python docker client.  However, that is just another optional dependency that we will have to enforce if we allow for this.  I could add

```
#Check if docker is available on the host machine
    try:
        import docker
        docker_available = True
    except:
        docker_available = False
```

Build and push:
```
if docker_available:
	client = docker.from_env()
	#pass in credentials from existing synapse credentials
	client.login(username,password, registry="http://docker.synapse.org")
	client.images.build(path = "path/to/directory/containing/Dockerfile",tag ="docker.synapse.org/syn1234/myrepo")
	for line in client.images.push('docker.synapse.org/syn1234/myrepo', stream=True):
		print line
else:
	print("Install docker and python docker client")
```

PR #409, opened 2017-06-08
Retries of download will fetch new preSignedURLs instead of retrying on the same url until failure(SYNPY-484)

zimingd:


synapseclient/client.py
```diff
@@ -1769,16 +1768,20 @@ def _downloadFileHandle(self, url, destination, fileHandle, retries=5):
 
         :returns: path to downloaded file
         """
+        #TODO:z fix documentation for this
```
zimingd: do what the TODO says and fix documentation

synapseclient/client.py
```diff
@@ -1915,8 +1915,7 @@ def _download(self, url, destination, fileHandleId=None, expected_md5=None):
 
         ## check md5 if given
         if expected_md5 and actual_md5 != expected_md5:
-            scheme = urlparse(url).scheme
-            if  scheme != 'file' and os.path.exists(destination):
+            if  urlparse(url).scheme != 'file' and os.path.exists(destination):
```
larssono: I just realized another possible problem.

What happens if the file is downloaded already; modified by the user locally then refetched - would we delete it?  What happens if they modify the file during transfer?

PR #420, opened 2017-06-16
Allow for an entity be passed into Link class 

thomasyu888:
https://sagebionetworks.jira.com/browse/SYNPY-494


thomasyu888:
Can we pull this in?

tests/integration/integration_test_Entity.py
```diff
@@ -141,7 +141,7 @@ def test_Entity():
 
     newfolder = Folder('Testing Folder', parent=project)
     newfolder = syn.store(newfolder)
-    link = Link(newfolder.id, parent = folder.id)
```
kdaily: should this be removed, or another test created?

thomasyu888: So this shouldn't matter anymore, as I fixed the Link entity to accept both Synapse ids or entities.

kdaily: I meant should you be testing both (passing in ID or an entity object)? I'm fine with not just wondering what we do in other tests that create Synapse objects.

thomasyu888: If you notice https://github.com/Sage-Bionetworks/synapsePythonClient/pull/420/files/83281f034455f150013c0de58177b6f89a09f3dd#diff-7c91124835c610ac5387325cc1920f6dR123

I do a link out to a file entity by using the id.  I passed in the newfolder entity to test the feature I added in the Link entity.

kdaily: Aha - didn't expand far enough to see that. My bad!

thomasyu888: Merge?

PR #427, opened 2017-06-28
fix spelling mistake in metadata manifest file name

kdaily:
`METEDATA` should be `METADATA`


brucehoff:
Good job!  Thanks, Kenny.

PR #436, opened 2017-08-09
renamed _findEntityIdByNameAndParent() to findEntityId() (SYNPY-506)

zimingd:


larssono:
All of our public functions are camel case if remember correctly.  I don't think we should start mixing them in the public interface.

PR #443, opened 2017-08-21
Update sync.py to allow file handles instead of paths

kdaily:


synapseutils/sync.py
```diff
@@ -216,19 +216,27 @@ def readManifestFile(syn, manifestFile):
...
     sys.stdout.write('OK\n')
 
-    sys.stdout.write('Validating that all paths exist')
-    df.path = df.path.apply(_check_path_and_normalize)
+    if 'dataFileHandleId' in df.columns and 'path' in df.columns:
+        raise ValueError("Cannot have 'path' and 'dataFileHandleId' columns in the same manifest.")
 
-    sys.stdout.write('OK\n')
+    if 'path' in df.columns:
+        sys.stderr.write('Validating that all paths exist')
```
zimingd: stdout instead of stderr

synapseutils/sync.py
```diff
@@ -216,19 +216,27 @@ def readManifestFile(syn, manifestFile):
...
+        df.path = df.path.apply(_check_path_and_normalize)
+    else:
+        sys.stderr.write('Validating that all file handles exist')
```
zimingd: stdout

synapseutils/sync.py
```diff
@@ -216,19 +216,27 @@ def readManifestFile(syn, manifestFile):
...
-    sys.stdout.write('Validating provenance...')
-    df = _sortAndFixProvenance(syn, df)
-    sys.stdout.write('OK\n')
+    if 'path' in df.columns:
+        sys.stdout.write('Validating that all files are unique...')
+        if (len(df.path)!=len(set(df.path))):
+            raise ValueError("All rows in manifest must contain a unique file to upload")
+        sys.stdout.write('OK\n')
+
+    if 'path' in df.columns:
```
zimingd: the if statement can be combined with the one above

synapseutils/sync.py
```diff
@@ -334,11 +342,13 @@ def syncToSynapse(syn, manifestFile, dryRun=False, sendMessages=True, retries=MA
 
     """
     df = readManifestFile(syn, manifestFile)
-    sizes = [os.stat(os.path.expandvars(os.path.expanduser(f))).st_size for f in df.path if not is_url(f)]
-    #Write output on what is getting pushed and estimated times - send out message.
-    sys.stdout.write('='*50+'\n')
-    sys.stdout.write('We are about to upload %i files with a total size of %s.\n ' %(len(df), utils.humanizeBytes(sum(sizes))))
-    sys.stdout.write('='*50+'\n')
+
+    if 'dataFileHandleId' not in df.columns:
```
zimingd: what about a case with a mix of dataFileHandleId and path? will nothing be printed but we still perform uploads?

synapseutils/sync.py
```diff
@@ -216,19 +216,27 @@ def readManifestFile(syn, manifestFile):
             raise ValueError("Manifest must contain a column of %s" %field)
     sys.stdout.write('OK\n')
 
-    sys.stdout.write('Validating that all paths exist')
-    df.path = df.path.apply(_check_path_and_normalize)
+    if 'dataFileHandleId' in df.columns and 'path' in df.columns:
+        raise ValueError("Cannot have 'path' and 'dataFileHandleId' columns in the same manifest.")
 
-    sys.stdout.write('OK\n')
+    if 'path' in df.columns:
```
zimingd: what happens in the edge case when the 'path' header is specified in the TSV but the rest of the column is left blank?

synapseutils/sync.py
kdaily: There's another open issue (Jira) to move all output to stderr. Using stdout means you can't easily redirect any output, and should be standard all the way across our client.

synapseutils/sync.py
thomasyu888:
```suggestion
```

synapseutils/sync.py
```diff
@@ -216,19 +216,27 @@ def readManifestFile(syn, manifestFile):
...
 
-    sys.stdout.write('Validating provenance...')
-    df = _sortAndFixProvenance(syn, df)
-    sys.stdout.write('OK\n')
+    if 'path' in df.columns:
+        sys.stdout.write('Validating that all files are unique...')
+        if (len(df.path)!=len(set(df.path))):
+            raise ValueError("All rows in manifest must contain a unique file to upload")
+        sys.stdout.write('OK\n')
+
```
thomasyu888:
```suggestion
```

PR #454, opened 2017-09-05
removed usage of 'totalNumberOfResult' for paginated results (SYNPY-512)

zimingd:


synapseclient/client.py
```diff
@@ -2525,18 +2524,12 @@ def _GET_paginated(self, uri, limit=20, offset=0):
...
-
-            ## platform bug PLFM-3589 causes totalNumberOfResults to be too large,
-            ## by counting evaluations to which the current user does not have access.
-            ## So, we need to check for empty results and bail if we see that.
-            if len(results) == 0:
-                totalNumberOfResults = offset
+            prev_num_results = len(results)
 
             for result in results:
                 offset += 1
```
larssono: why not offset+=len(results)?

PR #458, opened 2017-09-12
_getFileHandleDownload() raises Errors if response contains failureCode (SYNPY-408)

zimingd:


synapseclient/client.py
```diff
@@ -1855,10 +1855,16 @@ def _getFileHandleDownload(self, fileHandleId,  objectId, objectType='FileEntity
...
-        results = self.restPOST('/fileHandle/batch', body=json.dumps(body),
+        response = self.restPOST('/fileHandle/batch', body=json.dumps(body),
                                 endpoint=self.fileHandleEndpoint)
-        return results['requestedFiles'][0]
+        result = response['requestedFiles'][0]
+        failure = result.get('failureCode')
+        if failure == 'NOT_FOUND':
+            raise exceptions.SynapseFileNotFoundError("The fileHandleId %s could not be found" % fileHandleId)
+        elif failure == "UNAUTHORIZED":
+            raise exceptions.SynapseError("You are not authorized to access fileHandleId %s associated with the synapse Entity %s" % (fileHandleId, objectId))
```
larssono: Change error message to include object Id and the last part of object type.

synapseclient/client.py
```diff
@@ -1812,10 +1812,16 @@ def _getFileHandleDownload(self, fileHandleId,  objectId, objectType='FileEntity
...
-        results = self.restPOST('/fileHandle/batch', body=json.dumps(body),
+        response = self.restPOST('/fileHandle/batch', body=json.dumps(body),
                                 endpoint=self.fileHandleEndpoint)
-        return results['requestedFiles'][0]
+        result = response['requestedFiles'][0]
+        failure = result.get('failureCode')
+        if failure == 'NOT_FOUND':
+            raise exceptions.SynapseFileNotFoundError("The fileHandleId %s could not be found" % fileHandleId)
+        elif failure == "UNAUTHORIZED":
+            raise exceptions.SynapseError("You are not authorized to access fileHandleId %s associated with the Synapse %s: %s" % (fileHandleId, objectType, objectId))
```
kimyen: We only have 2 failure code in the backend now, but this could change. The client should handle a default case when failure is not None.

PR #464, opened 2017-09-22
Synr 1125

brucehoff:


synapseclient/doze.py
```diff
@@ -0,0 +1,29 @@
+'''
+Created on Sep 21, 2017
+
+@author: bhoff
+
+sleep while checking registered listeners
+'''
+import time
+
+listeners=[]
```
brucehoff: _listeners

synapseclient/doze.py
```diff
@@ -0,0 +1,29 @@
...
+
+def add_listener(listener):
+    global listeners
```
brucehoff: global -> nonlocal ??

tests/unit/unit_test_doze.py
```diff
@@ -0,0 +1,24 @@
...
+
+@author: bhoff
+'''
+import synapseclient.doze as doze
+
+def setup():
+    global counter
+    counter = 0
+    
+def counter_inc():
```
brucehoff: define as an inner function to test_doze.  Then 'global' is not needed.

synapseclient/doze.py
```diff
@@ -0,0 +1,29 @@
+'''
+Created on Sep 21, 2017
```
brucehoff: use different names for module and function

synapseclient/doze.py
zimingd: I was wrong about the use of `nonlocal` during our previous. `nonlocal` would not work because it is a Python 3 only feature and we have to also support Python 2.

Using `del listeners[:]` instead of `listeners=[]` avoids having to use `global` since you are no longer reassigning `listeners` so there would not be a scoping issue

synapseclient/doze.py
```diff
@@ -0,0 +1,29 @@
...
+
+@author: bhoff
+
+sleep while checking registered listeners
+'''
+import time
+
+listeners=[]
+
+def add_listener(listener):
```
zimingd: add a safety check so we dont end up calling` listener()` on a non callable variable during `doze`
```
if not callable(listener):
    raise ValueError("listener is not callable")
```

PR #466, opened 2017-09-26
SYNR-1134: extend Table() to take a dict as values

kimyen:


synapseclient/table.py
```diff
@@ -813,6 +813,7 @@ def Table(schema, values, **kwargs):
       - a list of lists (or tuples) where each element is a row
       - a string holding the path to a CSV file
       - a Pandas `DataFrame <http://pandas.pydata.org/pandas-docs/stable/api.html#dataframe>`_
+      - a dict where each key is a column
```
kimyen: - a dict which will be wrapped by a pandas DataFrame
include a link to pandas DataFrame

tests/unit/unit_test_tables.py
```diff
@@ -181,6 +181,26 @@ def test_as_table_columns():
     except ImportError as e1:
         sys.stderr.write('Pandas is apparently not installed, skipping test_as_table_columns.\n\n')
 
+def test_dict_to_table():
+    try:
+        import pandas as pd
```
kimyen: use a helper method:
```
def _try_import_pandas():
try:
  import pandas as pd
  return pd
except ImportError:
  raise SkipTest()
```

then in the test:
```
pd = _try_import_pandas()
<other commands>
```

PR #472, opened 2017-10-03
allow non-threaded version of multipart upload

kdaily:
https://sagebionetworks.jira.com/browse/SYNPY-553

### DIFF
--- a/synapseclient/multipart_upload.py
+++ b/synapseclient/multipart_upload.py
@@ -294,11 +294,26 @@ def _upload_chunk(part, completed, status, syn, filename, get_chunk_function,
             sys.stderr.write(str(ex1))
             sys.stderr.write("Encountered an exception: %s. Retrying...\n" % str(type(ex1)))
 
-class FakePool(Object):
+class DummyPool():
     def map(self, *args, **kwargs):
         return map(*args, **kwargs)
     def terminate(self):
         return None
+
+class DummyValue():
+
+    def __init__(self, value):
+        self.value = value
+
+    def get_lock(self):
+        return DummyLock()
+
+class DummyLock():
+    def __enter__(self):
+        pass
+
+    def __exit__(self ,type, value, traceback):
+        return False
 
 def _multipart_upload(syn, filename, contentType, get_chunk_function, md5, fileSize,
                       partSize=None, storageLocationId = None,**kwargs):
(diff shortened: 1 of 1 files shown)


synapseclient/multipart_upload.py
```diff
@@ -329,18 +348,34 @@ def _multipart_upload(syn, filename, contentType, get_chunk_function, md5, fileS
...
     retries=0
-    mp = Pool(8)
+
+    try:
+        mp = Pool(8)
+        expiredValue = Value(c_bool, False)
+        singleThreaded = False
+    except OSError:
+        mp = DummyPool()
+        expiredValue = DummyValue(False)
```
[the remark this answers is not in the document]
zimingd: If you do the refactor suggested above,  just revert the changes to `expired`/`expiredValue `. The only issue reverting it would cause is repeated warning messages about url expiration (but upload will still work) when you have a large file upload (should be fixed by #467)

synapseclient/multipart_upload.py
```diff
@@ -295,8 +294,28 @@ def _upload_chunk(part, completed, status, syn, filename, get_chunk_function,
             sys.stderr.write(str(ex1))
             sys.stderr.write("Encountered an exception: %s. Retrying...\n" % str(type(ex1)))
 
+class DummyPool():
```
zimingd: Since this is a temporary workaround until AWS Lambda gets fixed we should try to make as little change to the current multipart_upload as possible.

Move the Dummy* classes into own module file. (I also recommend not having "dummy" in the module name since that might be confused with multiprocessing.dummy)

rename these classes to be Pool() and Value() and make their __init__() emulate that of the multiprocessing module classes. 
e.g. you can do fakemultiprocessing.Pool(8) but the 8 gets ignored. ('fakemultiprocessing' is just a placeholder for a module name that you think would fit)

then in the import section you can do a try block:
```
try:
   from multiprocessing import Value
   from multiprocessing.dummy import Pool
   Value('d',0) # just try to create a value and it should fail on AWS Lambda; no need to assign to a variable
except OSError:
   from .fakemultiprocessing import Pool, Value
```
Now there is no need for the if check on the `singleThreaded ` boolean and the unchanged logic in multipart_upload should work as is

synapseclient/multipart_upload.py
```diff
@@ -329,18 +348,34 @@ def _multipart_upload(syn, filename, contentType, get_chunk_function, md5, fileS
...
+        expiredValue = DummyValue(False)
+        singleThreaded = True
     try:
         while retries<MAX_RETRIES:
             ## keep track of the number of bytes uploaded so far
-            completed = Value('d', min(completedParts * partSize, fileSize))
 
-            printTransferProgress(completed.value, fileSize, prefix='Uploading', postfix=filename)
-            chunk_upload = lambda part: _upload_chunk(part, completed=completed, status=status, 
+            if singleThreaded:
```
zimingd: this if else block should no longer be necessary after the refactor changes suggested above.

zimingd:
We ran into a similar issue and implemented it here
#574

PR #488, opened 2017-11-06
Pass in followLink for recursive syncFromSynapse

thomasyu888:


thomasyu888:
https://sagebionetworks.jira.com/browse/SYNPY-595

synapseutils/sync.py
```diff
@@ -116,9 +116,10 @@ def generateManifest(syn, allFiles, filename):
     annotKeys = set()
     data = []
     for entity in allFiles:
-        row = {'parent': entity['parentId'], 'path': entity.path, 'name': entity.name,
+        row = {'parent': entity['parentId'], 'path': entity.get("path"), 'name': entity.name,
                'synapseStore': entity.synapseStore, 'contentType': allFiles[0]['contentType']}
-        row.update({key:val[0] for key, val in entity.annotations.items()})
+        row.update({key:val[0] for key, val in entity.annotations.items() if len(val) > 0})
+        row.update({key:"" for key, val in entity.annotations.items() if len(val) == 0})
```
zimingd: you can avoid iterating the annotations twice with
`row.update({key:(val[0] if len(val) > 0 else "") for key, val in entity.annotations.items()})`

zimingd:
Please add an integration test to the tests_synapseutils_sync.py file that uses Links recursively
Make sure to use `schedule_for_cleanup()` to clean up any downloaded files as a part of the test.

PR #490, opened 2017-11-15
Added PartialRowSet (SYNPY-597)

zimingd:


synapseclient/table.py
```diff
@@ -473,7 +473,8 @@ class SchemaBase(Entity, Versionable):
     _property_keys = Entity._property_keys + Versionable._property_keys + ['columnIds']
     _local_keys = Entity._local_keys + ['columns_to_store']
 
-    @abstractproperty  #forces subclasses to define _synapse_entity_type
+    @property
+    @abstractmethod  #forces subclasses to define _synapse_entity_type
     def _synapse_entity_type(self):
         pass
```
larssono: At first observation this seems non-pythonic.  I am sure I will change my mind when I get to the use of this.

synapseclient/table.py
```diff
@@ -692,8 +693,65 @@ def __init__(self, **kwargs):
...
+
+
+    def _synapse_store(self, syn):
+        """
+        Creates and POSTs an AppendableRowSetRequest_
+
+        .. AppendableRowSetRequest: http://docs.synapse.org/rest/org/sagebionetworks/repo/model/table/AppendableRowSetRequest.html
+        """
+        print(self)
+        append_rowset_request ={'concreteType': concrete_types.APPENDABLE_ROWSET_REQUEST,
```
larssono: Remove debug print statement.

synapseclient/table.py
```diff
@@ -692,8 +693,65 @@ def __init__(self, **kwargs):
...
+        response = syn._POST_table_transaction(self.tableId, append_rowset_request)
+        return response['results'][0]
+
 
-class RowSet(DictObject):
+
+class PartialRowset(AppendableRowset):
+    """
+    A set of Partial Rows used for Partially updating cells within a table
+
```
larssono: Should probably be described in more detail.  What does partial rows mean? Or how about "Partially updating cells"

synapseclient/table.py
```diff
@@ -692,8 +693,65 @@ def __init__(self, **kwargs):
...
+    :param schema: The :py:class:`Schema` of thee table to update or its tableId as a string
```
larssono: s/thee/the

synapseclient/table.py
```diff
@@ -692,8 +693,65 @@ def __init__(self, **kwargs):
...
+    :param rows: A list of PartialRows
```
larssono: What are partial rows?  What is the type? Can you show an example?

synapseclient/table.py
```diff
@@ -692,8 +693,65 @@ def __init__(self, **kwargs):
...
+    """
+
+    def __init__(self, schema, rows):
+        super(PartialRowset, self).__init__(schema)
+
```
larssono: Why all the spaces between lines.

synapseclient/table.py
```diff
@@ -692,8 +693,65 @@ def __init__(self, **kwargs):
...
+        self.concreteType = concrete_types.PARTIAL_ROW_SET
+
```
larssono: ditto.

synapseclient/table.py
```diff
@@ -777,6 +821,51 @@ def __init__(self, values, rowId=None, versionNumber=None, etag=None):
...
+class PartialRow(DictObject):
+    """
+    A partial update for a single row.
+    Note "values" is a :py:class:`dict` where:
+        - key is the columnId to the desired row
+        - value is the new desired value for that column
+    The easiest way to know which columnIds to use is to check the `headers` of your most recent query:"
+        query_result = syn.TableQuery("SELECT * FROM syn123")
+        name_to_col_id = {col.name:col.id for col in query_results.headers}
+
```
larssono: This seems pretty annoying for the user to deal with.  Why not pass in a query (i.e. TableQueryResult) instead of a name_to_col_id?

synapseclient/table.py
```diff
@@ -777,6 +821,51 @@ def __init__(self, values, rowId=None, versionNumber=None, etag=None):
...
+        row_id = #you to set this yourself#
+
```
larssono: You need a complete example here.

synapseclient/table.py
```diff
@@ -777,6 +821,51 @@ def __init__(self, values, rowId=None, versionNumber=None, etag=None):
...
+        values_to_change = {'My Row Name': "new value",
+                            'My Other Row': 42}
+        partial_row = PartialRow(values_to_change, row_id, nameToColumnId=name_to_col_id)
+
+    :param values: A dict where:
+                - key is the columnId to the desired row
+                - value is the new desired value for that column
+    :param rowId: THe id of the row to be updated
+    :param etag: used for updating File/Project Views. Not necessary if updating data Tables
```
larssono: Rephrase, a data Table is not clear - 'only needed for File and Project Views'  Also add link to these classes.

synapseclient/table.py
```diff
@@ -777,6 +821,51 @@ def __init__(self, values, rowId=None, versionNumber=None, etag=None):
...
+    :param nameToColumnId: Optional map column names to column Ids. If this is provided, the keys of your `values`
+                           dict will be replaced with the column ids in the `nameToColumnId` dict
+
```
larssono: Why is this optional? When is it optional? If it is why even include it?

synapseclient/table.py
```diff
@@ -777,6 +821,51 @@ def __init__(self, values, rowId=None, versionNumber=None, etag=None):
...
+    """
+    def __init__(self, values, rowId, etag=None, nameToColumnId = None):
+        super(PartialRow, self).__init__()
+        if not isinstance(values, dict):
+            raise ValueError("values must be a dict")
```
larssono: Why not any mapping?

synapseclient/table.py
```diff
@@ -777,6 +821,51 @@ def __init__(self, values, rowId=None, versionNumber=None, etag=None):
...
+
+        if isinstance(rowId, (six.integer_types,six.string_types)):
+            try:
+                rowId = int(rowId)
+            except ValueError:
+                raise ValueError("rowId must be an integer or string that represents an integer")
```
larssono: This seems completely unecessary.  Just do int(rowID)  this will throw a value error if it is not an int.

PR #491, opened 2017-11-21
SYNR-1027

brucehoff:


synapseclient/client.py
```diff
@@ -3407,7 +3407,9 @@ def restGET(self, uri, endpoint=None, headers=None, retryPolicy={}, **kwargs):
         Performs a REST GET operation to the Synapse server.
 
         :param uri:      URI on which get is performed
-        :param endpoint: Server endpoint, defaults to self.repoEndpoint
+        :param endpoint: Server endpoint, defaults to 'https://repo-prod.prod.sagebase.org/file/v1'.  Some
```
zimingd: The endpoint does not always default to 'https://repo-prod.prod.sagebase.org/file/v1'
users can change the endpoint using `syn.setEndpoints(**synapseclient.client.STAGING_ENDPOINTS)`, in which case `endpoint` would default to the staging endpoint instead of production endpoint

brucehoff: @zimingd  that is a good point.  How would you write the Py client documentation so that someone could use 'restGET' along with the APIs doc's?

zimingd: Instead of listing the URLS you can mention the two constant dicts synapseclient.client.STAGING_ENDPOINTS
synapseclient.client.PRODUCTION_ENDPOINTS

and have then access the defiend endpoints like 

```
syn.restGET('/some/api/', endpoint=synapseclient.client.PRODUCTION_ENDPOINTS['fileHandleEndpoint'])
```

Or if always using the full name is too long
```
from synapseclient.client import PRODUCTION_ENDPOINTS
syn.restGET('/some/api/', endpoint=PRODUCTION_ENDPOINTS['fileHandleEndpoint'])
```

The available keys in PRODUCTION_ENDPOINTS and STAGING_ENDPOINTS are:
```
fileHandleEndpoint
repoEndpoint
authEndpoint
portalEndpoint
```

synapseclient/client.py
```diff
@@ -3407,7 +3407,10 @@ def restGET(self, uri, endpoint=None, headers=None, retryPolicy={}, **kwargs):
         Performs a REST GET operation to the Synapse server.
 
         :param uri:      URI on which get is performed
-        :param endpoint: Server endpoint, defaults to self.repoEndpoint
+        :param endpoint: Server endpoint, defaults to self.repoEndpoint.   
+        The required endpoint is shown in each service's page in the REST API doc's (http://docs.synapse.org/rest).
+        Options are: 'https://repo-prod.prod.sagebase.org/repo/v1', 'https://file-prod.prod.sagebase.org/file/v1' 
+        or 'https://file-prod.prod.sagebase.org/auth/v1'.
         :param headers:  Dictionary of headers to use rather than the API-key-signed default set of headers
```
larssono: This would encourage people to remember the endopoints these are already stored as member variables of the Synapse object.  Why not refer to the as:  self.repo)Endpoint, self.authEndpoint,  self.fileHandleEndpoint or at least include both. (There is also:self.portalEndpoint used in onWeb)

brucehoff: @larssono the use case is that of programming against the Synapse web APIs.  Say the developer refers to this page: http://docs.synapse.org/rest/POST/file/multipart.html
They gather from the API doc' that they need to call `Synapse.restPOST()` and the URL is `https://repo-prod.prod.sagebase.org/file/v1/file/multipart`  They must understand that this URL comprises and endpoint `https://repo-prod.prod.sagebase.org/file/v1` and a URI `/file/multipart`.  They would then call
```
Synapse.restPOST("/file/multipart", "https://repo-prod.prod.sagebase.org/file/v1")
```
In short, it's not so much a matter of "remembering" the endpoints as it is reconciling the API doc's with the Python client documentation.

PR #502, opened 2018-01-10
Synr 1111

brucehoff:


brucehoff:
These changes were reviewed and merged as PR 501.  Not sure why they still appear as a diff but I am merging them again.

PR #510, opened 2018-02-23
table.cast_valuies() now handles columnType=='LINK'

kdaily:
Add link to column types to add

### DIFF
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -449,7 +449,7 @@ def cast_values(values, headers):
         ## convert field to column type
         if field is None or field=='':
             result.append(None)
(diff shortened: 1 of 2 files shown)


synapseclient/table.py
```diff
@@ -449,7 +449,7 @@ def cast_values(values, headers):
         ## convert field to column type
         if field is None or field=='':
             result.append(None)
-        elif columnType in {'STRING', 'ENTITYID', 'FILEHANDLEID', 'LARGETEXT', 'USERID'}:
```
zimingd: Please also modify the test to include this case:  test_cast_values() in tests/unit/unit_test_tables.py

PR #513, opened 2018-03-21
Synapse client now uses a requests.Session() object for HTTP calls instead of using the static 'requests' module functions (SYNPY-647)

zimingd:


kimyen:
Looks good.

PR #521, opened 2018-04-06
remove outdated code

kimyen:
I believe that @zimingd recent work was about exposing the ability to set whether or not a user wants RowId and RowVersion in their query results. (Default TRUE.) This variable is used in the request to tell PLFM whether or not to return RowId and RowVersion. We no longer need to check the returned data.

### DIFF
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -1397,20 +1397,6 @@ class CsvFileTable(TableAbstractBaseClass):
             header=header,
             includeRowIdAndRowVersion=includeRowIdAndRowVersion)
 
-        ## A dirty hack to find out if we got back row ID and Version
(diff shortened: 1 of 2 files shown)


synapseclient/table.py
```diff
@@ -1397,20 +1397,6 @@ def from_table_query(cls, synapse, query, quoteCharacter='"', escapeCharacter="\
             header=header,
             includeRowIdAndRowVersion=includeRowIdAndRowVersion)
 
-        ## A dirty hack to find out if we got back row ID and Version
-        ## in particular, we don't get these back from aggregate queries
```
kimyen: User may ask for RowId and RowVersion but the backend may override this if the query is an aggregate query. We need to double check the backend behavior.

PR #522, opened 2018-04-06
Refactor

kimyen:


synapseclient/table.py
```diff
@@ -1719,20 +1708,6 @@ def to_row_object(row, row_id_col=None, row_ver_col=None):
                       etag=self.etag,
                       rows=[to_row_object(row, row_id_col, row_ver_col) for row in self])
 
-    def setColumnHeaders(self, headers):
```
kimyen: I am not sure if this can be removed. @larssono, @zimingd and I are looking into table.py and found that some code are no longer needed. However, we need your input on whether or not `CsvFileTable .setColumnHeaders()` can be removed or made private?

PR #532, opened 2018-04-23
 Fixed errors in generateManifest: unicode error in Python 2 (SYNPY-692) and setting incorrect contentType when multiple File entities exist(SYNPY-693)

zimingd:


setup.py
```diff
@@ -44,12 +44,12 @@
         'future',
         'backports.csv',
         'keyring',
-        'keyrings.alt'
     ],
     extras_require = {
         'pandas':  ["pandas"],
         'pysftp': ["pysftp>=0.2.8"],
         'boto3' : ["boto3"],
+        ':sys_platform=="linux2" or sys_platform=="linux"': ['secretstorage<3.0.0', 'keyrings.alt'],
```
kimyen: why do we need to categorize `linux` and `linux2`? Can we combine them into `linux` or OS specific?

kimyen: Please add a comment mention the reason why we have to do this. If this is temporarily, please make sure that it's clear indicate what should happen instead.

synapseutils/sync.py
```diff
@@ -111,14 +113,18 @@ def generateManifest(syn, allFiles, filename):
 
     :param filename: file where manifest will be written
     """
-    keys = ['path', 'parent', 'name', 'synapseStore', 'contentType', 'used',
-            'executed', 'activityName', 'activityDescription']
+    keys, data = _process_manifest_rows(syn, allFiles)
+    _write_manifest_data(filename, keys, data)
+
+
+def _process_manifest_rows(syn, allFiles):
```
kimyen: Recommend adding 1 or more lines explaining what this helper function does?

synapseutils/sync.py
```diff
@@ -111,14 +113,18 @@ def generateManifest(syn, allFiles, filename):
...
     data = []
     for entity in allFiles:
         row = {'parent': entity['parentId'], 'path': entity.get("path"), 'name': entity.name,
-               'synapseStore': entity.synapseStore, 'contentType': allFiles[0]['contentType']}
-        row.update({key:(val[0] if len(val) > 0 else "") for key, val in entity.annotations.items()})
+               'synapseStore': entity.synapseStore, 'contentType': entity['contentType']}
+        row.update({key: (val[0] if len(val) > 0 else "") for key, val in entity.annotations.items()})
 
         annotKeys.update(set(entity.annotations.keys()))
         try:
```
kimyen: This block of code could be refactor to a helper method whose contract is to return a valid value. This helper should handle the error appropriately. There should be unit test for this helper.

tests/unit/unit_test_synapseutils_sync.py
```diff
@@ -121,3 +124,50 @@ def test_syncFromSynapse__non_file_Entity():
          patch.object(syn, "get", return_value = Schema(name="asssdfa", parent="whatever")):
         assert_raises(ValueError, synapseutils.syncFromSynapse, syn, table_schema)
 
+
+def test_write_manifest_data__unicode_characters_in_rows():
+    # SYNPY-693
+    if not pandas_available:
```
kimyen: This feature does not depend on pandas, the tests should not be depended on pandas. 
Our build system should make sure all tests passed. If pandas are not available, tests should fail.

setup.py
zimingd: From my understanding of the documentation, 'linux' is returned in Python 3. 'linux2' is returned in Python 2. https://docs.python.org/2/library/sys.html#sys.platform

we can't use the recommended `startswith('linux')` because sys_platform is not a variable. it is a magics string that is bound to a `lambda: sys.platform`. if we try to use `startswith` we get an error `Invalid environment marker: sys_platform.startswith("linux")`

kimyen:
looks good!

PR #541, opened 2018-05-15
SYNPY-688

kimyen:


.gitignore
```diff
@@ -15,3 +15,6 @@ notes.md
 notes/
 examples/
 jenkins/
+
+.idea/*
```
larssono: Is this folder specific for this project?  If you have your own files that you want to be gitignored I suggest using a ~/.gitignore_global.

synapseutils/sync.py
```diff
@@ -67,15 +67,26 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
...
-    results = syn.getChildren(id)
-    zero_results = True
-    for result in results:
-        zero_results = False
-        if is_container(result):
-            if path is not None:  #If we are downloading outside cache create directory.
-                new_path = os.path.join(path, result['name'])
+    if not is_container(entity):
+        raise ValueError("The provided id: %s is was neither a container nor a File" % id)
+
```
larssono: "is was" -> is

synapseutils/sync.py
```diff
@@ -67,15 +67,26 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
             print(f.path)
 
     """
+    # initialize the result list
     if allFiles is None: allFiles = list()
+
+    # perform validation check on user input
+    if is_synapse_id(entity):
+        entity = syn.get(entity)
+
```
larssono: What happens if this is a very large file?  Should we download it then?

.gitignore
kimyen: This folder `.idea` is added by pyCharm.

synapseutils/sync.py
kimyen: `syncFromSynapse` does not only work for a container (`project`/ `file`), it also works when users give it a `file` entity. And yes, the expected behavior is to download the file.

The documentation on `entity` parameter does not mention `file`, but the error and the integration test (integration.test_synapseutils.test_syncFromSynapse__given_file_id) indicates that this function should work for files. 

I can change the behavior if you think we should limit this function to only work for container exclusively.

synapseutils/sync.py
```diff
@@ -67,15 +67,30 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
             print(f.path)
 
     """
+    # initialize the result list
```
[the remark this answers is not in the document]
kimyen: This comment was not on the line that is fixed. The fix is done, so I am going to dismiss this comment.

PR #543, opened 2018-05-17
SYNPY-646

kimyen:


tests/unit/unit_test_client.py
```diff
@@ -381,3 +389,23 @@ def test_username_property__credentials_is_None():
     syn.credentials = None
     assert_is_none(syn.username)
 
+@patch('synapseclient.Synapse._getEntityBundle')
+@patch('synapseclient.Synapse._getWithEntityBundle')
+class TestGet():
```
kimyen: convert this class to use `setup` and `teardown` that managing patchers. See `TestLogin` class.

synapseclient/client.py
```diff
@@ -69,6 +69,7 @@
 import json
 from collections import OrderedDict
 import logging
+from numbers import Number
```
kimyen: remove this import

PR #546, opened 2018-05-23
SYNPY-713

kimyen:


tests/unit/unit_test_tables.py
```diff
@@ -184,7 +184,69 @@ def test_as_table_columns():
...
+            for row in data:
+                writer.writerow(row)
+
+        cols = as_table_columns(filename)
+
+        cols[0]['name'] == 'ROW_ID'
+        cols[0]['columnType'] == 'INTEGER'
+        cols[1]['name'] == 'ROW_VERION'
+        cols[1]['columnType'] == 'INTEGER'
+        cols[2]['name'] == 'Name'
```
kimyen: There should only be 4 columns returned. RowId and RowVersion should not be in this list.

PR #548, opened 2018-05-24
SYNPY-714

kimyen:


synapseclient/table.py
```diff
@@ -1134,13 +1144,14 @@ def build_table(name, parent, values):
...
 
-    if not isinstance(values, pd.DataFrame):
+    if not isinstance(values, pd.DataFrame) and not isinstance(values, six.string_types):
         raise ValueError("Values of type %s is not yet supported." % type(values))
     if not pandas_available:
         raise ValueError("pandas package is required.")
     cols = as_table_columns(values)
     schema = Schema(name=name, columns=cols, parent=parent)
-    return Table(schema, values)
+    headers = [SelectColumn.from_column(col) for col in cols]
```
kimyen: Per @zimingd, we should refactor the name of SelectColumn to something else that represents what the class is: type information for each column of the CSV file. SelectColumn used to be used for only for CSVs retrieved from synapse; in that case, SelectColumn represented the types of columns that were in the SQL SELECT clause (including aggregates).

tests/unit/unit_test_tables.py
```diff
@@ -736,18 +736,39 @@ def test_RowSetTable_len():
...
+                                       '2,d\n'
+                                       '3,e')
+    with patch.object(synapseclient.table, "as_table_columns",
+                      return_value = [Column(name = "a", columnType = "INTEGER"),
+                                      Column(name = "b", columnType = "STRING")]),\
+        patch.object(io, "open", return_value = string_io):
+        table = build_table("test", "syn123", "some_file_name")
+        assert len(table) == 3
+        for col, row in enumerate(table):
+            assert row[0] == (col + 1)
```
kimyen: use `assert_equals` instead because `assert` does not show the different between the 2 values

PR #551, opened 2018-06-07
SYNPY-729 Deprecate query() and chunkedQuery() functions

kimyen:


synapseclient/__main__.py
```diff
@@ -98,6 +98,9 @@ def query(args, syn):
         for row in reader:
             sys.stdout.write("%s\n" % ("\t".join(row)))
     else:
+        # chunkedQuery will be removed in 1.9.0
+        sys.stdout.write('This query is deprecated. Query should only be used for table and view.'
```
kimyen: ```sys.stdout.write``` -> ```sys.stderr.write```

PR #556, opened 2018-06-13
SYNPY-750

kimyen:


tests/unit/unit_test_tables.py
```diff
@@ -902,3 +901,66 @@ def test_iter_metadata__no_etag(self):
...
+               "1,2,etag1,\"I like trains\"\n" \
+                "5,1,etag2,\"weeeeeeeeeeee\"\n"
+        cols = as_table_columns(StringIOContextManager(data))
+        headers = [SelectColumn(name="ROW_ID", columnType="STRING"),
+                   SelectColumn(name="ROW_VERSION", columnType="STRING")] + \
+                  [SelectColumn.from_column(col) for col in cols]
+        with patch.object(io, "open", return_value=StringIOContextManager(data)):
+            table = CsvFileTable("syn123", "/fake/file/path", headers=headers)
+            for row in table:
+                assert_equals(3, len(row))
```
zimingd: I ran this test through the debugger
This has weird values when iterating:
`row` has the value of:
`['2', 'etag1', 'I like trains']`
`['1', 'etag2', 'weeeeeeeeeeee']`
which means you cut off only the ROW_ID column when iterating.
this does not seem like intended behavior

zimingd: As an additional note, since you are testing the `__iter__`method, all of your tests should assert the actual values you expect to get back from `__iter__`  instead of just the result's length.

zimingd: Sounds good.

There is also the possible case of  csv not contain Row Metadata, headers contain Row Metadata
This should only happen if the user provides a csv without Row Metadata but the `table` has `self.includeRowIdAndRowVersion==True`

tests/unit/unit_test_tables.py
```diff
@@ -902,3 +901,97 @@ def test_iter_metadata__no_etag(self):
...
+        # csv file has row metadata, self.headers does not
+        data = "ROW_ID,ROW_VERSION,col\n" \
+               "1,2,\"I like trains\"\n" \
+               "5,1,\"weeeeeeeeeeee\"\n"
+        cols = as_table_columns(StringIOContextManager(data))
+        headers = [SelectColumn.from_column(col) for col in cols]
+        with patch.object(io, "open", return_value=StringIOContextManager(data)):
+            table = CsvFileTable("syn123", "/fake/file/path", headers=headers)
+            i = 0
+            for row in table:
```
zimingd: use `zip()` to iterate 2 iterables at once
```
expected_rows = [["I like trains"], ["weeeeeeeeeeee"]]
for expected_row, table_row in zip(expected_rows,table): 
    assert_equals(expected_row, table_row)
```

you don't need to wait for full build again. just make sure these unit tests still pass once you modify the test

synapseclient/table.py
```diff
@@ -1811,17 +1811,25 @@ def setColumnHeaders(self, headers):
...
                     lineterminator=self.lineEnd,
                     quotechar=self.quoteCharacter)
-                if self.header:
-                    header = next(reader)
-
-                for row in reader:
-                    yield cast_values(row, headers)
+                csv_header = set(next(reader))
+                num_metadata_cols_diff = len(csv_header & row_metadata_headers) - num_row_metadata_in_headers
+                if num_metadata_cols_diff == 0 or (num_metadata_cols_diff > 0 and num_row_metadata_in_headers == 0):
```
zimingd: One thing I noticed during the reading the logic of the code was that I was converting booleans related to `num_metadata_cols_diff ` into the relationship between `num_row_metadata_in_headers` and `len(csv_header & row_metadata_headers)` (i.e. `num_row_metadata_columns_in_csv`).

In other words, the `num_metadata_cols_diff ` doesn't immediately relay what being `>0` or `<0` means; I must reread the variable declaration of the `num_metadata_cols_diff ` to figure it out. I feel it is easier to read/understand when you use `num_row_metadata_columns_in_csv > num_row_metadata_in_headers`. I would reserve the use of the `num_metadata_cols_diff ` only for `cast_values(row[num_metadata_cols_diff:], headers)`

PR #561, opened 2018-06-21
SYNPY-743 Lint the code base

kimyen:


synapseutils/sync.py
```diff
@@ -283,7 +282,7 @@ def readManifestFile(syn, manifestFile):
             container = syn.get(synId, downloadFile=False)
         except SynapseHTTPError as e:
             sys.stdout.write('\n%s in the parent column is not a valid Synapse Id\n' % synId)
-            raise(e)
+            raise e
```
zimingd: in this case you can remove the `as e` and simply call `raise`

PR #574, opened 2018-07-26
SYNPY-788

kimyen:


synapseclient/pool_provider.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def map(self, func, iterable):
+        for item in iterable:
+            func(item)
+
+    def terminate(self):
+        pass
+
+
+def PoolProvider(single_threaded=False):
```
zimingd: snake case this and rename to get_pool()

this is currently doing the same thing that we do with table which hides the fact that it's not actually a classs

synapseclient/multipart_upload.py
```diff
@@ -350,7 +352,7 @@ def _multipart_upload(syn, filename, contentType, get_chunk_function, md5, fileS
     syn.logger.debug("previously completed %d parts, estimated %d bytes" % (completedParts, previously_completed_bytes))
     time_upload_started = time.time()
     retries = 0
-    mp = Pool(8)
+    mp = PoolProvider(single_threaded=synapseclient.config.single_threaded)
```
zimingd: the pool_provider class should be able to check the config. multipart_upload doesn't need that responsibility

tests/unit/unit_test_multipart_upload.py
```diff
@@ -114,3 +116,36 @@ def chunk_upload(part):
 
         # assert _put_chunk was called at least once
         assert_greater_equal(len(mocked_put_chunk.call_args_list), 1)
+
+
+def test__multipart_upload():
```
zimingd: test for multipart shouldn't care if it gets a single thread or multi-thread pool back. it should just use whatever the provider gives

tests/unit/unit_test_multipart_upload.py
```diff
@@ -114,3 +116,36 @@ def chunk_upload(part):
...
+    mocked_get_chunk_function = MagicMock(side_effect=[1, 2, 3, 4])
+    file_size = 1*MB
+    filepath = make_bogus_binary_file(n=file_size)
+    md5 = md5_for_file(filepath).hexdigest()
+    status = {'partsState': {},
+              'uploadId': {},
+              'state': 'COMPLETED'}
+    with patch.object(syn, "restPOST", return_value=status),\
+            patch.object(SingleThreadPool, "map", return_value=None) as single_thread_pool_map,\
```
zimingd: these patch.objects would no longer be needed. just make the pool_provider.get_pool return a mock object

PR #589, opened 2018-08-24
SYNPY-195

kimyen:


synapseclient/entity.py
```diff
@@ -750,12 +750,9 @@ def is_versionable(entity):
     if isinstance(entity, Versionable):
```
kimyen: add type check for input

kimyen: also check all usages on `is_versionable()`

synapseclient/retry.py
```diff
@@ -112,11 +112,8 @@ def _get_message(response):
     getting body.
     """
     if _is_json(response.headers.get('content-type', None)):
-        try:
-            json = response.json()
-            return json.get('reason', None)
-        except (AttributeError, ValueError):
-            pass
```
kimyen: The response can be truncated. So in this except we should return None. `pass` means that the function returns None here. But having a return statement will make the code more readable.

synapseclient/utils.py
```diff
@@ -154,19 +154,17 @@ def extract_user_name(profile):
...
-        pass
-    try:
+    if hasattr(obj, key):
+        return getattr(obj, key)
+    if hasattr(obj, 'properties') and key in obj.properties:
+        return obj.properties[key]
+    try :
         if key in obj:
             return obj[key]
         elif 'properties' in obj and key in obj['properties']:
```
kimyen: try:
    if key in obj:
        return obj[key]
    else:
        return obj['properties'][key]
except (KeyError, TypeError):
    # we cannot get the key from this obj, so this case will be treated as key not found
    pass

synapseclient/entity.py
kimyen: `is_versionable()` is only used in `client.py`'s `_updateEntity`

PR #596, opened 2018-09-10
SYNPY-812 Github issue template

kimyen:


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
+## Operating system
```
kdaily: Add in a header saying what this is for (bugs) and what it's not, like 'If you have a question, please don't use this form, but instead ask on the Synapse Help Forum <https://www.synapse.org/#!SynapseForum:default>'.

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,23 @@
...
+## Client version
+*Output of:*
+```
+import synapseclient
+synapseclient.__version__
+```
+
+## Description of the problem
+
+*Provide a description of the problem, and if possible a minimal reproducible example.*
```
kdaily: More information on what a minimal reproducible example is. Might be things like don't share private data with us.
I hesitate to put 'don't include extra imports, etc.' because many people may stop from filing if they have to do too much work, but I could be convinced.

kimyen: I think in some cases, the problem happens when using the package with other libraries. In those cases, we can't repro without extra imports. So I mentions "often do not requires extra imports."

ISSUE_TEMPLATE.md
```diff
@@ -19,9 +19,9 @@ synapseclient.__version__
 *Provide a description of the problem, and if possible a minimal reproducible example.*
 
 A minimal reproducible example:
-* uses a minimal dataset, necessary to reproduce the error;
-* the minimal runnable code necessary to reproduce the error, which can be run on the given dataset;
-* and often do not requires extra imports.
+* uses a minimal dataset necessary to reproduce the error
+* the minimal runnable code necessary to reproduce the error, which can be run on the given dataset
+* often do not include extra imports
```
kdaily: 'should not include extra imported packages'

ISSUE_TEMPLATE.md
```diff
@@ -19,9 +19,9 @@ synapseclient.__version__
 *Provide a description of the problem, and if possible a minimal reproducible example.*
 
 A minimal reproducible example:
-* uses a minimal dataset, necessary to reproduce the error;
-* the minimal runnable code necessary to reproduce the error, which can be run on the given dataset;
-* and often do not requires extra imports.
+* uses a minimal dataset necessary to reproduce the error
+* the minimal runnable code necessary to reproduce the error, which can be run on the given dataset
```
kdaily: 'includes the minimal code necessary...'

kdaily:
there are some grammatical changes required here.

PR #597, opened 2018-09-10
SYNPY-810 Simplify example

kimyen:


synapseclient/table.py
```diff
@@ -204,7 +204,7 @@
 Query for the rows you want to delete and call syn.delete on the results::
 
     results = syn.tableQuery("select * from %s where Chromosome='2'" %table.schema.id)
```
kdaily: can you fix spacing after the `%table.schema.id` to `% table.schema.id`?

PR #601, opened 2018-09-12
SYNPY-694

kimyen:


tests/unit/unit_test_synapseutils_copy.py
```diff
@@ -0,0 +1,50 @@
...
+import synapseutils
+
+
+def setup(module):
+    module.syn = unit.syn
+
+
+def test_copyWiki_empty_Wiki():
+    entity = {"id": "syn123"}
+    with patch.object(syn, "getWikiHeaders", return_value=[]), \
```
zimingd: should test for getWikiHeaders() returning None

tests/unit/unit_test_synapseutils_copy.py
```diff
@@ -0,0 +1,50 @@
...
+            patch.object(syn, "get", return_value=entity), \
+            patch.object(syn, "getWiki"):
```
zimingd: unnecessary patch

tests/integration/test_synapseutils_copy.py
```diff
@@ -201,241 +198,122 @@ def test_copy():
...
-    subwiki = Wiki(owner=project_entity, title='A sub-wiki', 
-                   markdown='%s' % file_entity.id, parentWikiId=wiki.id)
-    subwiki = syn.store(subwiki)
-
-    second_md = """
-    Testing internal links
-    ======================
-
-    [test](#!Synapse:%s/wiki/%s)
+class test_copyWiki:
```
zimingd: camel case now that it is a class

synapseutils/copy.py
```diff
@@ -439,7 +439,7 @@ def _updateInternalLinks(newWikis, wikiIdMap, entity, destinationId):
         newWikiId = wikiIdMap[oldWikiId]
         newWiki = newWikis[newWikiId]
         print("\tUpdating internal links for Page: %s\n" % newWikiId)
-        s = newWiki.markdown
+        s = newWiki.get("markdown")
```
zimingd: since we are under the assumption that `markdown` will always exist. we should use `['markdown']` instead of 'get()'
so that an error will be thrown if an invalid value is used.

synapseutils/copy.py
```diff
@@ -500,20 +508,24 @@ def copyWiki(syn, entity, destinationId, entitySubPageId=None, destinationSubPag
...
+    if not oldWikiHeaders:
+        return []
+
     for wikiHeader in oldWikiHeaders:
-        wiki = syn.getWiki(oldOwn, wikiHeader.id)
-        print('Got wiki %s' % wikiHeader.id)
-        if not wiki['attachmentFileHandleIds']:
+        wiki = syn.getWiki(oldOwn, wikiHeader.get('id'))
+        print('Got wiki %s' % wikiHeader.get('id'))
+        if "attachmentFileHandleIds" not in wiki or not wiki['attachmentFileHandleIds']:
```
zimingd: `if not wiki.get('attachmentFileHandleIds'):`

PR #606, opened 2018-09-20
SYNPY-819 Support view with more entity types

kimyen:


synapseclient/client.py
```diff
@@ -3290,15 +3290,21 @@ def _build_table_download_file_handle_list(self, table, columns):
...
+                      'viewTypeMask': view_type_mask}
         columns = []
+        next_page_token = None
         while True:
-            response = self.restPOST('/column/view/scope', json.dumps(view_scope))
+            if next_page_token:
+                url = '/column/view/scope?nextPageToken=%s' % next_page_token
+            else:
+                url = '/column/view/scope'
+            response = self.restPOST(url, json.dumps(view_scope))
```
kimyen: instead of rewriting the url, use the `params=` optional arugment to include nextpagetoken. It may also be beneficial to expose `params=` in the `rest...()` methods. currently it would only work because `**kwargs` are forwarded to the `requests` package

synapseclient/table.py
```diff
@@ -771,11 +821,11 @@ def _before_synapse_store(self, syn):
         # get the default EntityView columns from Synapse and add them to the columns list
         additional_columns = []
         if self.addDefaultViewColumns:
-            additional_columns.extend(syn._get_default_entity_view_columns(self.type))
+            additional_columns.extend(syn._get_default_entity_view_columns(self.get('viewTypeMask')))
```
kimyen: use bracket notation `[]`. at this point we should be guaranteed that viewTypeMask is set. If it isnt due to an bug, we want this part of the code to error out.

synapseclient/table.py
```diff
@@ -720,19 +768,21 @@ class EntityViewSchema(SchemaBase):
     
         project_or_folder = syn.get("syn123")  
         schema = syn.store(EntityViewSchema(name='MyTable', parent=project, scopes=[project_or_folder_id, 'syn123'],
-         view_type='file'))
+         includeEntityTypes=['file']))
```
kimyen: remember to update the docs

synapseclient/table.py
```diff
@@ -334,6 +334,47 @@
...
 
 
+# View Type Mask
+FILE = 0x01
+PROJECT = 0x02
+TABLE = 0x04
+FOLDER = 0x08
+VIEW = 0x10
+DOCKER = 0x20
+VIEW_TYPE_MASK = {
```
kimyen: use enum

synapseclient/table.py
```diff
@@ -698,8 +732,15 @@ class EntityViewSchema(SchemaBase):
     :param columns:                         a list of :py:class:`Column` objects or their IDs. These are optional.
     :param parent:                          the project in Synapse to which this table belongs
     :param scopes:                          a list of Projects/Folders or their ids
-    :param type:                            the type of EntityView to display: either 'file','project' or
-                                            'file_and_table'. Defaults to 'file'.
+    :param type:                            This field is deprecated. Please use `includeEntityTypes`
+    :param includeEntityTypes:              a list of entity types to include in the view. Supported entity types are:
```
kimyen: This needs to be updated.

synapseclient/table.py
```diff
@@ -334,6 +335,39 @@
...
+    PROJECT = 0x02
+    TABLE = 0x04
+    FOLDER = 0x08
+    VIEW = 0x10
+    DOCKER = 0x20
+
+
+def _get_view_type_mask(types_to_include):
+    if not types_to_include:
+        raise ValueError("Please include at least one of the following entity types specified in EntityType.")
```
kimyen: "Please include at least one of the entity types specified in EntityType."

PR #607, opened 2018-09-20
SYNPY-824 Remove query()

kimyen:


synapseclient/__main__.py
```diff
@@ -131,7 +131,7 @@ def _getIdsFromQuery(queryString, syn):
 
         ids = [x['id'] for x in csv.DictReader(open(tbl.filepath))]
     else:
-        for item in syn.chunkedQuery(queryString):
+        for item in syn.__chunkedQuery(queryString):
```
kimyen: rename this

tests/integration/test_caching.py
```diff
@@ -190,9 +190,8 @@ def sleep_for_a_bit():
 
 def get_all_ids_from_Project():
     """Fetches all currently available Synapse IDs from the parent Project."""
-    
-    others = syn.chunkedQuery('select id from entity where parentId=="%s"' % syn.test_parent.id)
-    return [result['entity.id'] for result in others]
+    others = list(syn.getChildren(syn.test_parent.id))
+    return [result['id'] for result in others]
```
zimingd: you can avoid generating the list of children. just use

`for result in syn.getchildren(...)`

PR #625, opened 2018-11-22
Submission of Docker

thomasyu888:


synapseclient/client.py
```diff
@@ -2170,7 +2170,7 @@ def getTeamMembers(self, team):
         for result in self._GET_paginated('/teamMembers/{id}'.format(id=id_of(team))):
             yield TeamMember(**result)
 
-    def submit(self, evaluation, entity, name=None, team=None, silent=False, submitterAlias=None, teamName=None):
+    def submit(self, evaluation, entity, name=None, team=None, silent=False, submitterAlias=None, teamName=None, dockerTag="latest"):
```
kimyen: Please add this parameter to the function documentation. The parameter documentation should include the description about this parameter and its default value.

synapseclient/client.py
```diff
@@ -2263,6 +2262,17 @@ def submit(self, evaluation, entity, name=None, team=None, silent=False, submitt
...
 
+        if isinstance(entity, synapseclient.DockerRepository):
+            submission['dockerRepositoryName'] = entity.repositoryName
+            docker_commits = self.restGET('/entity/{}/dockerCommit'.format(entity.id)) 
+            docker_digest = None
+            for commit in docker_commits['results']:
+                if dockerTag  == commit['tag']:
+                    docker_digest = commit['digest']
+                    submission['dockerDigest'] = docker_digest
+            if docker_digest is None:
```
kimyen: This check should be done before making the REST call.

synapseclient/client.py
```diff
@@ -2263,6 +2262,17 @@ def submit(self, evaluation, entity, name=None, team=None, silent=False, submitt
         elif team:
             submission['submitterAlias'] = team.name
 
+        if isinstance(entity, synapseclient.DockerRepository):
+            submission['dockerRepositoryName'] = entity.repositoryName
+            docker_commits = self.restGET('/entity/{}/dockerCommit'.format(entity.id)) 
+            docker_digest = None
+            for commit in docker_commits['results']:
```
kimyen: Looking for a digest given a tag can be refactored to its own function.

synapseutils/copy.py
```diff
@@ -398,7 +398,7 @@ def _getSubWikiHeaders(wikiHeaders, subPageId, mapping=None):
     """
     Function to assist in getting wiki headers of subwikipages
     """
-    subPageId = str(subPageId)
+    subPageId = str(int(subPageId)) if isinstance(subPageId, float) else str(subPageId)
```
kimyen: This commit is not related to submitting docker entity, please put it in a separate pull request. Otherwise, please include full test coverage for this change.

kimyen:
Thanks @thomasyu888 for submitting the pull request! Please also provide the following:
* Unit test that demonstrates submitting a docker entity. This test fails without your changes, and shall pass with your change. We can discuss more about the test when it's in place. Please let me know if you need help writing the test(s).

kimyen:
I accidentally click "close and comment" instead of "comment". "reopen" now.

synapseutils/copy.py
thomasyu888: Oh! I didn't see this!  Sorry about that.  I removed it

synapseclient/client.py
thomasyu888: But the rest call list the commits (tag/digest pairs) for the given Docker repository.  I wouldn't be able to check this prior to the rest call.

kimyen: Oh I misread. Please check for dockerTag (the input parameter) is None before making the rest call.

kimyen: I got confused because of the error message. For dockerTag is None, the error message should explain what is wrong and what is the suggested behavior. For example: "A dockerTag is required to submit a DockerEntity."
For digest not found, the error message should first explain what went wrong, then what is the suggested behavior: "Disgest not found for tag: <xxx>. Please specify a docker tag that exists. The dockerTag 'latest' is used as default."

kimyen: Furthermore, the default value in the method signature does not guarantee that the value is not None in the function body. One can specify `submit( ..., dockerTag=None)` in their code. This seems to be an edge case, but our code should handle it anyway.

thomasyu888: Added these.

thomasyu888:
Still need to add unit tests to test for specifying the "team" parameter

synapseclient/client.py
```diff
@@ -2198,87 +2255,58 @@ def submit(self, evaluation, entity, name=None, team=None, silent=False, submitt
         """
 
         evaluation_id = id_of(evaluation)
+        entity = self.get(entity)
```
thomasyu888: download = False

thomasyu888:
Need help with using setup in unit test.

thomasyu888:
New PR

PR #626, opened 2018-11-22
change the subpage id to an integer than a string

thomasyu888:
I think there was a Jira ticket on the R synapseutils copy wiki.

### DIFF
--- a/synapseutils/copy.py
+++ b/synapseutils/copy.py
@@ -398,7 +398,7 @@ def _getSubWikiHeaders(wikiHeaders, subPageId, mapping=None):
     """
     Function to assist in getting wiki headers of subwikipages
     """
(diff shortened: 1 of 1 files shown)


kimyen:
@thomasyu888 thank you for the separated pull request! I really appreciate it.

Please correct me if I am wrong. From the changes, I assume that the function now works for both subpage Id of type string and of type float. If so, please add unit tests to verify that:
* subpage Id as float doesn't throw error
* subpage Id as string but represent a number doesn't throw TypeError nor ValueError (it may throw NotFoundException, but this is the backend error; and if we haven't test for this, it should be in a separated unit test case.)
* subpage Id of different type throw TypeError or ValueError.

PR #635, opened 2018-12-12
SYNPY-695 Copy File Views

thomasyu888:


thomasyu888:
Until we decide whether the copy function will actually live in the client, there isn't a need to add copy features.  This can be resolved by just not copying entityViews.

PR #636, opened 2018-12-12
SYNPY-888 Exclude coping Docker Repositories and EntityViews

thomasyu888:
In addition to Docker repositories not being able to be copied by the synapse api, also exclude entity views for now.  Closed #635 for now.


synapseutils/copy.py
```diff
@@ -198,7 +198,8 @@ def _copyRecursive(syn, entity, destinationId, mapping=None, skipCopyAnnotations
         if not isinstance(syn.get(destinationId), Project):
             raise ValueError("You must give a destinationId of a new project to copy projects")
         copiedId = destinationId
-        entities = syn.getChildren(entity)
+        #Projects include Docker repos, and Docker repos cannot be copied with the Synapse rest API
+        entities = syn.getChildren(entity, includeTypes=['folder', 'file', 'table', 'link', 'entityview'])
```
zimingd: add test

thomasyu888:
This is not really an urgent issue as the copy function does take `excludeTypes` which determines entity types to not copy.  But it is true that inherintly

PR #637, opened 2018-12-12
SYNPY-881 Restrict copying of some entities with only READ or have access restrictions

thomasyu888:


synapseutils/copy.py
```diff
@@ -194,24 +195,31 @@ def _copyRecursive(syn, entity, destinationId, mapping=None, skipCopyAnnotations
...
-        copiedId = _copyFolder(syn, ent.id, destinationId, mapping=mapping, skipCopyAnnotations=skipCopyAnnotations,
-                               **kwargs)
-    elif isinstance(ent, File) and "file" not in excludeTypes:
-        copiedId = _copyFile(syn, ent.id, destinationId, version=version, updateExisting=updateExisting,
-                             setProvenance=setProvenance, skipCopyAnnotations=skipCopyAnnotations)
-    elif isinstance(ent, Link) and "link" not in excludeTypes:
-        copiedId = _copyLink(syn, ent.id, destinationId, updateExisting=updateExisting)
-    elif isinstance(ent, Schema) and "table" not in excludeTypes:
-        copiedId = _copyTable(syn, ent.id, destinationId, updateExisting=updateExisting)
+    profile_username = syn.getUserProfile()['userName']
```
kdaily:
```suggestion
    profile_username = syn.username
```

synapseutils/copy.py
```diff
@@ -185,6 +185,7 @@ def _copyRecursive(syn, entity, destinationId, mapping=None, skipCopyAnnotations
         raise ValueError("Excluded types can only be a list of these values: file, table, and link") 
 
     ent = syn.get(entity, downloadFile=False)
```
kdaily: replace this with a get entity bundle, which removes need for two more rest calls later on

synapseutils/copy.py
```diff
@@ -194,24 +195,31 @@ def _copyRecursive(syn, entity, destinationId, mapping=None, skipCopyAnnotations
...
-        copiedId = _copyLink(syn, ent.id, destinationId, updateExisting=updateExisting)
-    elif isinstance(ent, Schema) and "table" not in excludeTypes:
-        copiedId = _copyTable(syn, ent.id, destinationId, updateExisting=updateExisting)
+    profile_username = syn.getUserProfile()['userName']
+    permissions = syn.getPermissions(ent, profile_username)
+
+    access_requirements = syn.restGET('/entity/%s/accessRequirement' % ent.id)
+
+    #Do not copy anything with access requirements
+    if access_requirements['totalNumberOfResults'] == 0:
```
kdaily: this will not handle if an AR is added to a team, only directly to an entity or container.

synapseutils/copy.py
```diff
@@ -194,24 +195,31 @@ def _copyRecursive(syn, entity, destinationId, mapping=None, skipCopyAnnotations
...
+        elif isinstance(ent, Folder):
+            copiedId = _copyFolder(syn, ent.id, destinationId, mapping=mapping, skipCopyAnnotations=skipCopyAnnotations,
+                                   **kwargs)
+        elif isinstance(ent, File) and "file" not in excludeTypes and "DOWNLOAD" in permissions:
+            copiedId = _copyFile(syn, ent.id, destinationId, version=version, updateExisting=updateExisting,
+                                 setProvenance=setProvenance, skipCopyAnnotations=skipCopyAnnotations)
+        elif isinstance(ent, Link) and "link" not in excludeTypes:
+            copiedId = _copyLink(syn, ent.id, destinationId, updateExisting=updateExisting)
+        elif isinstance(ent, Schema) and "table" not in excludeTypes and "DOWNLOAD" in permissions:
+            copiedId = _copyTable(syn, ent.id, destinationId, updateExisting=updateExisting)
```
kdaily: separate PR for this

kdaily: for other PR, probably can just check DOWNLOAD in permissions once for any entity type

synapseutils/copy.py
zimingd: note: requesting access_requirements in a bundle is deprecated and will always return an empty list

synapseutils/copy.py
zimingd: prefer early return so that the main chunk of the code is not forced to be extra indented 
```
move line 178 `copiedId = None` down here. No usage of that variable until we reach this point

if access_requirements['totalNumberOfResults'] == 0:
   print("%s not copied" % ent.id)
   return mapping

//rest of the if/elif logic
```

this also would allow you to remove the `if copiedId is not None` on line 224

PR #638, opened 2018-12-12
SYNPY-389 Submission of Synapse Docker Repository

thomasyu888:


synapseclient/client.py
```diff
@@ -2011,23 +2011,25 @@ def getTeam(self, id):
         """
         Finds a team with a given ID or name.
 
-        :param id:  The ID or name of the team to retrieve
+        :param id:  The ID or name of the team or a Team object to retrieve
```
kimyen: what is the use case to extend the API to take a team object?

synapseclient/client.py
```diff
@@ -2039,7 +2041,24 @@ def getTeamMembers(self, team):
...
-    def submit(self, evaluation, entity, name=None, team=None, silent=False, submitterAlias=None, teamName=None):
+    def _get_docker_digest(self, entity, docker_tag):
+        '''
+        Get matching Docker sha-digest of a DockerRepository given a Docker tag
+        
+        :param entity:      Synapse id of Docker repository
+        :param docker_tag:  Docker tag
+        :returns: Docker digest matching Docker tag
+        '''
+        docker_tags = self.restGET('/entity/{entityId}/dockerTag'.format(entityId=entity.id))
```
kimyen: This API returns 1 page of the result. This implementation cause the client not able to retrieve the tag in the case that the tag is in the 2+th page.

synapseclient/client.py
```diff
@@ -2039,7 +2041,24 @@ def getTeamMembers(self, team):
...
+        '''
+        Get matching Docker sha-digest of a DockerRepository given a Docker tag
+        
+        :param entity:      Synapse id of Docker repository
+        :param docker_tag:  Docker tag
+        :returns: Docker digest matching Docker tag
+        '''
+        docker_tags = self.restGET('/entity/{entityId}/dockerTag'.format(entityId=entity.id)) 
+        docker_digest = None
+        for tag in docker_tags['results']:
```
kimyen: The result of the rest call should be a collection of DockerCommit. Each item in the list should be a docker_commit (not tag) which include the tag and the digest: https://docs.synapse.org/rest/GET/entity/id/dockerTag.html

synapseclient/client.py
```diff
@@ -2039,7 +2041,24 @@ def getTeamMembers(self, team):
...
+            if docker_tag == tag['tag']:
+                docker_digest = tag['digest']
+        if docker_digest is None:
+            raise ValueError("Docker tag {docker_tag} not found.  Please specify a docker tag that exists. The dockerTag 'latest' is used as default.".format(docker_tag=docker_tag))
```
kimyen: The error indicates that the tag "latest" is used as default. However the method signature doesn't show this contract.

synapseclient/client.py
```diff
@@ -2039,7 +2041,24 @@ def getTeamMembers(self, team):
...
+        return(docker_digest)
+
+    def submit(self, evaluation, entity, name=None, dockerTag="latest", team=None, silent=False, submitterAlias=None, teamName=None):
```
kimyen: It's dangerous to add a parameter in the middle of the method signature. Old code that depends on the key word argument to appear at the right position will break.

synapseclient/client.py
```diff
@@ -2049,6 +2068,7 @@ def submit(self, evaluation, entity, name=None, team=None, silent=False, submitt
                                 In the absent of this parameter, the entity name will be used.
         :param team:            (optional) A :py:class:`Team` object, ID or name of a Team that is registered for the
                                 challenge
+        :param dockerTag:       (optional) The Docker tag must be specified if the entity is a DockerRepository. Defaults to "latest".
```
kimyen: Argument description should suggest how to use the argument. Having both "optional" and "must be specified" can be confusing. Maybe "(optional) Set the docker tag to use in submitting a DockerRepository. Default to 'latest'."  In this case, it's optional because the code taking care of the case when the user doesn't specify a value. So "must be specified" is not true here.

synapseclient/client.py
```diff
@@ -2097,12 +2124,20 @@ def submit(self, evaluation, entity, name=None, team=None, silent=False, submitt
...
+            docker_digest = self._get_docker_digest(entity, dockerTag)
+        else:
+            docker_digest = None
+
         submission = {'evaluationId': evaluation_id,
                       'name': name,
                       'entityId': entity_id,
                       'versionNumber': entity_version,
+                      'dockerDigest': docker_digest,
+                      'dockerRepositoryName':docker_repository,
```
kimyen: Please see pep8 for space after colon (:) https://www.python.org/dev/peps/pep-0008/#variable-annotations

synapseclient/client.py
```diff
@@ -2097,12 +2124,20 @@ def submit(self, evaluation, entity, name=None, team=None, silent=False, submitt
...
                       'teamId': team_id,
-                      'contributors': contributors,
+                      'contributors':contributors,
```
kimyen: pep8

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -263,6 +263,12 @@ def setup(self):
             'etag': 'etag',
             'name': 'entity name'
         }
+        self.docker = {
+            'versionNumber': 7,
```
kimyen: docker is not version-able entity. The test should ensure that without versionNumber in a mock docker entity, the function still executed as expected.

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -263,6 +263,12 @@ def setup(self):
...
+            'id': 'syn1009',
+            'etag': 'etag',
+            'name': 'entity name'
+        }
```
kimyen: docker should have an `repositoryName` -- this is used in the code. Is it possible for a docker entity to not have this field? If the backend says no, then the client should throw an exception.

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -354,6 +366,19 @@ def test_team_not_eligible(self):
...
+    def test_submit_docker(self):
+        docker_entity = DockerRepository("foo", parentId = "syn1000001")
+        docker_entity.id = "syn123"
+        docker_entity.etag = "Fake etag"
+
+        docker_tag = {'results': [{'tag': 'latest',
+                                   'digest': 'sha256:6b079ae764a6affcb632231349d4a5e1b084bece8c46883c099863ee2aeb5cf8'}]}
+
+        with patch.object(syn, "get", return_value=docker_entity) as syn_get_entity, \
+             patch.object(syn, "restGET", return_value=docker_tag) as syn_store_patch:
```
kimyen: this test should mock the private method.

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -354,6 +366,19 @@ def test_team_not_eligible(self):
...
+            assert_raises(ValueError, syn.submit,'9090', syn_get_entity, "George", dockerTag=None)
+            assert_raises(ValueError, syn.submit,'9090', syn_get_entity, "George", dockerTag="foo")
+            submission = syn.submit('9090', syn_get_entity, name='George')
```
kimyen: Needs a set of unit test for the private get digest.

kimyen:
Pep8 speak app also added a comment with all the pep8 suggestions. Please check that out.

synapseclient/client.py
thomasyu888: I was just updating the docstring, because the function does
``` 
teamid = id_of(id)
```

synapseclient/client.py
thomasyu888: This is specified in 
```
    def submit(self, evaluation, entity, name=None, dockerTag="latest", team=None, silent=False, submitterAlias=None, teamName=None):

```

thomasyu888:
Thanks for initial review @kimyen!  I provided updates.

synapseclient/client.py
```diff
@@ -2097,10 +2132,18 @@ def submit(self, evaluation, entity, name=None, team=None, silent=False, submitt
...
+            docker_digest = self._get_docker_digest(entity, dockerTag)
+        else:
+            docker_digest = None
+
         submission = {'evaluationId': evaluation_id,
                       'name': name,
                       'entityId': entity_id,
                       'versionNumber': entity_version,
+                      'dockerDigest': docker_digest,
+                      'dockerRepositoryName':docker_repository,
```
thomasyu888:
```suggestion
                      'dockerRepositoryName': docker_repository,
```

PR #654, opened 2019-01-31
SYNPY-859 define public APIs - aborting

kimyen:


synapseclient/__init__.py
```diff
@@ -266,28 +266,44 @@
...
+    'EntityViewType',
+    # constants
+    'PUBLIC', 'AUTHENTICATED_USERS', 'ROOT_ENTITY', 'DEBUG_LOGGER_NAME', 'DEFAULT_LOGGER_NAME']
 
+
+# ensure user-agent is set to track Synapse Python client usage
 import requests
 USER_AGENT = {'User-Agent': 'synapseclient/%s %s' % (__version__, requests.utils.default_user_agent())}
 
-from synapseclient.core import logging_setup as __logingsetup
```
zimingd: We import logging setup to monkey-patch logging behavior. Perhaps there is a better way to do this, but it should not be removed altogether

synapseclient/core/upload/__init__.py
```diff
@@ -0,0 +1,8 @@
+# These are exposed functions and objects from the `synapseclient.core.upload` package.
+# However, these functions and objects are not public APIs for the Synapse Python client.
+# The Synapse Engineering team is free to change their signatures and implementations anytime.
+# Please use them at your own risk.
+
+from .upload_functions import upload_file_handle, upload_synapse_s3
+from .multipart_upload import multipart_upload_file, multipart_upload_string, MIN_PART_SIZE
```
zimingd: kind of weird that modules outside of the package would need to know what MIN_PART_SIZE is. pleas double check what is using it and whether we can refactor that logic back inside this module

setup.py
```diff
@@ -64,5 +84,4 @@
         'Topic :: Software Development :: Libraries',
         'Topic :: Scientific/Engineering',
         'Topic :: Scientific/Engineering :: Bio-Informatics'],
-    platforms=['any'],
```
zimingd: in the "classifier" section, please listed supported OSes: Windows, Mac OS X, (specific linux distros?)

synapseclient/__main__.py
```diff
@@ -225,7 +226,7 @@ def associate(args, syn):
     for fp in files:
         try:
             ent = syn.get(fp, limitSearch=args.limitSearch)
-        except SynapseFileNotFoundError:
+        except synapseclient.core.models.SynapseFileNotFoundError:
```
zimingd: let's pulll the Erros into their own package (`synapseclient.errors.SynapseError`)

synapseclient/__main__.py
```diff
@@ -55,20 +55,21 @@
 A few more commands (cat, create, update, associate)
 
 """
+
```
zimingd: let's follow pipenv's convention for imports:
use `import` for external imports. For imports w/ lots of `.` alias it with `_` (e.g. `import urllib3.util as urllib3_util`)
for local imports, use relative paths and `from`

PR #661, opened 2019-02-08
SYNPY-855 move value to pool_provider

kimyen:


synapseclient/pool_provider.py
```diff
@@ -29,8 +31,25 @@ def terminate(self):
...
 
+class SingleValue:
+
+    value = None
+
+    def __init__(self, type, value):
+        self.value = value
+
+    def get_lock(self):
+        pass
```
zimingd: This will not work in a context manager, `with value.get_lock():`, which multipart_upload uses. 
`get_lock` should return an instance of a class that implements `__enter__` and `__exit__` that `pass`
http://book.pythontips.com/en/latest/context_managers.html#implementing-a-context-manager-as-a-class

my test:
```
In [18]: class SingleValue:
        value = None
        def __init__(self, type, value):
                self.value = value
        def get_lock(self):
                pass
   ....:

In [19]: v = SingleValue('d', 5)

In [20]: with v.get_lock():print(v.value)
AttributeError                            Traceback (most recent call last)
<ipython-input-20-ecf6bbb23070> in <module>()
----> 1 with v.get_lock():print(v.value)

AttributeError: __exit__
```

zimingd: By default all of our CIs are using the multi-threaded versions since most machines support that, so please also add tests that force multipart_upload to be used in a single threaded context.

PR #664, opened 2019-02-08
SYNPY-917 add documentation

kimyen:


zimingd:
Why are R code examples being put in the Python documentation?

PR #671, opened 2019-03-25
SYNPY-943 update the dockerfile to use current github code

kdaily:
This will now install the appropriate branch/tag/etc when built locally or in Dockerhub.

Fixes https://sagebionetworks.jira.com/browse/SYNPY-943

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,92 @@
+# Git
+.git
+.gitignore
+
+# CI
+.codeclimate.yml
+.travis.yml
+
+# Docker
+docker-compose.yml
+.docker
+
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*/__pycache__/
+*/*/__pycache__/
+*/*/*/__pycache__/
+*.py[cod]
+*/*.py[cod]
+*/*/*.py[cod]
+*/*/*/*.py[cod]
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+env/
+build/
+develop-eggs/
+dist/
+downloads/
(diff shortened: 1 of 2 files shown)


thomasyu888:
One particular thing I don't like about this approach is that if you build this locally, it will move anything from your local directory to the docker container.  We should add a `.dockerignore` file or state explicitly which synapseclient files to move over into the docker container.

kdaily:
There shouldn't be anything in your current directory (assuming you are building the docker image using standard workflows) other than the thing you are intending to build?

We should add explicit directions to the README regarding the docker file though.

thomasyu888:
I totally agree with you.  If you are building the docker image through the automated build, this isn't an issue at all. 

However, if I were to test this dockerfile locally, I could unintentionally add things to a dockerhub that I didn't want to.  (Accidentally docker pushing...)

kdaily:
I think the `.dockerignore` is the right path. There are standard practices around for building Docker images for Python packages, and we follow closely. It is part of a code review that should prevent accidental pushes to remote sources.

The only branches we should build are `master`, `develop`, and tagged and versioned releases. All of these will go through a code review before triggering a build in Docker Hub. No force pushes, temporary or feature branches, or other individuals should push to the Docker Hub repository - only triggered through GitHub continuous deployment processes.

kdaily:
@thomasyu888 and I talked in person, and I think (?) agreed. He raises an important point that we should determine developer best practices for Docker/Docker Hub to prevent accidental inclusion of data or files that should not be shared. I would suggest another document or SOP :-)

PR #695, opened 2019-10-08
add no force version flag and test

kdaily:


tests/integration/test_command_line_client.py
```diff
@@ -148,6 +149,26 @@ def test_command_line_client():
...
+                 '--noForceVersion',
+                 '--id',
+                 file_entity_id,
+                 filename)
+
+    # Get the File again and check that the version did not change
+    new_file = syn.get(file_entity_id, downloadFile=False)
+    new_version = new_file.versionNumber
+    assert_equals(current_version, new_version)
+
```
thomasyu888: The version number should be bumped if the file uploaded is different.  Test that even if force_version is False that the version number is increased.

kdaily: I handled this by updating the unit tests for Entity.

PR #697, opened 2019-10-09
SYNPY-881 Don't copy entities that don't have DOWNLOAD permission

thomasyu888:
Currently the permissions of entities aren't checked, so copy function attempts to copy entities with VIEW permissions and fails.


zimingd:
had to do a minor merge conflict resolution after merging #696. waiting CI to finish before merging

thomasyu888:
Had to modify some tests

PR #703, opened 2019-10-13
SYNPY-851 invite user/email to team

thomasyu888:
This functionality includes 4 convenience functions that wrap rest calls
* get team open invitations (GET)
* get membership status of a user (GET)
* delete membership invitation (DELETE)
* membership invitation (POST)

It also has a helper function that allows users to invite a Synapse user or an email to a Synapse team.  If the user is already part of the team, it will not send an invitation.  If an open invitation exists, the user will not be sent an invitation unless `force=True`

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2235,20 +2235,20 @@ class Synapse(object):
                                  body=json.dumps(invite_request))
         return response
 
-    def invite_to_team(self, team, force=False, **kwargs):
+    def invite_to_team(self, team, inviteeId=None, inviteeEmail=None,
+                       message=None, force=False):
         """Invite user to a Synapse team via Synapse username or email
         (choose one or the other)
 
         :param syn: Synapse object
         :param team: A :py:class:`synapseclient.team.Team` object or a
                      team's ID.
-        :param force: If an open invitation exists for the invitee,
-                      the old invite will be cancelled. Default to False.
         :param inviteeId: Synapse username or profile id of user
         :param inviteeEmail: Email of user
         :param message: Additional message for the user getting invited to the
                         team. Default to None.
-
+        :param force: If an open invitation exists for the invitee,
+                      the old invite will be cancelled. Default to False.
 
         :returns: MembershipInvitation or None if user is already a member
         """
@@ -2265,18 +2265,22 @@ class Synapse(object):
             userid = self.getUserProfile(user)['ownerId']
             membership_status = self.get_membership_status(userid, teamid)
             is_member = membership_status['isMember']
-            open_invites = [invitation for invitation in open_invitations
-                            if invitation.get('inviteeId') == user]
+            open_invites_to_user = [invitation
+                                    for invitation in open_invitations
+                                    if invitation.get('inviteeId') == user]
         else:
-            open_invites = [invitation for invitation in open_invitations
-                            if invitation.get('inviteeEmail') == email]
+            open_invites_to_user = [invitation
+                                    for invitation in open_invitations
+                                    if invitation.get('inviteeEmail') == email]
         # Only invite if the invitee is not a member and
         # if invitee doesn't have an open invitation unless force=True
-        if not is_member and (not open_invites or force):
+        if not is_member and (not open_invites_to_user or force):
             # Delete all old invitations
-            for invite in open_invites:
+            for invite in open_invites_to_user:
                 self._delete_membership_invitation(invite['id'])
-            return self.send_membership_invitation(teamid, **kwargs)
+            return self.send_membership_invitation(teamid, inviteeId=inviteeId,
+                                                   inviteeEmail=inviteeEmail,
+                                                   message=message)
         # Return None if no invite is sent.  Should an error should be thrown?
         return None
 
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2177,7 +2177,8 @@ class Synapse(object):
         https://docs.synapse.org/rest/GET/team/id/member/principalId/membershipStatus.html
 
         :param user: Synapse userid
-        :param team: A :py:class:`synapseclient.team.Team` object or a team's ID.
+        :param team: A :py:class:`synapseclient.team.Team` object or a
+                     team's ID.
 
         :returns: dict of TeamMembershipStatus"""
         teamid = id_of(team)
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2172,6 +2172,20 @@ class Synapse(object):
         for result in self._GET_paginated('/teamMembers/{id}'.format(id=id_of(team))):
             yield TeamMember(**result)
 
+    def get_team_open_invitations(self, team):
+        """Retrieve the open requests submitted to a Team
+        https://docs.synapse.org/rest/GET/team/id/openInvitation.html
+
+        :param team: A :py:class:`synapseclient.team.Team` object or a
+                     team's ID.
+
+        :returns: generator of MembershipRequest
+        """
+        teamid = id_of(team)
+        request = "/team/{team}/openInvitation".format(team=teamid)
+        open_requests = self._GET_paginated(request)
+        return open_requests
+
     def get_membership_status(self, userid, team):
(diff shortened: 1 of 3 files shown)


synapseclient/client.py
```diff
@@ -2172,6 +2172,62 @@ def getTeamMembers(self, team):
         for result in self._GET_paginated('/teamMembers/{id}'.format(id=id_of(team))):
             yield TeamMember(**result)
 
+    def get_membership_status(self, userid, team):
```
kdaily: What if they already have an open invitation? Should you check that too?

https://docs.synapse.org/rest/GET/user/id/openRequest.html

synapseclient/client.py
```diff
@@ -2172,6 +2172,62 @@ def getTeamMembers(self, team):
...
+        if email is None:
+            userid = self.getUserProfile(user)['ownerId']
+            membership_status = self.get_membership_status(userid, teamid)
+            is_member = membership_status['isMember']
+            invite['inviteeId'] = str(userid)
+        else:
+            invite['inviteeEmail'] = email
+
+        if not is_member:
+            invite = self.restPOST("/membershipInvitation",
```
kdaily: make a post membership invitation function and then build up the body here and pass `**my_body` to the `POST` function.

synapseclient/client.py
thomasyu888: What we want is: https://docs.synapse.org/rest/GET/team/id/openInvitation.html.  Unfortunately we can't exactly just check for a user's open invitations because the rest call doesn't allow you to put in email addresses.

synapseclient/client.py
```diff
@@ -2172,6 +2172,62 @@ def getTeamMembers(self, team):
...
+        teamid = id_of(team)
+        is_member = False
+        # Message can be None
+        invite = {'teamId': str(teamid),
+                  'message': message}
+
+        if email is None:
+            userid = self.getUserProfile(user)['ownerId']
+            membership_status = self.get_membership_status(userid, teamid)
+            is_member = membership_status['isMember']
```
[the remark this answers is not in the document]
thomasyu888: Agreed, we should definitely file a Jira for this.

synapseclient/client.py
```diff
@@ -2172,6 +2172,113 @@ def getTeamMembers(self, team):
...
+        return membership_status
+
+    def _delete_membership_invitation(self, invitationid):
+        """Delete open membership invitation
+
+        :param invitationid: Open invitation id
+        """
+        self.restDELETE("/membershipInvitation/{id}".format(id=invitationid))
+
+    def membership_invitation(self, teamId, inviteeId=None, inviteeEmail=None,
```
zimingd: `send_membership_invitation`

synapseclient/client.py
```diff
@@ -2172,6 +2172,113 @@ def getTeamMembers(self, team):
...
+        if inviteeEmail is not None:
+            invite_request['inviteeEmail'] = str(inviteeEmail)
+        if inviteeId is not None:
+            invite_request['inviteeId'] = str(inviteeId)
+
+        response = self.restPOST("/membershipInvitation",
+                                 body=json.dumps(invite_request))
+        return response
+
+    def invite_to_team(self, team, force=False, **kwargs):
```
zimingd: use defined parameters instead of kwargs

synapseclient/client.py
```diff
@@ -2172,6 +2172,113 @@ def getTeamMembers(self, team):
...
+
+        teamid = id_of(team)
+        is_member = False
+        open_invitations = self.get_team_open_invitations(teamid)
+
+        if user is not None:
+            userid = self.getUserProfile(user)['ownerId']
+            membership_status = self.get_membership_status(userid, teamid)
+            is_member = membership_status['isMember']
+            open_invites = [invitation for invitation in open_invitations
```
zimingd: rename to `open_invites_to_user` since this is a filtered list of `open_invitations `

synapseclient/client.py
```diff
@@ -2172,6 +2172,113 @@ def getTeamMembers(self, team):
...
+                            if invitation.get('inviteeEmail') == email]
+        # Only invite if the invitee is not a member and
+        # if invitee doesn't have an open invitation unless force=True
+        if not is_member and (not open_invites or force):
+            # Delete all old invitations
+            for invite in open_invites:
+                self._delete_membership_invitation(invite['id'])
+            return self.membership_invitation(teamid, **kwargs)
+        # Return None if no invite is sent.  Should an error should be thrown?
```
zimingd: remove the `Should an error should be thrown?` comment. Returning none makes sense

```suggestion
        # Return None if no invite is sent.
```

tests/unit/unit_test_client.py
```diff
@@ -821,3 +821,146 @@ def test_update(self):
...
+                                      inviteeEmail=self.email,
+                                      message=self.message)
+            patch_rest_post.assert_called_once_with("/membershipInvitation",
+                                                    body=json.dumps(invite_body))
+
+    def test_bothuseremail_invite_to_team(self):
+        """Raise error when user and email is passed in"""
+        assert_raises(ValueError, syn.invite_to_team, self.team,
+                      inviteeId=self.userid, inviteeEmail=self.email)
+
```
zimingd: add test case where both `inviteeId` and `inviteeEmail` are `None`

synapseclient/client.py
```diff
@@ -2172,6 +2172,113 @@ def getTeamMembers(self, team):
...
+        :param message: Additional message for the user getting invited to the
+                        team. Default to None.
+
+
+        :returns: MembershipInvitation or None if user is already a member
+        """
+        user = kwargs.get("inviteeId")
+        email = kwargs.get("inviteeEmail")
+        if email is not None and user is not None:
+            raise ValueError("Must only specify 'user' or 'email'")
```
zimingd: callers of this will be inputting `inviteeId` and `inviteeEmail`
```suggestion
            raise ValueError("Must only specify one of 'inviteeId' or 'inviteeEmail'")
```

tests/unit/unit_test_client.py
```diff
@@ -821,3 +821,88 @@ def test_update(self):
...
+    def setup(self):
+        self.team = synapseclient.Team(id=222)
+        self.userid = 123
+        self.email = "test@email.com"
+        self.member_status = {'isMember': True}
+        self.response = {'inviteeId': self.userid}
+        self.message = "custom message"
+        self.profile = {'ownerId': self.userid}
+
+    def test_teamid_get_membership_status(self):
```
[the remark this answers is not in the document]
zimingd: same, especially since other tests use `patch_rest_delete` and `patch_rest_post`

tests/unit/unit_test_client.py
```diff
@@ -821,3 +821,146 @@ def test_update(self):
         self.mock_restPUT.assert_called_once_with('/projectSettings', body=json.dumps(new_location))
         self.mock_restPOST.assert_not_called()
 
+
+class TestMembershipInvitation:
```
zimingd: as a general style, test names should be:
`test_<function/method name>__<case being tested>` 
to make it easier to identify the function/method under test
the double underscore helps to separate the name from being confused with the case being tested

tests/unit/unit_test_client.py
```diff
@@ -821,3 +821,146 @@ def test_update(self):
...
+                       'inviteeEmail': self.email}
+        with patch.object(syn, "get_team_open_invitations",
+                          return_value=[]) as patch_get_invites,\
+             patch.object(syn, "getUserProfile",
+                          return_value=self.profile) as patch_get_profile,\
+             patch.object(syn, "membership_invitation",
+                          return_value=self.response) as patch_invitation:
+            invite = syn.invite_to_team(self.team, inviteeEmail=self.email,
+                                        message=self.message)
+            assert invite == self.response
```
zimingd: I think nosetests does not handle `assert` keyword, use `assert_equals()`

allaway:
This is amaaazing. I had to write a bunch of custom REST calls to handle in September. Thanks for adding this to the client.

PR #718, opened 2019-12-11
Initial multithreaded download in current client, no tests

minneker:
For initial meeting with @zimingd at 3 PM today, tests are actively being developed locally and PR will be updated ASAP.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -29,10 +29,8 @@ See also the `Synapse API documentation <https://docs.synapse.org/rest/>`_.
 """
 import configparser
 import collections
-import os
 import errno
 import sys
-import time
 import hashlib
 import webbrowser
 import shutil
@@ -41,7 +39,6 @@ import mimetypes
 import tempfile
 import warnings
 import getpass
-import json
 import logging
 import urllib.parse as urllib_urlparse
 
@@ -55,6 +52,7 @@ from .evaluation import Evaluation, Submission, SubmissionStatus
 from .table import SchemaBase, Column, TableQueryResult, CsvFileTable
 from .team import UserProfile, Team, TeamMember, UserGroupHeader
 from .wiki import Wiki, WikiAttachment
+from .producer_consumer_download_threads import *
 from synapseclient.core import cache, exceptions
 from synapseclient.core.constants import config_file_constants
 from synapseclient.core.constants import concrete_types
@@ -93,7 +91,7 @@ PUBLIC = 273949  # PrincipalId of public "user"
 AUTHENTICATED_USERS = 273948
 DEBUG_DEFAULT = False
 REDIRECT_LIMIT = 5
-
+NUM_THREADS = 14
 
 # Defines the standard retry policy applied to the rest methods
 # The retry period needs to span a minute because sending messages is limited to 10 per 60 seconds.
@@ -199,6 +197,7 @@ class Synapse(object):
         self.table_query_backoff = 1.1
         self.table_query_max_sleep = 20
         self.table_query_timeout = 600  # in seconds
+        self.multi_threaded = False
 
         # TODO: remove once most clients are no longer on versions <= 1.7.5
         cached_sessions.migrate_old_session_file_credentials_if_necessary(self)
@@ -540,12 +539,14 @@ class Synapse(object):
     #                   Get / Store methods                    #
     ############################################################
 
-    def get(self, entity, **kwargs):
+    def get(self, entity, multi_threaded=False, **kwargs):
         """
         Gets a Synapse entity from the repository service.
 
         :param entity:           A Synapse ID, a Synapse Entity object, a plain dictionary in which 'id' maps to a
                                  Synapse ID or a local file that is stored in Synapse (found by the file MD5)
+        :param multi_threaded:   Whether to use multiple threads for downloads.
+                                 Defaults to False.
         :param version:          The specific version to get.
                                  Defaults to the most recent version.
         :param downloadFile:     Whether associated files(s) should be downloaded.
@@ -580,6 +581,8 @@ class Synapse(object):
            print(syn.getProvenance(entity))
 
         """
+        # Sets multi_threaded to True if passed (optional parameter), else False (by default)
+        self.multi_threaded = multi_threaded
 
         # If entity is a local file determine the corresponding synapse entity
         if isinstance(entity, str) and os.path.isfile(entity):
@@ -1605,8 +1608,20 @@ class Synapse(object):
                                                                     fileHandle['fileKey'], destination,
                                                                     profile_name=profile)
                 else:
-                    downloaded_path = self._download_from_URL(fileResult['preSignedURL'], destination, fileHandle['id'],
-                                                              expected_md5=fileHandle.get('contentMd5'))
+                    if self.multi_threaded:
+                        expected_md5 = fileHandle.get('contentMd5')
+                        url = fileResult['preSignedURL']
+                        downloaded_path = self._download_from_url_multi_threaded(fileHandleId,
+                                                                                 objectId,
+                                                                                 objectType,
+                                                                                 destination,
+                                                                                 url,
+                                                                                 expected_md5=expected_md5)
+                    else:
+                        downloaded_path = self._download_from_URL(fileResult['preSignedURL'],
+                                                                  destination,
+                                                                  fileHandle['id'],
+                                                                  expected_md5=fileHandle.get('contentMd5'))
                 self.cache.add(fileHandle['id'], downloaded_path)
                 return downloaded_path
             except Exception as ex:
@@ -1622,6 +1637,59 @@ class Synapse(object):
 
         raise Exception("should not reach this line")
 
+    def _download_from_url_multi_threaded(self,
+                                          file_handle_id,
+                                          object_id,
+                                          object_type,
+                                          destination,
+                                          url,
+                                          expected_md5=None):
+        destination = os.path.abspath(destination)
+        actual_md5 = None
+        redirect_count = 0
+        delete_on_md5_mismatch = True
+        scheme = urllib_urlparse.urlparse(url).scheme
+        if scheme == 'http' or scheme == 'https':
+            request = DownloadRequest(file_handle_id=int(file_handle_id),
+                                      object_id=object_id,
+                                      object_type=object_type,
+                                      path=destination)
+            download_files(self, [request], NUM_THREADS)
+        else:
+            while redirect_count < REDIRECT_LIMIT:
+                redirect_count += 1
+                if scheme == 'file':
+                    delete_on_md5_mismatch = False
+                    destination = utils.file_url_to_path(url, verify_exists=True)
+                    if destination is None:
+                        raise IOError("Local file (%s) does not exist." % url)
+                    break
+                elif scheme == 'sftp':
+                    username, password = self._getUserCredentials(url)
+                    destination = SFTPWrapper.download_file(url, destination, username, password)
+                    break
+                elif scheme == 'ftp':
+                    urllib_urlparse.urlretrieve(url, destination)
+                    break
+                else:
+                    self.logger.error('Unable to download URLs of type %s' % scheme)
+                    return None
+
+            else:  # didn't break out of loop
+                raise SynapseHTTPError('Too many redirects')
+
+        if actual_md5 is None:  # if md5 not set (should be the case for all except http download)
+            actual_md5 = utils.md5_for_file(destination).hexdigest()
+
+        # check md5 if given
+        if expected_md5 and actual_md5 != expected_md5:
+            if delete_on_md5_mismatch and os.path.exists(destination):
+                os.remove(destination)
+            raise SynapseMd5MismatchError("Downloaded file {filename}'s md5 {md5} does not match expected MD5 of"
+                                          " {expected_md5}".format(filename=destination, md5=actual_md5,
+                                                                   expected_md5=expected_md5))
+        return destination
+
     def _download_from_URL(self, url, destination, fileHandleId=None, expected_md5=None):
         """
         Download a file from the given URL to the local file system.
@@ -1635,7 +1703,6 @@ class Synapse(object):
 
         :returns: path to downloaded file
         """
-
         destination = os.path.abspath(destination)
         actual_md5 = None
         redirect_count = 0
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -41,7 +41,8 @@ import warnings
 import getpass
 import logging
 import urllib.parse as urllib_urlparse
-
+import json
+import os
 
 import synapseclient
 from .annotations import from_synapse_annotations, to_synapse_annotations
@@ -197,7 +198,7 @@ class Synapse(object):
         self.table_query_backoff = 1.1
         self.table_query_max_sleep = 20
         self.table_query_timeout = 600  # in seconds
-        self.multi_threaded = False
+        self.multi_threaded = False  # if set to True, multi threaded download will be used for http and https URLs
 
         # TODO: remove once most clients are no longer on versions <= 1.7.5
         cached_sessions.migrate_old_session_file_credentials_if_necessary(self)
@@ -539,14 +540,12 @@ class Synapse(object):
     #                   Get / Store methods                    #
     ############################################################
 
-    def get(self, entity, multi_threaded=False, **kwargs):
+    def get(self, entity, **kwargs):
         """
         Gets a Synapse entity from the repository service.
 
         :param entity:           A Synapse ID, a Synapse Entity object, a plain dictionary in which 'id' maps to a
                                  Synapse ID or a local file that is stored in Synapse (found by the file MD5)
-        :param multi_threaded:   Whether to use multiple threads for downloads.
-                                 Defaults to False.
         :param version:          The specific version to get.
                                  Defaults to the most recent version.
         :param downloadFile:     Whether associated files(s) should be downloaded.
@@ -581,9 +580,6 @@ class Synapse(object):
            print(syn.getProvenance(entity))
 
         """
-        # Sets multi_threaded to True if passed (optional parameter), else False (by default)
-        self.multi_threaded = multi_threaded
-
         # If entity is a local file determine the corresponding synapse entity
         if isinstance(entity, str) and os.path.isfile(entity):
             bundle = self._getFromFile(entity, kwargs.pop('limitSearch', None))
@@ -1646,7 +1642,6 @@ class Synapse(object):
                                           expected_md5=None):
         destination = os.path.abspath(destination)
         actual_md5 = None
-        redirect_count = 0
         delete_on_md5_mismatch = True
         scheme = urllib_urlparse.urlparse(url).scheme
         if scheme == 'http' or scheme == 'https':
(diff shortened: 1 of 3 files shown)


synapseclient/client.py
```diff
@@ -540,12 +539,14 @@ def printEntity(self, entity, ensure_ascii=True):
     #                   Get / Store methods                    #
     ############################################################
 
-    def get(self, entity, **kwargs):
+    def get(self, entity, multi_threaded=False, **kwargs):
```
minneker: Do not let get() change the state of the client.

synapseclient/client.py
```diff
@@ -540,12 +539,14 @@ def printEntity(self, entity, ensure_ascii=True):
...
         """
         Gets a Synapse entity from the repository service.
 
         :param entity:           A Synapse ID, a Synapse Entity object, a plain dictionary in which 'id' maps to a
                                  Synapse ID or a local file that is stored in Synapse (found by the file MD5)
+        :param multi_threaded:   Whether to use multiple threads for downloads.
+                                 Defaults to False.
```
minneker: Get rid of this when taking out.

synapseclient/client.py
```diff
@@ -1622,6 +1637,59 @@ def _downloadFileHandle(self, fileHandleId, objectId, objectType, destination, r
...
+        redirect_count = 0
+        delete_on_md5_mismatch = True
+        scheme = urllib_urlparse.urlparse(url).scheme
+        if scheme == 'http' or scheme == 'https':
+            request = DownloadRequest(file_handle_id=int(file_handle_id),
+                                      object_id=object_id,
+                                      object_type=object_type,
+                                      path=destination)
+            download_files(self, [request], NUM_THREADS)
+        else:
```
minneker: Call original _download_from_URL

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+import json
+import os
+import requests
+
+# constants
+BUF_SIZE = 20
+MAX_RETRIES = 5
+MB = 2**20
+SYNAPSE_DEFAULT_DOWNLOAD_PART_SIZE = 8 * MB
+SYNAPSE_DEFAULT_FILE_ENDPOINT = "https://repo-prod.prod.sagebase.org/file/v1"
```
minneker: Allow this to be configurable, base this off of client's current endpoint is.

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+SYNAPSE_URL_PATH_DOWNLOAD_GET_BATCH_PRE_SIGNED_URL = '/fileHandle/batch'
+ISO_AWS_STR_FORMAT = '%Y%m%dT%H%M%SZ'
+PARTIAL_CONTENT_CODE = 206
+CONNECT_FACTOR = 3
+BACK_OFF_FACTOR = 0.5
+
+# transfer progress parameters
+transferred = 0
```
minneker: Explore using "multiprocessing.value" to pass these values to threads and allow for safe mutation. Check if analogous resource for threads only with less overhead.

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+                if not _url_is_valid(url):
+                    _, pre_signed_url_new = _get_pre_signed_batch_request_json(self.client, self.request)
+                    url = pre_signed_url_new
+                # request the chunk and capture the response
+                response = self.session.get(url, headers=headers, stream=True)
+                # try request until successful or out of retries
+                try_counter = 0
+                while response.status_code != PARTIAL_CONTENT_CODE and try_counter < MAX_RETRIES:
+                    try_counter += 1
+                    if not _url_is_valid(url):
```
minneker: Package making of request in method (with refresh)

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+                        _, pre_signed_url_new = _get_pre_signed_batch_request_json(self.client, self.request)
+                        url = pre_signed_url_new
+                    response = self.session.get(url, headers=headers, stream=True)
+
+                if response.status_code == PARTIAL_CONTENT_CODE:
```
minneker: Consider simulating do-while loop to condense logic here.

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+                with open(path, "r+b") as file_write:
+                    global transferred
+                    global t_first_write
+                    global previously_transferred
+                    if transferred == 0:
+                        t_first_write = time.time()
+                    file_write.seek(start)
+                    file_write.write(data)
+
+                    transferred = len(data) + previously_transferred
```
minneker: Look into removing one of these, i.e. previously transferred.

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+            # Get the chunk data from the data queue
+            start, file_name, path, data = self.data_queue.get()
+            try:
+                # write data to file
+                with open(path, "r+b") as file_write:
+                    global transferred
+                    global t_first_write
+                    global previously_transferred
+                    if transferred == 0:
+                        t_first_write = time.time()
```
minneker: Initialize before thread creation, pass information to threads to use in progress bar.

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+            finally:
+                self.data_queue.task_done()
+
+
+def download_files(client, download_requests: typing.Sequence[DownloadRequest], num_threads):
+    """
+    Main driver for the multi-threaded download. Uses the producer-consumer with Queue design pattern as described
+    in Effective Python Item 39.
+
+    :param client: A SynapseBaseClient
```
minneker: change to "client"

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+    :param download_requests: A batch of DownloadRequest objects specifying what Synapse files to download
+    :param num_threads: The number of download threads
+    :return: Map between each DownloadRequest in download_requests object and the corresponding DownloadResponse object
+    """
+
+    session = requests.Session()
```
minneker: Package this into a create new session helper

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+            consumer_worker = ConsumerDownloadThread(data_queue)
+            producer_threads_local.append(producer_worker)
+            consumer_threads_local.append(consumer_worker)
+
+        for producer_thread, consumer_thread in zip(producer_threads_local, consumer_threads_local):
+            producer_thread.start()
+            consumer_thread.start()
+
+        for producer_thread, consumer_thread in zip(producer_threads_local, consumer_threads_local):
+            producer_threads.append(producer_thread)
```
minneker: use extend here

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+            consumer_threads.append(consumer_thread)
+
+        global transferred
+        global t_first_write
+        global previously_transferred
+        transferred = t_first_write = previously_transferred = 0
+
+        file_name, pre_signed_url = _get_pre_signed_batch_request_json(client, request)
```
minneker: Bundle up into get file size helper

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+        pre_signed_url_chunk_generator = _get_chunk_pre_signed_url(file_size,
+                                                                   file_name,
+                                                                   pre_signed_url,
+                                                                   request.path)
+        global to_be_transferred
+        to_be_transferred = file_size
+        _create_empty_file(file_size, request.path)
+        for chunk in pre_signed_url_chunk_generator:
+            range_queue.put(chunk)
+    range_queue.join()
```
minneker: Could this allow consumers to terminate if producer is too slow to populate consumer queue, i.e. what is termination condition for consumer queue?

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+                                        "fileHandleId": request.file_handle_id,
+                                        "associateObjectId": request.object_id,
+                                        "associateObjectType": request.object_type
+                                    }
+                                ],
+                                "includePreSignedURLs": True,
+                                "includeFileHandles": True,
+                                "includePreviewPreSignedURLs": False
+                            }
+    pre_signed_url_batch = client.restPOST(uri=SYNAPSE_URL_PATH_DOWNLOAD_GET_BATCH_PRE_SIGNED_URL,
```
minneker: Try to replace with client method that gets pre signed urls

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+
+    :param url: A pre-signed download url from AWS
+    :return: True if url is not expired (i.e. valid), False otherwise
+    """
+    parsed_url_list = split('[&|=]', url)
+    time_made = parsed_url_list[parsed_url_list.index('X-Amz-Date') + 1]
+    time_made_datetime = datetime.strptime(time_made, ISO_AWS_STR_FORMAT)
+    expires = parsed_url_list[parsed_url_list.index('X-Amz-Expires') + 1]
+    time_delta_seconds = (datetime.utcnow() - time_made_datetime).total_seconds()
+    return time_delta_seconds < int(expires)
```
minneker: Add 2 second buffer here

synapseclient/producer_consumer_download_threads.py
```diff
@@ -0,0 +1,282 @@
...
+def _url_is_valid(url: str) -> bool:
+    """
+    Checks if url is expired
+
+    :param url: A pre-signed download url from AWS
+    :return: True if url is not expired (i.e. valid), False otherwise
+    """
+    parsed_url_list = split('[&|=]', url)
+    time_made = parsed_url_list[parsed_url_list.index('X-Amz-Date') + 1]
+    time_made_datetime = datetime.strptime(time_made, ISO_AWS_STR_FORMAT)
```
minneker: Try to use url parser to get url parameters instead of splitting and parsing.

synapseclient/producer_consumer_download_threads.py
minneker: Can do this (will do for now), however it is accessing a protected member of the client class which is not good style/discouraged. Since we know that accessing this protected member won't mutate the state of the client we will use it in this patch, in v2 no protected members should be accessed outside of the class they are declared.

synapseclient/producer_consumer_download_threads.py
minneker: Got rid of previously_transferred, simply incremented transferred: transferred += len(data)

synapseclient/producer_consumer_download_threads.py
minneker: Refactored, still using global variable, incorporated a lock.

synapseclient/producer_consumer_download_threads.py
thomasyu888: Is this different from `https://file-prod.prod.sagebase.org/file/v1`?  Because this is in the client:
```
synapseclient.client.PRODUCTION_ENDPOINTS
{'repoEndpoint': 'https://repo-prod.prod.sagebase.org/repo/v1',
 'authEndpoint': 'https://auth-prod.prod.sagebase.org/auth/v1',
 'fileHandleEndpoint': 'https://file-prod.prod.sagebase.org/file/v1',
 'portalEndpoint': 'https://www.synapse.org/'}
```

When you get a synapse connection you get 
```syn.fileHandleEndpoint
'https://file-prod.prod.sagebase.org/file/v1'```

Comment (minneker) [reply synapseclient/producer_consumer_download_threads.py]: It can be different. Notice that the client also has staging endpoints. I hardcoded this endpoint for testing, but I have since changed it to use whatever file handle endpoint is being used by the client performing the download. It will be one of the file handle endpoints, not a repo one. Sorry for the confusion.

Comment (thomasyu888) [reply synapseclient/producer_consumer_download_threads.py]: No problem, I was just curious.

File: synapseclient/producer_consumer_download_threads.py (hunk 17 above)
Comment (minneker) [reply synapseclient/producer_consumer_download_threads.py]: Changed to use urlparse and parse_qs from urllib.parse

**Comment by zimingd** [comment]
@minneker I see the code review comments have been marked as resolved, but there are no additional commits on this pull request. Did you forget to push or are they on a separate branch?

File: synapseclient/producer_consumer_download_threads.py (hunk 14 above)
Comment (minneker) [reply synapseclient/producer_consumer_download_threads.py]: Added a CloseableQueue class which allows producers to signal when they are done producing, this way consumers will terminate only when there is no more data yet to be produced.

**Comment by minneker** [comment]
@zimingd Sorry about that, just committed my most recent changes. Unit tests are not finished yet, will update PR with unit tests soon.

**Comment by thomasyu888** [comment]
Looks pretty cool!.  So the usage would be:

```python
import synapseclient
syn = synapseclient.Synapse(multi_threaded=True)
syn.login()
syn.get(....)
```

minneker:
Thanks, @thomasyu888 ! One slight modification:

```
import synapseclient
syn = synapseclient.Synapse()
syn.multi_threaded = True
syn.login()
syn.get(....)
```

I.e. the Synapse class is initialized with `multi_threaded = False`, and it must be changed after instantiation for this patch.

I wrote the patch such anything that calls `_downloadFileHandle` will experience the speed-up. Note this speed-up applies only to fileHandles that are not external. 

So, the short answer is yes it should work seamlessly with `tableQuery`. I have yet to try it, but I definitely will...

---Update---

I decided to try it and here's what I found! I created a dummy csv table in synapse with about 2 million lines (~72 MB). I downloaded it 3 times using single threaded (default) and 3 times using multithreaded (after the csv was built). I'm on campus WiFi which `fast.com` reports is `87 Mbps`. I experienced `2-3x speed-up` using `tableQuery` with `multi_threaded = True`.  

Please try it on some tables of your own and let me know what you think! Being hard-wired to the Sage internet via an ethernet cable should improve performance as well.

zimingd:
@minneker, are there any unfinished/unaddressed changes that have not yet been committed? If not, I will fork from your branch to add tests and get it merged into the develop branch.

minneker:
@zimingd, all local changes are committed. Please let me know if you have any questions or concerns!

thomasyu888:
Using this:

```python
import synapseclient
import time
import os


def download(syn):
    timestart = time.time()
    temp = syn.get("syn7851250")
    timeend = time.time()
    duration = timeend - timestart
    os.unlink(temp.path)
    return duration


if __name__ == "__main__":
    syn = synapseclient.Synapse()
    syn.multi_threaded = True
    syn.login()
    duration = download(syn)
    print(f"multi-threaded: {duration}")

    syn = synapseclient.Synapse()
    syn.multi_threaded = False
    syn.login()
    duration = download(syn)
    print(f"Not multi-threaded: {duration}")
```

Here are my results when I downloaded a 125MB file 6 times.  3 times with `multi_threaded=True` and 3 times with `multi_threaded=False`.

```
multi-threaded: 16.338900804519653
Not multi-threaded: 27.90092897415161

multi-threaded: 16.75437879562378
Not multi-threaded: 32.26206588745117

multi-threaded: 17.848705768585205
Not multi-threaded: 16.326812982559204
```

How much does this feature depend on internet connection?  I found that using this feature on spotty internet actually causes the `multi_threaded=True` become slower.   I have also experienced the download itself just stalling.

minneker:
@thomasyu888, this feature is highly dependent on the strength of internet connection. Downloading files based on pre-signed urls is an I/O bound processes (as opposed to CPU bound, for example). The infrastructure required to perform multithreaded download on a slow network will actually be the same/slightly slower than the single-threaded download because of this extra overhead. 

I would recommend using multithreaded download for large files (or large batches of small files) on a strong (preferably wired) internet connection. This is where the feature provides the greatest speed-ups and is most useful. If you happen to see the same behavior under these conditions please let me know ASAP and I would be happy to look into it!

thomasyu888:
Thanks @minneker for the explanation.  I do actually see what you're talking about when I connect to a weaker connection.

So on a solid internet connection:

* How large of a file do you recommend me to test?
* Do you recommend me using `synapseutils.syncFromSynapse()` to test the download of large batches of small files.   Or, can i just loop through, lets say 100 files, and do `syn.get` operation?

minneker:
No problem, @thomasyu888! Great to hear the behavior is the same on your end.

Regarding your questions while on a solid internet connection:

- You should be able to see considerable speed-up using a ~300 MB file
- You can use `synapseutils.syncFromSynapse()` or loop through the files using `syn.get()` because `synapseutils.syncFromSynapse()` uses `syn.get()` to download the files.

If you're interested, the speed tests I performed are located in this Jira: https://sagebionetworks.jira.com/browse/SYNPY-682

For reference, 'PCQ' is the multithreaded approach implemented in this PR.

thomasyu888:
Thanks again @minneker.    I am now on an ec2 instance.  I did the below experiment and still found that using the multi_threading feature is slower.  Here is what i did

1.  Using a file that is 2.9GB, i ran my code above 3 times and here are my results:
```
# first run
Downloading  [####################]100.00%   2.9GB/2.9GB (103.4MB/s) Homo_sapiens.GRCh37.75.dna.primary_assembly.fa Done...
    multi-threaded: 52.61539626121521

Downloading  [####################]100.00%   2.9GB/2.9GB (78.8MB/s) Homo_sapiens.GRCh37.75.dna.primary_assembly.fa Done...
    Not multi-threaded: 38.44852542877197

# second run
Downloading  [####################]100.00%   2.9GB/2.9GB (105.0MB/s) Homo_sapiens.GRCh37.75.dna.primary_assembly.fa Done...
    multi-threaded: 52.17721891403198

Downloading  [####################]100.00%   2.9GB/2.9GB (95.5MB/s) Homo_sapiens.GRCh37.75.dna.primary_assembly.fa Done...
    Not multi-threaded: 31.734225511550903

# third run
Downloading  [####################]100.00%   2.9GB/2.9GB (104.5MB/s) Homo_sapiens.GRCh37.75.dna.primary_assembly.fa Done...
    multi-threaded: 52.27965712547302

Downloading  [####################]100.00%   2.9GB/2.9GB (95.8MB/s) Homo_sapiens.GRCh37.75.dna.primary_assembly.fa Done...
    Not multi-threaded: 31.99653124809265
```

My observation is that there seems to be a long period (10 - 20 seconds) after the file is downloaded using the multi-threading.  Is this because the filechunks are being collected?

2. Using `syncFromSynapse` I download download files from a directory that has 108 small files (around 3kb to 2MB each). Running the same experiment above.  I deleted all the files each time.

```
# first run
multi-threaded: 51.000094413757324
Not multi-threaded: 26.47758913040161

# second run
multi-threaded: 47.510202169418335
Not multi-threaded: 27.255042791366577

# third run
multi-threaded: 45.50248980522156
Not multi-threaded: 26.645621299743652
```

minneker:
Thanks for your update, @thomasyu888! There are a couple of things that could be happening, it will help me if you could answer these questions:

- What is your internet speed according to fast.com?
- Where are the files you're downloading located? (i.e. internal or external file handles?)
- Are you still using the same script you commented above, or something else? Could you provide it here?
- Are the files being downloaded and written to the ec2 local file system? 
- What are the specs for the ec2 you're using?

According to the download progress bars, the multithreaded download _is_ downloading the file at a higher rate (~10 MB/s faster) than the single threaded download. I wrote the download in such a way that it writes to file while receiving data (producer threads fetch the data chunks and consumer threads write those to file). So I am unsure why there is this hang-up after the download is performed. Hopefully with your answers I can figure out what is going on.

thomasyu888:
Thanks @minneker. 

*  I am running the script on the ec2, so i dont think my internet speed is relevant. But just for reference my internet speed is 200 Mbps (perhaps this would be helpful for my previous experiment)
* The files are internal file handles.
* I updated the script from above.  I gave you download access to the synapse id to the 2.9GB file but I can't give you access to the folder containing 108 files.
* The files are all being downloaded and written to the synapse cache which is on the ec2 local file system
* The ec2 is a `t3.mirco` instance that i am using.  More information: https://aws.amazon.com/ec2/instance-types/t3/

PR #720, opened 2019-12-24
SYNPY-924: Raise error when a list is passed into syn.delete

thomasyu888:
This function was throwing a silent error, because the error was not raised after catching the `AttributeError`


thomasyu888:
@zimingd , please take a look.

synapseclient/client.py
```diff
@@ -1104,16 +1104,19 @@ def delete(self, obj, version=None):
...
         """
         # Handle all strings as the Entity ID for backward compatibility
         if isinstance(obj, str):
             if version:
-                self.restDELETE(uri='/entity/%s/version/%s' % (id_of(obj), version))
+                self.restDELETE(uri='/entity/{}/version/{}'.format(id_of(obj),
+                                                                   version))
             else:
-                self.restDELETE(uri='/entity/%s' % id_of(obj))
+                self.restDELETE(uri='/entity/{}'.format(id_of(obj)))
```
zimingd: we can use f-strings now that we are requiring python 3.6

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -767,6 +767,46 @@ def test_move():
...
+        patch_rest_delete.assert_called_once_with(uri="/entity/syn1235")
+
+
+def test_delete__string_version():
+    with patch.object(syn, "restDELETE") as patch_rest_delete:
+        syn.delete("syn1235", version=1)
+        patch_rest_delete.assert_called_once_with(uri="/entity/syn1235/version/1")
+
+
+class mock_synapse_delete:
```
zimingd: Remove this class and just pass in a new `mock_obj=MagicMock()` object in the test.

has_attr() will always return true for MagicMock

you can then assert that was called `mock_obj._synapse_delete.assert_called_once()`

PR #721, opened 2019-12-24
Synpy 791: Specify explicit dependency versions

thomasyu888:
Add in `boto3` dependency


setup.py
```diff
@@ -41,7 +41,7 @@
     extras_require={
         'pandas': ["pandas==0.25.0"],
         'pysftp': ["pysftp>=0.2.8"],
-        'boto3' : ["boto3"],
+        'boto3' : ["boto3==2.49.0"],
```
thomasyu888, line 44:
```suggestion
        'boto3': ["boto3==2.49.0"],
```

thomasyu888:
Unsure if its necessary to lock down the version here?

thomasyu888:
Tackled by #762

PR #723, opened 2020-01-29
call different API to retrieve fileHandle information (SYNPY-1036)

zimingd:


thomasyu888:
Looks good, but is an integration test needed to upload file as user and then upload the same file as another user?

synapseclient/client.py
```diff
@@ -980,7 +980,10 @@ def store(self, obj, **kwargs):
             # update the file_handle metadata if the FileEntity's FileHandle id has changed
             if '_file_handle' in local_state \
                     and properties['dataFileHandleId'] != local_state['_file_handle'].get('id', None):
-                local_state['_file_handle'] = self._getFileHandle(properties['dataFileHandleId'])
+                local_state['_file_handle'] = find_data_file_handle(
+                    self._getEntityBundle(properties['id'], bitFlags=0x800 | 0x1)
```
thomasyu888: Nice, the reason both the filehandle and entity bigflag are specified is because `find_data_file_handle` requires both.

zimingd: I think a better fix would to be also to switch over to https://rest-docs.synapse.org/rest/POST/entity/id/bundle2.html for entity bundles as well so that we can avoid using bitflags. Now that i have more time to work on the client I'll do that before checking this in

zimingd:
We'll definitely need to add a second user for a full integration test. I'll continue this work after the multi-thread download is done

thomasyu888:
Are you going to switch this into `develop` as well?

thomasyu888:
Why is a annotations refactor part of this PR?

zimingd:
EntityBundle v2 contains Annotations v2 so to use EntityBundleV2 we have to also change the Annotations translation code

thomasyu888:
Just out of curiosity, is there documentation on the differences between annotation v1 and v2?

zimingd:
Well annotation v1 had zero documentation because its schema wasn't auto-generatable.

synapseclient/client.py
```diff
@@ -1046,34 +1044,44 @@ def _createAccessRequirementIfNone(self, entity):
...
-            DOI                        = 0x8000
-            FILE_NAME                  = 0x10000
-            THREAD_COUNT               = 0x20000
-            RESTRICTION_INFORMATION    = 0x40000
+        :parameter requestedObjects:    A dict indicating settings for what to include
 
-        For example, we might ask for an entity bundle containing file handles, annotations, and properties::
+        default value for requestedObjects is::
+
+            requestedObjects = {'includeEntity': True,
```
jkiang13: Are we likely to pass False for any of these values (i.e. does the backend API itself have defaults for some of these values)? If not then a set of keys may be simpler than a dictionary of True values.

synapseclient/client.py
```diff
@@ -811,7 +813,8 @@ def _resolve_download_path_collisions(self, downloadLocation, file_name, ifcolli
                                  'for "ifcollision"' % ifcollision)
         return downloadPath
 
-    def store(self, obj, **kwargs):
+    def store(self, obj, *, createOrUpdate=True, forceVersion=True, versionLabel=None, isRestricted=False,
```
jkiang13: Not an issue w/ this PR, but the casing inconsistency in some of the public functions especially is not ideal. This just unrolls the current kwargs so no change but if we did want to clean this up at some point without immediate breaking changes some kwargs to catch old style params and convert them could be usefu.

synapseclient/client.py
zimingd: document that unneeded keys (set to False) can be omitted.

synapseclient/annotations.py
```diff
@@ -248,37 +221,122 @@ def set_privacy(annotations, key, is_private=True, value_types=['longAnnos', 'do
...
+def to_synapse_annotations(annotations: Annotations) -> typing.Dict[str, typing.Any]:
+    """Transforms a simple flat dictionary to a Synapse-style Annotation object.
+    https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/annotation/v2/Annotations.html
+    """
+
+    if is_synapse_annotations(annotations):
+        return annotations
+    synapse_annos = {}
+
+    if not hasattr(annotations, 'id') or not hasattr(annotations, 'etag'):
```
zimingd: isinstance

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -1132,3 +1140,23 @@ def test_invite_to_team__force_invite(self):
...
+        def test_not_annotation(self):
+            with patch.object(syn, "restPUT") as mock_rest_put:
+                # pass in non-annotation object
+                assert_raises(ValueError, syn.set_annotations, {})
+                mock_rest_put.assert_not_called()
+
+        def test_with_annotations(self):
+            with patch.object(syn, "restPUT") as mock_rest_put:
+                # pass in non-annotation object
+                assert_raises(ValueError, syn.set_annotations,
```
zimingd: remove assert_raises

PR #724, opened 2020-01-31
SYNPY-1039 - Convert TYPE_LIST table columns into dataframe columns with lists

thomasyu888:
Currently, when you have list type columns, when you do a query you get:

```
syn = synapseclient.login()
temp = syn.tableQuery('SELECT * FROM syn21093721')
temp.asDataFrame()['example'][0]
'["store", "me", "now"]'
```

It would be nice if these were returned as actual lists


thomasyu888:
Ready for review @jkiang13

PR #725, opened 2020-02-19
SYNPY-1041: Create snapshot convenience function

thomasyu888:
Creating snapshots


tests/unit/synapseclient/unit_test_client.py
```diff
@@ -1687,3 +1688,70 @@ def test_get_submission_with_annotations():
...
+        )
+
+
+def test__async_table_update():
+    """Async table update"""
+    snapshot = {'snapshotVersionNumber': 2}
+    with patch.object(syn, 'restPOST', return_value=snapshot) as restpost:
+        syn._async_table_update("syn1234", create_snapshot=True,
+                                   comment="foo", label="new_label",
+                                   activity=2)
```
thomasyu888, line 1712:
```suggestion
        syn._async_table_update("syn1234", create_snapshot=True,
                                comment="foo", label="new_label",
                                activity=2)
```

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -1687,3 +1688,70 @@ def test_get_submission_with_annotations():
...
+
+def test__create_table_snapshot():
+    """Testing creating table snapshots"""
+    snapshot = {'snapshotVersionNumber': 2}
+    with patch.object(syn, 'restPOST', return_value=snapshot) as restpost:
+        syn._create_table_snapshot("syn1234", comment="foo", label="new_label",
+                                   activity=2)
+        restpost.assert_called_once_with(
+            "/entity/syn1234/table/snapshot",
+            body = '{"snapshotComment": "foo", "snapshotLabel": "new_label", '
```
thomasyu888, line 1701:
```suggestion
            body='{"snapshotComment": "foo", "snapshotLabel": "new_label", '
```

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -1687,3 +1688,70 @@ def test_get_submission_with_annotations():
...
+def test__async_table_update():
+    """Async table update"""
+    snapshot = {'snapshotVersionNumber': 2}
+    with patch.object(syn, 'restPOST', return_value=snapshot) as restpost:
+        syn._async_table_update("syn1234", create_snapshot=True,
+                                comment="foo", label="new_label",
+                                activity=2)
+        restpost.assert_called_once_with(
+            "/entity/syn1234/table/transaction/async/start",
+            body = '{"changes": [], "createSnapshot": true, '
```
thomasyu888, line 1715:
```suggestion
            body='{"changes": [], "createSnapshot": true, '
```

synapseclient/client.py
```diff
@@ -3075,6 +3075,65 @@ def getColumns(self, x, limit=100, offset=0):
...
+        """Creates Table snapshot
+
+        :param table:  The schema of the Table
+        :param comment:  Optional snapshot comment.
+        :param label:  Optional snapshot label.
+        :param activity:  Optional activity ID applied to snapshot version.
+
+        :return:  Snapshot Response
+        """
+        snapshot_body = {"snapshotComment": comment,
```
jkiang13, line 3107: Could we only include the dictionary keys if the value is not default/None. We aren't consistent about that in the client now, but I'm trying to move towards absence of a key for a default/null value since I think there is less ambiguity about what that means to the backend.

synapseclient/client.py
```diff
@@ -3075,6 +3075,65 @@ def getColumns(self, x, limit=100, offset=0):
...
+        :param comment:  Optional snapshot comment.
+        :param label:  Optional snapshot label.
+        :param activity:  Optional activity ID applied to snapshot version.
+
+        :return:  Snapshot Response
+        """
+
+        table_update_body = {'changes': changes,
+                             'createSnapshot': create_snapshot,
+                             'snapshotOptions': {'snapshotComment': comment,
```
jkiang13, line 3130: Same here for default/None.

jkiang13 (commented):
Looks good.

synapseclient/client.py
thomasyu888: Thanks @jkiang13 .  Do you propose doing something like this:

```
{key:value for key, value in snapshot_body.items() if value}
```

Also should i create a convenience function?

```
def _create_snapshot_body(comment=None, label=None, activity=None)
        snapshot_body = {"snapshotComment": comment,
                         "snapshotLabel": label,
                         "snapshotActivityId": activity}
        new_body = {key:value for key, value in snapshot_body.items() if value}
        return new_body)
```

synapseclient/client.py
```diff
@@ -3124,14 +3125,17 @@ def _async_table_update(self, table, changes=[], create_snapshot=False,
...
+                            'snapshotLabel': label,
+                            'snapshotActivityId': activity}
+        new_snapshot = {key:value for key, value in snapshot_options.items() if value}
         table_update_body = {'changes': changes,
                              'createSnapshot': create_snapshot,
-                             'snapshotOptions': {'snapshotComment': comment,
-                                                 'snapshotLabel': label,
-                                                 'snapshotActivityId': activity}}
+                             'snapshotOptions': new_snapshot}
+        new_table_update_body = {key:value for key, value in table_update_body.items() if value}
```
jkiang13, line 3135: I think none of these are optional so we don't need to prune for this https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/TableUpdateTransactionRequest.html

synapseclient/client.py
```diff
@@ -3107,15 +3107,16 @@ def _create_table_snapshot(self, table, comment=None, label=None, activity=None)
         snapshot_body = {"snapshotComment": comment,
                          "snapshotLabel": label,
                          "snapshotActivityId": activity}
+        new_body = {key:value for key, value in snapshot_body.items() if value}
```
jkiang13, line 3110: `if value is not None` here since empty strings should probably be passed for these.

synapseclient/client.py
```diff
@@ -3124,14 +3125,17 @@ def _async_table_update(self, table, changes=[], create_snapshot=False,
 
         :return:  Snapshot Response
         """
-
+        snapshot_options = {'snapshotComment': comment,
+                            'snapshotLabel': label,
+                            'snapshotActivityId': activity}
+        new_snapshot = {key:value for key, value in snapshot_options.items() if value}
```
jkiang13, line 3131: `if value is not None` here since empty strings should probably be passed for these.

thomasyu888:
One thing I ran into is that the snapshot version isn't actually returned.  For Table snapshots, its easier because the snapshot version is returned: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/SnapshotResponse.html.  However, with View snapshots, a token is returned, then you call this to return a snapshot version: https://rest-docs.synapse.org/rest/GET/entity/id/table/transaction/async/get/asyncToken.html.  

Thoughts?

synapseclient/client.py
```diff
@@ -3075,6 +3075,67 @@ def getColumns(self, x, limit=100, offset=0):
...
+        :param comment:  Optional snapshot comment.
+        :param label:  Optional snapshot label.
+        :param activity:  Optional activity ID applied to snapshot version.
+
+        :return:  Snapshot Response
+        """
+        snapshot_body = {"snapshotComment": comment,
+                         "snapshotLabel": label,
+                         "snapshotActivityId": activity}
+        new_body = {key:value for key, value in snapshot_body.items() if value is not None}
```
thomasyu888, line 3110:
```suggestion
        new_body = {key: value for key, value in snapshot_body.items() if value is not None}
```

synapseclient/client.py
```diff
@@ -3075,6 +3075,67 @@ def getColumns(self, x, limit=100, offset=0):
...
+        :param comment:  Optional snapshot comment.
+        :param label:  Optional snapshot label.
+        :param activity:  Optional activity ID applied to snapshot version.
+
+        :return:  Snapshot Response
+        """
+        snapshot_options = {'snapshotComment': comment,
+                            'snapshotLabel': label,
+                            'snapshotActivityId': activity}
+        new_snapshot = {key:value for key, value in snapshot_options.items() if value is not None}
```
thomasyu888, line 3131:
```suggestion
        new_snapshot = {key: value for key, value in snapshot_options.items() if value is not None}
```

jkiang13 (approved):

jkiang13:
> @jkiang13 One thing I ran into is that the snapshot version isn't actually returned when calling `create_snapshot`. For Table snapshots, its easier because the snapshot version is returned: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/SnapshotResponse.html. However, with View snapshots, a token is returned, then you call this to return a snapshot version: https://rest-docs.synapse.org/rest/GET/entity/id/table/transaction/async/get/asyncToken.html.
> 
> Thoughts?

Are you wanting the version for purposes not shown in this PR (which doesn't return either response)? If so we perhaps wrap optionally wrap the async table update call with a _waitForAsync depending on if the invoker needed the synchronous response of the version.

thomasyu888:
@jkiang13 , I was just wondering if we _should_ return the version number.  Currently when we run `create_snapshot`, nothing is returned to the user, so maybe users will wonder if the code worked or not?

jkiang13:
> @jkiang13 , I was just wondering if we _should_ return the version number. Currently when we run `create_snapshot`, nothing is returned to the user, so maybe users will wonder if the code worked or not?



> @jkiang13 , I was just wondering if we _should_ return the version number. Currently when we run `create_snapshot`, nothing is returned to the user, so maybe users will wonder if the code worked or not?

I do not think there is anything wrong with a method that triggers a side effect and doesn't return anything, provided it is documented (perhaps we should update the docstring to be clear about that however). The fact is that we can't return the version consistently unless we make this a blocking method pending the completion of the async task (if the version itself is not valuable then I don't think we'd want to impose that wait on the caller).

As the reporter of the original issue, perhaps @kelshmo can weigh in on whether the version of the created snapshot would be valuable to the invoker.

kelshmo:
I think it is valuable to return the version of the snapshot. The returned synID.version could then be referenced in Provenance. Without it, it is not easy to identify the relevant version (could be obscured or lost in the version history if an automatic process is creating snapshots).

thomasyu888:
Thanks @kelshmo .  @jkiang13, I agree with Kelsey about returning the version as it would be helpful for adding provenance.  That being said, table versions won't be as much of an issue because we don't need to wait for the async task. It will only be fileviews that will be more troublesome.  Thoughts?

jkiang13:
What if we were to add an argument to the create_snapshot function you have in this PR, e.g. wait=True. If it's just a table snapshot then this argument will have no effect and the snapshot version would always be returned.  If we would fall into the async table update and wait=True then the implementation of that method would internally use _waitForAsync to block until the version was available to be returned. If wait=False then it would just return the async token instead immediately.

thomasyu888:
Thanks @jkiang13 .  That is certainly interesting.   The function would then return 2 different types depending on this `wait` parameter?  

Would I return a dictionary: `{"version": 8}` or `{"token": "aaaaa"}`?  Or should i just return the value itself and if it isn't an integer, then people would infer that it is the token that is returned?

jkiang13:
I agree that a heterogenous return type (in this case probably a dictionary with different keys) might be confusing (especially since the a schema snapshot would never return an async token even if wait=False), but I think wait=True would be the default and so someone using the function in it's default form would either get the version they expected or a e.g. a timeout exception rather than a different return value, and hopefully would read the docstring before using wait=False.

Other options:
1. Always return None if wait=False (probably even if no async table update involved). The semantics would be simpler, and easier to document, although slightly less flexible to the invoker (no access to the token).
2. Don't have the wait keyword arg and just always wait on the async table update and always then return the version (or raise an Exception). This would require someone invoking the method who didn't want to block on the version to deal with it (or invoke the raw API call themselves).
3. Return a Future. This would involve running a thread within the client to resolve the Future and I think might be confusing to someone not realizing they were invoking a threaded context. I don't think this one is worth the complexity.

What do you think of #1?

thomasyu888:
I like the first option you put there.  I guess another option is just to have different "create_snapshot" functions for the views vs tables.

1. create_table_snapshot
2. create_view_snapshot

And create_view_snapshot could have a `wait` parameter?  I do think its nice that the `create_snapshot` can create either a version of a Table or View, but due to their inherent differences, maybe splitting them up is better? Happy to go with option 1 as well.

jkiang13:
I like having them in one as well, and I think option 1 covers the 99% use case, I doubt there are that any callers who actually might want to make a non blocking call and then also make use of an async token.

thomasyu888:
Thanks.  After investigating some more, I realize some of these functions already exist, but are written a little differently.  Specifically:

* `syn._POST_table_transaction`:  This does the `/entity/{id}/table/transaction/async` but doesnt have the `snapshotVersion` in the request body
* `syn._waitForAsync`: Appends `/start` to the uri above - starts and waits for the async job to be complete.

Should I just refactor `_POST_table_transaction` to include the ability to create snapshots?  That being said, theres currently no option of `wait=False` if I use this function.

jkiang13:
Good point about the existing _POST_table_transaction.

I made a PR to your branch with a possible implementation including wait.

thomasyu888:
Aweomse.  I'm going to add support for submission views as well.  Did you want me to add the `typing` to each parameter?

jkiang13:
I not sure I understand what you mean about the type parameter, can you elaborate?

I did remove _POST_table_transaction in that PR to your fork. I replaced the references to it with your new _async_table_update method, and hopefully as an underscore function it doesn't have any outside callers... 🤞

thomasyu888:
Done. Thanks for the help!

jkiang13:
Looks good.

PR #727, opened 2020-03-11
Multi-threaded download of S3 filehandle presigned-urls (SYNPY-682)

zimingd:


synapseclient/client.py
```diff
@@ -93,7 +93,7 @@
 AUTHENTICATED_USERS = 273948
 DEBUG_DEFAULT = False
 REDIRECT_LIMIT = 5
-
+NUM_THREADS = os.cpu_count() + 4
```
thomasyu888: Why + 4?

synapseclient/core/multithread_download/download_threads.py
```diff
@@ -0,0 +1,393 @@
...
+                for bytes in response.iter_content(MAX_CHUNK_WRITE_SIZE):
+                    self.data_queue.put((start, bytes))
+                    start += len(bytes)
+            except QueueClosedException:
+                # the data_queue was closed so stop retrieving data chunks
+                response.close()
+                break
+
+    def _get_response_with_retry(self, start:int, end:int) -> Response:
+        range_header = {'Range': f'bytes={start}-{end}'}
```
thomasyu888: Just checking to see if we are supporting python 3.6+?  f-interpolation doesn't work in python 3.5.

zimingd: we are supporting 3.6+ since the latest version is 3.8.

synapseclient/client.py
zimingd: additional threads for io bound tasks even if core count is low

synapseclient/core/multithread_download/download_threads.py
thomasyu888: Will need to make sure this is updated: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/setup.py#L35

synapseclient/client.py
```diff
@@ -199,6 +199,7 @@ def __init__(self, repoEndpoint=None, authEndpoint=None, fileHandleEndpoint=None
         self.table_query_backoff = 1.1
         self.table_query_max_sleep = 20
         self.table_query_timeout = 600  # in seconds
+        self.multi_threaded = False  # if set to True, multi threaded download will be used for http and https URLs
```
jkiang13: Do we need to expose this as settable from a command line option? I see how one using synapse as a Python library could set it but if someone was trying to use this feature from the command line is there a way to set it?

synapseclient/core/multithread_download/download_threads.py
```diff
@@ -0,0 +1,396 @@
...
+BACK_OFF_FACTOR: float = 0.5
+
+
+class DownloadRequest(NamedTuple):
+    """
+    A request to download a file from Synapse
+
+    ...
+
+    Attributes
```
thomasyu888: Inconsistent documentation method in the client.  This is a `numpy` style, but we use the `sphinx` style.  Although personally, I like the google/numpy style docs better.

zimingd: The core package code is not user facing and won't have sphinx docs generated for it.

PR #732, opened 2020-03-21
Release-notes 2.0.0

zimingd:


zimingd:
Rendered version in sphinx docs:

image

brucehoff:
Looks great to me.

kdaily:
Why is it not:

```
syn = synapseclient.Synapse(multi_threaded=True)
```

thomasyu888:
> Why is it not:
> 
> ```
> syn = synapseclient.Synapse(multi_threaded=True)
> ```

View https://github.com/Sage-Bionetworks/synapsePythonClient/pull/718#issuecomment-586000993

zimingd:
@kdaily 
We were planning to make it the default implementation once it's been field-tested and deemed stable so we didn't want to introduce new constructor argument and then remove a few versions later.

I'm not quite sure how many people prefer to use `syn = synapseclient.login()` versus `syn = synapseclient.Synapse();syn.login()` so I opted for the less verbose one.

kdaily:
We have many users using `synapseclient.Synapse();syn.login()` because there are other features that have no parameter (e.g., verbosity of logging) or that are using other user/pass management options.

kdaily:
Typo in first bullet point "Syanpse"

kdaily:
Rewrite the sentence starting with "Multi-threaded download currently will work..." to:

"Currently, multi-threaded download works with files stored in AWS S3 where most files on Synapse reside."

Does this mean that any custom storage location also benefits from this speedup? If not, this is a major thing to note.

kdaily:
`synapseutils.copy` - change 'nor' to 'or'  ('nor' should only be used along with 'neither')

kdaily:
This section is more about submissions than it is about Docker object:

"- Synapse Docker Repository(:code:`synapseclient.DockerRepository`) objects can be used as the :code:`entity` argument in :code:`synapseclient.Synapse.submit()`. Optional argument :code:`docker_tag="latest"` has also been added to :code:`synapseclient.Synapse.submit()`"

How about:

"- Synapse Docker Repository(:code:`synapseclient.DockerRepository`) objects can now be submitted to Synapse evaluation queues using the :code:`entity` argument in :code:`synapseclient.Synapse.submit()`. An optional argument :code:`docker_tag="latest"` has also been added to :code:`synapseclient.Synapse.submit()`" to designate which tagged Docker image to submit."

kdaily:
"A full list of issues addressed in this release are below."

kdaily:
At the beginning: "Python 2 is no longer supported in this release." (need full stop at the end of the statement to make it a sentence)

zimingd:
@kdaily 
Is this the link to include for access requirements?

https://docs.synapse.org/articles/access_controls.html

kelshmo:
- I suggest fixing the formatting in the Bug, Task, Improvement section to be consistent with the notes above (e.g how functions are formatted, quotes etc.) All of these could use some edits to provide more context but I understand these populate directly from the Jira issue itself.

- columnTypes is a major change - can you add content to the highlights? 

- I suggest: ":code:synapseutils.copy() now has limitations on what can be copied. A user must have [download permissions](https://docs.synapse.org/articles/accounts_certified_users_and_profile_validation.html) on the entity they want to copy. Users cannot copy any entities that are [controlled use](https://docs.synapse.org/articles/contribute_and_access_controlled_use_data.html#how-do-you-access-data-with-conditions-for-use)." Although I'm wondering if this is even true about the AR/Controlled Use piece, since the JIRA issue for the `copy` function looks like it will copy if the user has public download permissions. @thomasyu888 looks like you wrote it, can you clarify?

- other places you can link - [Synapse Docker Registry](https://docs.synapse.org/articles/docker.html)

thomasyu888:
To clear up some confusion, there are two separate tickets
* One which was to not copy any files with AR
* The next was to fix a bug that wasn't copying files that you had download permissions to because I was just checking the user's ACL and not if the user had download permissions on a file (if the user is on a team)

The AR check will always precede any other check, so even if you have download permissions on a file, you will **not** be able to download it if there is AR on it.

PR #737, opened 2020-03-27
Multipart upload refactor/performance SYNPY-1029

jkiang13:
Refactoring and performance improvements surrounding S3 based multipart uploads.

The main changes are in tweaking the default of how many threads we allocate based on the number of cores on the machine, and improving the use of requests.Sessions which will allow for greater use of persistent http connections.

https://sagebionetworks.jira.com/browse/SYNPY-1029


synapseclient/core/upload/multipart_upload.py
```diff
@@ -3,385 +3,526 @@
...
+                    prefix='Uploading',
+                    postfix=self._dest_file_name,
+                    dt=time.time() - time_upload_started,
+                    previouslyTransferred=previously_transferred,
+                )
+            except Exception as cause:
+                with self._lock:
+                    self._aborted = True
+
+                if isinstance(cause, KeyboardInterrupt):
```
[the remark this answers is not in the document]
zimingd: Not sure if this is the case but if higher-level code is expecting to catch KeyboardInterrupt to do any cleanup, re-raise it.

synapseclient/core/pool_provider.py
```diff
@@ -50,13 +67,29 @@ def __init__(self, type, value):
...
+def get_executor(thread_count=DEFAULT_POOL_SIZE):
+    """
+    Provides an Executor as defined by the client config suitable
+    for running tasks work as defined by the client config.
+
+    :param thread_count:        number of concurrent threads
+
+    :return: an Executor
+    """
+    if config.single_threaded:
```
zimingd: we should get rid of the config and just attempt to detect if threading is possible.
I'm not sure if ThreadPoolExecutor handles dummy_threading or not already, but if it does then we have don't have to do anything. https://docs.python.org/3.6/library/dummy_threading.html

synapseclient/core/upload/multipart_upload.py
```diff
@@ -3,385 +3,526 @@
...
+            'fileName': self._dest_file_name,
+            'generatePreview': self._preview,
+            'contentType': self._content_type,
+            'partSizeBytes': self._part_size,
+            'fileSizeBytes': self._file_size,
+            'storageLocationId': self._storage_location_id,
+        }
+
+        return self._syn.restPOST(
+            "/file/multipart?forceRestart={}".format(
```
zimingd: we can use f-strings now that we require 3.6+

synapseclient/core/upload/multipart_upload.py
```diff
@@ -3,385 +3,526 @@
...
-        raise ValueError('Minimum part size is %d MB.' % (min_part_size/MB))
-    if int(math.ceil(float(fileSize) / partSize)) > max_parts:
-        raise ValueError('A part size of %0.1f MB results in too many parts (%d).'
-                         % (float(partSize)/MB, int(math.ceil(fileSize / partSize))))
-    return partSize
 
+thread_local = threading.local()
+
+
+class UploadAbortedException(Exception):
```
thomasyu888: Should these be in synapseclient.core.exceptions?

synapseclient/core/upload/multipart_upload.py
```diff
@@ -3,385 +3,526 @@
...
+                if ex.response.status_code == 403 and retry < 1:
+                    # we interpret this to mean our pre_signed url expired.
+                    self._syn.logger.debug(
+                        "The pre-signed upload URL for part {} has expired."
+                        "Refreshing urls and retrying.\n".format(part_number)
+                    )
+
+                    # we refresh all the urls and obtain this part's
+                    # specific url for the retry
+                    pre_signed_part_url = self._refresh_pre_signed_part_urls(
```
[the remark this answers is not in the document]
jkiang13: Whenever there is a refresh it fetches new urls for all the outstanding parts, not just the part that encountered the 403, so the first thread to hit an expired url will prevent further calls to refresh these urls (until the next expiration period elapses).

synapseclient/core/upload/multipart_upload.py
```diff
@@ -3,385 +3,526 @@
...
+        self._storage_location_id = storage_location_id
+        self._max_workers = max_workers
+        self._force_restart = force_restart
+
+        self._lock = threading.Lock()
+        self._aborted = False
+
+        # populated later
+        self._upload_id: str = None
+        self._pre_signed_part_urls: OrderedDict[int, str] = None
```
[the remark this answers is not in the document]
jkiang13: You're right. At one point I was using the keys in the dict to order the fetches and it was useful to have an OrderedDict for testing determinism but this is no longer the case.

synapseclient/core/pool_provider.py
jkiang13: I agree the config setting is currently strange. It's documented here thoughhttps://python-docs.synapse.org/build/html/news.html?highlight=single_thread#id18, so I'm thinking at least one release to announce it's being removed and maybe some stderr warning when setting it would be nice rather than removing it.

I think a user exposed option to set the number of threads specifically i.e. aws client's max_concurrent_requests would be good.

synapseclient/core/upload/multipart_upload.py
jkiang13: Good point.

synapseclient/core/upload/multipart_upload.py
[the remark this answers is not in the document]
jkiang13: Good point.

synapseclient/core/upload/multipart_upload.py
jkiang13: In this case the string that's being inserted is an instance variable that's being lowered, inline in the format so it seems a bit awkward to use an f-string here.

PR #738, opened 2020-04-03
maxThreads configurable

jkiang13:
1. Extends the threading configurability to allow you to specify a value for S3 downloads as well as uploads since it seems nicely symmetrical that way.
2. Adds a [transfer] section to the synapseConfig with a maxThreads option for specifying the setting so you don't have to always specify it on each call. 
3. Makes the public facing API param for maxThreads mixed case since that matches the existing convention on those calls. Internal function params may still be snake_cased per PEP8.
4. Adds some doc for this to the release notes.

### DIFF
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -8,12 +8,18 @@ Release Notes
 Highlights:
 ----------------
 
-- The Synapse client object's :code:`store` and :code:`get` methods have been enhanced with an optional
-  :code:`maxThreads` keyword parameter that allows explicit specification of the number of concurrent threads
-  and connections that will be used to perform uploads and downloads. Currently these parameters only apply
-  to files whose underlying storage is AWS S3, and will be ignored if passed for other operations. If no parameter
-  is passed reasonable defaults will be chosen. If local resources allow adjusting these values higher can reduce
-  transfer times.
+- A :code:`max_threads` property of the Synapse object has been added to customize the number of concurrent threads
+  that will be used to transfer files.
+
+  .. code-block:: python
+
+    import synapseclient
+    syn = synapseclient.login()
+    syn.max_threads = 20
+
+  If not customized the default value is (cpu count + 4). Adjusting this value
+  higher may speed up file transfers if the underlying system resources can take advantage of the higher setting.
+
 
   Alternately, a value can be stored in the synapseConfig configuration file that will automatically apply
   as the default.
@@ -21,7 +27,7 @@ Highlights:
   .. code-block::
 
      [transfer]
-     maxThreads=16
+     max_threads=16
 
 A full list of issues addressed in this release are below.
 
--- a/docs/news.rst
+++ b/docs/news.rst
@@ -2,6 +2,36 @@
 Release Notes
 =============
 
+2.1.0 (2020-04-0X)
+==================
+
+Highlights:
+----------------
+
+- The Synapse client object's :code:`store` and :code:`get` methods have been enhanced with an optional
+  :code:`maxThreads` keyword parameter that allows explicit specification of the number of concurrent threads
+  and connections that will be used to perform uploads and downloads. Currently these parameters only apply
+  to files whose underlying storage is AWS S3, and will be ignored if passed for other operations. If no parameter
+  is passed reasonable defaults will be chosen. If local resources allow adjusting these values higher can reduce
+  transfer times.
+
+  Alternately, a value can be stored in the synapseConfig configuration file that will automatically apply
+  as the default.
+
+  .. code-block::
+
+     [transfer]
+     maxThreads=16
+
+A full list of issues addressed in this release are below.
+
+Bug
+---
+
+-  [`SYNPY-1036 <https://sagebionetworks.jira.com/browse/SYNPY-1036>`__] -
+   different users storing same file to same folder results in 403
+
+
 2.0.0 (2020-03-23)
 ==================
 **Python 2 is no longer supported as of this release.** This release requires Python 3.6+.
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -396,7 +396,7 @@ class Synapse(object):
 
     def _get_transfer_config_max_threads(self):
         # note RawConfigParser lowercases so maxThreads -> maxthreads
-        max_threads =  self._get_config_section_dict('transfer').get('maxthreads')
+        max_threads = self._get_config_section_dict('transfer').get('maxthreads')
         if max_threads:
             try:
                 return int(max_threads)
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -565,6 +565,8 @@ class Synapse(object):
         :param limitSearch:      a Synanpse ID used to limit the search in Synapse if entity is specified as a local
                                  file.  That is, if the file is stored in multiple locations in Synapse only the ones
                                  in the specified folder/project will be returned.
+        :param max_threads:      The maximum number of threads to use when downloading the file (currently only
+                                 applies to S3 uploads)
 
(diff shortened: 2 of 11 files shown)


synapseclient/client.py
```diff
@@ -394,6 +394,16 @@ def _get_client_authenticated_s3_profile(self, endpoint, bucket):
...
         return self._get_config_section_dict(config_section).get("profile_name", "default")
 
+    def _get_transfer_config_max_threads(self):
+        # note RawConfigParser lowercases so maxThreads -> maxthreads
+        max_threads = self._get_config_section_dict('transfer').get('maxthreads')
+        if max_threads:
+            try:
+                return int(max_threads)
+            except ValueError:
+                self.logger.warning("Invalid transfer.maxThreads config setting (%s), igoring.", max_threads)
```
zimingd, line 404: It would be better to raise an intelligible error so that users can fix their config instead of it failing silently since, by default, no config entry is set for maxThreads

synapseclient/client.py
```diff
@@ -689,6 +701,7 @@ def _getWithEntityBundle(self, entityBundle, entity=None, **kwargs):
         submission = kwargs.pop('submission', None)
         followLink = kwargs.pop('followLink', False)
         path = kwargs.pop('path', None)
+        max_threads = kwargs.pop('maxThreads', None) or self._get_transfer_config_max_threads()
```
zimingd, line 704: kwargs was originally used to avoid accidentally setting the wrong value for optional parameters in in positional args

Since we moved away from python 2, we can stop using kwargs in favor of Keyword-Only Arguments.

https://www.python.org/dev/peps/pep-3102/

synapseclient/client.py
```diff
@@ -942,7 +957,7 @@ def store(self, obj, *, createOrUpdate=True, forceVersion=True, versionLabel=Non
                                                 md5=local_state_fh.get('contentMd5'),
                                                 file_size=local_state_fh.get('contentSize'),
                                                 mimetype=local_state_fh.get('contentType'),
-                                                max_threads=max_threads)
+                                                max_threads=maxThreads or self._get_transfer_config_max_threads())
```
zimingd, line 960: Cache the value from config to avoid re-reading config file on every store.

docs/news.rst
```diff
@@ -2,6 +2,36 @@
 Release Notes
 =============
 
+2.1.0 (2020-04-0X)
+==================
+
+Highlights:
+----------------
+
+- The Synapse client object's :code:`store` and :code:`get` methods have been enhanced with an optional
```
zimingd, line 11: What are your thoughts on using a client attribute (e.g. `syn.num_threads`) instead of args in both get and store?

jkiang13: It would be somewhat less flexible in that you wouldn't be able to make a per call selection but it would be simpler.

I was actually thinking of combining the config object (where the current single_threaded value is stored) with the reading of the synapseConfig fil which would affect a few of your other comments as well. I would prefer it be on a separate module than the Synapse client itself, as I think there are likely to be places you'd want to reference those values as well as having an import to Synapse itself and having them stored directly on Synapse could then lead to a circular dependency issue.

If you are amenable to that I will make that change part of this PR.

thomasyu888: Make sure the right people are contacted about the docs for this change: https://docs.synapse.org/articles/client_configuration.html#customize-the-synapse-configuration-file

synapseclient/client.py
thomasyu888: Agreed, I think if a user is being intentional about setting the maxThreads, an error should be thrown it it is set incorrectly.

thomasyu888:
Is there a "risk" in setting the maxThreads too high?  If so, should we cap the value at which maxThreads can be set?

jkiang13:
> Is there a "risk" in setting the maxThreads too high? If so, should we cap the value at which maxThreads can be set?

Each thread has memory overhead and thread context switches have performance costs so allocating more threads than the CPUs on your system can reasonably run will just slow things down. It’s hard to make an assumption about what any user’s system can handle however, so any max should be pretty high (the equivalent was-cli setting max_concurrent_requests appears unbounded). I am fine with a high sanity cap though, 128 seems reasonable.

docs/news.rst
jkiang13: Looking closer config.single_thread is the oddity and there are already many other config properties already set directly on the Synapse object, so adding it there as Ziming suggested. Also added some caching on reading the config file once since we already parse it multiple times per init.

jkiang13:
Note there were some failing integration tests already merged into develop that are fixed in this branch. I mistakenly thought in my earlier PR that integration tests were auto run and all PRs by Travis. Those changes can be viewed in isolation [here](https://github.com/Sage-Bionetworks/synapsePythonClient/compare/develop...jkiang13:fix-upload-integration-tests) but I merged them into this branch because here would have been merge conflicts anyway with some of the suggested changes.

synapseclient/client.py
```diff
@@ -204,6 +209,7 @@ def __init__(self, repoEndpoint=None, authEndpoint=None, fileHandleEndpoint=None
         self.table_query_max_sleep = 20
         self.table_query_timeout = 600  # in seconds
         self.multi_threaded = False  # if set to True, multi threaded download will be used for http and https URLs
```
zimingd, line 211: having `multi_threaded` and `max_threads` in a case where `multi_threaded` only affects download is confusing. I would be ok with removing `max_threads` completely for the next release, and defaulting to multi_threaded implementation for download as well.

synapseclient/client.py
```diff
@@ -176,6 +177,10 @@ def __init__(self, repoEndpoint=None, authEndpoint=None, fileHandleEndpoint=None
 
         cache_root_dir = cache.CACHE_ROOT_DIR
 
+        # cache reading the config file at an instance level (i.e.
+        # the config will be read once per instantiated Synapse object).
+        self.getConfigFile = functools.lru_cache()(self._getConfigFile)
```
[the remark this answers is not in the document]
zimingd: I tested on Python 3.6 and verified that using the annotation works per instance
```
>>> import functools
>>> class A:
...      @functools.lru_cache()
...      def do(self):
...          print('using func')
...          return 'a'
...
...
>>> a = A()
>>> a2 = A()
>>> a2.do()
using func
'a'
>>> a.do()
using func
'a'
>>> a.do()
'a'
>>> a2.do()
'a'
>>> a3 = A()
>>> a3.do()
using func
'a'
>>>
```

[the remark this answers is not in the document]
jkiang13: I must have been thinking of a different memoizing decorator I have used in the past which had to be used in this way to support instance level caching vs static caching. Agree cleaner as a decorator.

synapseclient/client.py
jkiang13: I agree it's a bit confusing to have both values. I do think it's useful to let the user customize the value, and think it is cleaner if it applies to both upload and download. We just added the multiThreaded property so I don't think we want to immediately yank it though, but I think the doc you put in suggests it may be pulled in a future release. We could make it a deprecated property however.

zimingd:
Changes look good. Let me know if this is ready to merge and I'll hit the button.

As an aside, I really like your idea of writing the release highlights as we check in code. We're less likely to accidentally omit anything important this way. I'll be doing the same in my future pull requests.

jkiang13:
I think it's ready to go, although I haven't been able to successfully run the integration tests on the current commit I think due to the issue with dev. I last ran them successfully on Sat though and the changes since then are mostly code quality vs functional.

jkiang13:
Integration tests pass now that dev is in better shape so I'm going to merge this now.

PR #739, opened 2020-04-09
Support parallelized integration tests from Linux

jkiang13:
Includes changes to the integration tests to support the nose multiprocessing plugin to support faster integration test runs via parallelization. https://nose.readthedocs.io/en/latest/plugins/multiprocess.html

The ```test/integration/__init__.py``` fixtures are changed to support being a shared fixture across processes. This necessitates creating separate Synapse objects after the fork for child processes (to avoid SSL issues arising from multiple connections from forked identical Synapse object's http connection pools) as well as moving the scheduled cleanups for resources associated with individual modules into the teardowns of those modules specifically, since the parent process cleanup can no longer be responsible. This is better anyway since it more tightly couples test module cleanup with their associated tests and leaves less cruft behind if a build should be aborted early for any reason.

On my own machine this drops an integration test run from ~6.5 minutes to ~2.5 minutes, and on Travis the Linux tests go from about 10 minutes to about 5 minutes, which makes it quicker to verify changes. This isn't supported on OSX unfortunately (outside of a VM or container) but the tests can continue to run as before there without use of the multi processsing params.

### DIFF
--- a/tests/integration/synapseutils/test_synapseutils_sync.py
+++ b/tests/integration/synapseutils/test_synapseutils_sync.py
@@ -107,48 +107,6 @@ def test_syncToSynapse():
                 assert_equals(set(orig_list), set(new_list))
 
 
-def test_syncFromSynapse():
-    """This function tests recursive download as defined in syncFromSynapse
-    most of the functionality of this function are already tested in the
-    tests/integration/test_command_line_client::test_command_get_recursive_and_query
-
-    which means that the only test if for path=None
-    """
-    # Create a Project
-    project_entity = syn.store(Project(name=str(uuid.uuid4())))
-    schedule_for_cleanup(project_entity.id)
-
-    # Create a Folder in Project
-    folder_entity = syn.store(Folder(name=str(uuid.uuid4()), parent=project_entity))
-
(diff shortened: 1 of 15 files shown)


tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -105,51 +103,9 @@ def test_syncToSynapse():
                 orig_list = ['%s.%s' % (i.id, i.versionNumber) if isinstance(i, Entity) else i
                              for i in syn._convertProvenanceList(orig.split(';'))]
                 new_list = ['%s.%s' % (i.id, i.versionNumber) if isinstance(i, Entity) else i
-                             for i in syn._convertProvenanceList(new.split(';'))]
+                            for i in syn._convertProvenanceList(new.split(';'))]
                 assert_equals(set(orig_list), set(new_list))
 
-        
-def test_syncFromSynapse():
```
jkiang13, line 112: This removed test is being shadowed by a subsequent (almost) identical test of the same name.

jkiang13:
This was nose based, we've since switched to pytest.

PR #741, opened 2020-04-10
py-2.1 release notes

jkiang13:
Rendered sphinx notes:


thomasyu888:
Looks good. Can we distribute this to the docs group?

kdaily:
Looks pretty good! I would:

1. Change 'cpu' to 'CPU'.
1. Link to our docs article on how to use a synapse config file (https://docs.synapse.org/articles/client_configuration.html)

kdaily:
Is there some baseline guidance on what 'system resources' would benefit from a value greater than 'cpu count + 4'?

kdaily:
Does the thread increase currently apply to both upload and download?

jkiang13:
> Does the thread increase currently apply to both upload and download?

Yes it does. Downloads are still influenced by the separate multi_threaded attribute which we added last release which may be a little confusing (for multi threaded downloads you must still set that attribute) after which it will obey the max_threads value.

We mentioned in the docs last release that the multi_threaded setting was meant to be temporary as all downloads would move to use multiple threads/connections and I think we will remove it in the next release, but having just added it thought it should remain in this release.

jkiang13:
> Is there some baseline guidance on what 'system resources' would benefit from a value greater than 'cpu count + 4'?

It's tricky since it depends on different factors, not just the number of CPUs, in fact mostly on the user's bandwidth. If they are already maxing out their upload bandwidth then adding all the concurrent requests in the world won't help. I think we've tried to pick a sensible default but then leave it to the user to set another value if they would benefit from it.

If we wanted to describe this more thoroughly we could borrow from Amazon's language here (max_concurrent_requests) which this setting is analogous to:
https://docs.aws.amazon.com/cli/latest/topic/s3-config.html

PR #742, opened 2020-04-17
fix KeyError when uploading file that already exists in synapse (SYNPY-1036)

zimingd:


synapseclient/client.py
```diff
@@ -972,8 +972,9 @@ def store(self, obj, *, createOrUpdate=True, forceVersion=True, versionLabel=Non
                 properties['dataFileHandleId'] = bundle['entity']['dataFileHandleId']
 
             # update the file_handle metadata if the FileEntity's FileHandle id has changed
-            if '_file_handle' in local_state \
```
zimingd, line 975: bug was caused by `'_file_handle'` in `local_state` being initialized elsewhere with `{'id':None , ...}`, which causes `properties['dataFileHandleId'] != local_state['_file_handle'].get('id', None)` to evaluate as `True` when an existing FileHandle was already found on Synapse causing a value to be set for `properties['dataFileHandleId']`

zimingd:
The `continuation line over-indented for visual indent` from the bot is wrong

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -1285,3 +1285,29 @@ def test_get_transfer_config_max_threads():
...
+                                        'concreteType': 'org.sagebionetworks.repo.model.file.S3FileHandle'}]
+                       }
+    with patch.object(syn, '_getEntityBundle', return_value=returned_bundle), \
+         patch.object(synapseclient.client, 'upload_file_handle', return_value=returned_file_handle), \
+         patch.object(syn.cache, 'contains', return_value=True), \
+         patch.object(syn, '_createEntity'), \
+         patch.object(syn, 'set_annotations'), \
+         patch.object(Entity, 'create'), \
+         patch.object(syn, 'get'):
+
```
jkiang13, line 1310: This level of patching reflects how gnarly this store fn is.

PR #743, opened 2020-04-21
SYNPY-1056 - expectation new style annotations breaking getSubmission call

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1056

syn.getSubmission pre fetches the associated entity, and in doing so passes along annotations returned in its entityBundleJSON from the Evaluations api which are old style (stringAnnotations, dateAnnotations, etc.). With recent changes getWithEntityBundle expects new style annotations as used with Entity Bundle Services V2.

This attempts to translate between the two for the purposes of making the getWithEntityBundle call.

### DIFF
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -333,3 +333,31 @@ def from_synapse_annotations(raw_annotations: typing.Dict[str, typing.Any]) -> A
         annos[key] = [conversion_func(v) for v in value_and_type['value']]
 
     return annos
+
+
+def convert_old_annotation_json(annotations):
+    """Transforms a parsed JSON dictionary of old style annotations
+    into a new style consistent with the entity bundle v2 format."""
+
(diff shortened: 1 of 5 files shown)


synapseclient/annotations.py
```diff
@@ -333,3 +333,31 @@ def from_synapse_annotations(raw_annotations: typing.Dict[str, typing.Any]) -> A
...
+    converted = {k: v for k, v in annotations.items() if k in ('id', 'etag')}
+    converted_annos = converted['annotations'] = {}
+
+    type_mapping = {
+        'doubleAnnotations': 'DOUBLE',
+        'stringAnnotations': 'STRING',
+        'longAnnotations': "LONG",
+        'dateAnnotations': 'TIMESTAMP_MS',
+
+        # TODO what to do with blobAnnotations
```
jkiang13, line 351: There doesn't seem to be an equivalent of blobAnnotations in the newer form... not sure if that is an issue or not.

PR #744, opened 2020-04-22
Travis/MacOS/cryptography workaround

jkiang13:
Workaround for https://github.com/pyca/cryptography/issues/5215 on Travis/MacOS.

Probably don't need to merge unless that is not resolved quickly...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -57,7 +57,9 @@ install:
     sudo apt-get install gnome-keyring
   fi
 - pip install -U setuptools
-- travis_wait pip install pysftp pandas boto3
+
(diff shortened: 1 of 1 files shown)


thomasyu888:
Just out of curiosity, how do we use `cryptography`?

jkiang13:
> Just out of curiosity, how do we use `cryptography`?

It's a dependency via pysftp, the sftp integration tests fail on Travis under MacOS at the moment because of the issue in the latest version of cryptography.

jkiang13:
This is fixed upstream.

PR #745, opened 2020-05-11
SYNPY-1059: Change collections to collections.abc where appropriate

thomasyu888:
My understanding is the `collections.namedtuple()` and `collections.OrderedDict()` will be obsolete as well in Python 3.10, but there is still some time for that.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -28,6 +28,7 @@ See also the `Synapse API documentation <https://docs.synapse.org/rest/>`_.
 
 """
 import collections
+import collections.abc
 import configparser
 import deprecated
 import functools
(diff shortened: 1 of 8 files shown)


synapseclient/client.py
```diff
@@ -3237,7 +3238,7 @@ def _build_table_download_file_handle_list(self, table, columns):
         col_indices = [i for i, h in enumerate(table.headers) if h.name in columns]
         # see: http://docs.synapse.org/rest/org/sagebionetworks/repo/model/file/BulkFileDownloadRequest.html
         file_handle_associations = []
-        file_handle_to_path_map = collections.OrderedDict()
+        file_handle_to_path_map = collections.abc.OrderedDict()
```
thomasyu888, line 3241:
```suggestion
        file_handle_to_path_map = collections.OrderedDict()
```

PR #746, opened 2020-05-19
SYNPY-1049 STS storage location support

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1049

Adds STS storage location support to the python client.

- Adds functions in the client and command line for retrieving a token and outputting it.
- Adds an optional setting via synapseConfig configuration to automatically obtain STS tokens for storage locations that are enabled for STS it and uses those to upload and download files via boto3 (if it is installed).  Presently this works for downloading from any STS enabled storage location, but only works for uploading for custom external storage locations due to https://sagebionetworks.jira.com/browse/PLFM-6227 If preconditions do not allow the files are transferred as normal.
- Tokens are cached for reuse in memory so that long running processes with repeated access to the same storage locations don't need to re-fetch STS tokens for each transfer
- Makes use of https://github.com/Sage-Bionetworks/synapsedev-infra/pull/56 to backfill external storage location integration tests as well as add a new integration test for STS storage locations


synapseclient/core/constants/concrete_types.py
```diff
@@ -10,6 +10,7 @@
 
 # Concrete types for FileHandles
 EXTERNAL_OBJECT_STORE_FILE_HANDLE = "org.sagebionetworks.repo.model.file.ExternalObjectStoreFileHandle"
+EXTERNAL_FILE_HANDLE = 'org.sagebionetworks.repo.model.file.ExternalFileHandle'
```
thomasyu888, line 13: Would it be preferable to be able to retrieve these via a rest call, or is maintaining this list here fine?

synapseclient/core/utils.py
```diff
@@ -880,3 +880,8 @@ def require_param(param, name):
     if param is None:
         raise ValueError("%s parameter is required." % name)
 
+
+def snake_case(string):
+    """Convert the given string from CamelCase to snake_case"""
+    # https: // stackoverflow.com / a / 1176023
```
thomasyu888, line 886: Interesting, is this a contention for commenting stack overflow urls?

tests/unit/synapseclient/core/unit_test_retry.py
```diff
@@ -59,3 +59,21 @@ def foo():
     assert_raises(SynapseError, with_retry, function, **retryParams)
     assert_equals(function.call_count, 1 + 4 + 3 + 4 + 1)
 
+
+def test_with_retry__no_status_code():
```
thomasyu888, line 63: This is awesome!  So this means that if a function throws any exception, it will retry?

tests/unit/synapseclient/unit_test_commandline.py
```diff
@@ -2,8 +2,8 @@
 
 """
 
-from nose.tools import assert_equals, assert_true, assert_false
-from mock import patch
+from nose.tools import assert_equal, assert_true, assert_false
+from mock import patch, Mock
```
thomasyu888, line 6: Small thing, but mock is now part of `unittest`. Eventually the backport will probably go away, so all imports will most likely be 

```
from unittest.mock....
```

docs/news.rst
```diff
@@ -29,7 +29,41 @@ Highlights:
...
+
+    # get an STS token for a previously enabled storage location and use with boto
+    import boto3
+    boto_credentials = synapseclient.get_sts_storage_token(folder_id, 'read_only', output_format='boto')
+    s3_client = boto3.client('s3', **boto_credentials)
+    s3_client.download_file(<bucket>, <key>, <filename>)
+
+  .. code-block::
+
+    # from the command line output commands that can be used to supply credentials to the awscli
```
thomasyu888, line 49: Awesome, so with this enabled, the client would check if a folder has STS enabled?   If the token is expired, would the user be prompted to get a new token?

synapseclient/__main__.py
```diff
@@ -75,9 +76,9 @@
 
 import synapseclient
 import synapseutils
-from . import Activity
-from .wiki import Wiki
-from .annotations import Annotations
+from synapseclient import Activity
```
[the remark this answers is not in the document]
jkiang13: In general I prefer absolute imports, but in this case it was just that it made it nice to be able to create a run configuration from my IDE with a right click without an installed entry point (the equivalent of “python synapseclient/__main__.py”) which doesn’t work with relative imports. “python -m synapseclient.__main__” works from the command line either way of course.

synapseclient/core/constants/concrete_types.py
jkiang13: These concrete types are spec’ed as part of the API so I think it’s fine to have them defined in client constants and I wouldn’t expect them to dynamically change any more than I would expect the endpoint urls to be dynamically change. If they were fetched with a remote call there would need to be a bootstrap request from the client.

synapseclient/core/utils.py
jkiang13: Not sure how those extra spaces got in there, probably just a copy and paste artifact. Fixed.

tests/unit/synapseclient/core/unit_test_retry.py
jkiang13: The change to synapseclient.core.retry.with_retry in this PR allows it to be used also with functions whose return value is something other than a requests http response with a status code (in which case it still benefits from the Exception handling and exponential backoff functionality). I needed this functionality to use with boto3 calls that don’t have the same shape and it made sense to modify the existing utility to be compatible rather than duplicate it.

tests/unit/synapseclient/unit_test_commandline.py
jkiang13: Yes this is true, good point. The current project uses the backport throughout so I will make a separate pull request to search and replace all uses of the backport (after which we can remove it as a test_requires dependency).

docs/news.rst
jkiang13: If the user is using boto3 transparently via the synapse client (i.e. they have an STS storage location configured and have use_boto_sts=true in their .synapseConfig) then they don’t need to manage tokens, uploads and downloads will automatically favor using boto to conduct the transfer, and tokens will be cached (in memory for the life of the process) and refreshed as needed.

If they are exporting the token to use themselves e.g. from the command line and the awscli then they will need to get a new token every 12 hours.

docs/news.rst
```diff
@@ -29,7 +29,43 @@ Highlights:
      [transfer]
      max_threads=16
 
-  This release also includes other optimizations associated with file uploads.
+- This release includes support for directly accessing
```
jkiang13, line 32: Add example of getting bucket name and key of a storage location.

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+                token = token_cache[entity_id] = self._fetch_token(syn, entity_id, permission)
+
+        return token
+
+    @staticmethod
+    def _fetch_token(syn, entity_id, permission):
+        return syn.restGET(f'/entity/{entity_id}/sts?permission={permission}')
+
+
+_TOKEN_STORE = _StsTokenStore()
```
jkiang13, line 99: Move to an instance variable of the Synapse object

tests/integration/synapseclient/core/test_external_storage.py
```diff
@@ -0,0 +1,190 @@
...
+        with_retry(
+            lambda: s3_read_client.get_object(
+                Bucket=bucket_name,
+                Key=remote_key
+            ),
+            retry_exceptions=[s3_read_client.exceptions.NoSuchKey]
+        )
+
+        # create an external file handle so we can read it via synapse
+        file_handle = syn._createExternalS3FileHandle(
```
jkiang13, line 178: Provide an example of doing this in a public facing way.

tests/integration/synapseclient/core/test_external_storage.py
```diff
@@ -0,0 +1,190 @@
...
+            Key=f"{key_prefix}/owner.txt"
+        )
+
+        return s3_client
+
+    def _configure_storage_location(self, sts_enabled=False):
+        folder_name = str(uuid.uuid4())
+        s3_client = self._prepare_bucket_location(folder_name)
+
+        folder = syn.store(Folder(name=folder_name, parent=integration.project))
```
jkiang13, line 81: Create a convenience function for creating storage location and enabling it for STS?

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+        # e.g. expiration
+        value = {"aws_{}".format(snake_case(k)): value[k] for k in (
+            'accessKeyId', 'secretAccessKey', 'sessionToken'
+        )}
+
+    elif output_format == 'shell':
+        # make output in the form of commands that will set the credentials into the user's
+        # environment such that they can e.g. run awscli commands
+
+        if platform.system() == 'Windows' and 'bash' not in os.environ.get('SHELL', ''):
```
jkiang13, line 121: Powershell?

tests/unit/synapseclient/unit_test_commandline.py
```diff
@@ -56,3 +56,18 @@ def test_get_multi_threaded_flag():
...
+
+
+def test_get_sts_token():
+    """Test getting an STS token."""
+    folder_id = 'syn_1'
+    permission = 'read_write'
+    syn = Mock()
+
+    expected_output = 'export foo=bar'
+    syn.get_sts_storage_token.return_value = expected_output
```
zimingd, line 68: expected output never asserted

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+    Check if the boto/STS transfers are enabled in the Synapse configuration.
+    If enabled then synapseclient will attempt to automatically use boto to upload
+    and download from supported storage locations that are sts enabled.
+
+    :param syn:         A Synapse client
+
+    :returns: True if STS if enabled, False otherwise
+    """
+
+    use_boto_sts = syn._get_config_section_dict('transfer').get('use_boto_sts', '')
```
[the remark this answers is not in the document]
jkiang13: Made a settable value on the Synapse client.
`syn.use_boto_sts_transfers`

synapseclient/core/remote_file_storage_wrappers.py
```diff
@@ -51,24 +69,53 @@ def download_file(bucket, endpoint_url, remote_file_key, download_file_path, pro
...
+                    prefix='Downloading',
+                )
+
+            s3_obj.download_file(
+                download_file_path,
+                Callback=progress_callback,
+                Config=transfer_config,
+            )
+
+            # why return what we were passed...?
```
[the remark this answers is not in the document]
jkiang13: Good point these are both opt-in at the moment, but the opt in for this on a sts enabled endpoint should be higher precedence.

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+export AWS_SECRET_ACCESS_KEY="{value['secretAccessKey']}"
+export AWS_SESSION_TOKEN="{value['sessionToken']}"
+"""
+    elif output_format != 'json':
+        raise ValueError(f'Unrecognized output_format {output_format}')
+
+    return value
+
+
+def with_boto_sts_credentials(fn, *args, **kwargs):
```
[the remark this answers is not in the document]
jkiang13: I think a context manager doesn't fit here since we're operating on the result of the call (i.e. the exception) not providing wrapping context to the call. A decorator could work but it would need to be pretty proscriptive, in that the wrapped function would have to fit the credentials, and at that point I think it doesn't really become too useful as general decorator since you'd probably have to create a special closure to fit the case anyway.

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+
+    # the passed fn takes boto style credentials
+    token_kwargs = dict(kwargs)
+    token_kwargs['output_format'] = 'boto'
+
+    for attempt in range(2):
+        credentials = get_sts_credentials(*args, **token_kwargs)
+        try:
+            response = fn(**credentials)
```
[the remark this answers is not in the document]
jkiang13: Done.

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+    DEFAULT_MIN_LIFE = datetime.timedelta(hours=1)
+
+    def __init__(self, max_token_cache_size=DEFAULT_TOKEN_CACHE_SIZE):
+        self._tokens = {p: _TokenCache(max_token_cache_size) for p in STS_PERMISSIONS}
+        self._lock = threading.Lock()
+
+    def get_token(self, syn, entity_id, permission, min_remaining_life: datetime.timedelta = None):
+        min_remaining_life = min_remaining_life if min_remaining_life is not None else self.DEFAULT_MIN_LIFE
+
+        utcnow = datetime.datetime.utcnow()
```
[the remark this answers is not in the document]
jkiang13: Done.

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+    @staticmethod
+    def _fetch_token(syn, entity_id, permission):
+        return syn.restGET(f'/entity/{entity_id}/sts?permission={permission}')
+
+
+_TOKEN_STORE = _StsTokenStore()
+
+
+def get_sts_credentials(syn, entity_id, permission, output_format='json', **kwargs):
+    """Get STS credentials for the given entity_id and permission, outputting it in the given format
```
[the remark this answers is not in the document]
jkiang13: Removed kwargs, defined parameters explicitly.

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+    # we won't hand out previously retrieved tokens that have less than this amount of
+    # time left on them. we don't know exactly when they'll be used so we don't want to
+    # hand out an about-to-expire cached token.
+    DEFAULT_MIN_LIFE = datetime.timedelta(hours=1)
+
+    def __init__(self, max_token_cache_size=DEFAULT_TOKEN_CACHE_SIZE):
+        self._tokens = {p: _TokenCache(max_token_cache_size) for p in STS_PERMISSIONS}
+        self._lock = threading.Lock()
+
+    def get_token(self, syn, entity_id, permission, min_remaining_life: datetime.timedelta = None):
```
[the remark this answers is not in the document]
jkiang13: Made it a keyword only arg on function that calls this, and made it a required positional arg here since this is an internal call anyway.

tests/unit/synapseclient/unit_test_commandline.py
jkiang13: Fixed.

synapseclient/core/sts_transfer.py
```diff
@@ -0,0 +1,210 @@
...
+    on each additional insert."""
+
+    def __init__(self, max_size):
+        super().__init__()
+        self.max_size = max_size
+
+    def __setitem__(self, key, value):
+        super().__setitem__(key, value)
+        self._prune()
+
```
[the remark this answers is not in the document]
jkiang13: It's not a FIFO cache since it is implicitly ordered by the expiration of the tokens. If it were an LRU cache then it could expire a token with longer life that hadn't been more recently used before a token that was about to expire. It's not necessarily obvious to me that that is preferable.

synapseclient/client.py
```diff
@@ -1693,17 +1692,43 @@ def _downloadFileHandle(self, fileHandleId, objectId, objectType, destination, r
...
                                                                     profile_name=profile)
-                elif self.multi_threaded and fileHandle['concreteType'] == concrete_types.S3_FILE_HANDLE:
+
+                elif self.multi_threaded and concreteType == concrete_types.S3_FILE_HANDLE:
                     downloaded_path = self._download_from_url_multi_threaded(fileHandleId,
                                                                              objectId,
                                                                              objectType,
                                                                              destination,
                                                                              expected_md5=fileHandle.get('contentMd5'))
+                elif sts_transfer.is_boto_sts_transfer_enabled(self) and \
```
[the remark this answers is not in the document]
jkiang13: Good point these are both opt in right now, but it probably makes sense that the opt in on the more specific sts enabled end point takes higher precedence.

synapseclient/core/sts_transfer.py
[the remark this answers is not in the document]
jkiang13: It will function in the edge case just generating new tokens and re-adding them, as you say. It will never not return a token for the requested location due to being full. If it were to switch to FIFO then you'd need to inspect all the tokens every prune to see if they had expired, not just check until you found one that hadn't (or additionally keep a separate data structure).

docs/build/html/S3Storage.html
```diff
@@ -0,0 +1,296 @@
...
+</div>
+</div></blockquote>
+</div>
+<div class="section" id="automatic-transfers-to-from-sts-storage-locations-using-boto3-with-synapseclient">
+<h3>Automatic transfers to/from STS storage locations using boto3 with synapseclient<a class="headerlink" href="#automatic-transfers-to-from-sts-storage-locations-using-boto3-with-synapseclient" title="Permalink to this headline">¶</a></h3>
+<p>The Python Synapse client can be configured to automatically use STS tokens to perform uploads and downloads to enabled
+storage locations using an installed boto3 library rather than through the traditional Synapse client APIs.
+This can improve performance in certain situations, particularly uploads of large files, as the data transfer itself
+can be conducted purely against the AWS S3 APIs, only invoking the Synapse APIs to retrieve the necessary token and
+to update Synapse meta data in the case of an upload. Once configured to do so, retrieval of STS tokens for supported
```
kelshmo, line 225:
```suggestion
to update Synapse metadata in the case of an upload. Once configured to do so, retrieval of STS tokens for supported
```

docs/build/html/S3Storage.html
```diff
@@ -0,0 +1,296 @@
...
+</ol>
+<blockquote>
+<div><div class="highlight-default notranslate"><div class="highlight"><pre><span></span><span class="n">pip</span> <span class="n">install</span> <span class="n">boto3</span>
+</pre></div>
+</div>
+</div></blockquote>
+<ol class="arabic simple" start="2">
+<li><p>Update your Synapse client configuration file (typically “.synapseConfig” in your $HOME directory unless otherwise
+configured) with the following (adding the [transfer] section if it is not already present), or configure
+it on a per Synapse client object basis by setting the <strong>use_boto_sts_transfers</strong> property.</p></li>
```
kelshmo, line 239:
```suggestion
<li><p>To enable automatic transfers on all uploads and downloads, update your Synapse client configuration file (typically “.synapseConfig” in your $HOME directory, unless otherwise
configured) with the [transfer] section, if it is not already present. To leverage STS/boto3 transfers on a per Synapse client object basis, set the <strong>use_boto_sts_transfers</strong> property.</p></li>
```

docs/build/html/S3Storage.html
```diff
@@ -0,0 +1,296 @@
...
+</div>
+<div class="section" id="automatic-transfers-to-from-sts-storage-locations-using-boto3-with-synapseclient">
+<h3>Automatic transfers to/from STS storage locations using boto3 with synapseclient<a class="headerlink" href="#automatic-transfers-to-from-sts-storage-locations-using-boto3-with-synapseclient" title="Permalink to this headline">¶</a></h3>
+<p>The Python Synapse client can be configured to automatically use STS tokens to perform uploads and downloads to enabled
+storage locations using an installed boto3 library rather than through the traditional Synapse client APIs.
+This can improve performance in certain situations, particularly uploads of large files, as the data transfer itself
+can be conducted purely against the AWS S3 APIs, only invoking the Synapse APIs to retrieve the necessary token and
+to update Synapse meta data in the case of an upload. Once configured to do so, retrieval of STS tokens for supported
+operations occurs automatically without any change in synapseclient usage.</p>
+<p>To enable STS/boto3 transfers, do the following:</p>
```
kelshmo, line 227:
```suggestion
<p>To enable STS/boto3 transfers on all `get` and `store` operations, do the following:</p>
```

docs/build/html/S3Storage.html
```diff
@@ -0,0 +1,296 @@
...
+<span class="n">use_boto_sts</span><span class="o">=</span><span class="n">true</span>
+
+<span class="c1"># alternatively set on a per instance basis within python code</span>
+<span class="n">syn</span><span class="o">.</span><span class="n">use_boto_sts_transfers</span> <span class="o">=</span> <span class="kc">True</span>
+</pre></div>
+</div>
+</div></blockquote>
+<p>Once this is done any get or store operation through the configured client to or from STS enabed storage locations
+will be automatically conducted using the boto3 library rather than through the normal Synapse APIs.
+Note that if boto3 is not installed then these settings will have no effect.</p>
```
kelshmo, line 253:
```suggestion
<p>Note that if boto3 is not installed, then these settings will have no effect.</p>
```
I removed the leading sentence as I think it is well-stated above.

PR #749, opened 2020-06-02
SYNPY-1036 Handle id-less entity updates without a failed create

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1036

https://github.com/Sage-Bionetworks/synapsePythonClient/pull/723 prevented a 403 that occurred while attempting to retrieve the file handle during a store, but another 403 is raised when "creating" the entity later in the store operation if the user is different from the original file handle creator.

Currently we seem to handle every store that doesn't include the entity id as a create (POST /entity) only switching to handling it as an update (PUT /entity/[id]) after the back end returns a 409 conflict error for the create. This mostly works when the user invoking the store is the creator of the file handle, but returns an uncaught 403 for other users.

This change is so that if the the bundle fetch earlier in the store operation returns an entity bundle, we process the store as an update (merging the passed properties and annotations with those of found bundle) rather than as a failed create followed by an update. Doing it this way succeeds for any user that has permission to update the File, and avoids the extra failed create call.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -857,6 +857,15 @@ class Synapse(object):
                                  'for "ifcollision"' % ifcollision)
         return downloadPath
 
+    def _merge_update_with_bundle(self, bundle, properties, annotations):
+        existing_entity = bundle['entity']
+        existing_annos = from_synapse_annotations(bundle['annotations'])
+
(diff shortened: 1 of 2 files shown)


jkiang13:
Closing to open against v2.1.0-rc instead.

PR #750, opened 2020-06-02
SYNPY-1036 Handle id-less entity updates without a failed create

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1036

https://github.com/Sage-Bionetworks/synapsePythonClient/pull/723 prevented a 403 that occurred while attempting to retrieve the file handle during a store, but another 403 is raised when "creating" the entity later in the store operation if the user is different from the original file handle creator.

Currently we seem to handle every store that doesn't include the entity id as a create (POST /entity) only switching to handling it as an update (PUT /entity/[id]) after the back end returns a 409 conflict error for the create. This mostly works when the user invoking the store is the creator of the file handle, but returns an uncaught 403 for other users.

This change is so that if the the bundle fetch earlier in the store operation returns an entity bundle, we process the store as an update (merging the passed properties and annotations with those of the found bundle) rather than as a failed create followed by an update. Doing it this way succeeds for any user that has permission to update the File, and avoids the extra failed create call.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -857,14 +857,6 @@ class Synapse(object):
                                  'for "ifcollision"' % ifcollision)
         return downloadPath
 
-    def _merge_update_with_bundle(self, bundle, properties, annotations):
-        existing_entity = bundle['entity']
-        existing_annos = from_synapse_annotations(bundle['annotations'])
-
-        merged_props = {**existing_entity, **properties}
-        merged_annos = {**existing_annos, **annotations}
-
-        return merged_props, merged_annos
 
     def store(self, obj, *, createOrUpdate=True, forceVersion=True, versionLabel=None, isRestricted=False,
               activity=None, used=None, executed=None, activityName=None, activityDescription=None):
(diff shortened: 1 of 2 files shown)


synapseclient/client.py
```diff
@@ -952,6 +961,15 @@ def store(self, obj, *, createOrUpdate=True, forceVersion=True, versionLabel=Non
             bundle = self._getEntityBundle(entity)
 
             if bundle:
+                if createOrUpdate:
+                    # update our properties from the existing bundle so that we have
```
zimingd, line 965: This would not work if we deleted some annotation entries since we would end up merging the ones we deleted back in.

zimingd: We shouldn't rely on the createOrUpdate flag being set to alter this behavior. I'm pretty sure most users don't even know that it exists.

synapseclient/client.py
```diff
@@ -1040,17 +1058,14 @@ def store(self, obj, *, createOrUpdate=True, forceVersion=True, versionLabel=Non
...
+                        annotations,
+                    )
 
-                    # Update the conflicting Entity
-                    existing_entity.update(properties)
-                    properties = self._updateEntity(existing_entity, forceVersion, versionLabel)
+                    properties = self._updateEntity(properties, forceVersion, versionLabel)
 
-                    # Merge new annotations with existing annotations
-                    existing_annos = from_synapse_annotations(bundle['annotations'])
```
zimingd, line 1051: In this case since an ID didn't previously exist, we don't treat missing annotations as deletions since they might have just created an new Entity object w/ no annotations initialized.

synapseclient/client.py
jkiang13: Changed it to only merge annotations when resolving a 409.

PR #751, opened 2020-06-03
SYNPY-1061 - clean up all PEP8 issues

jkiang13:
* Cleans up all outstanding pep8 issues
* Adds flake8 configuration to make it easy to confirm checks from the command line in the future without having to submit a PR to see pep8speaks results
* Modifies pep8speaks configuration to use the the flake8 rules for consistent results
* Enforces flake8 as part of Travis build
* Removes usage of mock backport in favor of unittest.mock


synapseclient/client.py
```diff
@@ -3317,7 +3338,7 @@ def downloadTableColumns(self, table, columns, **kwargs):
 
         """
 
-        FAILURE_CODES = ["NOT_FOUND", "UNAUTHORIZED", "DUPLICATE", "EXCEEDS_SIZE_LIMIT", "UNKNOWN_ERROR"]
+        # FAILURE_CODES = ["NOT_FOUND", "UNAUTHORIZED", "DUPLICATE", "EXCEEDS_SIZE_LIMIT", "UNKNOWN_ERROR"]
```
zimingd, line 3341: remove if unused

synapseclient/core/utils.py
```diff
@@ -547,7 +547,7 @@ def query_limit_and_offset(query, hard_limit=1000):
     """
     # Regex a lower-case string to simplify matching
     tempQueryStr = query.lower()
-    regex = '\A(.*\s)(offset|limit)\s*(\d*\s*)\Z'
+    regex = '\A(.*\s)(offset|limit)\s*(\d*\s*)\Z'  # noqa double backslashes?
```
zimingd, line 550: remove comment if resolved

synapseclient/table.py
```diff
@@ -1242,7 +1242,7 @@ def Table(schema, values, **kwargs):
     try:
         import pandas as pd
         pandas_available = True
-    except:
+    except:  # noqa
```
zimingd, line 1245: except ImportError

synapseutils/sync.py
```diff
@@ -184,20 +185,26 @@ def _checkProvenace(item, path):
                     bundle = syn._getFromFile(item)
                     return bundle
                 except SynapseFileNotFoundError:
-                    SynapseProvenanceError(("The provenance record for file: %s is incorrect.\n"
-                                            "Specifically %s is not being uploaded and is not in Synapse."
-                                            % (path, item)))
+                    # TODO absence of a raise here appears to be a bug and yet tests fail if this is raised
```
zimingd, line 188: looks like the intention was to raise the exception here. we should fix tests to account for this error being thrown

tests/integration/synapseclient/core/test_download.py
```diff
@@ -38,7 +39,7 @@ def test_download_check_md5():
 
 def test_resume_partial_download():
     original_file = utils.make_bogus_data_file(40000)
-    original_md5 = utils.md5_for_file(original_file).hexdigest()
+    utils.md5_for_file(original_file).hexdigest()
```
zimingd, line 42: remove unused

tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -234,46 +235,10 @@ def test_write_manifest_data__unicode_characters_in_rows():
         assert_equals(datarow['col_B'], dfrow.col_B)
 
 
-def test_syncFromSynapse():
```
zimingd, line 237: why was this deleted?

jkiang13: It's shadowing another test of the same name:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/67db15cf5f32fcb4e7953396df192712f82ff733/tests/integration/synapseutils/test_synapseutils_sync.py#L113

synapseclient/table.py
jkiang13: There is another [import pandas ](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/67db15cf5f32fcb4e7953396df192712f82ff733/synapseclient/table.py#L365) in this file that implies that an import can somehow raise other errors, the associated [bug](https://sagebionetworks.jira.com/browse/SYNPY-177) implies an AttributeError is possible, which seems strange, but given that a previous bug exists for this I was not confident in narrowing the Exception.

synapseutils/sync.py
jkiang13: I agree it should be fixed, but will need to look at it further given that the change is to the actual code and not just the tests. I've created an [issue](https://sagebionetworks.jira.com/browse/SYNPY-1069) for it and will address it separately.

PR #754, opened 2020-06-08
SYNPY-1070 - Add support for submission view

thomasyu888:
@jkiang13 
Due to the potential plan on creating sub-clients, i can also add this into `challengeutils` for now.  I just wanted this to be reviewed to make sure I am implementing it the way you would.  If `challengeutils` becomes `synapseclient-challenge` this is most likely something that could live in that module.  Although it could also live under `table / view` services...  

* Adds ability to create submission views via the client.
* Example below on how to add submissionAnnotation using the existing `Annotation` class
* Use `async` call to get all annotation keys found in a entity or submission view and return them as column headers.

```
import synapseclient
from synapseclient.table import SubmissionViewSchema

syn = synapseclient.login()
project = syn.get("syn4990358")

# Create submission view schema
submission_view = SubmissionViewSchema(
    name='Testing Sub View', parent=project, scopes=['9614543']
)
schema = syn.store(submission_view)

# Query schema view
# syn22141919
temp = syn.tableQuery(f"select * from {schema.id}")
subviewdf = temp.asDataFrame()
print(subviewdf)

# Store new style annotations
sub_status = syn.getSubmissionStatus("9704911")
add_annotations = synapseclient.Annotations(
    sub_status.id, sub_status.etag, {"testing": "wow"}
)
to_add = synapseclient.annotations.to_synapse_annotations(add_annotations)

sub_status['submissionAnnotations'] = to_add
syn.store(sub_status)

# Add columns
test_column = syn.store(synapseclient.Column(name='testing', columnType='STRING'))
schema.addColumn(test_column)
schema = syn.store(schema, forceVersion=False)

```

TODO: Add test


synapseclient/table.py
```diff
@@ -866,9 +866,142 @@ def _filter_duplicate_columns(self, syn, columns_to_add):
...
+        if self.get('scopeIds') is None:
+            self.scopeIds = []
+
+        # add the scopes last so that we can append the passed in scopes to those defined in properties
+        if scopes is not None:
+            self.add_scope(scopes)
+
+    def add_scope(self, entities):
+        """
+        :param entities: a Project or Folder object or its ID, can also be a list of them
```
thomasyu888, line 932: Update this docstring

synapseclient/client.py
```diff
@@ -3449,6 +3449,31 @@ def _get_default_entity_view_columns(self, view_type_mask):
...
+    def _get_annotation_submission_view_columns(self, scope_ids):
+        view_scope = {'scope': scope_ids,
+                      'viewEntityType': 'submissionview'}
+        columns = []
+        next_page_token = None
+        while True:
+            params = {}
+            if next_page_token:
+                params = {'nextPageToken': next_page_token}
+            response = self.restPOST('/column/view/scope', json.dumps(view_scope), params=params)
```
thomasyu888, line 3470: Use async API call

synapseclient/evaluation.py
```diff
@@ -195,12 +199,40 @@ def getURI(cls, id):
         return '/evaluation/submission/%s/status' % id
 
     def __init__(self, **kwargs):
+        annotations = kwargs.get('submissionAnnotations')
+        if annotations is None:
+            annotations = {}
```
thomasyu888, line 204:
```suggestion
        annotations = kwargs.get('submissionAnnotations', {})
```

synapseclient/table.py
```diff
@@ -866,9 +869,142 @@ def _filter_duplicate_columns(self, syn, columns_to_add):
...
+        self.addAnnotationColumns = addAnnotationColumns and is_from_normal_constructor
+
+        if self.get('scopeIds') is None:
+            self.scopeIds = []
+
+        # add the scopes last so that we can append the passed in scopes to those defined in properties
+        if scopes is not None:
+            self.add_scope(scopes)
+
+    def add_scope(self, entities):
```
thomasyu888, line 933: There is a number of functions that repeat in `SubmissionViewSchema` that appear in `EntityViewSchema`. Should there be a `ViewBase` class?

synapseclient/client.py
```diff
@@ -3444,27 +3443,67 @@ def _build_table_download_file_handle_list(self, table, columns):
...
 
+    def _get_default_view_columns(self, view_type, view_type_mask=None):
+        """Get default view columns"""
+        uri = f"/column/tableview/defaults?viewEntityType={view_type}"
+        if view_type_mask:
+            uri += f"&viewTypeMask={view_type_mask}"
+        return [Column(**col)
+                for col in self.restGET(uri)['list']]
+
     @memoize
```
thomasyu888, line 3454: Not sure what to do with the `memoize` functions.  But i refactored the functions that get default view columns and the existing annotation headers for views.

synapseclient/client.py
```diff
@@ -3444,27 +3443,67 @@ def _build_table_download_file_handle_list(self, table, columns):
...
+        """Get submission view columns"""
+        columns = self._get_annotation_view_columns(scope_ids, "submissionview")
+        return columns
+
+    def _get_annotation_entity_view_columns(self, scope_ids, view_type_mask):
+        """Get entity view columns"""
+        columns = self._get_annotation_view_columns(scope_ids, "entityview",
+                                                    view_type_mask=view_type_mask)
+        return columns
+
```
thomasyu888, line 3506: Not sure these are required as I can just call `_get_annotation_view_columns`

synapseclient/evaluation.py
```diff
@@ -182,6 +187,28 @@ def deleteURI(self):
         return '/evaluation/submission/%s' % self.id
 
 
+def _convert_to_annotation_cls(
```
thomasyu888, line 190: Not sure if this function is necessary

synapseclient/evaluation.py
```diff
@@ -195,12 +222,33 @@ def getURI(cls, id):
         return '/evaluation/submission/%s/status' % id
 
     def __init__(self, **kwargs):
+        annotations = kwargs.get('submissionAnnotations', {})
```
thomasyu888, line 225: Since the `__init__` will most likely be filled in by the rest call, it will definitely have synapse style annotations.  Unsure if the conversion is necessary here?

synapseclient/evaluation.py
```diff
@@ -195,12 +222,33 @@ def getURI(cls, id):
...
+
         super(SubmissionStatus, self).__init__(kwargs)
 
     def postURI(self):
         return '/evaluation/submission/%s/status' % self.id
 
     def putURI(self):
+        # If not synapse annotations, turn them into synapseclient.Annotations
+        # must have id and etag to turn into synapse annotations
+        if not is_synapse_annotations(self.submissionAnnotations):
```
thomasyu888, line 243: An alternative here is to fail the store function is not a synapse annotation OR not the `Annotation` class.

synapseclient/table.py
```diff
@@ -866,9 +869,142 @@ def _filter_duplicate_columns(self, syn, columns_to_add):
         return valid_columns
 
 
+class SubmissionViewSchema(SchemaBase):
```
thomasyu888, line 872: Unsure how i should test this?

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -931,28 +931,60 @@ def test_get_unsaved_entity():
...
+
+def test_get_annotation_submission_view_columns():
+    expected = Mock()
+    with patch.object(syn, "_get_annotation_view_columns") as get_annotation:
+        get_annotation.return_value = expected
+        col = syn._get_annotation_submission_view_columns(['1234'])
+        get_annotation.assert_called_once_with(
+            ['1234'], "submissionview"
+        )
+        assert_equal(col, expected)
```
thomasyu888, line 988:
```suggestion


```

synapseclient/evaluation.py
jkiang13: I think it would be nice if the the entire time the object was in the client it was expected to be of a certain shape, in this case the Annotations typed dictionary, so I think it's useful to convert it on the deserialization from the backend.

synapseclient/evaluation.py
jkiang13: I know we talked about this on Slack, but I think I didn’t realize then that the type of the submissionAnnotations would actually be different as a side effect of the putURI (I thought it was protecting against the user setting a synapse dicitonary style submissionAnnotations attribute onto the SubmissionStatus after init but before the store).

It seems like a strange side effect, i.e. calling putURI which is presumably just a getter causes the object to morph. Could we instead override the SubmissionStatus’ json() method and put the annotation translation there? That way the SubmissionStatus itself has a consistent shape and it's only the serialization when it's washed through the store method where the translation occurs.

synapseclient/table.py
jkiang13: They seem pretty much identical other than a few constants and how the EntityView deals with the entity types/viewmask. If we weren’t dealing with a potentially breaking change if we modified the current EntityViewSchema I I would suggest they don’t even need to be separate classes if properly parameterized. But a common base or maybe just a function that spits out an appropriately parameterized class instead of the duplication would be good here.

synapseclient/table.py
jkiang13: If we can consolidate it per your suggestion above then it lessens the need for separate tests.

Adding a usage of a submission view to an integration test in test_evaluations.py would be good though.

synapseclient/evaluation.py
```diff
@@ -195,12 +222,33 @@ def getURI(cls, id):
...
+        # synapseclient.Annotations
+        kwargs['submissionAnnotations'] = _convert_to_annotation_cls(
+            id=kwargs['id'],
+            etag=kwargs['etag'],
+            values=annotations
+        )
+
         super(SubmissionStatus, self).__init__(kwargs)
 
     def postURI(self):
```
thomasyu888, line 237: This rest call actually doesn't even exist....

synapseclient/evaluation.py
thomasyu888: I think I understand this.  But the issue is that there is nothing stopping people from doing:

```
sub_status = syn.getSubmissionStatus(123455)
sub_status.submissionAnnotations = {"foo": "bar"}
```

In this scenario above, I think you also convert dictionary annotations into `synapseclient.annotations` for entities.  I will overload the `json` function, but I'm wondering if we should throw an error if people do specify non `synapseclient.Annotation` object in `submissionAnnotation` or should I add this chunk of code in to allow for the example I posted.

synapseclient/table.py
thomasyu888: I think I prefer to have `EntityViewSchema` and `SubmissionViewSchema` just because they are technically two different types.  Submissions aren't entities.  I will create a `ViewBase` class.

synapseclient/client.py
```diff
@@ -3444,27 +3443,67 @@ def _build_table_download_file_handle_list(self, table, columns):
...
+                uri='/column/view/scope/async',
+                request=view_scope
+            )
             columns.extend(Column(**column) for column in response['results'])
             next_page_token = response.get('nextPageToken')
             if next_page_token is None:
                 break
         return columns
 
+    def _get_annotation_submission_view_columns(self, scope_ids):
```
thomasyu888, line 3496: I dont think there is a need for this function and we can just go ahead and depcrecate `_get_annotation_entity_view_columns`?

synapseclient/evaluation.py
jkiang13: Oh, I was thinking a side effect-less overload of json rather than shifting the side effects to json(). What if we did something like the below. Then the SubmissionStatus object itself never changes unexpectedly as a side effect of storing it, just the serialized dictionary that is stored is customized for the format needed by the backend.

```python
    def json(self, ensure_ascii=True):
        json_dict = self

        if not is_synapse_annotations(self.submissionAnnotations):
            json_dict = self.copy()

            annotations = _convert_to_annotation_cls(
                id=self.id, etag=self.etag,
                values=self.submissionAnnotations
            )

            # Turn into synapse annotation
            json_dict['submissionAnnotations'] = to_synapse_annotations(
                annotations
            )

        return json.dumps(json_dict, sort_keys=True, indent=2, ensure_ascii=ensure_ascii)
```

thomasyu888: I did do that

thomasyu888:
Its also irritating that when you do add a column:

```
test_column = syn.store(synapseclient.Column(name='testing', columnType='STRING'))
schema.addColumn(test_column)
schema = syn.store(schema, forceVersion=False)
```

You must have the `forceVersion=False` flag on, because there is a different way of versioning Table/Views

synapseclient/table.py
```diff
@@ -702,7 +702,100 @@ def __init__(self, name=None, columns=None, parent=None, properties=None, annota
...
+        # get the default EntityView columns from Synapse and add them to the columns list
+        additional_columns = []
+        if self._synapse_entity_type == "org.sagebionetworks.repo.model.table.SubmissionView":
+            view_type = "submissionview"
+            mask = None
+        elif self._synapse_entity_type == "org.sagebionetworks.repo.model.table.EntityView":
+            view_type = "entityview"
+            mask = self['viewTypeMask']
+        else:
+            raise ValueError("_synapse_entity_type not set")
```
jkiang13, line 775: Sorry I thought I mentioned this previously on this PR but I guess I only did in Slack. It seems like we could make this so that the super class doesn't know about the specific view type. I think these lines could be replaced with something like the below or alternately some additional abstract properties defined in the subclasses?

```
        view_type = self._synapse_entity_type.split(".")[-1].lower()
        mask = self.get('viewTypeMask')
```

tests/integration/synapseclient/test_evaluations.py
```diff
@@ -128,18 +127,20 @@ def test_evaluations():
...
-            for kvp in status.annotations['longAnnos']:
-                if kvp['key'] == 'bogosity':
-                    assert_false(kvp['isPrivate'])
+            # a = from_submission_status_annotations(status.annotations)
+            a = status.submissionAnnotations
+            assert_equals(a['foo'], ['bar'])
+            assert_equals(a['bogosity'], [bogosity[submission.id]])
+            # for kvp in status.annotations['longAnnos']:
+            #     if kvp['key'] == 'bogosity':
+            #         assert_false(kvp['isPrivate'])
```
jkiang13, line 143: Seems like we can just remove these lines (and above).

tests/integration/synapseclient/test_evaluations.py
```diff
@@ -128,18 +127,20 @@ def test_evaluations():
             bogosity[submission.id] = b
             a = dict(foo='bar', bogosity=b)
             b += 123
-            status['annotations'] = to_submission_status_annotations(a)
-            set_privacy(status['annotations'], key='bogosity', is_private=False)
+            # status['annotations'] = to_submission_status_annotations(a)
+            # set_privacy(status['annotations'], key='bogosity', is_private=False)
```
jkiang13, line 131: Remove commented lines for older annotations?

tests/integration/synapseclient/test_evaluations.py
```diff
@@ -110,13 +110,13 @@ def test_evaluations():
...
-            status.score = random.random()
+            # status.score = random.random()
             if submission['name'] == 'Submission 01':
                 status.status = 'INVALID'
-                status.report = 'Uh-oh, something went wrong!'
+            #    status.report = 'Uh-oh, something went wrong!'
             else:
                 status.status = 'SCORED'
-                status.report = 'a fabulous effort!'
+            #    status.report = 'a fabulous effort!'
```
jkiang13, line 118: Are these fields and above no longer used?

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -931,27 +931,60 @@ def test_get_unsaved_entity():
...
+#     with patch.object(syn, "_get_default_view_columns") as mock_get:
+#         syn._get_default_entity_view_columns(mask)
+#         mock_get.assert_called_with("entityview",
+#                                     view_type_mask=mask)
+
+
+# def test_get_default_submission_view_columns():
+#     with patch.object(syn, "_get_default_view_columns") as mock_get:
+#         syn._get_default_submission_view_columns()
+#         mock_get.assert_called_with("submissionview")
```
jkiang13, line 945: Are these finished?

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -931,27 +931,60 @@ def test_get_unsaved_entity():
...
+
+# def test_get_annotation_submission_view_columns():
+#     expected = Mock()
+#     with patch.object(syn, "_get_annotation_view_columns") as get_annotation:
+#         get_annotation.return_value = expected
+#         col = syn._get_annotation_submission_view_columns(['1234'])
+#         get_annotation.assert_called_once_with(
+#             ['1234'], "submissionview"
+#         )
+#         assert_equal(col, expected)
```
jkiang13, line 987: Are these finished?

tests/integration/synapseclient/test_evaluations.py
thomasyu888: Ya, `status.report` is now deprecated

synapseclient/table.py
thomasyu888: Do i need to raise a ValueError here?

PR #760, opened 2020-06-19
synapseclient/reticulate documentation

jkiang13:
Examples in vignette style for using the Python synapseclient with R through reticulate.

As rendered:
[synapseclient_reticulate.pdf](https://github.com/Sage-Bionetworks/synapsePythonClient/files/4806205/synapseclient_reticulate.pdf)


thomasyu888:
This would be an awesome thing to review at the code_review meeting

jkiang13:
virutalenv conda instructions

jkiang13:
Table asDataFrame issues

jkiang13:
other use cases dependencies, e.g. shiny apps with multiple users

jkiang13:
Using it in the context of an R package that is separately packaged, ensure things are loaded properly.

brucehoff:
I suggest adding a section on how to use with Shiny.  In particular, it's important that the Synapse client has authenticated as the user making a request to the Shiny app'. (If the app' author fails to do this then users can end up accessing each other's Synapse content.)  An example by @karawoo :
https://github.com/Sage-Bionetworks/dccvalidator/blob/aade4ace5a3f7e11ff289c74444f32c8745b5306/R/app-server.R#L34

docs/reticulate.rst
```diff
@@ -0,0 +1,517 @@
...
+
+Annotating Synapse Entities
+===========================
+
+This illustrates adding annotations to a Synapse entity.
+
+  .. code-block::
+
+    # first retrieve the existing annotations object
+    annotations = syn$get_annotations(project)
```
brucehoff, line 209: In R it's customary to use "<-" rather than "=" for assignment.

docs/reticulate.rst
```diff
@@ -0,0 +1,517 @@
...
+
+  .. code-block::
+
+    filePath <- tempfile()
+    connection <- file(filePath)
+    writeChar("some test", connection, eos = NULL)
+    close(connection)
+
+    file = synapseclient$File(filePath, name="provenance_file.txt", parent=project)
+    file <- syn$store(, activity = act)
```
brucehoff, line 241: Is this supposed to be `file <- syn$store(file, activity = act)` ?

docs/reticulate.rst
```diff
@@ -0,0 +1,517 @@
...
+
+    # build a Synapse table from the data frame.
+    # a schema is automatically generated
+    # note that reticulate will automatically convert from an R data frame to Pandas
+    table <- synapseclient$build_table("My Favorite Genes", project, genes)
+
+    table <- syn$store(table)
+
+Alternately the schema can be specified. At this time when using date values it is necessary
+to use string or millisecond values and explicitly specify the type in the schema due to how
```
brucehoff, line 275: by "string or millisecond values" did you mean "millisecond values either as string or integer"?

docs/reticulate.rst
```diff
@@ -0,0 +1,517 @@
...
+    submissions <- syn$getSubmissionBundles(eval)
+
+    # submissions are returned as a generator
+    list(iterate(submissions))
+
+Retrieving submission by id:
+
+  .. code-block::
+
+    # Not evaluating this section because of SYNPY-235
```
brucehoff, line 393: https://sagebionetworks.jira.com/browse/SYNPY-235 is closed.  Can we remove this line?

docs/reticulate.rst
```diff
@@ -0,0 +1,517 @@
...
+  .. code-block::
+
+    submissionStatus <- syn$getSubmissionStatus(submission)
+    submissionStatus
+
+Query an evaluation:
+
+  .. code-block::
+
+    queryString <- sprintf("query=select * from evaluation_%s LIMIT %s OFFSET %s'", eval$id, 10, 0)
```
brucehoff, line 407: We should update this, as well as the Python doc's, to use the new Evaluation View queries.

brucehoff:
Good work.  It will be interesting to see if this only complements or completely replaces synapser as the way to access Synapse from R.

PR #761, opened 2020-06-22
SYNPY-1075 handle Submission entityBundleJSON w/ v2 annotations

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1075

In stack 316 archived entityBundleJSON on a Submission is saved/returned using v2 annotations. This allows the client to handle either form, as either form might be returned depending on when the submission was made and the entity JSON was serialized.

### DIFF
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -339,17 +339,28 @@ def convert_old_annotation_json(annotations):
     """Transforms a parsed JSON dictionary of old style annotations
     into a new style consistent with the entity bundle v2 format."""
 
-    converted = {k: v for k, v in annotations.items() if k in ('id', 'etag')}
-    converted_annos = converted['annotations'] = {}
+    meta_keys = ('id', 'etag')
 
     type_mapping = {
         'doubleAnnotations': 'DOUBLE',
         'stringAnnotations': 'STRING',
         'longAnnotations': "LONG",
         'dateAnnotations': 'TIMESTAMP_MS',
-
-        # TODO what to do with blobAnnotations
     }
+
+    annos_v1_keys = set(meta_keys) | set(type_mapping.keys())
+
+    # blobAnnotations appear to be little/unused and there is no mapping defined here but if they
+    # are present on the annos we should treat it as an old style annos dict
+    annos_v1_keys.add('blobAnnotations')
+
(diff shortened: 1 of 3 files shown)


synapseclient/annotations.py
```diff
@@ -339,18 +339,29 @@ def convert_old_annotation_json(annotations):
...
         'dateAnnotations': 'TIMESTAMP_MS',
-
-        # TODO what to do with blobAnnotations
     }
 
+    annos_v1_keys = set(meta_keys) | set(type_mapping.keys())
+
+    # blobAnnotations appear to be little/unused and there is no mapping defined here but if they
+    # are present on the annos we should treat it as an old style annos dict
+    annos_v1_keys.add('blobAnnotations')
```
zimingd, line 355: blob annotations are completely removed now and it will never be returned from the backend

jkiang13: Could they be returned in old entityBundleJSON that was serialized at the time of a submission that occurred long ago? Or were they just never used at all?

thomasyu888: Will these still exist for older submissions?

zimingd: I did a scan of the DB before removing that from the API and there was actually 0 bytes of blob annotations in our database.

jkiang13: @thomasyu888 is right, looks like older submission may have the serialized blobAnnotations key even if it doesn't have any actual blobAnnotations. Also I added the creationDate and uri system properties which could also be there.

zimingd: We still accept it for backwards compatibility, but the backend will ignore it and it won't be stored nor returned as a JSON key on GETs.

What you're seeing is a result of the python client still using the Annotations v1 representation

zimingd: @jkiang13 there was never used at all and will never be returned by the back-end. Before I removed blobannotations from the API, I used a script to scan the serialized data in the database to verify no blob-annotations were being used by anyone.

zimingd:
@thomasyu888 @jkiang13 
`uri` will never be returned from our APIs as well. I think this is the python client formulating it and inserting it for the user. If it ever gets sent to the backend it will be ignored anyways.

PR #762, opened 2020-06-23
SYNPY-1076 unpin specific dependency versions

jkiang13:
This is to address [SYNPY-1076](https://sagebionetworks.jira.com/browse/SYNPY-1076) which results from a stale pandas dependency breaking when compiled against the latest released version of numpy.

But rather than just upgrade the pinned version this unpins specific package version dependencies instead and allows the installed version to float above a minimum.

My thoughts roughly mirror [these](https://stackoverflow.com/a/44938662). In a controlled deployment environment where all dependencies (including downstream dependencies) can be frozen by an environment manager such as pipenv, then it makes sense to pin all specific versions. But synapseclient is a library that is installed and used in other applications and environments and typically installed through a simple pip install. We cannot guarantee what dependencies might already exist or be needed in that environment, and being unnecessarily specific means we might overwrite a specific version that actually is needed in an environment.

Note that this is a different approach than that expressed in [SYNPY-791](https://sagebionetworks.jira.com/browse/SYNPY-791).

Furthermore since synapseclient is typically installed via pip install, the fact that we pin some versions can result in some unstable/untested combinations of downstream dependencies that we can't control unless we also define and pin the downstream dependencies of our dependencies, an immediate example of this being this issue. In these cases it's better to let contemporary versions be installed as they will have been tested with their own contemporary dependencies.

Without pinned versions we do become vulnerable to a breaking change in a new library release, but we are already vulnerable to that to some degree as illustrated above, and we don't foster the use of stale dependencies that may have bugs or security issues, or have stale versions of dependencies being installed alongside new versions of downstream dependencies as in this issue.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -38,13 +38,13 @@ setuptools.setup(
     python_requires='>=3.6.*',
     install_requires=[
         'requests>=2.22.0',
-        'keyring==12.0.2',
-        'deprecated==1.2.4',
+        'keyring>=12.0.2',
+        'deprecated>=1.2.4',
     ],
     extras_require={
-        'pandas': ["pandas==0.25.0"],
+        'pandas': ["pandas>=0.25.0"],
         'pysftp': ["pysftp>=0.2.8"],
-        'boto3': ["boto3"],
+        'boto3': ["boto3>=1.7.0"],
         'tests': test_deps,
         ':sys_platform=="linux2" or sys_platform=="linux"': ['keyrings.alt==3.1'],
     },
--- a/setup.py
+++ b/setup.py
@@ -38,7 +38,7 @@ setuptools.setup(
     python_requires='>=3.6.*',
     install_requires=[
         'requests>=2.22.0,<3.0',
-        'keyring>=12.0.2',
+        'keyring==12.0.2',
         'deprecated>=1.2.4,<2.0',
     ],
     extras_require={
--- a/setup.py
+++ b/setup.py
@@ -37,14 +37,14 @@ setuptools.setup(
     # requirements
     python_requires='>=3.6.*',
     install_requires=[
-        'requests>=2.22.0',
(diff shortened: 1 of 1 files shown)


thomasyu888:
Also fixes SYNPY-791

setup.py
```diff
@@ -37,14 +37,14 @@
     # requirements
     python_requires='>=3.6.*',
     install_requires=[
-        'requests>=2.22.0',
-        'keyring==12.0.2',
-        'deprecated==1.2.4',
+        'requests>=2.22.0,<3.0',
+        'keyring>=12.0.2',
```
zimingd, line 41: Some later versions of keying had breaking changes, which is why we required a specific version we knew worked. I would recommend verifying that the latest version works as expected on Windows, Linux (We used Ubuntu distro for testing previously), and Mac before requiring the latest version.

Specifically on Linux, there's no default credential storage, unlike Mac and Windows, so we would also need to verify that it can work (defaulting to a text file storage) on ones without a keychain (I believe Ubuntu comes with GNOME keyring). This might occur on something like a EC2 AmazonLinux instance or a minimal linux install. 

We've also had issues with headless connections (e.g. SSH) on some version of keyring version >12 when a cred storage was available on the Linux machine, but the user was unable to unlock the cred store w/out a pop-up GUI password prompt. I'm unsure if this was addressed in the latest version of keyring.

jkiang13: I've re-pinned keyring given that it may have issues beyond what is covered in unit and integration tests and created an issue to investigate upgrading it in the future. https://sagebionetworks.jira.com/browse/SYNPY-1077

keyring 12.0.2 has no downstream dependencies that aren't already maxxed out (SecretStorage < 3 on Linux) so at least its downstream dependencies shouldn't shift underneath it if it is pinned.

PR #763, opened 2020-06-30
SYNPY-1078 syncFromSynapse manifest performance

jkiang13:
syncFromSynapse when synced to an external location outside the cache generates manifest file(s) in the synced folder structure. For a deep folder structure there are performance issues related to interaction with the recursion and the manifest generation that are more completely described here:
https://sagebionetworks.jira.com/browse/SYNPY-1078

This speeds things up by changing the following:
1. A manifest file that is written to each subdirectory now only includes the files that are rooted in that subdirectory, rather than all files synced so far. The previous behavior of writing those files that happen to have been synced so far doesn't make much sense and seems to be a byproduct of the recursive implementation of the function rather than intentional behavior. It's not clear to me that we actually even want a manifest at anything other than the root of the sync, although this preserves that for now (with the more sensible set of files).
2. Caching the provenance fetches for each file.  When we write each manifest row we fetch the provenance via a call to Synapse. Because we include each file in potentially multiple manifests depending on how deeply it is located in the directory structure, this meant multiple provenance fetches for each file. This caches the provenance data so the provenance for each entity is only fetched once per sync.

The above two issues previously interacted with each other to result in severe slowdowns with all of the repeated provenance fetches.

Also adds a utility decorator that can be used when a keyword parameter is deprecated from a remaining public signature.

### DIFF
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -25,6 +25,7 @@ import uuid
 import importlib
 import numbers
 import urllib.parse as urllib_parse
+import warnings
 
 
 UNIX_EPOCH = datetime.datetime(1970, 1, 1, 0, 0)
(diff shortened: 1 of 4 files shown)


jkiang13:
pep8speaks isn't handling the flake8 extend-ignore directive in setup.cfg. This will be fixed separately in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/764 which replaces the dynamically added module variables in nose fixtures and eliminates the need for the ignore.

PR #764, opened 2020-07-08
SYNPY-1085 pytest

jkiang13:
Replaces nose with pytest

nose (as opposed to nose2) has been unmaintained for years. In switching to a new test framework, between pytest and nose2, pytest is far more widely used and frequently updated (6k GH stars vs 600).

Assertions were converted automatically using https://pypi.org/project/nose2pytest/ where syntactically supported, other changes were manual.

Previous module "syn" and "project" fixtures wired in nose's setup_module function are moved to pytest session fixtures as defined in separate [conftest](https://www.tutorialspoint.com/pytest/pytest_conftest_py.htm) files for each unit and integration tests.

The only new functionality in this PR that is not a port is a change to ensure that there are no remote connections made during unit tests. Previously unit tests were making some outbound requests (mostly version checks implicit when creating a Synapse object). Now a pytest fixture enforces that any socket connection will raise an error (it is assumed that any remote calls in a unit test should be mocked, or otherwise replaced with an integration test), and existing sources of requests are suppressed (e.g. `Synapse(skip_checks=True)`)


tests/unit/conftest.py
```diff
@@ -9,6 +11,35 @@
...
+    def block_socket_side_effect(*args, **kwargs):
+        raise ValueError(_BLOCKED_CONNECTION_MESSAGE)
+
+    block_socket_patcher = mock.patch(
+        'socket.socket',
+        side_effect=block_socket_side_effect
+    )
+    block_socket_patcher.start()
+
+    request.addfinalizer(block_socket_patcher.stop)
```
jkiang13, line 34: This is a new fixture in this PR, not ported from a previous nose fixture.

PR #768, opened 2020-07-14
SYNPY-1082: urlretrieve Use urllib_request

thomasyu888:
`urlparse.urlretrieve` doesn't exist anymore.


thomasyu888:
Should probably have a test for this.

jkiang13:
Thanks. Given that this broke without any test failures we should have a test for it (I created a PR against your fork branch if you'd like to use that one).

PR #772, opened 2020-07-23
SYNPY-1084 allow anonymous calls

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1084

Per SYNPY-1084 we always enforce that calls made to Synapse are from a logged in client, even for endpoints that don't require auth. This removes the client preemptively raising a SynapseAuthenticationError with a logged out client and instead allows the call to go through. If the call returns a 401 or 403 status code then the client will then raise a SynapseAuthenticationError chained from the error raised as a result of the call.

Also unifies the implementation of the various Synapse#rest methods which were largely duplicative so that they could be more easily tested.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -3620,6 +3620,17 @@ class Synapse(object):
 
             raise
 
+    def _rest_call(self, method, uri, data, endpoint, headers, retryPolicy, requests_session, **kwargs):
+        uri, headers = self._build_uri_and_headers(uri, endpoint, headers)
+        retryPolicy = self._build_retry_policy(retryPolicy)
+        requests_session = requests_session or self._requests_session
+
+        requests_method_fn = getattr(requests_session, method)
+        response = with_retry(lambda: requests_method_fn(uri, data=data, headers=headers, **kwargs),
+                              verbose=self.debug, **retryPolicy)
+        self._handle_synapse_http_error(response)
+        return response
+
     def restGET(self, uri, endpoint=None, headers=None, retryPolicy={}, requests_session=None, **kwargs):
         """
         Sends an HTTP GET request to the Synapse server.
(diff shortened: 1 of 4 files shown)


synapseclient/client.py
```diff
@@ -3591,18 +3591,16 @@ def sendMessage(self, userIds, messageSubject, messageBody, contentType="text/pl
     #                   Low level Rest calls                   #
     ############################################################
 
-    def _generateSignedHeaders(self, url, headers=None):
+    def _generateHeaders(self, url, headers=None):
```
thomasyu888, line 3594: Any reason for the camel case?  If the function is being renamed, should we just do `_generate_headers`

jkiang13: Good point.

tests/integration/synapseclient/test_evaluations.py
```diff
@@ -188,12 +188,15 @@ def test_teams():
     team = syn.store(Team(name=name, description="A fake team for testing..."))
     schedule_for_cleanup(team)
 
-    found_team = syn.getTeam(team.id)
+    # not logged in, teams are public
+    anonymous_syn = Synapse()
```
thomasyu888, line 192: Is a "not signed in error" thrown when trying to use a function that requires authentication?

jkiang13: This is pretty much why I marked this as a draft so far and not done yet. Right now you will get the error that the back end serves, which may not be user friendly as the current behavior of telling the user to log in.

I think the options are:
1. Just surface the back end error (as it is now).
2. On any 403 status code when the client is not logged in raise a SynapseAuthenticationError indicating you are not logged in (although I think there are endpoints which may not return a 403 even when the underlying cause is auth, for example the batch entity API I think will return a 201 always with individual failed responses, and it's also a bit specious because you might have tried to access something you don't have permission to even when logged in).
3. Keep a registry of known endpoints in the client that do not require auth, and raise an error before submitting the request if the endpoint is in that registry.

Let me know if you have suggestions per the above or others.

jkiang13: FWIW I went with #2 of the above in subsequent commits.

Issue #767, opened 2020-07-12
Allow overriding the cache location via Synapse() constructor


keller-mark:
# Bug Report

## Operating system

macOS

## Client version
2.1.0

## Description of the problem

*Provide a description of the problem, and if possible a minimal reproducible example.*

I would like to download one dataset to one directory, and a second dataset to a different directory, but these need to be downloaded using the `syn.downloadTableColumns` function.

This function automatically downloads to the cache location, so this is not possible without updating the config file in between.

It would be great if the cache location could be set in the `Synapse()` constructor directly.

jkiang13:
@keller-mark Thanks for the suggestion. So I understand the general need, you'd like to be able to use `downloadTableColumns` to download to an alternate path than the default cache. Would something like an optional `downloadLocation` parameter on `downloadTableColumns` similar to `syn.get` also satisfy this requirement?

In the meantime, something like the following utility method might be useful as a workaround:

```python
import configparser
import synapseclient
import tempfile


def synapse_custom_cache_dir(cache_dir):
    """Return a new Synapse instance with the given cache directory as its configured cache root"""

    config = configparser.RawConfigParser()
    config.read(synapseclient.client.CONFIG_FILE)

    if not config.has_section('cache'):
        config.add_section('cache')
    config['cache']['location'] = cache_dir

    with tempfile.NamedTemporaryFile(mode='w') as tempconfig:
        config.write(tempconfig)
        tempconfig.flush()

        return synapseclient.Synapse(configPath=tempconfig.name)
```

Sample usage:

```python
syn = synapse_custom_cache_dir('/path/to/location1')
syn.login()
results = syn.tableQuery('SELECT * FROM syn123 LIMIT 5')
syn.downloadTableColumns(results, ['foo', 'bar'])

syn = synapse_custom_cache_dir('/path/to/location2')
syn.login()
results = syn.tableQuery('SELECT * FROM syn456 LIMIT 5')
syn.downloadTableColumns(results, ['foo', 'bar'])
```

keller-mark:
A downloadLocation parameter would be great. Thanks for the workaround.

PR #775, opened 2020-07-30
SYNPY-1088 download table query csv to download location

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1088

Currently when we download the files associated with a table query when running e.g. `synapse get -q "select id from syn123"` we download the csv to the synapse cache, and the files to the working directory or downloadLocation if specified.

Per the JIRA issue there is a desire to have the CSV file available in the download location, similar to how we output the manifest of a `synapse get -r` command.

Also resolves  https://sagebionetworks.jira.com/browse/SYNPY-1081, adds optional downloadLocation arg to Synapse.downloadTableColumns which should resolve https://github.com/Sage-Bionetworks/synapsePythonClient/issues/767


thomasyu888:
Looks good

kelshmo:
Looks good!

jkiang13:
> Looks good, are we going to change from using `downloadLocation` to `download_location` at some point?

We can't break it immediately obviously, we can transition this and other args perhaps over several versions. I added this decorator a while back with the plan to use it start doing that https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/utils.py#L894, but at the moment in interferes with Sphinx autodoc and so can't be used on any function that we autodoc.

PR #776, opened 2020-08-19
SYNPY-1090 Allow use of api key with command line client

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1090

Allows an api key to be used where the password is in the command line client. It is used in place of a password instead of as a separate command line argument in order to simplify things for the user, in particular when inputting the password from the console input prompt  (don't want them to have to specify which they are using or enter an empty password followed by an api key for example).

### DIFF
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -6,25 +6,26 @@ Utility Functions
 Utility functions useful in the implementation and testing of the Synapse client.
 
 """
-import collections.abc
-import os
-import sys
-import hashlib
-import re
+import base64
 import cgi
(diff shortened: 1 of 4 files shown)


synapseclient/core/utils.py
```diff
@@ -6,25 +6,26 @@
 Utility functions useful in the implementation and testing of the Synapse client.
 
 """
-import collections.abc
-import os
-import sys
-import hashlib
-import re
+import base64
```
jkiang13, line 9: base64 is a new import the rest are just reordered alphabetically

PR #781, opened 2020-09-25
SYNPY-1063 GitHub Actions build

jkiang13:
Uses GitHub Actions to replace the following:

1. 	Travis unit/integration tests on Linux/Mac (py 3.6, 3.7, 3.8)
2. 	AppVeyor unit/integration tests on Windows (py 3.6, 3.7, 3.8)
3. 	Jenkins prod and staging deployment jobs (python_prod_deploy, python_stagng_deploy, python_pypi_check)

All branch pushes and pull requests will trigger unit tests (integration tests will run if the appropriate secrets are available, which they will not be in PRs from forks).

Release events trigger deployments to test.pypi.org or pypi.org or for Releases marked as prerelease or not, respectively.

(Note this PR shows AppVeyor tests failing since the AppVeyor config is removed). The AppVeyor integration would be disabled after this is merged.


.github/workflows/build.yml
```diff
@@ -0,0 +1,249 @@
...
+        with:
+          python-version: 3.6
+
+      - name: set-release-env
+        shell: bash
+        run: |
+          RELEASE_TAG="${{ github.event.release.tag_name }}"
+          if [[ $RELEASE_TAG =~ ^v?([[:digit:]\.]+)(-rc)? ]]; then
+            VERSION="${BASH_REMATCH[1]}"
+            if [[ "${{ github.event.release.prerelease}}" == "true" ]]; then
```
zimingd, line 108: if -rc is found and not pre-release fail the build

.github/workflows/build.yml
```diff
@@ -0,0 +1,249 @@
...
+          if [[ -n "$VERSION" ]]; then
+            sed "s|\"latestVersion\":.*$|\"latestVersion\":\"$VERSION\",|g" synapseclient/synapsePythonClient > temp
+            rm synapseclient/synapsePythonClient
+            mv temp synapseclient/synapsePythonClient
+          fi
+
+      - id: build-package
+        shell: bash
+        run: |
+          python3 -m pip install setuptools==38.6.0
```
zimingd, line 135: are we able to use later versions?

.github/workflows/build.yml
```diff
@@ -0,0 +1,249 @@
...
+        with:
+          upload_url: https://uploads.github.com/repos/${{ github.event.repository.full_name }}/releases/${{ github.event.release.id }}/assets?name=${{ env.PACKAGE_NAME }}
+          asset_name: ${{ env.PACKAGE_NAME }}
+          asset_path: dist/${{ env.PACKAGE_NAME }}
+          asset_content_type: application/gzip
+
+      - name: deploy-to-pypi
+        shell: bash
+        run: |
+          python3 -m pip install twine==1.11.0
```
zimingd, line 166: version upgrade?

.github/workflows/build.yml
```diff
@@ -0,0 +1,249 @@
...
+
+          # it can take some time for the packages to become available in pypi after uploading
+          for i in 5 10 20 40; do
+            if pip3 install --index-url $PYPI_INDEX_URL --extra-index-url https://pypi.org/simple "synapseclient==$VERSION"; then
+              ACTUAL_VERSION=$(synapse --version)
+
+              if [ -n "$(echo "$ACTUAL_VERSION" | grep -oF "$VERSION")" ]; then
+                # found the expected version
+                exit 0
+              fi
```
zimingd, line 242: fail early if we've retrieved from pypi but version mismatch

zimingd (approved):

.github/workflows/build.yml
jkiang13: The reason for the pinned twine was that it happens to have the same keyring dependency as synapseclient, so it could be installed into the same python environment that was used to build the client package itself.

I split the package and deploy into separate jobs, the deploy job now redownloads the complete artifact rather than building it so doesn't share any dependencies, allowing the specific version to be removed.

.github/workflows/build.yml
jkiang13: Added a check confirming that -rc is present on prerelease/test deployments.

.github/workflows/build.yml
jkiang13: Done.

PR #782, opened 2020-09-29
SYNPY-1099 Replace usage of deprecated entity version endpoint

jkiang13:
Replaces usage of ```PUT /entity/{id}/version``` per [SYNPY-1098](https://sagebionetworks.jira.com/browse/SYNPY-1098)


thomasyu888 (commented):
What was the motivation behind this update?

thomasyu888 (approved):

synapseclient/client.py
```diff
@@ -3557,24 +3557,30 @@ def _updateEntity(self, entity, incrementVersion=True, versionLabel=None):
...
         if is_versionable(entity):
-            if incrementVersion or versionLabel is not None:
-                uri += '/version'
+            if versionLabel:
+                entity['versionLabel'] = str(versionLabel)
+                incrementVersion = True
+
+            elif incrementVersion:
                 if 'versionNumber' in entity:
                     entity['versionNumber'] += 1
```
zimingd, line 3577: Does setting this do anything? Can we just set label to current version + 1?
We don't seem to increment the version number if a versionLabel is provided since this is in an elif block

zimingd: If we don't need the version bump, then this logic can be simplified to deciding what label to use

jkiang13: You're right, setting the entity version number doesn't have any affect, so we don't need to do that. I modified the conditionals accordingly.

zimingd (approved):

PR #788, opened 2020-10-09
py-2.2.2 release notes

jkiang13:
Release notes for Python client 2.2.2

Can be previewed here https://jkiang13.github.io/synapsePythonClient/build/html/news.html#xx


docs/news.rst
```diff
@@ -2,6 +2,36 @@
...
 =============
 
+2.2.2 (2020-10-XX)
+==================
+
+Highlights
+----------
+
+- This version addresses an issue with downloads being retried unsuccessfully after encountering certain types of errors.
+- A `create_snapshot <index.html#synapseclient.Synapse.create_snapshot>`__ function is added for making table and view snapshots.
```
thomasyu888, line 12: I just realize that the term "snapshot" deviates from "version".  I used `create_snapshot` because the rest call said "snapshot". Thoughts?

jkiang13: The rest response uses the "snapshotVersionNumber" key, would **create_snapshot_version** be clearer?

thomasyu888: Well it's slightly confusing, because "snapshots" isn't a term that is used widely in Synapse.  When you upload a new file, you create a new version.  We are creating table versions here, but we are calling them snapshots.  I wanted to call them `create_table_version` but then that wouldn't encapsulate views, so I just stuck with `create_snapshot` (also to mimic the rest call).  Thoughts @kelshmo ?

thomasyu888: Perhaps just adding to the documentation saying that `create_snapshot` is essentially creating versions of table or view  ?

kelshmo: I think this is an important point - I do think `create_snapshot_version` is more clarifying than `create_snapshot`.

docs/build/html/Client.html
```diff
@@ -206,6 +206,28 @@ <h2>Synapse<a class="headerlink" href="#synapse" title="Permalink to this headli
...
+<dd><p>Creates Table or EntityView snapshots</p>
```
kelshmo, line 212:
```suggestion
<dd><p>Create a new Table Version or a new View version.</p>
```

kelshmo: Verbiage straight from the UI.

kelshmo:
I tested both tables and views - works beautifully! I suggest updating the description to be harmonious with the UI and perhaps renaming the function to `create_snapshot_version` so there is less ambiguity.

docs/news.rst
jkiang13: Renamed.

thomasyu888 (approved):

PR #791, opened 2020-10-27
SYNPY-1102 - filter out empty string annotations on syncToSynapse

jkiang13:
syncToSynapse will parse and set an empty string as an annotation value from a manifest TSV file if an upload row doesn't contain a value for that annotation. We'd prefer that on a manifest sync it set no annotation for that key if there is no value rather than an empty string.

https://sagebionetworks.jira.com/browse/SYNPY-1102

### DIFF
--- a/tests/unit/synapseutils/unit_test_synapseutils_sync.py
+++ b/tests/unit/synapseutils/unit_test_synapseutils_sync.py
@@ -364,6 +364,14 @@ def test_manifest_upload(syn):
         'anno_2': ['v3', 'v4', ''],
     }
 
+    # any empty annotations that result from any empty csv column should not
+    # be included in the upload
+    expected_anno_data = [
+        {'anno_2': 'v3'},
+        {'anno_1': 'v1', 'anno_2': 'v4'},
+        {'anno_1': 'v2'}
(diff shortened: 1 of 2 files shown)


tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -351,6 +351,53 @@ def test_extract_file_entity_metadata__ensure_correct_row_metadata(syn):
...
+        expected_name = data['name'][i]
+        expected_used = data['used'][i]
+        expected_executed = data['executed'][i]
+
+        expected_annos = {}
+        expected_annos.update({'anno_1': data['anno_1'][i]})
+        expected_annos.update({'anno_2': data['anno_2'][i]})
+
+        # empty strings values should not be uploaded as annotations
+        expected_annos = {k: v for k, v in expected_annos.items() if v != ''}
```
zimingd, line 386: This test generates expected annotations in the same way as the code being tested. Prefer an explicit dict.

Copy the `data` dict and overwrite keys/value pairs that should have changed

zimingd (changes requested):

PR #800, opened 2021-01-20
SYNPY-885 cache root dir

linchiahuisage:
https://sagebionetworks.jira.com/browse/SYNPY-885
this PR allows user to customized the cahe root directory, or keep it as the default directory same as core.Cache

### DIFF
--- a/tests/unit/synapseclient/unit_test_client.py
+++ b/tests/unit/synapseclient/unit_test_client.py
@@ -2547,29 +2547,20 @@ class TestTableQuery:
             assert (mock_download_result, expected_path) == actual_result
 
 
-class TestInitCachePath:
+def test_init_change_cache_path():
+    cache_root_dir = '.synapseCache'
+    silent_logger_name = 'synapseclient_silent'
+    fanout = 1000
+    file_handle_id = '-1337'
 
-    @pytest.fixture(autouse=True, scope='function')
-    def init_syn(self, syn):
-        self.syn = syn
+    syn = Synapse(debug=False, skip_checks=True)
+    expected_cache_path = os.path.join(str(Path.home()), cache_root_dir,
+                                       str(int(file_handle_id) % fanout), str(file_handle_id))
+    assert syn.cache.get_cache_dir(file_handle_id) == expected_cache_path
 
-    def setup(self):
-        self.cache_root_dir = '.synapseCache'
-        self.fanout = 1000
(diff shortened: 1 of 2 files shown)


tests/unit/synapseclient/unit_test_client.py
```diff
@@ -2543,3 +2544,21 @@ def test_query_table_csv(self, downloadLocation, syn):
             )
 
             assert (mock_download_result, expected_path) == actual_result
+
+
+def test_init_change_cache_path():
+    cache_root_dir = '.synapseCache'
```
jkiang13, line 2550: It's good to have a docstring as the first line of a test that describes what it is doing.

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -2543,3 +2544,21 @@ def test_query_table_csv(self, downloadLocation, syn):
...
+    fanout = 1000
+    file_handle_id = '-1337'
+
+    syn = Synapse(debug=False, skip_checks=True)
+    expected_cache_path = os.path.join(str(Path.home()), cache_root_dir,
+                                       str(int(file_handle_id) % fanout), str(file_handle_id))
+    assert syn.cache.get_cache_dir(file_handle_id) == expected_cache_path
+
+    temp_dir = tempfile.TemporaryDirectory()
```
jkiang13, line 2559: Since this creates a temporary directory, we should clean it up after the test. We can delete after it is used with the cleanup method, or by using it as a context manager as described in the documentation here: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

e.g.
```
with tempfile.TemporaryDirectory() as temp_dir_name:
  # use temp_dir_name
```

A context manager assures that it is deleted no matter how the function exits.

jkiang13 (changes requested):

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -2547,6 +2547,10 @@ def test_query_table_csv(self, downloadLocation, syn):
 
 
 def test_init_change_cache_path():
+    """
+    Verify that the user can customize the cache path.
+    The cache path would set to be default value if cache_root_dir argument is None.
```
jkiang13, line 2552: _The cache path is set to the default value if cache_root_dir argument is None._

PR #803, opened 2021-01-28
Fix invalid version constraint on sphinx-argparse

BrunoGrandePhD:
I get the following warning when installing `synapseclient` using Poetry. I think the one-line change in this PR addresses the issue. 

```
PackageInfo: Invalid constraint (sphinx-argparse<.3,>=0.2 ; extra == "docs") found in synapseclient-2.2.2 dependencies, skipping
```


jkiang13:
Interesting, I've not seen that issue elsewhere, but looks good. Thanks.

BrunoGrandePhD:
Thanks, @jkiang13! I think Poetry is just picky about version formatting.

PR #804, opened 2021-01-29
SYNPY-1058 auth token support

jkiang13:
Supports logging into the python client using a bearer token such as a personal access token.

Programmatically the token can be passed during login instead of the password or apiKey, e.g.:

`synapseclient.login(<username>, authToken='<token>')`

From the command line an authToken can be passed as if it were a password (i.e. using the -p parameter or by entering it via stdin when prompted), similar to the current behavior of accepting either the password or api key.

`synapse login -u <username> -p <token>`

The SynapseCredentials object is refactored into a requests.auth.AuthBase implementation so that it can be used directly with the Python requests **auth** parameter where needed.

The "rememberMe" when using an authToken will store the token in the keyring directly (it will not store the api key obtained after fetching it from the services as happens when "rememberMe" is used with a password). It stores the token under a different key than the api key. The logic for storing into the keyring is moved to the credential implementation rather than maintained separately with the cached_sessions logic.


synapseclient/__main__.py
```diff
@@ -983,22 +983,39 @@ def login_with_prompt(syn, user, password, rememberMe=False, silent=False, force
...
+    # which type of secret we had, but the alternative of defining separate commands/parameters for the different
+    # types of secrets would pollute the top level argument space and make the stdin input option more complex
+    # for the user (e.g. first ask them to specify which type of secret they have rather than just an input prompt)
+
+    def default_password_filter(password):
+        return True
+    login_attempts = (
+        ('password', default_password_filter),
+        ('apiKey', utils.is_base64_encoded),  # an api key is base64 encoded so we can exclude strings that aren't
+        ('authToken', default_password_filter),  # although tokens are technically encoded, the client treats
```
zimingd, line 999: try auth token above api key

tests/unit/synapseclient/unit_test_commandline.py
```diff
@@ -191,6 +195,42 @@ def login_side_effect(*args, **kwargs):
...
+        login.side_effect = SynapseAuthenticationError()
+
+        # simulate failure both as password and as api key
+        with pytest.raises(SynapseAuthenticationError):
+            cmdline._authenticate_login(syn, username, auth_token, **login_kwargs)
+
+        assert expected_login_calls == login.call_args_list
+        login.reset_mock()
+
+        # now simulate success when used as an api key
```
zimingd, line 225:
```suggestion
        # now simulate success when used as an auth token
```

zimingd: If we do change the order such that auth_token > api_key, then the test for api_token should have a simulated auth_token failure

synapseclient/core/credentials/cached_sessions.py
```diff
@@ -75,7 +49,7 @@ def migrate_old_session_file_credentials_if_necessary(syn):
             if key == "<mostRecent>":
                 set_most_recent_user(value)
             else:
-                set_api_key(key, value)
+                SynapseApiKeyCredentials(key, value).store_to_keyring()
```
zimingd, line 52: we can potentially remove this old migration code now. This was implemented around 2-3 years ago for a 1-time migration of older client versions.

tests/integration/synapseclient/integration_test.py
```diff
@@ -50,17 +51,20 @@ def test_login(syn):
...
-            syn.login(username, password, rememberMe=True, silent=True)
+        # remember login info in cache
+        syn.login(username, password, rememberMe=True, silent=True)
 
-            # login using cached info
-            syn.login(username, silent=True)
-            syn.login(silent=True)
+        # login using cached info
+        syn.login(username, silent=True)
+        syn.login(silent=True)
```
zimingd, line 67: assert username is as expected when logging in w/out a username

zimingd (changes requested):
minor changes

synapseclient/core/credentials/credential_provider.py
```diff
@@ -22,7 +22,8 @@ def _get_auth_info(self, syn, user_login_args):
 
         :param ``synapseclient.client.Synapse`` syn:        Synapse client instance
         :param ``cred_data.UserLoginArgs`` user_login_args: subset of arguments passed during syn.login()
-        :return: tuple of (username, password, api_key, token), any of these values could None if it is not available.
+        :return: tuple of (username, password, api_key, bearer auth token (e.g. a personal access token),
```
zimingd, line 25: missing a )
```suggestion
        :return: tuple of (username, password, api_key, bearer auth token (e.g. a personal access token)),
```

synapseclient/core/credentials/credential_provider.py
```diff
@@ -100,7 +102,7 @@ def _get_auth_info(self, syn, user_login_args):
             username = config_username
             password = config_dict.get('password')
             api_key = config_dict.get('apikey')
-            auth_token = config_dict.get('authToken')
+            auth_token = config_dict.get('authtoken')
```
zimingd, line 105: Could you also add an entry for `authtoken` to the example config file and remove the one for `apikey`: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/.synapseConfig


Sorry for missing this in the last review!

zimingd (approved):

synapseclient/core/credentials/credential_provider.py
```diff
@@ -62,10 +71,16 @@ class UserArgsSessionTokenCredentialsProvider(SynapseCredentialsProvider):
...
+        password = None
+        api_key = None
+        auth_token = None
+
         if user_login_args.session_token:
-            return syn.getUserProfile(sessionToken=user_login_args.session_token)['userName'], None,\
-                   syn._getAPIKey(user_login_args.session_token)
-        return None, None, None
+            username = syn.getUserProfile(sessionToken=user_login_args.session_token)['userName']
+            api_key = syn._getAPIKey(user_login_args.session_token)
```
zimingd, line 81: i wonder if we should request a Access token instead of the API key. we can do this in a future issue.

jkiang13: Yes, I'd like to do this also but it's a bit trickier because it would mean generating a new access token with each rememberMe login, each of which would be saved, and which are limited to 100, so it requires some additional interaction to support if happening automatically on a login. I'm thinking it should probably be a discrete step rather than happening automatically on login.

synapseclient/core/credentials/credential_provider.py
jkiang13: Created a separate PR for the doc https://github.com/Sage-Bionetworks/synapsePythonClient/pull/806

synapseclient/core/credentials/credential_provider.py
jkiang13: fixed parens

PR #805, opened 2021-02-01
Synpy 929(834) silent parameter functions

linchiahuisage:
change

- add silent mode option for command line and pass in as an argument to create Synapse instance
-  If silent is true, the logger will change to silent user and won't print out the transfer progress on terminal
- add unit test for commandline.py/ client.py/ remote_storage_file_wrappers.py


jkiang13:
Since the change here are already part of https://github.com/Sage-Bionetworks/synapsePythonClient/pull/807 we can just review/merge that PR instead.

PR #808, opened 2021-02-04
SYNPY-1112 re-use matching file handles during a storage migration

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1112

The storage location migration utility will attempt to copy a new file handle for every entity that it copies. If two entities/versions share the same file handle and an attempt is made to copy them concurrently, it produces a DuplicateKeyException in the backend. However the utility should only be copying each file handle once and re-using the newly copied handle for each entity/version that shares the same source file handle id.

### DIFF
--- a/synapseutils/migrate_functions.py
+++ b/synapseutils/migrate_functions.py
@@ -615,6 +615,8 @@ def migrate_indexed_files(
             # ordering by synapse id. there can be multiple file handles associated with a particular
             # synapse id (i.e. multiple file entity versions or multiple table attached files per table),
             # so the ordering and where clause need to account for that.
+            # we also include in the query any unmigrated files that were skipped previously through
+            # the query loop that share a file handle with a file handle id that is now finished.
             version = key.version if key.version is not None else -1
             row_id = key.row_id if key.row_id is not None else -1
             col_id = key.col_id if key.col_id is not None else -1
@@ -685,7 +687,7 @@ def migrate_indexed_files(
                     # from another record in this batch then we defer it
                     continue
 
-                to_file_handle_id = _check_file_handle_exists(cursor, from_file_handle_id)
+                to_file_handle_id = _check_file_handle_exists(conn.cursor(), from_file_handle_id)
                 if not to_file_handle_id:
                     pending_file_handle_ids.add(from_file_handle_id)
 
--- a/synapseutils/migrate_functions.py
+++ b/synapseutils/migrate_functions.py
@@ -621,6 +621,25 @@ def migrate_indexed_files(
             row_id = key.row_id if key.row_id is not None else -1
             col_id = key.col_id if key.col_id is not None else -1
 
+            query_kwargs = {
+                'indexed_status': _MigrationStatus.INDEXED.value,
+                'id': key.id,
+                'file_type': _MigrationType.FILE.value,
+                'table_type': _MigrationType.TABLE_ATTACHED_FILE.value,
+                'version': version,
+                'row_id': row_id,
+                'col_id': col_id,
+
+                # ensure that we aren't ever adding more items to the shared executor than allowed
+                'limit': min(batch_size, max_concurrent_file_copies - len(futures)),
+            }
+
+            # we can't use both named and positional literals in a query, so we use named
+            # literals and then inline a string for the values for our file handle ids
+            # since these are a dynamic list of values
+            pending_file_handle_in = "('" + "','".join(pending_file_handle_ids) + "')"
+            completed_file_handle_in = "('" + "','".join(completed_file_handle_ids) + "')"
+
             results = cursor.execute(
                 f"""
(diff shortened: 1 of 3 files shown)


tests/integration/synapseutils/test_synapseutils_migrate.py
```diff
@@ -72,12 +72,21 @@ def test_migrate_project(request, syn, schedule_for_cleanup, storage_location_id
     file_1 = synapseclient.File(name=file_1_name, path=file_1_path, parent=folder_1_entity)
     file_1_entity = syn.store(file_1)
 
-    file_2_path = _create_temp_file()
+    # file 2 shares a file handle with file 1
+    file_2_path = file_1_path
     schedule_for_cleanup(file_2_path)
     file_2_name = "{}-{}".format(test_name, 2)
     file_2 = synapseclient.File(name=file_2_name, path=file_2_path, parent=folder_1_entity)
+    file_2.dataFileHandleId = file_1_entity.dataFileHandleId
```
zimingd, line 80: 1 and 3 share file handle id
2 is different file

tests/unit/synapseutils/unit_test_synapseutils_migrate.py
```diff
@@ -1369,6 +1375,182 @@ def test_continue_on_error__false(self, syn, db_path):
             assert 'boom' in str(ex)
 
 
+def test_create_new_file_version(syn):
```
zimingd, line 1378: break up into 2 test cases

synapseutils/migrate_functions.py
```diff
@@ -608,9 +633,18 @@ def migrate_indexed_files(
...
                     where
                         status = ?
-                        and ((id > ? and type in (?, ?))
-                            or (id = ? and type = ? and version is not null and version > ?)
-                            or (id = ? and type = ? and (row_id > ? or (row_id = ? and col_id > ?))))
+                        and (
+                                (
+                                    ((id > ? and type in (?, ?))
+                                    or (id = ? and type = ? and version is not null and version > ?)
+                                    or (id = ? and type = ? and (row_id > ? or (row_id = ? and col_id > ?))))
```
zimingd, line 640: inline the static values such as "type" so it is easier to see which case is being handled

zimingd (changes requested):

synapseutils/migrate_functions.py
```diff
@@ -350,16 +350,23 @@ def _ensure_schema(cursor):
     # we get counts grouping on status
     cursor.execute("create index if not exists ix_status on migrations(status)")
 
+    # we check to see if there is already a migrated copy of a file handle before doing a copy
+    cursor.execute(
+        "create index if not exists ix_file_handle_ids on migrations(from_file_handle_id, to_file_handle_id)"
```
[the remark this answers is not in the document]
jkiang13: It isn't 1-1 since we're multiple entity/versions can share a file handle id. It's probably the case that the index could only have the from value since the number of entities sharing a file handle is probably not large, but the code does query on both these values (from_file_handle_id matches and to_file_handle_id is not null in order to find the first row where the file handle is filled in).

synapseutils/migrate_functions.py
```diff
@@ -608,9 +633,18 @@ def migrate_indexed_files(
                     from migrations
                     where
                         status = ?
-                        and ((id > ? and type in (?, ?))
-                            or (id = ? and type = ? and version is not null and version > ?)
-                            or (id = ? and type = ? and (row_id > ? or (row_id = ? and col_id > ?))))
+                        and (
+                                (
+                                    ((id > ? and type in (?, ?))
+                                    or (id = ? and type = ? and version is not null and version > ?)
```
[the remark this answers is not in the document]
jkiang13: Can use named or positional but not both, which is a bit inconvenient since there are a couple dynamically sized lists of values that isn't easily named. Using named literals with some inlined values, the inlined values are file handle ids so should have known formatting and not result in formatting issues due to unexpected characters etc.

synapseutils/migrate_functions.py
jkiang13: Used a mix of named args and inline as above.

synapseutils/migrate_functions.py
```diff
@@ -350,16 +350,23 @@ def _ensure_schema(cursor):
...
     completed, futures = concurrent.futures.wait(futures, return_when=return_when)
+    completed_file_handle_ids = set()
 
     for completed_future in completed:
 
         to_file_handle_id = None
         ex = None
         try:
-            key, to_file_handle_id = completed_future.result()
+            key, from_file_handle_id, to_file_handle_id = completed_future.result()
```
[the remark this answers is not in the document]
jkiang13: it's used on line 392

zimingd (approved):

PR #810, opened 2021-02-08
Synpy 638 after date to cache purge

linchiahuisage:
Change

- add after_date argument in the purge method from cache.py
- check the valid arguments each after_date or before_date
- add unit test for purge method and private cache_dirs


synapseclient/core/cache.py
```diff
@@ -281,22 +280,34 @@ def _cache_dirs(self):
...
+        if before_date is None and after_date is None:
+            raise ValueError("Either before date or after date should be provided")
+
         if isinstance(before_date, datetime.datetime):
             before_date = utils.to_unix_epoch_time_secs(before_date)
+        if isinstance(after_date, datetime.datetime):
+            after_date = utils.to_unix_epoch_time_secs(after_date)
+
+        if before_date and after_date and before_date > after_date:
+            raise ValueError("Before date can not be larger than after date")
```
jkiang13, line 299: Is this backwards? Shouldn't we allow before_date > after_date but not the reverse?

If we set before_date to datetime.datetime(2021, 2, 9) and after_date to datetime.datetime(2021, 2, 7) then I think we'd want to delete everything between 2/7 and 2/9, which seems like we should allow. If we allow after_date > before_date then it will purge everything except what is between those two dates, which seems a bit confusing.

synapseclient/core/cache.py
```diff
@@ -281,22 +280,34 @@ def _cache_dirs(self):
...
         for cache_dir in self._cache_dirs():
             # _get_modified_time returns None if the cache map file doesn't
             # exist and n > None evaluates to True in python 2.7(wtf?). I'm guessing it's
             # OK to purge directories in the cache that have no .cacheMap file
+
             last_modified_time = _get_modified_time(os.path.join(cache_dir, self.cache_map_file_name))
-            if last_modified_time is None or before_date > last_modified_time:
+            if last_modified_time is None or \
+                    (before_date and before_date > last_modified_time) or \
+                    (after_date and after_date < last_modified_time):
```
jkiang13, line 310: I think maybe this should be:

```
if last_modified_time is None or (
                    (not before_date or before_date > last_modified_time) and 
                    (not after_date or after_date < last_modified_time)):
```

That way if both dates were supplied then anything between those dates would be deleted?

synapseclient/core/cache.py
```diff
@@ -281,22 +280,34 @@ def _cache_dirs(self):
                     if os.path.isdir(path2) and re.match('\\d+', item2):
                         yield path2
 
-    def purge(self, before_date, dry_run=False):
+    def purge(self, before_date=None, after_date=None, dry_run=False):
         """
```
jkiang13, line 284: Can you add parameter description to the docstring here:

e.g.

        :param before_date: Files last updated before this date will be deleted

etc.

jkiang13 (changes requested):

synapseclient/core/cache.py
jkiang13: Could you add the docstrings for these parameters? both the new after_date and the existing before_date and dry_run parameters?

jkiang13 (changes requested):

synapseclient/core/cache.py
```diff
@@ -281,22 +280,43 @@ def _cache_dirs(self):
...
         the cache.
+
+        :param before_date: To specify the date that all the files in Synapse caches which creating date before it will
+                            be removed.
+        :param after_date:  To specify the date that all the files in Synapse caches which creating date after it will
+                            be removed.
+        :param dry_run:     To determine purge method actually remove the files. If dry_run is True then just print out
+                            the files that will be removed.
+
+        :returns: count how many file will be removed
```
jkiang13, line 297: How about the following:
        Either the before_date or after_date must be specified. If both are passed, files between the two dates are selected for removal.

        :param before_date: if specified, all files before this date will be removed
        :param after_date:  if specified, all files after this date will be removed
        :param dry_run:     if dry_run is True, then the selected files are printed rather than removed
        :returns: the number of files selected for removal

jkiang13 (changes requested):

synapseclient/core/cache.py
jkiang13: I think you missed part of this comment on that last commit. I was also suggesting that the :param: strings get updated in this comment, I think they not quite clear enough as is.

jkiang13 (changes requested):

jkiang13 (approved):

PR #811, opened 2021-02-16
Synpy 660 validate filename sync

linchiahuisage:
Change

- when using Synapseutils sync.syncToSynapse method, add two validation.
- first one is checking the size of each file is not empty
- second one is checking the file name is set properly
- add unit tests for each private method when reading the manifest


synapseutils/sync.py
```diff
@@ -919,3 +933,27 @@ def _manifest_upload(syn, df):
...
+
+
+def _check_size_each_file(df):
+    for f in df.path:
+        if is_url(f):
+            continue
+        single_file_size = os.stat(os.path.expandvars(os.path.expanduser(f))).st_size
+        if single_file_size == 0:
+            print('\nAll the files uploaded cannot be 0 byte.')
+            raise ValueError('All the files uploaded cannot be 0 byte.')
```
jkiang13, line 959: You don't need to print and raise an error, raising an error should be enough. But the message should have the file name in it. Each "File {} is empty, empty files cannot be uploaded to Synapse"

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+
+    manifest = StringIO(header + row1 + row2 + row3 + row4)
+    mock_os.path.isfile.side_effect = [True, True, True, False]
+    mock_os.path.abspath.side_effect = [path1, path3]
+    mock_stat = MagicMock(spec='st_size')
+    mock_os.stat.return_value = mock_stat
+    mock_stat.st_size = 5
+
+    # mock syn.get() to return a project because the final check is making sure parent is a container
+    # mock isfile() to always return true to avoid having to create files in the home directory
```
jkiang13, line 808: This comment mentions isfile but isfile is not mocked

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+    row2 = f"{path2}\t{project_id}\tTest_file-name`s(1).txt\n"
+    row3 = f"{path3}\t{project_id}\t\n"
+
+    manifest = StringIO(header + row1 + row2 + row3)
+    mock_os.path.isfile.side_effect = [True, True, True]
+    mock_os.path.abspath.side_effect = [path1, path2, path3]
+    mock_os.path.basename.return_value = 'file3.txt'
+
+    # mock syn.get() to return a project because the final check is making sure parent is a container
+    # mock isfile() to always return true to avoid having to create files in the home directory
```
jkiang13, line 867: This comment mentions isfile but isfile is not mocked

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+    path3 = os.path.abspath(os.path.expanduser('~/file3.txt'))
+    path4 = os.path.abspath(os.path.expanduser('~/file4.txt'))
+
+    row1 = f"{path1}\t{project_id}\tTest_file_name.txt\n"
+    row2 = f"{path2}\t{project_id}\tTest_file-name`s(1).txt\n"
+    row3 = f"{path3}\t{project_id}\t\n"
+    row4 = f"{path4}\t{project_id}\tTest_file_name_with_#.txt\n"
+
+    manifest = StringIO(header + row1 + row2 + row3 + row4)
+    mock_os.reset_mock()
```
jkiang13, line 892: Why do you need to reset_mock here?

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+    mock_os.path.isfile.side_effect = [True, True, True, True]
```
jkiang13, line 893: Since the return value of this never changes could it be isfile.return_value = True rather than using a side_effect?

synapseutils/sync.py
```diff
@@ -770,6 +771,18 @@ def readManifestFile(syn, manifestFile):
...
 
+    # Check each size of uploaded file
+    sys.stdout.write('Validating that all the files are not empty...')
+    _check_size_each_file(df)
+    sys.stdout.write('OK\n')
+
+    # check the name of each file should be store on Synapse
+    name_column = 'name'
+    if name_column in df.columns:
+        sys.stdout.write('Validating that all name column are filled... \n')
```
jkiang13, line 782: Maybe "Validating file names..." since it is doing more than just checking that the names are filled (checking length, characters etc).

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+    path1 = os.path.abspath(os.path.expanduser('~/file1.txt'))
+    path2 = os.path.abspath(os.path.expanduser('~/file2.txt'))
+    path3 = os.path.abspath(os.path.expanduser('~/file3.txt'))
+
+    row1 = f"{path1}\t{project_id}\tTest_file_name.txt\n"
+    row2 = f"{path2}\t{project_id}\tTest_file-name`s(1).txt\n"
+    row3 = f"{path3}\t{project_id}\t\n"
+
+    manifest = StringIO(header + row1 + row2 + row3)
+    mock_os.path.isfile.side_effect = [True, True, True]
```
jkiang13, line 862: side_effects are all the same, can we use return_value instead?

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+                     'test_filename_too_long_test_filename_too_long_test_filename_too_long_test_filename_too_long_'
+
+    row1 = f"{path1}\t{project_id}\tTest_file_name.txt\n"
+    row2 = f"{path2}\t{project_id}\tTest_file-name`s(1).txt\n"
+    row3 = f"{path3}\t{project_id}\t\n"
+    row4 = f"{path4}\t{project_id}\t{long_file_name}\n"
+
+    manifest = StringIO(header + row1 + row2 + row3 + row4)
+    mock_os.reset_mock()
+    mock_os.path.isfile.side_effect = [True, True, True, True]
```
jkiang13, line 928: return_value vs side_effect?

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+                     'test_filename_too_long_test_filename_too_long_test_filename_too_long_test_filename_too_long_' \
+                     'test_filename_too_long_test_filename_too_long_test_filename_too_long_test_filename_too_long_'
+
+    row1 = f"{path1}\t{project_id}\tTest_file_name.txt\n"
+    row2 = f"{path2}\t{project_id}\tTest_file-name`s(1).txt\n"
+    row3 = f"{path3}\t{project_id}\t\n"
+    row4 = f"{path4}\t{project_id}\t{long_file_name}\n"
+
+    manifest = StringIO(header + row1 + row2 + row3 + row4)
+    mock_os.reset_mock()
```
jkiang13, line 927: Why reset_mock here?

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+    mock_os.path.isfile.side_effect = [True, True, True, True]
+    mock_os.path.abspath.side_effect = [path1, path2, path3, path4]
+    mock_os.path.basename.return_value = 'file3.txt'
```
jkiang13, line 930: Why file3.txt here?

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -773,3 +777,160 @@ def test_get_file_entity_provenance_dict__error_not_404(self):
...
+    path3 = os.path.abspath(os.path.expanduser('~/file3.txt'))
+
+    row1 = f"{path1}\t{project_id}\tTest_file_name.txt\n"
+    row2 = f"{path2}\t{project_id}\tTest_file-name`s(1).txt\n"
+    row3 = f"{path3}\t{project_id}\t\n"
+
+    manifest = StringIO(header + row1 + row2 + row3)
+    mock_os.path.isfile.side_effect = [True, True, True]
+    mock_os.path.abspath.side_effect = [path1, path2, path3]
+    mock_os.path.basename.return_value = 'file3.txt'
```
jkiang13, line 864: Why file3.txt?

jkiang13 (changes requested):

tests/unit/synapseutils/unit_test_synapseutils_sync.py
linchiahuisage: Because I didn't separate each unit test, now it's no need to use reset_mock anymore

tests/unit/synapseutils/unit_test_synapseutils_sync.py
linchiahuisage: Because we still have one name column is empty which path is path3 = os.path.abspath(os.path.expanduser('~/file3.txt'))

linchiahuisage: if the row['name'] is empty we will call os.path.basename to get the file name on the path

jkiang13:
I think there are a few unresolved comments from above? It might be easy to miss the because they are hidden under "5 hidden conversations" when I view the PR.

tests/unit/synapseutils/unit_test_synapseutils_sync.py
linchiahuisage: Because we still have one name column is empty which path is path3 = os.path.abspath(os.path.expanduser('~/file3.txt'))
if the row['name'] is empty we will call os.path.basename to get the file name on the path

jkiang13 (changes requested):

jkiang13 (approved):

PR #812, opened 2021-02-17
SYNPY-1058 username not required with authtoken

jkiang13:
Per [validation feedback ](https://sagebionetworks.jira.com/browse/SYNPY-1058?focusedCommentId=125040), this makes it so that username is not needed with an auth token login, and furthermore logging in with a username and an auth token will raise an error if the token doesn't match the user (rather than ambiguously logging in to the token user ignoring the username).

Per feedback, we also now require a view scoped token since lacking a view scope can fail in unexpected ways as the client makes many meta data calls.


BrunoGrandePhD:
I tried using a PAT that only had the `download` scope because I didn't really need the `view` scope, but I ran into an error indicating that the `view` scope is required. If this is indeed the case, I would add a note to the documentation. 

```
SynapseAuthenticationError: You are not logged in and do not have access to a requested resource.
  caused by SynapseHTTPError: 403 Client Error: 
insufficient_scope. Request lacks scope(s) required by this service: view
```

For context, I installed the version from this PR using the following command: 

```
python3 -m pipx install --force git+https://github.com/jkiang13/synapsePythonClient.git@SYNPY-1058-authtoken-username-not-required
```

BrunoGrandePhD:
After creating a new token with the `view` and `download` scopes, I ran into another error when trying to recursively download a Synapse folder. I've included the commands I've run and their output below, but I've masked my PAT with `[...]`. Interestingly, the `synapse get` command appears to initiate a download but fails partway. Any ideas? 

For context, I'm running these commands on an EC2 instance (Ubuntu 18.04) in the `sandbox` environment. 

```
$ synapse login --rememberMe -p "[...]"
Welcome, Bruno Grande!

Logged in as: bgrande (3413689)


$ synapse get -r syn22092368
 \ Downloaded 56.0MB (27.8MB/s)
ValueError: File download failed during sync
  caused by SynapseHTTPError: 400 Client Error: Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidArgument</Code><Message>Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified</Message><ArgumentName>Authorization</ArgumentName><ArgumentValue>Bearer [...]</ArgumentValue><RequestId>F15D2F667D46EE84</RequestId><HostId>MkfLU60NkZYnhKxGD6e9LYlaz+HsY48g9gTplDmMDzhk+aqm1V89hp3VuUvb8OqCsu+PVoBrzYk=</HostId></Error>
```

BrunoGrandePhD:
I'm trying to log in using my username and password to confirm that it works without using a PAT. I noticed the following error message when I mistyped my password. The last line assumes the user intended to provide a token. I would clarify that the issue could also be a mistyped password because the last line might throw off some users. 

```
$ synapse login --rememberMe 
Synapse username (leave blank if using an auth token): bgrande
Password or api key for user bgrande:

SynapseAuthenticationError: You are not logged in and do not have access to a requested resource.
  caused by SynapseHTTPError: 401 Client Error: 
Expected three sections of the token but found 1
```

BrunoGrandePhD:
I'm running into unusual behavior and I don't know whether it's expected. 

I deleted the Synapse session file and tried logging in using my username and password, which worked. However, when I try downloading the Synapse folder, I run into the same `Bad Request` error as above. Interestingly, when I delete the Synapse session file and go straight to `synapse get`, I'm prompted for my username and password and the download proceed as expected. 

I'm guessing that the PAT is memorized in a Linux keychain somewhere. Unfortunately, the [docs](https://python-docs.synapse.org/build/html/Credentials.html) don't specify where, so I'll have to do some more digging (perhaps by looking at the source code) to figure out where to delete the keychain entry, if one exists. 

That said, I'm still surprised that entering my username and password didn't override the PAT. Did this happen because a username wasn't specified when I entered a PAT? 

```
$ rm -r ~/.synapse*

$ synapse login --rememberMe
Synapse username (leave blank if using an auth token): bgrande
Password or api key for user bgrande:
Welcome, Bruno Grande!

Logged in as: bgrande (3413689)

$ synapse get -r syn22092368

ValueError: File download failed during sync
  caused by SynapseHTTPError: 400 Client Error: Bad Request
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidArgument</Code><Message>Only one auth mechanism allowed; only the X-Amz-Algorithm query parameter, Signature query string parameter or the Authorization header should be specified</Message><ArgumentName>Authorization</ArgumentName><ArgumentValue>Bearer [...]</ArgumentValue><RequestId>42DB9EADC84F17D8</RequestId><HostId>Fr+lzL6jT0G257NEgUXckg3w6YOV0mwgr83o+6n6AthYYM0t8IKMD1T5+r0mEEwiCE7nqirZQlw=</HostId></Error>


$ rm -r ~/.synapse*

$ synapse get -r syn22092368
Synapse username (leave blank if using an auth token): bgrande
Password or api key for user bgrande:
Welcome, Bruno Grande!

Downloading  [####################]100.00%
[...]
```

jkiang13:
Thanks @BrunoGrandePhD

In the order of your comments:

1. After conferring with the Synapse team we've decided to make the client reject tokens that don't have view scope. They would be a bit of a minefield because the client makes many view scoped calls while performing operations that the user may think are download or edit related. We're also going to change the access token web UI to warn that the programmatic clients minimally expect a view scoped token.

2. The failed download issue was caused by Synapse headers being passed to the actual object host, which can reject them. I had this fixed but apparently hadn't pushed that commit in this PR. I've added it now. Sorry for that confusion.

3. This is a good point. From the command line the client may attempt to use the credentials in a number of ways since the console interface passes the secret in the same way. It was returning the last errored attempt it encountered which in this case was the token error. It may be more intuitive to return the password attempt error. I've made this change in this PR.

4. Both credentials were stored in the keyring since there was no intervening logout. This is a good point and I've made a change such that on a rememberMe login it will clear all stored credentials, not just the previously stored credentials using the same login mechanism (i.e. password vs auth token). Without this then a bare login would use the first credentials it found in the keyring (in this case preferring the token to the password obtained api key).

synapseclient/client.py
```diff
@@ -3797,14 +3810,20 @@ def _rest_call(self, method, uri, data, endpoint, headers, retryPolicy, requests
         retryPolicy = self._build_retry_policy(retryPolicy)
         requests_session = requests_session or self._requests_session
 
+        auth = self.credentials
+        call_kwargs = kwargs
+        if 'auth' in kwargs:
+            call_kwargs = {**kwargs}
+            auth = call_kwargs.pop('auth')
```
zimingd, line 3817: `auth = call_kwargs.pop('auth', self.credentials)`

synapseclient/core/credentials/cred_data.py
```diff
@@ -103,14 +105,48 @@ class SynapseAuthTokenCredentials(SynapseCredentials):
...
+
+        # if for any reason we are not able to decode the token and check its scopes
+        # we do NOT raise an error. this is to accommodate the possibility of a changed
+        # token format some day that this version of the client may still be able to
+        # pass as a bearer token.
+        try:
+            token_body = json.loads(
+                str(
+                    base64.urlsafe_b64decode(
+                        token.split('.')[1] + '=='
```
zimingd, line 121: do we have to append the padding `==`

synapseclient/jkiang@kiang.org
```diff
@@ -0,0 +1 @@
+hello
```
zimingd, line 1: hi

synapseclient/core/credentials/credential_provider.py
```diff
@@ -120,8 +137,9 @@ def _get_auth_info(self, syn, user_login_args):
         if not user_login_args.skip_cache:
             username = user_login_args.username or cached_sessions.get_most_recent_user()
             if username:
-
                 api_creds = SynapseApiKeyCredentials.get_from_keyring(username)
+
+                # tokens are stored in the keyring under the username, so the username
```
zimingd, line 142: incomplete sentence

synapseclient/core/credentials/credential_provider.py
```diff
@@ -155,6 +173,10 @@ def get_credentials(self, syn, user_login_args):
                 return creds
         return None
 
+    def forget_credentials(self, syn, user_login_args):
```
zimingd, line 176: unused function?

tests/unit/synapseclient/core/credentials/unit_test_cred_provider.py
```diff
@@ -245,27 +274,40 @@ def teardown(self):
...
+            skip_cache=False,
+            auth_token=None,
+        )
 
         returned_tuple = self.provider._get_auth_info(self.syn, user_login_args)
 
         assert self.expected_return_tuple == returned_tuple
         self.mock_get_config_authentication.assert_called_once_with()
 
     def test_get_auth_info__user_arg_username_does_not_match_config(self):
```
zimingd, line 298: test name does not match intended test. 
I would assume that if a uncached username is being used, then no token should be returned

zimingd (changes requested):

synapseclient/core/credentials/cred_data.py
jkiang13: Added an explanation. We're basically always adding padding because Python b64decode will ignore any unnecessary padding so adding the maximum padding doesn't hurt (but we do need to add the padding as the tokens we get from the backend are unpadded). https://stackoverflow.com/a/49459036

synapseclient/core/credentials/credential_provider.py
jkiang13: rm'ed

tests/unit/synapseclient/core/credentials/unit_test_cred_provider.py
jkiang13: This was also masking a perhaps ambiguous case that is probably not correct. If you had a username and an auth token in your config and supplied a username at the login prompt and the login prompt did not match the config it would allow you to login using the config token. I've removed that ambiguity (i.e. if you supply a username arg and there is a username in the config file it won't use the config file at all). I don't think this is probably a likely case because there's not really any reason for someone to add a username and auth token to their config file now, but it seems better to have no ambiguity.

synapseclient/__main__.py
```diff
@@ -977,54 +977,66 @@ def login_with_prompt(syn, user, password, rememberMe=False, silent=False, force
         _authenticate_login(syn, user, password, silent=silent, rememberMe=rememberMe, forced=forced)
     except SynapseNoCredentialsError:
         # if there were no credentials in the cache nor provided, prompt the user and try again
-        while not user:
-            user = input("Synapse username: ")
```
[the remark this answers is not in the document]
jkiang13: I've updated the prompt slightly, if they supply a username then we don't prompt for one but the username appears as part of the password prompt (i.e. "Password, api key, or auth token for user {user}:"), and indicated that if they do happen to input a username they can still supply an auth token if they choose (although it will error if it doesn't match the username).

zimingd (approved):

PR #817, opened 2021-02-19
Synpy 1107 option to not create manifest

linchiahuisage:
Change

- Add one option for command-line input which is either '-supMani' or '--suppressManifest'. The default value is False
- Add unit test for suppressing manifest command-line option
- Add unit test for Synapseutils sync that allowing a client to determine whether to create the manifest file or not.

https://sagebionetworks.jira.com/browse/SYNPY-1107


jkiang13:
Since all of the commits for this PR are in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/822 we can review that one instead and close this.

PR #818, opened 2021-02-19
SYNPY-1114 containerized sftp integration tests

jkiang13:
This replaces the use of a fixed EC2 instance as a storage location for Python integration tests with a container that is created during the GH Actions CI run.

This is simpler and has no external dependency. The one downside is that Windows and Mac integration tests cannot use it because GH Actions doesn't support containers for those platforms, and you can't share a container across matrix jobs.

The SFTP integration tests look for an "SFTP_HOST" environment variable and will test against those, using matching credentials found in the synapseConfig.


jkiang13:
Shoot wrong bug # closing...

PR #822, opened 2021-02-24
Synpy 756 generate manifest function only

linchiahuisage:
Change

- For cmd-line get function, add one more arg option '-cm' or '--createManifest' which should specifically add one of three choices ['all', 'root', 'suppress'].
- The syncFromSynapse in Synapseutils.sync.py, adding two arguments which are manifest and downloadFile.
- Add unit test for unit_test_commandline.py
- Add unit test for unit_test_synapseutils_sync.py


synapseutils/sync.py
```diff
@@ -61,6 +62,8 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
     :param followLink:  Determines whether the link returns the target Entity.
                         Defaults to False
 
+    :param manifest:    Determines whether creating manifest file automatically.
```
jkiang13, line 65: Can you document the optional values here ("all", "root", "suppress") as part of the doc string?

jkiang13: Also it would be good to validate the value passed in is one of the options. The command line call will do that automatically when this is called from the command line, but if invoked programmatically there is no check.

synapseclient/__main__.py
```diff
@@ -577,6 +578,9 @@ def build_parser():
                             'is deemed fully stable and becomes the default implementation.')
     parser_get.add_argument('id', metavar='syn123', nargs='?', type=str,
                             help='Synapse ID of form syn123 of desired data object.')
+    # add no manifest option
+    parser_get.add_argument('--createManifest', type=str, choices=['all', 'root', 'suppress'],
```
jkiang13, line 582: The command line option is "createManifest", but the name of the argument to the function is just manifest. Maybe we can make them the same, just "manifest"?

jkiang13 (changes requested):

PR #829, opened 2021-03-12
SYNPY-1123 escape table query column names during migration index

jkiang13:
Fixes a few issues related to indexing table attached file handles when indexing for a migration:

1. Table column names could contain special characters and weren't being escaped.
2. Some older file handles don't include a storageLocationId, this shouldn't produce a KeyError (related [PLFM-6637](https://sagebionetworks.jira.com/browse/PLFM-6637)
3. If only indexing tables and not files, there's no reason to scan the folder hierarchy since folders only contain files.

### DIFF
--- a/synapseutils/migrate_functions.py
+++ b/synapseutils/migrate_functions.py
@@ -963,7 +963,7 @@ def _index_file_entity(
     if entity_versions:
         insert_values = []
         for (entity, version) in entity_versions:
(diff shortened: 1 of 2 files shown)


jkiang13:
Handled by https://github.com/Sage-Bionetworks/synapsePythonClient/pull/848

PR #830, opened 2021-03-12
make the get method arguments formally

linchiahuisage:
change

- change the kwargs argument to keyword arguments in Synapse.get method
- change private method _getWithEntityBundle from kwargs argument to keyword arguments


jkiang13:
Closing this, didn't turn out as needed.

PR #831, opened 2021-03-15
[SYNPY-1174] Add function to create Synapse config file

thomasyu888:
- `synapse -c ~/.mySynapseConfig config`

Since we don't split up credential configuration file, it is a bit too hacky to try to edit the file minimally in place.  Therefore, I will copy the old configuration file into a `.synapseConfig.backup` and write a new `.synapseConfig` file.  It's nice that we are pushing for people to only put their `authtoken` in the `.synapseConfig` file, are we motivating all other forms of login by using `synapse login --rememberMe`?

There are a couple of solutions if we wanted to add more to the config function:
- We could use some sort of templating technology (jinja, mustache), but that just adds a dependency to the Python client.
- If we don't care to maintain the comments in the new synapse config, we could read in the old config file and write out a new config file based on what already exists.


synapseclient/__main__.py
```diff
@@ -378,6 +379,46 @@ def storeTable(args, syn):
...
+    passwd = None
+    while not passwd:
+        # if the terminal is not a tty, we are unable to read from standard input
+        # For git bash using python getpass
+        # https://stackoverflow.com/questions/49858821/python-getpass-doesnt-work-on-windows-git-bash-mingw64
+        if not sys.stdin.isatty():
+            raise SynapseAuthenticationError(
+                "No password, key, or token was provided and unable to read from standard input")
+        else:
+            passwd = getpass.getpass(secret_prompt)
```
thomasyu888, line 395: This could be a convenience function.

synapseclient/__main__.py
```diff
@@ -378,6 +379,46 @@ def storeTable(args, syn):
...
+    # Read in configuration
+    cur_config_path = os.path.dirname(os.path.realpath(args.configPath))
+    # Make a copy of the existing config if it exists
+    if os.path.exists(args.configPath):
+        shutil.copyfile(
+            args.configPath,
+            os.path.join(cur_config_path, f"{args.configPath}.backup")
+        )
+
+    with open(os.path.join(script_dir, ".synapseConfig"), "r") as config_o:
```
thomasyu888, line 410: Should there be some sort of warning that says the old config file will be copied into a new file?  If there are existing configurations, they need to be copied into the newly created config file.

thomasyu888:
@jkiang13 This is ready - I added tests.

mattfazza (approved):
I'm gonna trust the conversation before me here because I lack a bit of context.  It looks like the Windows tests have been fixed.  Is the configuration file something we need to fix?

PR #835, opened 2021-03-18
Synpy 1033 store issue for blank version comment

linchiahuisage:
Change

- add checking logic to make properties['versionComment'] with the value or None
- add unit test to check update correctly

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -1023,8 +1023,11 @@ class Synapse(object):
         entity = obj
         properties, annotations, local_state = split_entity_namespaces(entity)
         bundle = None
-        # Anything with a path is treated as a cache-able item
+        # Explicitly set an empty versionComment property if none is supplied,
+        # otherwise an existing entity bundle's versionComment will be copied to the update.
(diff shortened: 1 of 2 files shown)


synapseclient/client.py
```diff
@@ -1024,6 +1024,7 @@ def store(self, obj, *, createOrUpdate=True, forceVersion=True, versionLabel=Non
         properties, annotations, local_state = split_entity_namespaces(entity)
         bundle = None
         # Anything with a path is treated as a cache-able item
+        properties['versionComment'] = properties['versionComment'] if 'versionComment' in properties else None
```
jkiang13, line 1027: Can you add a comment here that explains why we are doing this, e.g.
```Explicitly set an empty versionComment property if none is supplied, otherwise an existing entity bundle's versionComment will be copied to the update.```

jkiang13: Can you also move the above comment ```Anything with a path is treated as a cache-able item``` below this line, it seems like it applies to the next line not this line.

jkiang13 (changes requested):

jkiang13 (approved):

PR #836, opened 2021-03-19
Synpy 1064 speed up integration tests

linchiahuisage:
Change

- speed up the integration tests with pytest-xdist dependency which allows test platform run integration tests with multiple CPU


.github/workflows/build.yml
```diff
@@ -112,7 +134,7 @@ jobs:
             export EXTERNAL_S3_BUCKET_AWS_ACCESS_KEY_ID="${{secrets.EXTERNAL_S3_BUCKET_AWS_ACCESS_KEY_ID}}"
             export EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY="${{secrets.EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY}}"
 
-            pytest -sv tests/integration
+            pytest -sv tests/integration -n auto --dist loadscope
```
jkiang13, line 137: Can we add a comment here, e.g.
`use loadscope to avoid issues running tests concurrently that share scoped fixtures`

.github/workflows/build.yml
```diff
@@ -45,7 +45,30 @@ jobs:
...
 
+      - name: get-dependencies-location
+        shell: bash
+        run: |
+          SITE_PACKAGES_LOCATION=$(python -c "from distutils.sysconfig import get_python_lib; print(get_python_lib())")
+          SITE_BIN_DIR=$(python3 -c "import os; import platform; import distutils.sysconfig; pre = distutils.sysconfig.get_config_var('prefix'); bindir = os.path.join(pre, 'Scripts' if platform.system() == 'Windows' else 'bin'); print(bindir)")
+          echo "::set-output name=site_packages_loc::$SITE_PACKAGES_LOCATION"
+          echo "::set-output name=site_bin_dir::$SITE_BIN_DIR"
+          echo $SITE_PACKAGES_LOCATION
+          echo $SITE_BIN_DIR
```
jkiang13, line 56: we can remove these echoes I think, we probably just had them while we were testing

.github/workflows/build.yml
```diff
@@ -45,7 +45,30 @@ jobs:
...
+        id: get_dependencies
```
jkiang13, line 57: All the other steps use hyphenated names rather than underscores, can we change it to be consistent?

.github/workflows/build.yml
```diff
@@ -45,7 +45,30 @@ jobs:
...
+
+      - name: Cache py-dependencies
+        id: cache-venv
```
jkiang13, line 60: I think the id of this step was from when we were trying to use a virtual env, maybe just change the id to cache-dependencies

jkiang13 (changes requested):

.github/workflows/build.yml
```diff
@@ -68,7 +90,7 @@ jobs:
       - name: run-unit-tests
         shell: bash
         run: |
-          pytest -sv tests/unit
+          pytest -sv tests/unit -n auto
```
jkiang13, line 93: Do we need -n auto for unit tests, does it make a difference?

jkiang13 (changes requested):

.github/workflows/build.yml
```diff
@@ -10,6 +10,10 @@
 name: build
 
 on:
+  schedule:
```
jkiang13, line 13: We can remove this trigger, it was just for testing.

jkiang13 (changes requested):

jkiang13 (approved):

PR #837, opened 2021-03-20
add credential provider to check AWS Parameter Store for token (SC-260)

zimingd:


synapseclient/core/credentials/credential_provider.py
```diff
@@ -178,7 +217,9 @@ def get_credentials(self, syn, user_login_args):
     UserArgsSessionTokenCredentialsProvider(),  # This provider is DEPRECATED
```
zimingd, line 217: Aside: we can probably remove this one now.

synapseclient/core/credentials/credential_provider.py
```diff
@@ -145,11 +151,44 @@ def _get_auth_info(self, syn, user_login_args):
...
+            import botocore  # if can import boto3, can import botocore
+            try:
+                ssm_client = boto3.client('ssm')
+                result = ssm_client.get_parameter(
+                    Name=ssm_param_name,
+                    WithDecryption=True,
+                )
+                token = result['Parameter']['Value']
+            except botocore.exceptions.ClientError as e:
+                syn.logger.warn(f'{self.ENVIRONMENT_VAR_NAME} was defined as {ssm_param_name}, '
```
jkiang13, line 177: Is every ClientError here a missing SSM param? Also this will only log the ClientError message and not the stack. I'm ambivalent whether the full stack is useful here but if so e.g. syn.logger.warn(<message>, exc_info=True)

jkiang13 (changes requested):

jkiang13 (approved):

PR #839, opened 2021-03-30
SYNPY-1128 more robust download network connections

jkiang13:
Ensures that all data is read from the network socket from a retried closure. Previously an error streaming from the socket subsequent to the actual connection would not be retried. In doing so this eliminates streaming from the individual range responses, but the range responses themselves ensure that we aren't accumulating too much in memory (in fact since accumulated each range response in memory before writing it to disk anyway the memory footprint doesn't change).


synapseclient/core/retry.py
```diff
@@ -6,20 +6,55 @@
...
-    :param function:           A function with no arguments.  If arguments are needed, use a lambda (see example).
-    :param retry_status_codes: What status codes to retry upon in the case of a SynapseHTTPError.
-    :param retry_errors:       What reasons to retry upon, if function().response.json()['reason'] exists.
-    :param retry_exceptions:   What types of exceptions, specified as strings, to retry upon.
-    :param retries:            How many times to retry maximum.
-    :param wait:               How many seconds to wait between retries.
-    :param back_off:           Exponential constant to increase wait for between progressive failures.
+    :param function:                A function with no arguments.  If arguments are needed, use a lambda (see example).
+    :param retry_status_codes:      What status codes to retry upon in the case of a SynapseHTTPError.
+    :param retry_status_code_in:    True if should retry if the status code is in retry_status_codes,
```
zimingd, line 50: the name and documentation is slightly confusing. 
maybe `retry_status_code_in` -> `retry_on_status_code_match`?
replace `if should retry` -> ` - retry`
replace `is in` with `exists in`

also, using a string enumeration of retry types might be more clear about the intentions than `True`/`False`
`retry_type` could be : 
`ON_MATCH`  - retry if status code matches values in `retry_status_codes`
`ON_MISMATCH` - retry if status code does not match values in `retry_status_codes`


alternatively, we can use 2 lists:
`retry_status_codes` -  list of status codes to retry if received as a response
`expected_status_codes` - retries until response returns one of the expected codes.

zimingd (changes requested):

synapseclient/core/retry.py
```diff
@@ -91,8 +90,8 @@ def foo(a, b, c): return [a, b, c]
         # Check if we got a retry-able error
         if response is not None and hasattr(response, 'status_code'):
             if (
-                    (retry_status_code_in and response.status_code in retry_status_codes) or
-                    (not retry_status_code_in and response.status_code not in retry_status_codes)
+                (expected_status_codes and response.status_code not in expected_status_codes) or
+                response.status_code in retry_status_codes
```
zimingd, line 94: `None` could be passed as value to `retry_status_codes`

synapseclient/core/multithread_download/download_threads.py
```diff
@@ -340,8 +340,7 @@ def session_get():
             # but for now matching previous behavior.
             response = with_retry(
                 session_get,
-                retry_status_code_in=False,
-                retry_status_codes=(HTTPStatus.PARTIAL_CONTENT,),
+                expected_status_codes=(HTTPStatus.PARTIAL_CONTENT,),
```
zimingd, line 343: just double checking, we still want to retry on the usual `[429, 500, 502, 503, 504]` right?

zimingd (changes requested):

synapseclient/core/multithread_download/download_threads.py
jkiang13: With the expected_status_codes passed it short circuits the retry_status_codes, and in this case with it will retry on anything other than a 206 PARTIAL CONTENT which seems a bit strange, but that was the previous behavior (except we would do it rapidly with no backoff).

synapseclient/core/retry.py
jkiang13: Added a Falsey check to retry_status_codes before using it

zimingd (approved):

PR #843, opened 2021-04-02
Synpy 1139 pytest coverage

linchiahuisage:
Change

- add pytest-cov dependency that help to upload the html folder to show the coverage of whole project
- upload the zip file artifact for user to download and look at the coverage
- zip file artifact is across three platform


thomasyu888:
I used to do this as well for some of my packages but I found that it's a lot cleaner to just register a package with coveralls.  You can even then have a badge for coverage.

jkiang13:
Closing this, if needed could use an external service.

PR #852, opened 2021-04-23
SYNPY-1113-store_command_positional_arg_file

linchiahuisage:
Change

- check the args with file or FILE before handling the store command
- add unit test for test case

### DIFF
--- a/synapseclient/__main__.py
+++ b/synapseclient/__main__.py
@@ -134,7 +134,8 @@ def store(args, syn):
     # If both args.FILE and args.file specified raise error
     if args.file and args.FILE:
         raise ValueError('only specify one file')
-
+    if not args.file and not args.FILE:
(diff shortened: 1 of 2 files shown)


synapseclient/__main__.py
```diff
@@ -134,7 +134,8 @@ def store(args, syn):
     # If both args.FILE and args.file specified raise error
     if args.file and args.FILE:
         raise ValueError('only specify one file')
-
+    if args.type == 'File' and not args.file and not args.FILE:
+        raise ValueError(f'For the {args.subparser} command, put the positional FILE as the first arg')
```
jkiang13, line 138: I think this error message is a bit too specific. We could maybe just just say
"{args.subpaser} missing required FILE argument"

jkiang13 (changes requested):

jkiang13 (approved):

PR #856, opened 2021-05-04
Synpy 1142 generalize retry backoff logic

linchiahuisage:
Change

- wrap the with_retry method with "with_retry_network" and generalize the logic with back_off
- fix the unit test and integration test after this consolidation

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,7 +133,8 @@ jobs:
             export EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY="${{secrets.EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY}}"
 
              # use loadscope to avoid issues running tests concurrently that share scoped fixtures
-            pytest -sv tests/integration -n auto --dist loadscope
+#            pytest -sv tests/integration -n auto --dist loadscope
+            pytest -sv tests/integration/synapseclient/test_evaluations.py::test_teams -n auto --dist loadscope
           fi
 
   # on a GitHub release, build the pip package and upload it as a GitHub release asset
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,7 +133,7 @@ jobs:
             export EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY="${{secrets.EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY}}"
 
              # use loadscope to avoid issues running tests concurrently that share scoped fixtures
-#            pytest -sv tests/integration -n auto --dist loadscope
+            # pytest -sv tests/integration -n auto --dist loadscope
             pytest -sv tests/integration/synapseclient/test_evaluations.py::test_teams -n auto --dist loadscope
           fi
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -133,8 +133,8 @@ jobs:
             export EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY="${{secrets.EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY}}"
 
              # use loadscope to avoid issues running tests concurrently that share scoped fixtures
-            # pytest -sv tests/integration -n auto --dist loadscope
-            pytest -sv tests/integration/synapseclient/test_evaluations.py::test_teams -n auto --dist loadscope
+            pytest -sv tests/integration -n auto --dist loadscope
+            # pytest -sv tests/integration/synapseclient/test_evaluations.py::test_teams -n auto --dist loadscope
           fi
 
   # on a GitHub release, build the pip package and upload it as a GitHub release asset
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -134,7 +134,6 @@ jobs:
 
              # use loadscope to avoid issues running tests concurrently that share scoped fixtures
             pytest -sv tests/integration -n auto --dist loadscope
-            # pytest -sv tests/integration/synapseclient/test_evaluations.py::test_teams -n auto --dist loadscope
           fi
 
   # on a GitHub release, build the pip package and upload it as a GitHub release asset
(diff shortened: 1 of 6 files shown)


synapseclient/core/retry.py
```diff
@@ -80,44 +113,23 @@ def foo(a, b, c): return [a, b, c]
...
-                elif 'Please slow down.  You may send a maximum of 10 message' in response:
-                    retry = True
-                    wait = 16
-                    logger.debug("retrying " + response_message)
-
         # Check if we got a retry-able exception
         if exc is not None:
             if (exc.__class__.__name__ in retry_exceptions or
                     exc.__class__ in retry_exceptions or
+                    not retry_errors or
```
jkiang13, line 132: Is this to check if retry_errors is None? If so should it be (retry_errors and any([msg.lower() in str(exc_info[1]).lower() for msg in retry_errors])?

And if we need to check that for None should we also check the other fields?

.github/workflows/build.yml
```diff
@@ -134,6 +134,7 @@ jobs:
 
              # use loadscope to avoid issues running tests concurrently that share scoped fixtures
             pytest -sv tests/integration -n auto --dist loadscope
+            # pytest -sv tests/integration/synapseclient/test_evaluations.py::test_teams -n auto --dist loadscope
```
jkiang13, line 137: Remove this?

synapseclient/core/retry.py
```diff
@@ -38,9 +38,45 @@
...
+            #     logger.warning('%s...\n' % response_message)
+            #     logger.warning('Retrying in %i seconds' % wait)
+        elif response.status_code not in range(200, 299):
+            # For all other non 200 messages look for retryable errors in the body or reason field
+            response_message = _get_message(response)
+            if any([msg.lower() in response_message.lower() for msg in retry_errors]):
+                retry = True
+                logger.debug('retrying %s' % response_message)
+            # special case for message throttling
+            elif 'Please slow down.  You may send a maximum of 10 message' in response:
```
jkiang13, line 70: We should check if this response is still being returned from the back end. It's strange to have a 200 with an error message so it may have been a specific Synapse message that may or may not be used anymore...

tests/integration/synapseclient/test_evaluations.py
```diff
@@ -179,16 +180,18 @@ def test_teams(syn, project, schedule_for_cleanup):
...
-    while tries > 0:
-        try:
-            found_team = anonymous_syn.getTeam(name)
-            break
-        except ValueError:
-            tries -= 1
-            if tries > 0:
-                time.sleep(sleep_time)
-                sleep_time *= 2
+    # tries = 8
```
jkiang13, line 183: We can hopefully remove this commented out code now?

synapseclient/core/retry.py
```diff
@@ -155,3 +175,9 @@ def _get_message(response):
     except (AttributeError, ValueError):
         # The response can be truncated. In which case, the message cannot be retrieved.
         return None
+
+
+class SynapseThrottling(Exception):
```
jkiang13, line 180: Maybe rename it _SynapseThrottlingException (since it's internal to this module).

jkiang13 (changes requested):

synapseclient/core/retry.py
```diff
@@ -38,9 +38,45 @@
...
+            elif 'Please slow down.  You may send a maximum of 10 message' in response:
+                logger.debug("retrying " + response_message)
+                raise SynapseThrottling(response_message, wait=16)
+        return retry
+    return with_retry(function, _handle_status_code, verbose=verbose,
+                      retry_errors=retry_errors, retry_exceptions=retry_exceptions, retries=retries, wait=wait,
+                      back_off=back_off, max_wait=max_wait)
+
+
+def with_retry(function, retry_evaluator=None, verbose=False, retry_errors=[], retry_exceptions=[],
```
jkiang13, line 79: Can we also use this to generalize the retry logic for these cases with a custom evaluator function?

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/client.py#L3126

jkiang13 (changes requested):

synapseclient/core/retry.py
linchiahuisage: I want to implement the logic that if users didn't pass any exception as the argument, I will default retry cuz we default to accept all exception

jkiang13 (changes requested):

synapseclient/core/retry.py
```diff
@@ -38,10 +38,51 @@
 ]
 
 
-def with_retry(function, verbose=False,
-               retry_status_codes=[429, 500, 502, 503, 504], expected_status_codes=[],
-               retry_errors=[], retry_exceptions=[],
-               retries=DEFAULT_RETRIES, wait=DEFAULT_WAIT, back_off=DEFAULT_BACK_OFF, max_wait=DEFAULT_MAX_WAIT):
+def with_retry_network(function, verbose=False,
+                       retry_status_codes=[429, 500, 502, 503, 504], expected_status_codes=[], retry_errors=[],
+                       retry_exceptions=None, retries=3, wait=1, back_off=2, max_wait=30):
```
jkiang13, line 43: Doesn't retry_exceptions=None here change the default behavior of the old with_retry (which would only retry on the listed exceptions)?

jkiang13 (changes requested):

synapseclient/core/retry.py
```diff
@@ -155,3 +183,9 @@ def _get_message(response):
     except (AttributeError, ValueError):
         # The response can be truncated. In which case, the message cannot be retrieved.
         return None
+
+
+class _SynapseThrottlingException(Exception):
```
thomasyu888, line 188: Should this live in `exceptions.py`?

thomasyu888:
Just out of curiosity, why is this change an improvement?  Is it the possibility of passing in other `retry_evaluators` into `with_retry`?  When I saw the `with_retry` function within our client, I was a bit surprised, because it feels like something that should already written in other packages, but I didn't do much research - are there packages that implement a more standard retry solution?

jkiang13:
Closing this, didn't turn out as needed.

PR #860, opened 2021-05-14
[SYNPY-1011] Implement --recursive for `synapse store`

BrunoGrandePhD:
This is a draft PR.

### DIFF
--- a/synapseclient/__main__.py
+++ b/synapseclient/__main__.py
@@ -134,7 +134,49 @@ def sync(args, syn):
                                retries=args.retries)
 
 
-def store(args, syn):
+def _recursive_store(args, syn):
+    # Credit for the original design of this function goes to Larsson Omberg
+    # https://gist.github.com/larssono/db35917cf58440fe0b19
+
+    args.file = args.FILE if args.FILE is not None else args.file
+    args.FILE = None
+
+    if args.parentid is None:
+        raise ValueError('You must use --parentId when making a recursive upload.')
+    if not os.path.isdir(args.file):
+        raise ValueError('You can only recursively upload a folder.')
+
+    parents = {args.file: args.parentid}
+    for dirpath, dirnames, filenames in os.walk(args.file):
+        # Add the subfolders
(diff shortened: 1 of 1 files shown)


synapseclient/__main__.py
```diff
@@ -134,7 +134,71 @@ def sync(args, syn):
...
+    # Confirm that target entity on Synapse is a folder
+    if args.parentid is None:
+        raise ValueError('--parentId is required for recursive uploads.')
+    parent = syn.get(args.parentid, downloadFile=False)
+    parent_type = parent.concreteType.split(".")[-1]
+    if not parent_type == "Folder" and not parent_type == "Project":
+        raise ValueError('--parentId must refer to a folder or project.')
+
+    # Some CLI arguments are ignored with --recursive; set to None
+    ignored = ["name", "annotations", "replace", "used", "executed"]
```
thomasyu888, line 155: Should `id ` be added?

thomasyu888:
This looks ok to me, but in general, I know the `store` function is pretty heinous.  I also think that we could potentially add this as part of the python client itself instead of only adding it as a command line function.

synapseclient/__main__.py
BrunoGrandePhD: It could. Because `--parentId` and `--id` are mutually exclusive, I wouldn't expect `id` to be present past the conditional on line 147. That said, I'll add it because it's useful for developers to see the expectation. 

```suggestion
    ignored = ["id", "name", "annotations", "replace", "used", "executed"]
```

BrunoGrandePhD:
@thomasyu888: When I wrap up this PR, I can add a function in `synapseutils` that is essentially a copy of Larsson's [original code](https://gist.github.com/larssono/db35917cf58440fe0b19). Unfortunately, the core of the code will be duplicated because the `store` subcommand requires special handling.

BrunoGrandePhD:
I'm closing this PR in favor of a more modular `create_manifest` subcommand, which I'll propose in a separate PR.

PR #862, opened 2021-05-19
Synpy 1146 writing boolean from df into synapse table

linchiahuisage:
Change

- fix the issue that pandas infer the data type of coulmn
- add unit test to verify column type are match with Synapse table

### DIFF
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -300,6 +300,7 @@ import abc
 import enum
 import json
 from builtins import zip
+from pandas.api.types import infer_dtype
 
 from synapseclient.core.utils import id_of, from_unix_epoch_time
 from synapseclient.core.exceptions import SynapseError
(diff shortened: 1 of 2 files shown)


synapseclient/table.py
```diff
@@ -317,6 +318,17 @@
                      'S': 'STRING', 'U': 'STRING', 'O': 'STRING',
                      'a': 'STRING', 'p': 'INTEGER', 'M': 'DATE'}
 
+# default is STRING, only need to put the non-STRING keys in here
```
jkiang13, line 321: we could remove DTYPE_2_TABLETYPE above if it is now unused.

jkiang13 (changes requested):

synapseclient/table.py
```diff
@@ -300,6 +300,7 @@
 import enum
 import json
 from builtins import zip
+from pandas.api.types import infer_dtype
```
thomasyu888, line 303: Does this function read in the entire column of values to determine the type?

thomasyu888 (approved):

PR #863, opened 2021-05-20
[SYNPY-1156] dateTime_dataframe_store_to_Synapse

linchiahuisage:
Change

- if the column type is DATE, change the date format to unix epoch and trailing the 3 digit of millisecond

### DIFF
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -1748,6 +1748,16 @@ class CsvFileTable(TableAbstractBaseClass):
             df = df2
             includeRowIdAndRowVersion = True
 
+        def _trailing_date_time_millisecond(t):
+            if isinstance(t, str):
+                return t[:-3]
+
+        for col in schema.columns_to_store:
+            if col['columnType'] == 'DATE':
+                import pandas as pd
+                df[col.name] = pd.to_datetime(df[col.name], errors='coerce').dt.strftime('%s%f')
+                df[col.name] = df[col.name].apply(lambda x: _trailing_date_time_millisecond(x))
+
         f = None
         try:
             if not filepath:
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -1757,6 +1757,7 @@ class CsvFileTable(TableAbstractBaseClass):
             f = io.open(filepath, mode='w', encoding='utf-8', newline='')
 
             test_import_pandas()
+            import pandas as pd
             for col in schema.columns_to_store:
                 if col['columnType'] == 'DATE':
                     def _trailing_date_time_millisecond(t):
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -1748,16 +1748,6 @@ class CsvFileTable(TableAbstractBaseClass):
             df = df2
             includeRowIdAndRowVersion = True
 
-        def _trailing_date_time_millisecond(t):
-            if isinstance(t, str):
-                return t[:-3]
-
-        for col in schema.columns_to_store:
-            if col['columnType'] == 'DATE':
-                import pandas as pd
-                df[col.name] = pd.to_datetime(df[col.name], errors='coerce').dt.strftime('%s%f')
-                df[col.name] = df[col.name].apply(lambda x: _trailing_date_time_millisecond(x))
-
         f = None
         try:
             if not filepath:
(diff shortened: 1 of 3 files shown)


synapseclient/table.py
```diff
@@ -1756,6 +1756,15 @@ def from_data_frame(cls, schema, df, filepath=None, etag=None, quoteCharacter='"
 
             f = io.open(filepath, mode='w', encoding='utf-8', newline='')
 
+            import pandas as pd
```
thomasyu888, line 1759: This sort of makes pandas a required install right?  Should there be a check for if pandas is installed?

jkiang13: For consistency we should use the same check although it's in the from_data_frame which is likely only going to be called from a context where pandas was already used.

I'm not sure this is the best approach though, Godfrey was working on this as he left and didn't have a chance to finish it. At very least if we modify the df for the purposes of serializing it to a csv file we should copy it first. The to_csv function of the data frame has a date_format param on it to which we might be able to use instead although it I'm not sure there is an easy way to preserve millisecond accuracy using that string format alone...

tests/integration/synapseclient/test_tables.py
```diff
@@ -233,7 +233,10 @@ def test_tables_pandas(syn, project):
     df['string_'] = df['string_'].transform(str)
 
     # SYNPY-717
-    df['datetime64'] = df['datetime64'].apply(lambda x: pd.Timestamp(x).tz_localize('UTC'))
+    if os.name == 'nt':
+        df['datetime64'] = pd.to_datetime(df['datetime64'], utc=True)
+    else:
+        df['datetime64'] = pd.to_datetime(df['datetime64'], unit='ms', utc=True)
```
mattfazza, line 239: I would add a comment that this is a check for Windows.

synapseclient/table.py
mattfazza: I'm late to this conversation, but what's the problem of requiring pandas?

mattfazza (commented):
Left a couple of comments.  Everything looks fine though.

synapseclient/table.py
thomasyu888: @mattfazza The python package actually doesn't require the installation of pandas.  So the parts of the client where pandas is needed `def asDataFrame(...)` actually uses this: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/table.py#L1455

tests/integration/synapseclient/test_tables.py
```diff
@@ -233,7 +233,10 @@ def test_tables_pandas(syn, project):
     df['string_'] = df['string_'].transform(str)
 
     # SYNPY-717
-    df['datetime64'] = df['datetime64'].apply(lambda x: pd.Timestamp(x).tz_localize('UTC'))
+    if os.name == 'nt':
```
thomasyu888, line 236:
```suggestion
# This is a check for windows
    if os.name == 'nt':
```

synapseclient/table.py
thomasyu888, line 1759:
```suggestion
            test_import_pandas()
```

tests/integration/synapseclient/test_tables.py
```diff
@@ -233,7 +233,11 @@ def test_tables_pandas(syn, project):
     df['string_'] = df['string_'].transform(str)
 
     # SYNPY-717
-    df['datetime64'] = df['datetime64'].apply(lambda x: pd.Timestamp(x).tz_localize('UTC'))
+# This is a check for windows
+    if os.name == 'nt':
```
thomasyu888, line 237:
```suggestion
    # This is a check for windows
    if os.name == 'nt':
```

synapseclient/table.py
```diff
@@ -1756,6 +1756,15 @@ def from_data_frame(cls, schema, df, filepath=None, etag=None, quoteCharacter='"
 
             f = io.open(filepath, mode='w', encoding='utf-8', newline='')
 
+            test_import_pandas()
```
thomasyu888, line 1759:
```suggestion
            test_import_pandas()
            import pandas as pd
```

emmanmills (approved):

PR #866, opened 2021-06-01
Support DATE_LIST, ENTITYID_LIST, USERID_LIST table columns

thomasyu888:


mattfazza (approved):
Pretty straight forward PR (:

tests/unit/synapseclient/unit_test_tables.py
```diff
@@ -1160,13 +1163,19 @@ def test_as_data_frame__list_columns(self):
...
             'fill': [[], []],
         })
 
         headers = [
             SelectColumn(name='string_list', columnType='STRING_LIST'),
             SelectColumn(name='integer_list', columnType='INTEGER_LIST'),
             SelectColumn(name='boolean_list', columnType='BOOLEAN_LIST'),
+            SelectColumn(name='date_list', columnType='DATE_LIST'),
+            SelectColumn(name='entity_list', columnType='ENTITYID_LIST'),
+            SelectColumn(name='user_list', columnType='USERID_LIST'),
```
emmanmills, line 1178: Will these panda df column names ('string_list', 'integer_list' etc.) be re-used anywhere outside the unit test? I wonder if they should be saved as constants so it's easier to maintain in the future.

emmanmills: Can we do something like this for the list of items that need to be cast?
```
elif columnType == 'INTEGER_LIST':
    type = json.loads(field)
    result.append([int(item) for item in type])
```

thomasyu888: The values returned from synapse are already in integer format:
```
import json
json.loads('[1,2,3]')  
[1, 2, 3]
```

emmanmills: If the lists don't need to be cast, I don't see any issue with the function you mentioned. Could you elaborate more about your concern?

thomasyu888: It was just more a response to your question:

```
result.append([int(item) for item in type])
```

We don't need to do this because the values are cast to integers already.

emmanmills: Oh, I am actually curious about why you said previously: 

`this function makes it a bit difficult: https://github.com/thomasyu888/synapsePythonClient/blob/add-entity-user-list/synapseclient/table.py#L499-L536.`

thomasyu888: Ahh, I was hoping that by maintaining a global variable of all the "synapse list columns" I wouldn't need to update it manually in other areas, but due to this function, it seems like I still have to.

emmanmills: I see. We only need to update 2 places rather than many other places. So I think this is still an improvement.

emmanmills (approved):

Issue #875, opened 2021-08-03
Passing a pandas dataframe with a column called "read" breaks the type parsing in as_table_columns()


esurface:
# Bug Report

## Operating system

MacOSX

## Client version
```
2.3.1
```

## Description of the problem

Symptom:
Passing a Pandas Dataframe with the column labeled "read" to as_table_columns() throws a 'TypeError' when calling `_csv_to_pandas_df()`. 

Bug:
The code tries to parse the value as a string instead of a Pandas DF in this code here:
```
    # filename of a csv file
    # in Python 3, we can check that the values is instanceof io.IOBase
    # for now, check if values has attr `read`
    if isinstance(values, str) or hasattr(values, "read"):   <----- hasattr(values, "read") is True!
        df = _csv_to_pandas_df(values)                                <----- _csv_to_pandas_df() returns a TypeError
    # pandas DataFrame
    if isinstance(values, pd.DataFrame):
        df = values.                                                                <----- Should assign df here instead
```

Catching this in the debugger,  I see that the input parameter `values` has the attr `read` and so the code tries to parse it as a string in _csv_to_pandas_df:

```
>>>values["read"]
0    
Name: read, dtype: object
>>>isinstance(values, str)
False
>>>hasattr(values, "read")
True
```

### To Reproduce
Note, 'Table(schema, df)' calls as_table_columns() internally:
```
import pandas as pd
from synapseclient import Schema, Column, Table, Row, RowSet, as_table_columns, build_table, table

project = 'synXXXXXXXX'
df = pd.DataFrame([{'read': '0'}])
columns = []
for column in df.columns:
     columns.append(Column(name=column, columnType='STRING'))
schema = Schema('TEST_TABLE', columns, parent=project)
table = Table(schema, df)
```

### Expected behavior

Users should be able to pass a pandas dataframe with a column called "read" to the function

### Actual behavior

If you care to see the error:

```
  File "/Users/esurface/opt/miniconda2/envs/py3/lib/python3.9/site-packages/pandas/io/common.py", line 554, in get_handle
    if _is_binary_mode(path_or_buf, mode) and "b" not in mode:
  File "/Users/esurface/opt/miniconda2/envs/py3/lib/python3.9/site-packages/pandas/io/common.py", line 859, in _is_binary_mode
    return isinstance(handle, binary_classes) or "b" in getattr(handle, "mode", mode)
TypeError: argument of type 'method' is not iterable
```

[...]

PR #876, opened 2021-08-03
Check parameter is a dataframe first in as_table_comlums to avoid bug #875

esurface:
Fix to bug #875


thomasyu888:
Thanks for your contribution @esurface.  This fix should work, I just need to get the correct permissions on this repo to approve running the tests.

PR #880, opened 2021-09-21
[SYNPY-737] Fix integration tests

thomasyu888:


tests/integration/synapseclient/test_command_line_client.py
```diff
@@ -82,9 +82,9 @@ def test_command_line_client(test_state):
                  'synapse',
                  '--skip-checks',
                  'create',
-                 '-name',
+                 '--name',
                  str(uuid.uuid4()),
-                 '-description',
+                 '--description',
```
mattfazza, line 87: Does PEP8 have a standard on CL arguments?  I see what you're doing here, just wondering.

mattfazza (approved):
Simple PR!

tests/integration/synapseclient/test_command_line_client.py
thomasyu888: Source from here as command line standards: https://en.wikipedia.org/wiki/Command-line_interface#Arguments

PR #884, opened 2021-10-05
V2.5.0 rc

thomasyu888:


docs/news.rst
```diff
@@ -2,6 +2,68 @@
...
+
+        # from the command line
+        synapse manifest ./ --parentId syn123 --manifestFile ./manifest.tsv
+
+- Added ability to generate manifest file from local directory structure
+
+  .. code-block:: bash
+
+        # from the command line
+        synapse manifest ./ --parentId syn123 --manifestFile ./manifest.tsv
```
thomasyu888, line 25: @BrunoGrandePhD  What do you think creating a default for the `--manifestFile` parameter?  Also should we rename it `--manifest-file` as we are slowly trying to move away from camel case?  I can make these changes, just wanted your thoughts.

docs/news.rst
```diff
@@ -2,6 +2,68 @@
...
+
+- Added ability to describe csv and tsv files stored in Synapse.
+
+  .. code-block:: bash
+
+        # from python
+        import synapseutils
+        synapseutils.describe(syn=syn, entity="syn12345", mode="string")
```
thomasyu888, line 33: @mattfazza I just realized now that there isn't a check for when you put in a random string for mode.  Should there be a check for if a user inputs something other than "string" or "object"?

On another note, should we just leverage the synapse logging system and always return an object.  If the user doesn't want the output printed, they can instantiate the synapse connection with `silent=True`? Thoughts?

docs/news.rst
BrunoGrandePhD: I'm happy with renaming the argument name to kebab or snake case. As for a default value, I would like it to output to stdout by default. I can do this later this week once I get a high-priority task off of my plate, but I'm fine if you make the update.

PR #885, opened 2021-10-07
[SYNPY-1165]/[SYNPY-1188] Small generate manifest improvements

thomasyu888:
* stdout the manifest unless manifest path is specified
* Update cli parameter case

Example:
```
$ synapse manifest . --parentid syn4990358
path	parent
./vicc_assay.yaml	syn4990358
./assay_information.yaml	syn4990358
./temp.tsv	syn4990358

$
```


thomasyu888:
Actually, need to add tests still.

PR #886, opened 2021-10-07
[SYNPY-1184] Add `includeTypes` to `synapseutils.walk()`

BrunoGrandePhD:
A small but incredibly useful quality of life tweak. Can this be included in 2.5? Pretty please? 

In cases where a mix of non-folder entities are expected, this avoids the user from having to make individual requests for each Synapse ID to figure out the type.


BrunoGrandePhD:
Confirming that this will be slated for v2.6 so we can write proper tests and documentation in addition to adding a CLI argument.

thomasyu888:
Adding cli argument out of scope of this PR. open new JIRA for that.

tests/unit/synapseutils/unit_test_synapseutils_walk.py
```diff
@@ -0,0 +1,85 @@
...
+        {"id": "syn124", "concreteType": "org.sagebionetworks.repo.model.Folder", "name": "test_folder"}
+    ]
+    child_list = [
+        [{"id": "syn2222", "concreteType": "File", "name": "test_file"},
+         {"id": "syn124", "concreteType": "org.sagebionetworks.repo.model.Folder", "name": "test_folder"}],
+        [{"id": "syn22223", "conreteType": "File", "name": "test_file_2"}]
+    ]
+    expected = [
+        (('parent_folder', 'syn123'), [('test_folder', 'syn124')], [('test_file', 'syn2222')]),
+        (('parent_folder/test_folder', 'syn124'), [], [('test_file_2', 'syn22223')])
```
thomasyu888, line 49:
```suggestion
        ((os.path.join('parent_folder', 'test_folder'), 'syn124'), [], [('test_file_2', 'syn22223')])
```

tests/unit/synapseutils/unit_test_synapseutils_walk.py
```diff
@@ -0,0 +1,85 @@
+from unittest.mock import patch
```
thomasyu888, line 1:
```suggestion
import os
from unittest.mock import patch
```

tests/unit/synapseutils/unit_test_synapseutils_walk.py
thomasyu888: Must use `os` package to account for windows environment.

thomasyu888 (approved):

PR #887, opened 2021-10-11
[SYNPY-1167] Small improvements to synapseutils describe

thomasyu888:
* Change return type to be consistent
* log stats to stderr
* Update dtype to be a string that is returned instead of numpy `dtype` object
* Complete tests


thomasyu888:
> We do have a failing test in `FAILED tests/unit/synapseutils/unit_test_synapseutils_describe.py::TestDescribe::test_describe_with_mixed_series`

Interesting.  This is happening only in 3.6 due to a `TypeError: unhashable type: 'list'` error, therefore `related: {}`, but for python>3.6 there isn't any issue...

PR #890, opened 2021-10-25
[SYNPY-1196] Expose `forceVersion` parameter on wrapper function

BrunoGrandePhD:
Currently, if you use `changeFileMetaData()` to rename how a file is downloaded as, it creates a new version no matter what (even though the file contents haven't changed). I would prefer not having an initial version with the wrong name, so I would like to set `forceVersion` to `False`. 

What do others think? If there isn't any pushback, I can add relevant test cases and mark this PR as ready for review. 

This can wait until the next release.


thomasyu888:
I like this - the tests should be pretty straight forward to add.

BrunoGrandePhD:
I added a simple test that focuses on the new `forceVersion` argument. It can probably be fleshed out, but I don't have time to develop a full test suite for this function.

PR #891, opened 2021-11-01
[SYNPY-1197] Schema isn't always a entity

thomasyu888:
When storing tables, it's not always going to be a Schemas object that gets passed in.  A synapse id is also possible.

### DIFF
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -1336,7 +1336,7 @@ def Table(schema, values, **kwargs):
     Combine a table schema and a set of values into some type of Table object
     depending on what type of values are given.
 
-    :param schema: a table :py:class:`Schema` object
+    :param schema: a table :py:class:`Schema` object or Synapse Id of Table.
     :param values: an object that holds the content of the tables
                       - a :py:class:`RowSet`
                       - a list of lists (or tuples) where each element is a row
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -1776,13 +1776,14 @@ class CsvFileTable(TableAbstractBaseClass):
 
             test_import_pandas()
             import pandas as pd
-            for col in schema.columns_to_store:
-                if col['columnType'] == 'DATE':
-                    def _trailing_date_time_millisecond(t):
-                        if isinstance(t, str):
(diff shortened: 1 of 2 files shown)


synapseclient/table.py
```diff
@@ -1776,13 +1776,14 @@ def from_data_frame(cls, schema, df, filepath=None, etag=None, quoteCharacter='"
...
-                        if isinstance(t, str):
-                            return t[:-3]
-                    df[col.name] = pd.to_datetime(df[col.name], errors='coerce').dt.strftime('%s%f')
-                    df[col.name] = df[col.name].apply(lambda x: _trailing_date_time_millisecond(x))
+            if isinstance(schema, Schema):
+                for col in schema.columns_to_store:
+                    if col['columnType'] == 'DATE':
+                        def _trailing_date_time_millisecond(t):
+                            if isinstance(t, str):
+                                return t[:-3]
```
mattfazza, line 1784: Is this trimming the milliseconds?

thomasyu888:
Thanks @BrunoGrandePhD .  I added the instructions in the JIRA ticket.

synapseclient/table.py
thomasyu888: This is kind of a hacky function.  By default the units datetime unit returned is in nano seconds, so its basically a divide by 1000.

BrunoGrandePhD (commented):
@thomasyu888: Thanks for providing the reprex on the JIRA ticket! 

The code diff looks good to me, but shouldn't we update the docs accordingly? 

https://python-docs.synapse.org/build/html/Table.html#synapseclient.table.Table

thomasyu888:
@BrunoGrandePhD good suggestion, what should the docs say?

BrunoGrandePhD:
The `schema` function argument currently says "a table Schema object". This should be clarified to something along the lines of "a table Schema object or the Synapse ID of an existing table".

thomasyu888:
@BrunoGrandePhD Done.  There are actually lots of places in the documentation that could probably be changed.  (Eventually...)

BrunoGrandePhD (approved):

PR #894, opened 2021-12-07
[SYNPY-1199] Implement low-level functions for JSON Schema API

BrunoGrandePhD:
Example usage: https://gist.github.com/BrunoGrandePhD/481249a94196cbdd47e8524dab7ddbbc

Install using:
```console
 pip install --force-reinstall git+https://github.com/Sage-Bionetworks/synapsePythonClient@bgrande/SYNPY-1199/json-schemas-low-level
```


ychae:
@milen-sage are you still able to review this by EOD today?

BrunoGrandePhD:
It has come to my attention that the `displayName` field in the user profile has been deprecated. Hence, new Synapse users won't have it and thus the `_loggedIn()` [function](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/client.py#L505-L507) doesn't work as expected; it returns `None` instead of the display name, which evaluates to `True`. I've opened a Synapse Platform ticket to see how to perform the same check now that `displayName` has been deprecated.

milen-sage:
@ychae, @afwillia is going through the JSONSchema client features, since that aligns with his work for this sprint.

synapseclient/client.py
```diff
@@ -658,6 +660,31 @@ def _print_transfer_progress(self, *args, **kwargs):
...
+    _services = {
+        "json_schema": "JsonSchemaService",
+    }
+
+    def get_available_services(self):
+        services = self._services.keys()
+        return list(services)
+
+    def service(self, service_name: str):
+        import synapseclient.services
```
thomasyu888, line 676: Is there a reason for this import to be here and not part of the global imports?

synapseclient/services/json_schema.py
```diff
@@ -0,0 +1,613 @@
...
+        json_schema = JsonSchemaVersion.from_response(self, raw_version)
+        return json_schema
+
+
+class JsonSchemaService:
+    def __init__(self, synapse: Synapse = None) -> None:
+        self.synapse = synapse
+
+    @wraps(Synapse.login)
+    def login(self, *args, **kwargs):
```
thomasyu888, line 395: Hmmm - what is your rationale for providing the option for each service to have its own login function, but also allowing users to specify their own synapse object?

thomasyu888 (commented):
Some linting issues and comments.

synapseclient/client.py
```diff
@@ -506,6 +506,8 @@ def _loggedIn(self):
                 if user['displayName'] == 'Anonymous':
                     return False
                 return user['displayName']
+            elif 'userName' in user:
```
thomasyu888, line 509: Do you use this function in this PR?  If not, it may be faster to pull the bug fix in another PR.

synapseclient/client.py
BrunoGrandePhD: To avoid circular imports. When I revisit this PR, I hope to revisit the import order and [method](https://stackoverflow.com/a/37126790) in the hope of moving this import statement to the top.

synapseclient/services/json_schema.py
BrunoGrandePhD: Because users can theoretically create a `JsonSchemaService` instance without running `synapse.service("json_schema")`, I wanted to make sure there was a method to "log in" a `JsonSchemaService` instance after the fact. Previously, I had a `JsonSchemaService.set_synapse()` method, but I find this login method more elegant.

synapseclient/client.py
BrunoGrandePhD: This was intended to unblock Anthony (see [here](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/894#issuecomment-1034176703)). I don't know if this issue with `_loggedIn()` come up elsewhere.

vpchung (commented):
In Slack, @BrunoGrandePhD and I discussed adding attribute validation to the API so that users are aware of invalid values as soon as possible, e.g. [organizationName](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/schema/CreateOrganizationRequest.html) should be at least 6 characters long.  The current behavior does not catch this after a user tries to create:

```
>>> foo = js.JsonSchemaOrganization("foo")
>>> foo.create()
  File "<stdin>", line 1, in <module>
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/services/json_schema.py", line 288, in create
    already_exists = self.get()
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/services/json_schema.py", line 257, in get
    raise e
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/services/json_schema.py", line 251, in get
    response = self.service.get_organization(self.name)
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/services/json_schema.py", line 428, in wrapper
    result = func(self, *args, **kwargs)
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/services/json_schema.py", line 446, in get_organization
    response = self.synapse.restGET(
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/client.py", line 4091, in restGET
    response = self._rest_call('get', uri, None, endpoint, headers, retryPolicy, requests_session, **kwargs)
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/client.py", line 4075, in _rest_call
    self._handle_synapse_http_error(response)
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/client.py", line 4045, in _handle_synapse_http_error
    exceptions._raise_for_status(response, verbose=self.debug)
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/core/exceptions.py", line 160, in _raise_for_status
synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error: 
Organization name must be at least 6 characters
```

This is currently in my backlog TODOs.

PR #896, opened 2022-01-03
[SYNPY-1204, SYNPY-1206] Remove support for py 3.6 and add support for 3.10

thomasyu888:


mattfazza (approved):
As long as it builds fine, this is good!

PR #898, opened 2022-02-22
[SYNPY-1212] include datasets, submissionview, and materializedviews in getChildren

thomasyu888:


BrunoGrandePhD (approved):
I'm surprised there aren't other tests that need to be updated. From what I can tell, `getChildren()` returns dictionaries and not Synapse object instances, right?

thomasyu888:
I was surprised as well, but I did a search on `getChildren` and there didn't seem to be other places this needed to be updated.  I'll do another check.

Issue #899, opened 2022-02-23
Feature Request: Filter capability for syncFromSynapse


georggrab:
# Feature Request

Just started using Synapse for a dataset which contains data for multiple image analysis use-cases. Our team only needs a small subset of available data for the time being. As such, it would be great to have some sort of filtering capability in `syncFromSynapse` to only specify files that are needed. I will attach a pull request to elaborate.
Thanks!

### Expected behavior

`syncFromSynapse` allows for filtering capability:

```python
import synapseutils

match_only = ['instrument_instances.png', 'raw.png']
synapseutils.syncFromSynapse(syn, entity=project_id, path=local_folder, onlyMatching=match_only)
```

thomasyu888:
Thanks for your feature contribution!  We will discuss internally about this contribution!

PR #900, opened 2022-02-23
Add onlyMatching argument to syncFromSynapse to filter downloads

talkdirty:
This is to elaborate on issue #899 with a concrete PoC with desired functionality. I don't know if the approach is the desired / most efficient one!

### DIFF
--- a/synapseutils/sync.py
+++ b/synapseutils/sync.py
@@ -46,7 +46,7 @@ def _sync_executor(syn):
 
 
 def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFiles=None, followLink=False,
-                    manifest="all", downloadFile=True):
+                    manifest="all", downloadFile=True, onlyMatching=None):
     """Synchronizes all the files in a folder (including subfolders) from Synapse and adds a readme manifest with file
     metadata.
 
@@ -68,6 +68,11 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
 
     :param downloadFile Determines whether downloading the files.
                         Defaults to True
+
+    :param onlyMatching Determines list of regexes to be matched against files.
+                        Only if at least one file matches the regex, it will
+                        be downloaded.
+                        Defaults to None
 
     :returns: list of entities (files, tables, links)
 
@@ -104,7 +109,7 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
     # 2 threads always, if those aren't available then we'll run single threaded to avoid a deadlock
     with _sync_executor(syn) as executor:
         sync_from_synapse = _SyncDownloader(syn, executor)
-        files = sync_from_synapse.sync(entity, path, ifcollision, followLink, downloadFile, manifest)
+        files = sync_from_synapse.sync(entity, path, ifcollision, followLink, downloadFile, manifest, onlyMatching)
 
     # the allFiles parameter used to be passed in as part of the recursive implementation of this function
     # with the public signature invoking itself. now that this isn't a recursive any longer we don't need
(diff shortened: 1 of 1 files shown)


synapseutils/sync.py
```diff
@@ -69,6 +69,11 @@ def syncFromSynapse(syn, entity, path=None, ifcollision='overwrite.local', allFi
     :param downloadFile Determines whether downloading the files.
                         Defaults to True
 
+    :param onlyMatching Determines list of regexes to be matched against files.
+                        Only if at least one file matches the regex, it will
+                        be downloaded.
+                        Defaults to None
```
BrunoGrandePhD, line 75: What do you think of simplifying this argument to a single regex since multiple patterns can be combined using `|`? My rationale is that we can use [`re.compile()`](https://docs.python.org/3.10/library/re.html#re.compile) downstream on a single expression to be efficient. I'm guessing evaluating a single pre-compiled regex will be faster than compiling and matching a regex (which is what happens when you use `re.match()`) in a for-loop.

BrunoGrandePhD: We could either expect a pre-compiled regex or perform the compilation in `syncFromSynapse()`.

synapseutils/sync.py
```diff
@@ -256,22 +261,39 @@ def sync(self, entity, path, ifcollision, followLink, downloadFile=True, manifes
...
-                entity = self._syn.get(
-                    entity_id,
-                    downloadLocation=path,
-                    ifcollision=ifcollision,
-                    followLink=followLink,
-                    downloadFile=downloadFile,
-                )
+            
+                file_matches = True
+                if onlyMatching is not None:
```
BrunoGrandePhD, line 274: **Nit (_i.e._ [optional](https://google.github.io/eng-practices/review/reviewer/standard.html#:~:text=Reviewers%20should%20always%20feel%20free,they%20could%20choose%20to%20ignore.)):** This can be simplified to `if onlyMatching:` or perhaps more defensively as `if isinstance(onlyMatching, re.Pattern)` if we expect pre-compiled regex objects.

synapseutils/sync.py
```diff
@@ -256,22 +261,39 @@ def sync(self, entity, path, ifcollision, followLink, downloadFile=True, manifes
...
+                    file_matches = False
+                    entity_meta = self._syn.get(
+                        entity_id,
+                        downloadFile=False,
+                    )
+                    for regex in onlyMatching:
+                        if re.match(regex, entity_meta.name) is not None:
+                            file_matches = True
```
BrunoGrandePhD, line 282: As per my above comment, I think this part can be made more efficient if `re.compile()` was used upstream on a single regex (optionally, with multiple patterns separated by `|`) and then matched once here, like:

```
onlyMatching.match(entity_meta.name)
```

synapseutils/sync.py
```diff
@@ -256,22 +261,39 @@ def sync(self, entity, path, ifcollision, followLink, downloadFile=True, manifes
...
+
+                if file_matches:
```
BrunoGrandePhD, line 284: I don't think we need the `file_matches` flag. Couldn't we simply short-circuit this function with an early return statement if `onlyMatching` is set and there are no matches? This way, we avoid indenting the `self._syn.get(..., downloadFile=downloadFile)` bit. We would have to run the `parent_folder_sync.update()` statement you have below immediately before the return statement. 

I'll happily elaborate if my suggestion isn't clear.

synapseutils/sync.py
BrunoGrandePhD: Can we think of additional use cases other than filtering on the entity name? I'm just wondering if we can generalize this parameter to being a dictionary mapping entity properties/annotations (keys) to pre-compiled regular expressions (values)? 

@thomasyu888: What do you think?

thomasyu888:
Thanks @talkdirty for your initial contributions. If you have no objections, we have decided to take this on by incorporating Bruno's suggestions.   I will be merging this if we don't hear from you and completing the feature by adding tests etc.

synapseutils/sync.py
thomasyu888: Shifting conversation to here: https://sagebionetworks.jira.com/browse/SYNPY-1236.  @vpchung will be tackling this when she has the time.

PR #904, opened 2022-03-01
[SYNPY-1176] Add in check for unique entity name and parent

thomasyu888:
say for example someone has this manifest:

```
path    parent
/mypath1/file.txt    syn123
/mypath2/file.txt    syn123
```

Due to Synapse storing things within a folder via synapse entity name, you can see that even though the filepaths are technically unique, and those files could be different, both of these files will be stored in folder `syn123` as `file.txt`.  This is an issue as you can cause silent errors, or could cause the code to store to the same entity at the same time causing a concurrent update error.  To test:

1. Create a new Python virtualenv
2. Install this specific branch
    ```
    pip install pandas
    pip install https://github.com/thomasyu888/synapsePythonClient/archive/refs/heads/synpy-1176-imrpove-sync-error.zip
    ```
3. Add code here...


BrunoGrandePhD (approved):
🚀

thomasyu888:
Incorporated suggestions from [slack](https://sagebionetworks.slack.com/archives/CMSGC6ATD/p1647448008000049)

PR #905, opened 2022-03-01
[SYNPY-1189] Add documentation for maximumListLength

thomasyu888:


synapseclient/table.py
```diff
@@ -1026,21 +1026,24 @@ class Column(DictObject):
...
+    :var id:                  An immutable ID issued by the platform
+    :param columnType:        The column type determines the type of data that can be stored in a column. It can be any
+                              of: "STRING", "DOUBLE", "INTEGER", "BOOLEAN", "DATE", "FILEHANDLEID", "ENTITYID", "LINK",
+                              "LARGETEXT", "USERID". For more information, please see:
+                              https://docs.synapse.org/rest/org/sagebionetworks/repo/model/table/ColumnType.html
+    :param maximumSize:       A parameter for columnTypes with a maximum size. For example, ColumnType.STRINGs have a
+                              default maximum size of 50 characters, but can be set to a maximumSize of 1 to 1000
+                              characters.
+    :param maximumListLength: Required if using a columnType with a "_LIST" suffix. Describes the maximum number of
+                              values that will appear in that list. Value range 1-100 inclusive. Default 100
```
BrunoGrandePhD, line 1038: What happens if I set the value below the number of items in the list? Are the remaining items concatenated or dropped?

thomasyu888: The API throws an error if you try to set a maxlistlength that is less than the previous one you have set.   Similarly you can't set a number of values in the list greater than this limit.

BrunoGrandePhD: Sorry, I misunderstood the intent of this argument. I thought it was for retrieving list-columns, not specifying them in a table schema.

BrunoGrandePhD (approved):
🚀

PR #911, opened 2022-04-15
[SYNPY-1207] Add Dataset support

vpchung:
This PR will add initial support for the new Datasets entity (mentioned by this [JIRA ticket](https://sagebionetworks.jira.com/browse/SYNPY-1207)), including methods such as:

- adding a single DatasetItem
- adding a list of DatasetItems
- removing a single DatasetItem
- ~~adding a Synapse Folder to the Dataset~~
- ~~adding a list of Synapse Folders to the Dataset~~

### DIFF
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -842,6 +842,9 @@ class Dataset(SchemaBase):
         )
         if dataset_items:
             self.add_items(dataset_items)
+    
+    def __len__(self):
+        return len(self.properties.dataset_items)
 
     def add_item(self, dataset_item: Dict[str, str]):
         """
@@ -882,6 +885,12 @@ class Dataset(SchemaBase):
                     break
         else:
             raise ValueError("Not a Synapse ID: %s" % str(item_id))
+
+    def has_item(self, item_id):
+        """
+        :param item_id: a single dataset item Synapse ID
+        """
+        return any(item['entityId'] == item_id for item in self.properties.dataset_items)
 
     def _before_synapse_store(self, syn):
         # Remap `dataset_items` back to `items` before storing (since `items`
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -822,14 +822,14 @@ class Dataset(SchemaBase):
         # Add all Files within a Folder to the Dataset (recursive=True)
         dataset.addFolder("syn789", recursive=True)
 
-    To create a snapshot of the Dataset, use 
+    To create a snapshot of the Dataset, use
     :py:classmethod:`synapseclient.client.create_snapshot_version`.
 
     Example::
         syn = synapseclient.login()
         syn.create_snapshot_version(
-            dataset.id, 
-            label="v1.0", 
+            dataset.id,
+            label="v1.0",
             comment="This is version 1")
     """
     _synapse_entity_type: str = "org.sagebionetworks.repo.model.table.Dataset"
@@ -886,13 +886,17 @@ class Dataset(SchemaBase):
             raise ValueError("Not a synapse ID: %s" % str(item_id))
 
     def addFolderToDataset(self, folder_id: str, recursive: bool = False):
+        """Adds all File entities from a Folder Synapse ID.
+
+        Files added will default to the latest version. If specific version
+        is desired, use :py:classmethod::`Dataset.addItem` instead.
+
         :param folder_id:   Synapse Folder ID
         :param recursive:   Recursively add files (False by default)
         """
-        if recursive:
-            pass
-        else:
-            pass
+        items_to_add = []
+
+        self.addItems(items_to_add)
 
     def addFoldersToDataset(self, folder_ids: List[str], recursive: bool = False):
         """Adds all File entities from a list of Folder Synapse IDs.
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -842,6 +842,18 @@ class Dataset(SchemaBase):
         for dataset_item in dataset_items:
             self.addItem(dataset_item)
 
+    def removeItem(self, item_id):
+        """
+        :param item_id: a single dataset item Synapse ID
+        """
+        if isinstance(item_id, str) and item_id.startswith("syn"):
+            for i, curr_item in enumerate(self.properties.items):
+                if curr_item.get('entityId') == item_id:
+                    del self.properties.items[i]
+                    break
+        else:
+            raise ValueError("Not a synapse ID: %s" % str(item_id))
+
 
 class ViewBase(SchemaBase):
     """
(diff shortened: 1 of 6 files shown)


thomasyu888:
Thanks for the great contribution @vpchung! I'll take a closer look soon, but at first glance, I like the usage of your class methods to add single items and add folders to the dataset.

vpchung:
@thomasyu888 is there a plan to add type-hinting to the Python client? If so, what is the earliest version of Python we are supporting, a.k.a. would we need to import `typing`?

thomasyu888:
good question. Yes, we should support type hinting and we support 3.7 to 3.10

vpchung:
**To discuss:**
Should the client allow users to add private Files to a Dataset that they don't have permissions for?

Currently, adding a private File to a Dataset will work.  When printing from the client side, I can see items that I don't have access to (in this case, syn29461520):

```python
>>> dataset = syn.get("syn29430164")
>>> print(dataset)
Dataset: Pokedex (syn29430164)
  columns_to_store=[]
properties:
  ...
  items=[{'entityId': 'syn20685093', 'versionNumber': 1}, {'entityId': 'syn29461520', 'versionNumber': 1}]
  ...
annotations:
```

On the web UI, the private Files do not show (makes sense):
Screen Shot 2022-04-16 at 1 28 43 AM

But editing items _does_ show that there are private Files in the Dataset:
Screen Shot 2022-04-16 at 1 28 51 AM

vpchung:
Related to this [PR's comment](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/914#issuecomment-1101859500), the `items` property will no longer be used in this client.  Instead, files will be added to `datasetItems`, and will only convert to `items` when storing to Synapse.  Approaching it this way will enable users to all dictionary methods, including `items()`.

synapseclient/table.py
```diff
@@ -780,6 +781,112 @@ def __init__(self, name=None, columns=None, parent=None, definingSQL=None, prope
...
+        """
+        if isinstance(dataset_item, dict):
+            required_keys = {'entityId', 'versionNumber'}
+            if required_keys == dataset_item.keys():
+                self.properties.datasetItems.append(dataset_item)
+            elif required_keys - dataset_item.keys():
+                raise LookupError("DatasetItem missing a required property: %s" %
+                                  str(required_keys - dataset_item.keys()))
+
+            # FIXME? Having other key-value pair in the dict doesn't throw any
```
thomasyu888, line 861: Oh interesting.  So as an item, you can have:
```
{"bogus": "trial"}
```
And the API won't throw an error?

tests/integration/synapseclient/test_tables.py
```diff
@@ -227,6 +228,21 @@ def test_materialized_view(syn, project):
...
 
 
+def test_dataset(syn, project):
+    cols = [Column(name='id', columnType='ENTITYID'),
+            Column(name='name', columnType='STRING')]
+
+    dataset = Dataset(
+        name="Pokedex",
+        parent=project,
+        datasetItems=[{'entityId': "syn20685093", 'versionNumber': 1}],
```
thomasyu888, line 238: what is this entity id?

tests/unit/synapseclient/unit_test_tables.py
```diff
@@ -171,6 +172,17 @@ def test_materialized_view():
     assert mat_view.properties.columnIds == []
 
 
+def test_dataset():
+    dataset = Dataset(
```
thomasyu888, line 176: Do we need to test what happens when `items` is passed in?

vpchung: Agreed. The tests definitely need to be expanded.

tests/integration/synapseclient/test_tables.py
vpchung: It's to a sample file (which I realize is not viewable by the general public). Is there a list of official entities I can use for testing? Happy to just share the sample file as well.

synapseclient/table.py
vpchung: I think the API will throw an error if either `entityId` and/or `versionNumber` is missing. But it will not throw an error if there are extra item pairs, e.g.

```python
>>> dataset = syn.get("syn29430164")
>>> dataset.properties.datasetItems.append({'entityId': "syn123", 'versionNumber': 1, 'bogus': "trial"})
>>> dataset.datasetItems
[{'entityId': 'syn20685093', 'versionNumber': 6}, {'entityId': 'syn29461520', 'versionNumber': 1}, {'entityId': 'syn123', 'versionNumber': 1, 'bogus': 'trial'}]
>>> 
>>> 
>>> dataset = syn.store(dataset)
>>> dataset.datasetItems
[{'entityId': 'syn20685093', 'versionNumber': 6}, {'entityId': 'syn29461520', 'versionNumber': 1}, {'entityId': 'syn123', 'versionNumber': 1}]
```

Again, not sure if this case should even be handled, since the extras are not kept after storing.

thomasyu888: I think that's fine. We can just check for if the two requires keys are there but not for extras

synapseclient/table.py
```diff
@@ -780,6 +781,125 @@ def __init__(self, name=None, columns=None, parent=None, definingSQL=None, prope
...
+            raise ValueError("Not a Synapse ID: %s" % str(item_id))
+
+    def has_item(self, item_id):
+        """
+        :param item_id: a single dataset item Synapse ID
+        """
+        return any(item['entityId'] == item_id for item in self.properties.dataset_items)
+
+    def _before_synapse_store(self, syn):
+        super()._before_synapse_store(syn)
```
thomasyu888, line 896: Great use of super here.

synapseclient/table.py
```diff
@@ -780,6 +781,125 @@ def __init__(self, name=None, columns=None, parent=None, definingSQL=None, prope
...
+                    del self.properties.dataset_items[i]
+                    break
+        else:
+            raise ValueError("Not a Synapse ID: %s" % str(item_id))
+
+    def has_item(self, item_id):
+        """
+        :param item_id: a single dataset item Synapse ID
+        """
+        return any(item['entityId'] == item_id for item in self.properties.dataset_items)
```
thomasyu888, line 893: Can a dataset contain an entity of multiple versions? That is, can I have:

(Syn123, version1),(syn123,version2) 

In the dataset?

vpchung: No, it can't, which the API handles:

```python
synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error: 
Each dataset item must have a unique entity ID.  Duplicate: syn11111111
```

thomasyu888: Great. Thanks!

synapseclient/table.py
```diff
@@ -780,6 +781,119 @@ def __init__(self, name=None, columns=None, parent=None, definingSQL=None, prope
...
+        self.properties.setdefault('dataset_items', [])
+        super(Dataset, self).__init__(
+            name=name, columns=columns, properties=properties,
+            annotations=annotations, local_state=local_state, parent=parent,
+            **kwargs
+        )
+        if dataset_items:
+            self.add_items(dataset_items)
+
+    def __len__(self):
```
thomasyu888, line 846: Love this addition.

vpchung: Oh yay!  I wasn't sure if overwriting the dunder method was okay but in my mind, len() should return the number of items in the dataset, rather than the number of entity properties.

thomasyu888 (approved):

Issue #913, opened 2022-04-16
Update Entity class to be compatible with the new Dataset entity


vpchung:
One of the fields of the Dataset data model is named, `items`, which coincides with a Python builtin: `dict.items()`.  This results in a `TypeError` when trying to print a representation of the entity:

```python
>>> dataset = syn.get("syn123")
>>> repr(dataset)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/entity.py", line 434, in __repr__
    self.properties.items(),
TypeError: 'list' object is not callable
>>> 
```

Looking at the Entity class, the `__repr__()` should be updated as to not call on `self.properties.items()`, as to not confuse Python what it is we want (key-value pairs of `self.properties` vs. the actual property named `items`).

PR #914, opened 2022-04-16
[SYNPY-1219] update Entity __init__ to be compatible with Dataset

vpchung:
Due to there being a property named `items` in the Dataset schema, calling `self.properties.items()` results in a TypeError in the `__repr__()` method.

This PR will address #913 by utilizing the good ol' fashion way of getting a dictionary value by using its key instead.

**Before fix:**
```python
>>> dataset = syn.get("syn29430164")
>>> repr(dataset)
  File "<stdin>", line 1, in <module>
  File "/Users/vchung/Desktop/synapsePythonClient/synapseclient/entity.py", line 434, in __repr__
    self.properties.items(),
TypeError: 'list' object is not callable
>>> 
```

**After fix:**
```python
>>> repr(dataset)
"Entity(, modifiedBy='3393723', versionNumber=1, isLatestVersion=True, items=[{'entityId': 'syn20685093', 'versionNumber': 1}], parentId='syn20682803', name='Pokedex', etag='83c7c33b-253a-426a-b49d-e4cb57a67bdb', createdOn='2022-04-15T07:12:47.508Z', modifiedOn='2022-04-15T17:48:32.935Z', createdBy='3393723', columnIds=['81721', '81722', '81724'], concreteType='org.sagebionetworks.repo.model.table.Dataset', versionLabel='1', id='syn29430164', )"
```

If this is merged after #911 , then the proper entity type will be returned:

```python
>>> syn.get("syn29430164")
Dataset(columns_to_store=[], columnIds=['81721', '81722', '81724'], versionNumber=1, parentId='syn20682803', id='syn29430164', modifiedOn='2022-04-15T17:48:32.935Z', items=[{'entityId': 'syn20685093', 'versionNumber': 1}], name='Pokedex', createdOn='2022-04-15T07:12:47.508Z', createdBy='3393723', isLatestVersion=True, etag='83c7c33b-253a-426a-b49d-e4cb57a67bdb', concreteType='org.sagebionetworks.repo.model.table.Dataset', modifiedBy='3393723', versionLabel='1', )
```

Note: because of the updated code changes, a trailing comma will be present if the entity does not have annotations. Not sure if this is a big deal?

### DIFF
--- a/synapseclient/entity.py
+++ b/synapseclient/entity.py
@@ -139,6 +139,7 @@ See also:
 """
 
 import collections.abc
+import itertools
 import io
 import os
 import inspect
@@ -429,15 +430,12 @@ class Entity(collections.abc.MutableMapping):
         f.write(self.__class__.__name__)
         f.write("(")
         f.write(", ".join(
-            {"%s=%s" % (str(key), value.__repr__(),)
-             for key, value in list([k_v for k_v in self.__dict__.items()
-                                     if not (k_v[0] in ['properties', 'annotations'] or k_v[0].startswith('__'))])}))
-        f.write(", ")
-        f.write(", ".join(
-            {"%s=%s" % (str(key), self.properties[key].__repr__(),) for key in self.properties}))
-        f.write(", ")
-        f.write(", ".join(
-            {"%s=%s" % (str(key), value.__repr__(),) for key, value in self.annotations.items()}))
+            {"%s=%s" % (str(key), value.__repr__(),) for key, value in
+                itertools.chain(
+                    list([k_v for k_v in self.__dict__.items()
+                          if not (k_v[0] in ['properties', 'annotations'] or k_v[0].startswith('__'))]),
+                    self.properties.items(),
+                    self.annotations.items())}))
         f.write(")")
         return f.getvalue()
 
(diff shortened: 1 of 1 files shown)


vpchung:
After discussing with @thomasyu888 , we think the better approach is to not use the `items` property at all (at least, not in the client). The property name will instead be mapped to `datasetItems`.

thomasyu888 (approved):

synapseclient/entity.py
```diff
@@ -240,6 +240,9 @@ def __init__(self, properties=None, annotations=None, local_state=None, parent=N
                 if 'annotations' in properties and isinstance(properties['annotations'], collections.abc.Mapping):
                     annotations.update(properties['annotations'])
                     del properties['annotations']
+                if 'items' in properties:
```
thomasyu888, line 243: Can you add a comment like:
```
#HACK: items is returned from dataset API but items() is a method in python dict
```

vpchung: ahem, it's not a hack. imo, it's a pretty smart solution!

thomasyu888: Sorry - didn't mean for that to be interpreted as not a smart solution!  It is indeed!

PR #917, opened 2022-04-30
[SYNPY-1220] Add folder to dataset

vpchung:
This PR will add "advanced" features for the new Datasets entity (related PR: #911), including methods such as:

* adding a Synapse Folder to the Dataset (`add_folder()`)
* adding a list of Synapse Folders to the Dataset (`add_folders()`)
* truncating the Dataset of all current DatasetItems (`empty()`)

This PR also adds a new parameter, `force`, to the initial methods added in #911 that will enable force-adding of DatasetItems that already exist in the Dataset. (otherwise, error is thrown regarding duplicate IDs).


vpchung:
This might be easier to review once the other PR is merged, heh.

vpchung:
I tested `add_folder()` and `add_folders()` with two folders that has many sub-folders and files (n1=1734, n2=390), and they were relatively fast:

```python
>>> dataset.add_folder("syn29530156")
>>> start_time = time.time()
>>> dataset = syn.store(dataset)
>>> print(f"--- {time.time() - start_time} seconds ---")
--- 26.54198670387268 seconds ---
>>> 
>>> 
>>> dataset.empty()
>>> dataset = syn.store(dataset)
>>> dataset.add_folder("syn9850001")
>>> start_time = time.time()
>>> dataset = syn.store(dataset)
>>> print(f"--- {time.time() - start_time} seconds ---")
--- 2.8432188034057617 seconds ---
>>> 
>>> 
>>> dataset.empty()
>>> dataset = syn.store(dataset)
>>> dataset.add_folders(["syn29530156", "syn9850001"])
>>> start_time = time.time()
>>> dataset = syn.store(dataset)
>>> print(f"--- {time.time() - start_time} seconds ---")
--- 27.28242802619934 seconds ---
```

I also measured their memory usage with `tracemalloc` (since I'm using recursion to add the files) and that doesn't look too bad either:

```python
Top 3 lines for add_folder(...)
#1: json/decoder.py:353: 1119.0 KiB
    obj, end = self.scan_once(s, idx)
#2: synapseclient/table.py:979: 132.6 KiB
    files.append({
#3: http/client.py:1296: 35.9 KiB
    self.putrequest(method, url, **skips)
131 other: 270.7 KiB
Total allocated size: 1558.3 KiB

Top 3 lines for add_folders(..., force=True)
#1: json/decoder.py:353: 1371.7 KiB
    obj, end = self.scan_once(s, idx)
#2: synapseclient/table.py:979: 184.7 KiB
    files.append({
#3: requests/structures.py:43: 40.2 KiB
    self._store = OrderedDict()
140 other: 308.4 KiB
Total allocated size: 1904.9 KiB
```

Shoutout to @allaway and @avanlinden for supplying folders to test with!

synapseclient/table.py
```diff
@@ -829,75 +857,155 @@ class Dataset(SchemaBase):
...
+        self.__dict__.setdefault('folders_to_add', set())
+        self.__dict__.setdefault('force', False)
         super(Dataset, self).__init__(
             name=name, columns=columns, properties=properties,
             annotations=annotations, local_state=local_state, parent=parent,
             **kwargs
         )
+
+        # Add default Dataset columns if schema is not provided.
+        if not self.properties.columnIds and columns is None:
```
thomasyu888, line 884: Are default columns added in `_before_synaspe_store`?

synapseclient/table.py
```diff
@@ -829,75 +857,155 @@ class Dataset(SchemaBase):
...
+    _property_keys: List[str] = SchemaBase._property_keys + ['datasetItems']
+    _local_keys: List[str] = SchemaBase._local_keys + ['folders_to_add', 'force']
 
     def __init__(self, name=None, columns=None, parent=None, properties=None,
-                 annotations=None, local_state=None, dataset_items=None, **kwargs):
-        self.properties.setdefault('dataset_items', [])
+                 annotations=None, local_state=None, dataset_items=None,
+                 folder_ids=None, force=None, **kwargs):
+        self.properties.setdefault('datasetItems', [])
+        self.__dict__.setdefault('folders_to_add', set())
```
thomasyu888, line 875: Why the difference between `folders_to_add` and `folder_ids`

synapseclient/table.py
```diff
@@ -829,75 +857,155 @@ class Dataset(SchemaBase):
...
+            self.force = force
+
+    def add_folders(self, folder_ids: List[str], force: bool = False):
+        """
+        :param folder_ids:  a list of Synapse Folder IDs
+        :param force:       force add items from folders
+        """
+        if isinstance(folder_ids, list) or isinstance(folder_ids, set) or \
+                isinstance(folder_ids, tuple):
+            self.force = force
```
thomasyu888, line 972: Why add `force` as a class attribute?

synapseclient/table.py
vpchung: Mostly because `folder_ids` is a shorter variable for users to write :upside_down_face:

synapseclient/table.py
vpchung: Yep!  Its [super()](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/0f0316f57fe2c10e2d8153da67133588e9bb8046/synapseclient/table.py#L712) takes care of this.

synapseclient/table.py
vpchung: add_folder() and add_folders() do not actually add files into the dataset; it simply just tells the object what folders will need to be added.

The method that actually adds the file is the hidden helper function, _add_folder_files, which will need some way of knowing whether files should be overwritten.  Let me know if there's a better way to go around it.

synapseclient/table.py
vpchung: `folders` is even shorter, so maybe we should use that instead, heh.

synapseclient/table.py
thomasyu888: If it is handled in `super`, why should we add it here?

synapseclient/table.py
```diff
@@ -829,75 +857,155 @@ class Dataset(SchemaBase):
...
-            self.add_items(dataset_items)
+            self.add_items(dataset_items, force)
+        if folder_ids:
+            self.add_folders(folder_ids, force)
 
     def __len__(self):
-        return len(self.properties.dataset_items)
+        return len(self.properties.datasetItems)
+
+    def _check_needed_keys(syn, keys: List[str]):
```
thomasyu888, line 898: This seems like it can be a convenience function separate from the class. Why keep it in the class?

synapseclient/table.py
```diff
@@ -829,75 +857,155 @@ class Dataset(SchemaBase):
...
+
         syn = synapseclient.login()
         syn.create_snapshot_version(
             dataset.id,
             label="v1.0",
             comment="This is version 1")
     """
     _synapse_entity_type: str = "org.sagebionetworks.repo.model.table.Dataset"
-    _property_keys: List[str] = SchemaBase._property_keys + ['dataset_items']
+    _property_keys: List[str] = SchemaBase._property_keys + ['datasetItems']
```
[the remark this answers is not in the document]
vpchung: Unfortunately, I used `datasetItems` in the previous PR but had forgotten, so I had to address it here as well.  Otherwise, the methods would not work as expected.

synapseclient/table.py
vpchung: Interesting. So pass necessary keys as a param, then?

What other classes would need this check?

synapseclient/table.py
vpchung: I misunderstood your initial question - I thought you were asking whether these columns are added to the object during `_before_synapse_store`.

But no, none of the other table-like entities set a default schema.  For Dataset, I thought it would be necessary because otherwise, existing items would not show on the UI if a schema is not set.

synapseclient/table.py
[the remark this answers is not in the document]
vpchung: [Done](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/918).

synapseclient/table.py
thomasyu888: I like `folders`

synapseclient/entity.py
```diff
@@ -244,7 +244,7 @@ def __init__(self, properties=None, annotations=None, local_state=None, parent=N
                 # Re-map `items` to `datasetItems` to avoid namespace conflicts
                 # between Dataset schema and the items() builtin method.
                 if 'items' in properties:
-                    properties['datasetItems'] = properties['items']
+                    properties['dataset_items'] = properties['items']
```
thomasyu888, line 247: Is this a bug-fix with the current implementation?  We should add a test for this.

synapseclient/table.py
thomasyu888: I agree that the default columns should just be added since this is the behavior in the UI. You can use this function to actually get the default columns of a dataset:
```
syn._get_default_view_columns("datasets")
```

synapseclient/table.py
thomasyu888: I was referring to the fact that `_check_needed_keys(` didn't have a `self` as a parameter which makes it a little awkward in the class.  But, you're right, no other classes need this check.

synapseclient/table.py
thomasyu888: I see. Currently in the web client, the default is that dataset items are replaced so `force=True` by default.

synapseclient/table.py
vpchung: Ah, you are right (also - why did I pass `syn`....? ;;;).  Thanks for pointing it out - I'll make a note that this is a static method.

synapseclient/table.py
vpchung: Thinking it over, perhaps we should continue to offer the option to turn off `force` in the methods? I know we should be emulating the web UI's behavior, but if it were me and I was programmatically working with datasets, I think it'll be a convenient option to have.  Otherwise, the alternative would be to write a method that would turn on/off force-add, and requiring users to call that before adding any items.  If it were me, I would find that super inconvenient, as opposed to just doing `add_item(..., force=False)`.

Of course, another option is to not allow changing this behavior at all. That is, all duplicate items will overwrite the existing ones.  And if a user doesn't want that to happen, then they should use `has_item()` beforehand.  My only issue with this is that since the `add_item()` method already calls on `has_item()`, it'll be a little redundant (of course, the user wouldn't know that..... but we would :sweat_smile:)

synapseclient/table.py
vpchung: Oh, wow, didn't know that method existed, thanks!

thomasyu888 (approved):

PR #932, opened 2022-08-05
[SYNPY-226] Remove isInconsistent kwarg when results are returned as csv

thomasyu888:
Alternatively, an error can be thrown instead to let users know they have to specify resultsAs="rowset" to leverage the `isConsistent` parameter

This parameter is actually ignored in the backend.


danlu1 (commented):
Add deprecation warning to isConsistent parameter (line 3452)

danlu1 (approved):

PR #934, opened 2022-08-05
[SYNPY-560] set path if file downloaded

thomasyu888:
This probably requires some more thought.  For instance, there could probably be some minor refactor of the `_download_file_entity` to better support all of this.


thomasyu888:

PR #935, opened 2022-08-05
[SYNPY-620] allow list of user/team ids to be specified in setPermission

thomasyu888:
There needs to be more thought into how the code is configured here.  The nested for loops are not ideal.


thomasyu888:

PR #937, opened 2022-08-05
[SYNPY-1031] check activity

thomasyu888:
activity parameter expects Activity class, so fast fail if it is not.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -851,6 +851,10 @@ class Synapse(object):
                                       "permission. The file has NOT been downloaded."
                     self.logger.warning('\n' + '!'*len(warning_message)+'\n' + warning_message + '\n'
                                         + '!'*len(warning_message)+'\n')
+            else:
+                cached_file_path = self.cache.get(entity.dataFileHandleId, downloadLocation)
+                entity.path = os.path.normpath(cached_file_path)
+
         return entity
 
     def _ensure_download_location_is_directory(self, downloadLocation):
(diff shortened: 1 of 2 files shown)


danlu1 (approved):
Looks good to me.

synapseclient/client.py
```diff
@@ -851,6 +851,7 @@ def _getWithEntityBundle(self, entityBundle, entity=None, **kwargs):
                                       "permission. The file has NOT been downloaded."
                     self.logger.warning('\n' + '!'*len(warning_message)+'\n' + warning_message + '\n'
                                         + '!'*len(warning_message)+'\n')
+
```
thomasyu888, line 854:
```suggestion
```

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -2505,6 +2505,14 @@ def test_store__existing_no_update(syn):
         assert not mock_updatentity.called
 
 
+def test_store__wrong_activity(syn):
+    """Test that if function fails if activity object isn't passed in"""
+    ent = synapseclient.File("test", parentId="syn123")
+    with pytest.raises(ValueError) as ex_cm:
+        syn.store(ent, activity="foo")
+    assert "activity should be synapseclient.Activity object" ==  str(ex_cm.value)
```
thomasyu888, line 2513:
```suggestion
    assert "activity should be synapseclient.Activity object" == str(ex_cm.value)
```

PR #949, opened 2023-02-27
update name of utils function: `is_synapse_id`

### LINKED ISSUE #947: utils' `is_synapse_id` does not work as intended
# Bug Report

One of our collaborators for a challenge wants to implement a check in their workflow to ensure given IDs are valid Synapse IDs.  I suggested the utils function, `is_synapse_id`, but after looking at the implementation, I think the function needs to be revisited.

## Description of the problem

The current `is_synapse_id` function returns a substring that matches `r"(syn\d+)"`.

### Expected behavior

Honestly, because the function name starts with `is`, I expected a boolean to be returned, not a string.  Maybe we can rewrite the function as this?

```python
>>> def is_synapse_id(syn_id):
...   try:
...     syn.get(syn_id, downloadFile=False)
...   except synapseclient.core.exceptions.SynapseFileNotFoundError:
...     return False
...   except synapseclient.core.exceptions.SynapseHTTPError as err:
...     if err.response.status_code in (400, 404):
...       return False
...     # Valid ID but lacks permission
...     elif err.response.status_code == 403:
...       return True
...   else:
...     return True
... 
>>> is_synapse_id("test")
False
>>> is_synapse_id("123")
False
>>> is_synapse_id("syn123")
False
>>> is_synapse_id("syn123abc")
False
>>> is_synapse_id("syn28590455")
True
>>> is_synapse_id("syn22392179")
True
```

### Actual behavior

As noted above, any strings that begin with `syn` followed by numbers will be returned.  This will include strings that have letters that follow the digits and IDs that are actually not valid synIDs.

```python
>>> from synapseclient.core import utils
>>> utils.is_synapse_id("test")
>>> utils.is_synapse_id("123")
>>> utils.is_synapse_id("syn123")
'syn123'
>>> utils.is_synapse_id("syn123abc")
'syn123'
>>> utils.is_synapse_id("syn28590455")
'syn28590455'
>>> utils.is_synapse_id("syn22392179")
'syn22392179'
```

vpchung:
Fixes #947 , so that the utils function name is more descriptive.

The expected pattern has also been updated to be a little more stringent, that is, the synID should start with `syn` and end with a digit.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,6 +45,7 @@ install_requires = [
     "requests>=2.22.0,<3.0",
     "keyring>=15,<23.5",
     "deprecated>=1.2.4,<2.0",
+    "importlib-metadata<5.0",
 ]
 
 # on Linux specify a cryptography dependency that will not
--- a/setup.py
+++ b/setup.py
@@ -72,7 +72,7 @@ setuptools.setup(
     version=__version__,
     packages=setuptools.find_packages(exclude=["tests", "tests.*"]),
     # requirements
-    python_requires=">=3.7.*",
+    python_requires=">=3.7",
     install_requires=install_requires,
     extras_require={
         "pandas": ["pandas>=0.25.0,<2.0"],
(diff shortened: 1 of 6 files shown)


thomasyu888 (approved):

vpchung:
Thanks for the quick review, @thomasyu888 !  I can't merge unfortunately.

PR #950, opened 2023-03-01
SYNPY-1248: add `is_synapse_id` function to the client

### LINKED ISSUE #948: [Feature request] Add `is_synapse_id` to the client
# Feature Request

Add a helper function to the client that will allow checking whether a given string is a valid Synapse ID.  A "valid" synID must:

* have the expected format of `syn[0-9]+`
* be linked to an actual Synapse entity, that is, "syn123" is not a real Synapse entity

### Expected behavior

Return `True`/`False` whether a given synID is valid.  Something like this:

```python
>>> is_synapse_id("test")
False
>>> is_synapse_id("123")
False
>>> is_synapse_id("syn123")
False
>>> is_synapse_id("syn123abc")
False
>>> is_synapse_id("syn28590455")
True
>>> is_synapse_id("syn22392179")
True
```

### Request Motivation

One of our collaborators for a challenge wants to implement a check in their workflow to ensure given IDs are valid Synapse IDs. I suggested the utils function, `is_synapse_id`, but after some digging, realize that this function is more for testing the client, rather than for users to use.

vpchung:
Fixes #948

Aside from the unit tests, I also did tests with real Synapse entities:

```python
>>> # Not real IDs
>>> syn.is_synapse_id("test")
False
>>> syn.is_synapse_id("123")
False
>>> syn.is_synapse_id("syn123")
False
>>> # Real IDs
>>> syn.is_synapse_id("syn36025126")
True
>>> # Also works when user is logged out or lacks access
>>> syn.logout()
>>> syn.is_synapse_id("syn36025126")
True
>>> # Warning will be given if synID is not a string
>>> syn.is_synapse_id([123])
[WARNING] synID must be a string
False
```

Note: this should probably be merged after #949, since that PR contains fixes to the workflow.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -645,6 +645,7 @@ class Synapse(object):
                     return True
             return True
         self.logger.warn("synID must be a string")
+        return False
 
     def onweb(self, entity, subpageId=None):
         """Opens up a browser window to the entity page or wiki-subpage.
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -628,22 +628,24 @@ class Synapse(object):
 
     def is_synapse_id(self, syn_id):
         """Checks if given synID is valid (attached to actual entity?)"""
-        try:
+        if isinstance(syn_id, str):
             try:
-                self.get(syn_id, downloadFile=False)
-            except SynapseFileNotFoundError:
-                return False
-            except SynapseHTTPError as err:
-                if err.response.status_code in (400, 404):
+                try:
+                    self.get(syn_id, downloadFile=False)
+                except SynapseFileNotFoundError:
                     return False
-                # Valid ID but user lacks permission
-                elif err.response.status_code == 403:
+                except SynapseHTTPError as err:
+                    if err.response.status_code in (400, 404):
+                        return False
+                    # Valid ID but user lacks permission
+                    elif err.response.status_code == 403:
+                        return True
+            except SynapseAuthenticationError as err:
+                # Valid ID but user is not logged in
+                if err.__context__.response.status_code == 403:
                     return True
-        except SynapseAuthenticationError as err:
-            # Valid ID but user is not logged in
-            if err.__context__.response.status_code == 403:
-                return True
-        return True
+            return True
+        self.logger.warn("synID must be a string")
 
     def onweb(self, entity, subpageId=None):
         """Opens up a browser window to the entity page or wiki-subpage.
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -626,7 +626,7 @@ class Synapse(object):
                 return False
             raise
 
-    def is_synapse_id(self, syn_id):
+    def is_synapse_id(self, syn_id: str) -> bool:
         """Checks if given synID is valid (attached to actual entity?)"""
         if isinstance(syn_id, str):
             try:
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -626,6 +626,25 @@ class Synapse(object):
                 return False
             raise
 
+    def is_synapse_id(self, syn_id):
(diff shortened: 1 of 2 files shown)


synapseclient/client.py
```diff
@@ -626,6 +626,27 @@ def is_certified(self, user: typing.Union[str, int]) -> bool:
...
+                    self.get(syn_id, downloadFile=False)
+                except SynapseFileNotFoundError:
+                    return False
+                except SynapseHTTPError as err:
+                    if err.response.status_code in (400, 404):
+                        return False
+                    # Valid ID but user lacks permission
+                    elif err.response.status_code == 403:
+                        return True
+            except SynapseAuthenticationError as err:
```
thomasyu888, line 643: @vpchung what is the reason that there isn't just three `except` calls?

```
try:
   ...
except SynapseFileNotFoundError:
   ...
except SynapseHTTPError as err:
   ...
except SynapseAuthenticationError as err:
   ....
```

synapseclient/client.py
```diff
@@ -626,6 +626,27 @@ def is_certified(self, user: typing.Union[str, int]) -> bool:
                 return False
             raise
 
+    def is_synapse_id(self, syn_id):
+        """Checks if given synID is valid (attached to actual entity?)"""
+        if isinstance(syn_id, str):
```
thomasyu888, line 631: I forget but does `syn.get` first use the `utils.is_synapse_id_str`?

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -2999,15 +2999,25 @@ def test_is_certified__no_quiz_results(syn):
...
         is_certified = syn.is_certified("test")
     patch_get_user.assert_called_once_with("test")
     patch_get_cert.assert_called_once_with("foobar")
     assert is_certified is False
 
 
+def test_is_synapse_id(syn):
+    # Invalid IDs
+    assert syn.is_synapse_id("test") is False
+    assert syn.is_synapse_id("123") is False
```
thomasyu888, line 3014: Nit (Nitpicky): Is there a reason you don't do:
```
assert not syn.is_synapse_id("test")
assert not syn.is_synapse_id("123")
```

thomasyu888 (changes requested):
Thanks for the contribution @vpchung !  Some comments.

tests/unit/synapseclient/unit_test_client.py
vpchung: lol because it was Home Week and I forgot all proper coding rules...

synapseclient/client.py
vpchung: Ah, this is how I usually catch exception chains.  But yes, you are right -- for this particular scenario, we don't need to be this differential with the catches.

synapseclient/client.py
vpchung: It does, but passing any list-like collections will raise a TypeError.  I added this type check to handle those situations.

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -2999,15 +2999,25 @@ def test_is_certified__no_quiz_results(syn):
...
         is_certified = syn.is_certified("test")
     patch_get_user.assert_called_once_with("test")
     patch_get_cert.assert_called_once_with("foobar")
     assert is_certified is False
 
 
+def test_is_synapse_id(syn):
+    # Invalid IDs
+    assert not syn.is_synapse_id("test")
+    assert not syn.is_synapse_id("123")
```
thomasyu888, line 3014: Should add a test for passing in a non string.

thomasyu888 (approved):
🚀 LGTM! Thanks!

PR #959, opened 2023-05-23
[SYNPY-1266] Add support for GCP bucket

thomasyu888:
Allow for files to upload into external GCP bucket

- [x] add tests


tschaffter (approved):
Looks good!

PR #964, opened 2023-06-03
[SYNPY-557] Change return of downloadTableColumn

thomasyu888:
Change return of of downloadTableColumn to be consistent with query result


thomasyu888:

PR #966, opened 2023-06-03
[SYNPY-1236] add regex filter

thomasyu888:


thomasyu888:

PR #971, opened 2023-06-14
[Bug fix]: Fixed test related to pandas issue on window 2019

linglp:

PR #973, opened 2023-06-15
[SYNPY-1270] Deprecated parameter `date_parser` to support pandas2.0.0 and above

linglp:
## Changelog: 
old behavior: 
`_csv_to_pandas_df` takes in `date_parser` parameter and takes in function `datetime_millisecond_parser` to parse dates.

New behavior: 
Since pandas 2.0.0 deprecates `date_parser` parameter, we have to parse dates in a separate step. After reading the csv as a dataframe, the code would first change epoch time from string to integer and then apply pd.to_datetime function to date column. 

Also, since `dtype` and `freq` are not the actual parameter of pd.to_datetime function, I also reverted the changes of test in this PR: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/971


synapseclient/table.py
```diff
@@ -622,10 +622,6 @@ def _csv_to_pandas_df(
     test_import_pandas()
     import pandas as pd
 
-    # DATEs are stored in csv as unix timestamp in milliseconds
-    def datetime_millisecond_parser(milliseconds):
-        return pd.to_datetime(milliseconds, unit="ms", utc=True)
-
     if not date_columns:
```
thomasyu888, line 625: This if statement can probably be removed.

thomasyu888 (commented):
Nice work! I know you don't have windows, but can you paste the results of a pytest run on this specific test?

linglp:
> Nice work! I know you don't have windows, but can you paste the results of a pytest run on this specific test?

Sure Tom. Here's the result: 
```
(venv) lpeng@w153 synapsePythonClient % pytest tests/integration/synapseclient/test_tables.py
================================================= test session starts =================================================
platform darwin -- Python 3.11.3, pytest-7.3.2, pluggy-1.0.0
rootdir: /Users/lpeng/Documents/synapsepythonclient-fork/synapsePythonClient
configfile: setup.cfg
collected 13 items                                                                                                    

tests/integration/synapseclient/test_tables.py .............                                                    [100%]

================================================== warnings summary ===================================================
venv/lib/python3.11/site-packages/pkg_resources/__init__.py:121
  /Users/lpeng/Documents/synapsepythonclient-fork/synapsePythonClient/venv/lib/python3.11/site-packages/pkg_resources/__init__.py:121: DeprecationWarning: pkg_resources is deprecated as an API
    warnings.warn("pkg_resources is deprecated as an API", DeprecationWarning)

synapseclient/core/utils.py:10
  /Users/lpeng/Documents/synapsepythonclient-fork/synapsePythonClient/synapseclient/core/utils.py:10: DeprecationWarning: 'cgi' is deprecated and slated for removal in Python 3.13
    import cgi

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
===================================== 13 passed, 2 warnings in 146.78s (0:02:26) ======================================

```

synapseclient/table.py
thomasyu888: @linglp see this comment. I don't think you need to set the date_columns to an empty list any longer.

linglp: Thanks Tom! I have deleted the if not statement.

synapseclient/table.py
```diff
@@ -299,6 +299,8 @@
 import json
 from builtins import zip
 from typing import List, Dict
+import pandas as pd
```
thomasyu888, line 302: Since `pandas` isn't a required package to install, please follow the same format as other functions that use pandas / numpy.

```
    test_import_pandas()
    import pandas as pd
```

synapseclient/table.py
```diff
@@ -607,6 +609,18 @@ def join_column_names(columns):
     return ",".join(escape_column_name(c) for c in columns)
 
 
+def _convert_df_date_cols_to_datetime(df: pd.DataFrame, date_columns: List):
```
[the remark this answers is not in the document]
linglp: Thank you for the advice Tom! I have added some additional tests. In addition to what you mentioned, I also added a test to ensure that ValueError gets raised when date column contains invalid value that can't be changed to integer.

synapseclient/table.py
```diff
@@ -607,6 +610,32 @@ def join_column_names(columns):
     return ",".join(escape_column_name(c) for c in columns)
 
 
+def _convert_df_date_cols_to_datetime(df: DataFrameType, date_columns: List):
```
thomasyu888, line 613:
```suggestion
def _convert_df_date_cols_to_datetime(df: DataFrameType, date_columns: List) -> DataFrameType:
```

synapseclient/table.py
```diff
@@ -285,6 +285,7 @@
  - :py:meth:`synapseclient.Synapse.store`
  - :py:meth:`synapseclient.Synapse.delete`
 """
+from __future__ import annotations
```
thomasyu888, line 288: Is this still needed?

synapseclient/table.py
```diff
@@ -607,6 +610,32 @@ def join_column_names(columns):
...
+    test_import_pandas()
+    import pandas as pd
+    import numpy as np
+
+    # find columns that are in date_columns list but not in dataframe
+    diff_cols = list(set(date_columns) - set(df.columns))
+    if diff_cols:
+        raise ValueError("Please ensure that date columns are already in the dataframe")
+    try:
+        df[date_columns] = df[date_columns].astype(np.int64)
```
thomasyu888, line 627: What happens when there are blank values (`float('nan')`)?  Those are considered valid values

tests/unit/synapseclient/unit_test_tables.py
```diff
@@ -119,6 +121,106 @@ def test_cast_values__list_type():
...
+            ],
+            "string": ["str1", "str2", "str3", "str4", "str5"],
+        }
+    )
+    expected_date_df[["epoch_time1", "epoch_time2"]] = expected_date_df[
+        ["epoch_time1", "epoch_time2"]
+    ].astype("datetime64[ns, UTC]")
+
+    # convert epoch time to date time
+    test_df2 = _convert_df_date_cols_to_datetime(test_df, date_columns)
```
thomasyu888, line 214: I think you can just do:

```
assert_frame_equal(test_df2_dates, expected_date_df)
```

tests/unit/synapseclient/unit_test_tables.py
```diff
@@ -119,6 +121,106 @@ def test_cast_values__list_type():
...
 
+def test_convert_df_date_cols_to_datetime_empty_df():
+    test_df = pd.DataFrame()
+    date_columns = []
+    test_df_output = _convert_df_date_cols_to_datetime(test_df, date_columns)
+    assert test_df_output.empty
+
+
+def test_convert_df_date_cols_to_datetime_invalid_epoch_time():
+    test_df = pd.DataFrame(
```
thomasyu888, line 132: Explore using a `pytest.fixture` for this.

You can create one invalid dataframe that can be reused across these two tests. The first error will be the invalid column headers then it will test if the values in the columns are valid.

tests/unit/synapseclient/unit_test_tables.py
```diff
@@ -119,6 +121,106 @@ def test_cast_values__list_type():
...
+
+def test_convert_df_date_cols_to_datetime():
+    # construct test dataframe and get date columns
+    test_df = pd.DataFrame(
+        {
+            "row": [1, 2, 3, 4, 5],
+            "epoch_time1": [
+                "1107234000000",
+                "1107320400000",
+                "1107406800000",
```
thomasyu888, line 173: Add in a `float('nan')` and see what happens.

tests/unit/synapseclient/unit_test_tables.py
linglp: Yes that's a good point. Let me simplify the code further. Thanks for the suggestion.

tests/unit/synapseclient/unit_test_tables.py
linglp: Okay. I have added float('nan') in my test

tests/unit/synapseclient/unit_test_tables.py
linglp: Yes, that's a good idea. I have added fixture so that the dataframe can be used for multiple tests

synapseclient/table.py
linglp: I don't think so. I just removed it. Thanks for catching.

synapseclient/table.py
```diff
@@ -607,6 +609,37 @@ def join_column_names(columns):
...
+    import pandas as pd
+    import numpy as np
+
+    # find columns that are in date_columns list but not in dataframe
+    diff_cols = list(set(date_columns) - set(df.columns))
+    if diff_cols:
+        raise ValueError("Please ensure that date columns are already in the dataframe")
+    try:
+        df[date_columns] = (
+            df[date_columns].fillna(-1).astype(np.float64).astype(np.int64)
```
thomasyu888, line 629: I like the idea, but unfortunately -1 is a valid epochtime.  On the off chance someone actually puts in -1 as a value, you would replace it to a blank in line 636.  What are some other ways you would support NA values?

synapseclient/table.py
```diff
@@ -625,15 +625,13 @@ def _convert_df_date_cols_to_datetime(
     if diff_cols:
         raise ValueError("Please ensure that date columns are already in the dataframe")
     try:
-        df[date_columns] = (
-            df[date_columns].fillna(-1).astype(np.float64).astype(np.int64)
+        df[date_columns] = df[date_columns].applymap(
+            lambda x: np.int64(x) if pd.notnull(x) and str(x).isnumeric() else x
```
thomasyu888, line 629: This is a good solution. Couple things:

1. What if you provide it a number that is a float. Example: 1.1 <- That is technically a valid epoch time
2. I just noticed that you are doing try catch here, but I don't think its necessary here, since you're converting all the values that can be converted to integer and leaving the rest alone.

```
pd.to_datetime("test", unit="ms", utc=True)  # This will fail
```

thomasyu888: Keep the error message, but... maybe include valid epochtimes (integer, or doubles)

tests/unit/synapseclient/unit_test_tables.py
```diff
@@ -119,6 +140,111 @@ def test_cast_values__list_type():
...
+                "1107493200000",
+                "1107579600000",
+            ],
+            "epoch_time2": [
+                "1107234000000",
```
thomasyu888, line 206: Last thing, add some test cases here.

* Negative number.
* Decimal number (not sure what that will even look like for the expected - so this could be optional.

thomasyu888 (approved):
🔥 nicely done! I will merge soon!

tests/unit/synapseclient/unit_test_tables.py
linglp: Thanks for the feedbacks! I will add the test

PR #975, opened 2023-06-17
`pkg_resources` is a deprecated API - replace with importlib.resources

thomasyu888:
Replace `pkg_resources` call with `importlib.resources`


vpchung (approved):
Thanks for this!  This should help fix an import issue on one of our collab's end too.

PR #976, opened 2023-06-27
[SYNPY-967] deprecated memoize and added @lru_cache

linglp:
## Changelog
* Deprecated `memoize` and replaced it with `lru_cache` for Python 3.8 and above https://docs.python.org/3.8/library/functools.html#functools.lru_cache
* Since we are no longer using `memoize`, I also removed `refresh` parameter in login method
* Updated test
* Note: to address the pre-commit error below, I also updated `hashlib.md5()` by using `usedforsecurity=False` parameter:
```
[main]  INFO    running on Python 3.11.3
Run started:2023-06-27 22:10:20.620752

Test results:
>> Issue: [B324:hashlib] Use of weak MD5 hash for security. Consider usedforsecurity=False
   Severity: High   Confidence: High
   CWE: CWE-327 (https://cwe.mitre.org/data/definitions/327.html)
   More Info: https://bandit.readthedocs.io/en/0.0.0/plugins/b324_hashlib.html
   Location: synapseclient/client.py:2588:30
2587                            previouslyTransferred = 0
2588                            sig = hashlib.md5()
2589

--------------------------------------------------
>> Issue: [B324:hashlib] Use of weak MD5 hash for security. Consider usedforsecurity=False
   Severity: High   Confidence: High
   CWE: CWE-327 (https://cwe.mitre.org/data/definitions/327.html)
   More Info: https://bandit.readthedocs.io/en/0.0.0/plugins/b324_hashlib.html
   Location: synapseclient/core/utils.py:53:10
52
53          md5 = hashlib.md5()
54          with open(filename, "rb") as f:

--------------------------------------------------

```

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -118,7 +118,6 @@ from synapseclient.core.utils import (
     id_of,
     get_properties,
     MB,
-    memoize,
     is_json,
     extract_synapse_id_from_query,
     find_data_file_handle,
@@ -525,7 +524,7 @@ class Synapse(object):
             cached_sessions.set_most_recent_user(self.credentials.username)
 
         if not silent:
-            profile = self.getUserProfile(refresh=True)
+            profile = self.getUserProfile()
             # TODO-PY3: in Python2, do we need to ensure that this is encoded in utf-8
             self.logger.info(
                 "Welcome, %s!\n"
@@ -639,21 +638,19 @@ class Synapse(object):
         if self._is_logged_in():
             self.restDELETE("/secretKey", endpoint=self.authEndpoint)
 
-    @memoize
-    def getUserProfile(self, id=None, sessionToken=None, refresh=False):
+    @functools.lru_cache()
+    def getUserProfile(self, id=None, sessionToken=None):
         """
         Get the details about a Synapse user.
         Retrieves information on the current user if 'id' is omitted.
         :param id:           The 'userId' (aka 'ownerId') of a user or the userName
         :param sessionToken: The session token to use to find the user profile
-        :param refresh:      If set to True will always fetch the data from Synape otherwise will use cached information
         :returns: The user profile for the user of interest.
 
         Example::
             my_profile = syn.getUserProfile()
             freds_profile = syn.getUserProfile('fredcommo')
         """
-
         try:
             # if id is unset or a userID, this will succeed
             id = "" if id is None else int(id)
@@ -674,7 +671,6 @@ class Synapse(object):
                     else:  # no break
                         raise ValueError('Can\'t find user "%s": ' % id)
         uri = "/userProfile/%s" % id
-
         return UserProfile(
             **self.restGET(
                 uri, headers={"sessionToken": sessionToken} if sessionToken else None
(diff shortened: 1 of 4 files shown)


synapseclient/client.py
```diff
@@ -2589,7 +2585,7 @@ def _download_from_URL(
                     else:
                         mode = "wb"
                         previouslyTransferred = 0
-                        sig = hashlib.md5()
+                        sig = hashlib.new("md5", usedforsecurity=False)
```
thomasyu888, line 2588: @linglp can you speak to the error this is trying to fix?

linglp: Yeah.. If I don't make the change, I would get an error like this: 
```
[main]  INFO    running on Python 3.11.3
Run started:2023-06-27 22:10:20.620752

Test results:
>> Issue: [B324:hashlib] Use of weak MD5 hash for security. Consider usedforsecurity=False
   Severity: High   Confidence: High
   CWE: CWE-327 (https://cwe.mitre.org/data/definitions/327.html)
   More Info: https://bandit.readthedocs.io/en/0.0.0/plugins/b324_hashlib.html
   Location: synapseclient/client.py:2588:30
2587                            previouslyTransferred = 0
2588                            sig = hashlib.md5()
2589

--------------------------------------------------
>> Issue: [B324:hashlib] Use of weak MD5 hash for security. Consider usedforsecurity=False
   Severity: High   Confidence: High
   CWE: CWE-327 (https://cwe.mitre.org/data/definitions/327.html)
   More Info: https://bandit.readthedocs.io/en/0.0.0/plugins/b324_hashlib.html
   Location: synapseclient/core/utils.py:53:10
52
53          md5 = hashlib.md5()
54          with open(filename, "rb") as f:

--------------------------------------------------

```

synapseclient/client.py
```diff
@@ -118,7 +118,6 @@
     id_of,
     get_properties,
     MB,
-    memoize,
```
thomasyu888, line 121: Do you know if other pieces of code are using this function? If not, you can remove it all together.

synapseclient/client.py
```diff
@@ -639,21 +638,19 @@ def invalidateAPIKey(self):
         if self._is_logged_in():
             self.restDELETE("/secretKey", endpoint=self.authEndpoint)
 
-    @memoize
-    def getUserProfile(self, id=None, sessionToken=None, refresh=False):
```
thomasyu888, line 643: Is there a comparable parameter to refresh with the new function you are using?

thomasyu888 (commented):
Thanks for adding this! This looks great - just some comments

synapseclient/client.py
linglp: Thanks for the reminder Tom. I just checked again and it seems like this is the only place.

linglp:
@thomasyu888 Based on the current implementation, the cache could grow indefinitely. I think that is consistent with the old behavior, but do we want to set a maximum entry for the cache? Do we also want to add logic to evict cache based on space and time?

synapseclient/client.py
linglp: I searched and I can't see a parameter to refresh by using @lru_cache. Python also has `cache_clear()` and `cache_info()`. But I think the purpose of refresh is to grab the new data instead of using the cached result, which I think is different from clearing cache completely.

BryanFauble:
As a note: I made a mistake when I was trying to push to a remote I had permissions on and accidentally pushed to the main `develop` branch - I have since reverted the change and moved this PR over to: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/983

PR #981, opened 2023-10-10
SYNPY-1284: Fix location where url expiration is coming from

BryanFauble:
**Problem:**

- We stopped receiving a creation timestamp and a seconds until expire timestamp. Instead we now receive a expiration at x timestamp.
- We also needed to support the old method of receiving the expire timestamp

**Solution:**

- Updated the location where the expiration timestamp was coming from. We now no longer need to do any math to add the creation date to the expiration time as we directly get exactly when the expiration will occur.
- Kept the old path solution if we receive those fields in the URL

**Testing:**
1. I ran the affected integration tests
2. While running the test I verified that the correct UTC datetime object was created with the proper values for seconds since epoch

image

### DIFF
--- a/synapseclient/core/multithread_download/download_threads.py
+++ b/synapseclient/core/multithread_download/download_threads.py
@@ -186,8 +186,35 @@ def _pre_signed_url_expiration_time(url: str) -> datetime:
     :return: datetime in UTC of when the url will expire
     """
     parsed_query: dict = parse_qs(urlparse(url).query)
+    if (
+        parsed_query.get("X-Amz-Date") is not None
+        and parsed_query.get("X-Amz-Expires") is not None
+    ):
+        return _pre_signed_url_expiration_time_creaton_date_and_time_to_expire(
+            parsed_query
+        )
+    elif parsed_query.get("Expires") is not None:
+        return _pre_signed_url_expiration_time_expires_timestamp(parsed_query)
+
+    raise ValueError(
+        f"pre_signed_url does not contain an `Expires` field, or a `X-Amz-Date` field and a `X-Amz-Expires` field: {url}"
+    )
+
+
+def _pre_signed_url_expiration_time_expires_timestamp(parsed_query: dict) -> datetime:
     expires: int = int(parsed_query["Expires"][0])
     return datetime.datetime.utcfromtimestamp(expires)
+
+
+def _pre_signed_url_expiration_time_creaton_date_and_time_to_expire(
+    parsed_query: dict,
+) -> datetime:
+    time_made: str = parsed_query["X-Amz-Date"][0]
+    time_made_datetime: datetime.datetime = datetime.datetime.strptime(
+        time_made, ISO_AWS_STR_FORMAT
+    )
+    expires: str = parsed_query["X-Amz-Expires"][0]
+    return time_made_datetime + datetime.timedelta(seconds=int(expires))
 
 
 def _get_new_session() -> Session:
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 Well thats nice that we don't have to calculate it ourselves.

BryanFauble:
> 🔥 Well thats nice that we don't have to calculate it ourselves.

@thomasyu888 I thought I might be able to get away with this - I am talking back and forth on the jira ticket. This is from Marco & this will need to be addressed.

> Note that only pre-signed URLs from the synapse bucket CloudFront distribution have the Expires parameter, pre-signed URLs for files that are stored in other buckets are still from the s3 location so they will have the X-Amz-... parameters. Additionally, we also support google cloud buckets that might not have either parameter.

tests/unit/synapseclient/core/multithread_download/unit_test_download_threads.py
```diff
@@ -131,20 +131,24 @@ def test_generate_chunk_ranges():
 
 
 def test_pre_signed_url_expiration_time():
+    # GIVEN a pre signed URL with an expiration date of: Jul 22 2013 20:12:07 in seconds since epoch UTC
     url = (
         "https://s3.amazonaws.com/examplebucket/test.txt"
         "?X-Amz-Algorithm=AWS4-HMAC-SHA256"
         "&X-Amz-Credential=your-access-key-id/20130721/us-east-1/s3/aws4_request"
-        "&X-Amz-Date=20130721T201207Z"
-        "&X-Amz-Expires=86400"
```
thomasyu888, line 139: Although on second thought... we support external buckets within Synapse and those buckets won't support this feature...

BryanFauble: I updated the logic for this

synapseclient/core/multithread_download/download_threads.py
```diff
@@ -186,6 +186,29 @@ def _pre_signed_url_expiration_time(url: str) -> datetime:
     :return: datetime in UTC of when the url will expire
     """
     parsed_query: dict = parse_qs(urlparse(url).query)
+    if (
```
BWMac, line 189: NIT: I don't think you need any of the `is not None`'s in this function - the same outcome can be achieved with 
```
    if (
        parsed_query.get("X-Amz-Date")
        and parsed_query.get("X-Amz-Expires")
    ):
```
for example.

tests/unit/synapseclient/core/multithread_download/unit_test_download_threads.py
```diff
@@ -147,6 +147,48 @@ def test_pre_signed_url_expiration_time():
...
+    url = (
+        "https://s3.amazonaws.com/examplebucket/test.txt"
+        "?X-Amz-Algorithm=AWS4-HMAC-SHA256"
+        "&X-Amz-Credential=your-access-key-id/20130721/us-east-1/s3/aws4_request"
+        "&X-Amz-SignedHeaders=host"
+        "&X-Amz-Signature=signature-value"
+    )
+
+    # WHEN I get the expiration time from the pre signed URL
+    with pytest.raises(ValueError) as expiration_error:
```
BWMac, line 182: NIT: Rather than doing the `assert` to match the strings below, you can take care of that in this line.
```suggestion
    with pytest.raises(ValueError, match="pre_signed_url does not contain*") as expiration_error:
```
[source](https://stackoverflow.com/questions/23337471/how-do-i-properly-assert-that-an-exception-gets-raised-in-pytest/56569533#56569533)

BryanFauble:
Closing PR - We are going to take another approach to this problem at a later date. Platform is going to make their changes backwards compatible.

Spike for further changes: https://sagebionetworks.jira.com/browse/SYNPY-1288

PR #984, opened 2023-10-12
SYNPY-1285: Create pipfile

BryanFauble:
**Problem:**
Setting up the environment to start developing for the synapse python client was a mix of automated and manual steps. In order to stream line jumping into the project for future developers and to maintain dependencies for further development we want to leverge `pipenv` and virtual environments.

**Solution:**
I created a pipfile based off the versioning in the `setup.cfg`.

**Testing:**
1. I created a new virtual environment with `pipenv install` and verified I could use the CLI and any commands I wanted to use when importing the `synapseclient` into my python file.
2. I installed the dev dependencies with `pipenv install --dev` and verified that I could now run `pytest` - I ran the tests for both `unit` and `integration` tests on my local machine.
3. I verified with the dev dependencies installed that I could also generate new docs and html content


Pipfile
```diff
@@ -0,0 +1,29 @@
...
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[packages]
+boto3 = "1.28.62"
+cryptography = "3.3.2"
+deprecated = "1.2.14"
+keyring = "23.4.1"
+pandas = "2.0.3"
```
thomasyu888, line 11: There's some packages we don't require for users: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/2747b12a408729c8496fbf3fdf5d7411f8a8bba6/setup.cfg#L87-L94. but these are all relevant for development.

We can certainly lock the version down or use the `~=` function to be a bit more permissive.  Thoughts?

thomasyu888 (commented):
🤖 Nice, just a comment

Pipfile
BryanFauble: I'll move those packages over.

For versioning IMO I didn't like giving ranges that much when dealing with lock files - Since regenerating the lock file is going to update many dependencies that might not have been intentionally changed. Though I am ok with whichever direction.

thomasyu888: Could we follow something like this where the pipfile pulls from a setup.cfg and/or pyproject.toml: https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/e3ddde039e2776c6b8f053f7c502fc05dab6f9cd/Pipfile#L6-L13.  One concern I have is that the python community is moving away from setup.cfg as stated in pep-621: https://peps.python.org/pep-0621/.

I agree we shouldn't do ranges, but lets do something with this: https://peps.python.org/pep-0440/#compatible-release

BryanFauble: Great idea. I will explore this and add a new commit with the changes.

BryanFauble: I modified the pipfile to pull the dependencies from the `setup.cfg` like `py-orca` is doing - Is this what you had in mind?
https://github.com/Sage-Bionetworks/synapsePythonClient/pull/984/commits/fc0f18d43b840cb9cb7b2d6ad355b1df2d549106

thomasyu888 (approved):
🔥 LGTM!

PR #986, opened 2023-10-14
config from dev branch

xschildw:
This PR updates the encrypted config with that from the develop branch


thomasyu888 (approved):
I checked the md5 of the file against the one on the dev branch and it works. Thanks for doing this copy!  I forgot about the master branch.

PR #987, opened 2023-10-17
[SYNPY-1282] Adds Type Hinting to `client.py`

BWMac:
This PR adds type hints to several functions in `client.py` (I did more than 5 because most of them only have one or two variables).

I also added a quick note to CONTRIBUTING.md because I had an issue getting the `pipenv install` to work due to my having an out of date pipenv version.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -173,7 +173,7 @@ mimetypes.add_type("text/x-markdown", ".markdown", strict=False)
 DEFAULT_STORAGE_LOCATION_ID = 1
 
 
-def login(*args, **kwargs):
+def login(*args, **kwargs) -> synapseclient.Synapse:
     """
     Convenience method to create a Synapse object and login.
 
@@ -296,7 +296,7 @@ class Synapse(object):
         logging.getLogger("py.warnings").handlers = self.logger.handlers
 
     @property
-    def max_threads(self):
+    def max_threads(self) -> int:
         return self._max_threads
 
     @max_threads.setter
@@ -304,12 +304,12 @@ class Synapse(object):
         self._max_threads = min(max(value, 1), MAX_THREADS_CAP)
 
     @property
-    def username(self):
+    def username(self) -> Union[str, None]:
         # for backwards compatability when username was a part of the Synapse object and not in credentials
         return self.credentials.username if self.credentials is not None else None
 
     @functools.lru_cache()
-    def getConfigFile(self, configPath: str):
+    def getConfigFile(self, configPath: str) -> configparser.RawConfigParser:
         """
         Retrieves the client configuration information.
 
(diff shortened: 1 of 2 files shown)


synapseclient/client.py
```diff
@@ -296,20 +296,20 @@ def _init_logger(self):
         logging.getLogger("py.warnings").handlers = self.logger.handlers
 
     @property
-    def max_threads(self):
+    def max_threads(self) -> int:
         return self._max_threads
 
     @max_threads.setter
     def max_threads(self, value: int):
```
vpchung, line 303: @thomasyu888 Where do we currently stand with type-hinting void functions?  Are we omitting type hints altogether or is there a plan to use `-> None` (alternatively, `-> NoReturn` from [typing](https://docs.python.org/3/library/typing.html#typing.NoReturn))?

BryanFauble: From what I am reading `-> None` seems like the appropriate approach.

From typing it seems like it is implying that the function never returns anything.

Take this example:
```
def test_call():
    print("test_call()")

result = test_call()
print(result) # This prints `None`
```

vpchung: Yep! I read that it's only for cases like when a function only throws an exception or something similar.

TIL.

synapseclient/client.py
```diff
@@ -508,26 +508,26 @@ def login(
...
             return {}
 
-    def _get_config_authentication(self):
+    def _get_config_authentication(self) -> str:
         return self._get_config_section_dict(
             config_file_constants.AUTHENTICATION_SECTION_NAME
         )
 
-    def _get_client_authenticated_s3_profile(self, endpoint, bucket):
+    def _get_client_authenticated_s3_profile(self, endpoint: str, bucket: str) -> dict:
```
BryanFauble, line 524: I'm not sure that this one is a `-> dict` in this case. I think it might be a `-> str`. My reasoning:
This is called in these spots:

1. https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/upload/upload_functions.py#L280
2. https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/client.py#L2401

It's then getting passed over to: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/remote_file_storage_wrappers.py#L54

And it looks like it's being used as a `str`

synapseclient/client.py
```diff
@@ -654,7 +654,7 @@ def getUserProfile(
             )
         )
 
-    def _findPrincipals(self, query_string):
+    def _findPrincipals(self, query_string: str) -> list:
```
BryanFauble, line 657: To be more specific I think this would be what we want:

```suggestion
    def _findPrincipals(self, query_string: str) -> typing.List[UserGroupHeader]:
```

synapseclient/client.py
BWMac: Ah, yeah you're right. My mistake.

BWMac: Fixed in latest commit

synapseclient/client.py
BWMac: Made this change in my latest commit

BryanFauble (approved):
Changes LGTM!

PR #988, opened 2023-10-19
[SYNPY-1293] Update urllib3 version dependency

BryanFauble:
Background:
https://github.com/Sage-Bionetworks/synapsePythonClient/security/dependabot/7

Updating dependency a few patch version to push us out of the affected version range


BWMac (approved):
LGTM. It installs OK on my end.

PR #990, opened 2023-10-20
[SYNPY-1296] Config client error with api key or PAT

BryanFauble:
**Problem:**
The username is getting passed around in some cases as blank instead of None.

**Solution:**
Do a None and blank check on the username

**Testing:**
Tested using the CLI `synapse login`:
image

Tested with the CLI `synapse config` and verified I could do a `synapse login` after:
image

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -284,11 +284,11 @@
         },
         "urllib3": {
             "hashes": [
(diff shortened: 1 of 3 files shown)


thomasyu888 (approved):
🔥 thanks for the quick fix!

PR #992, opened 2023-10-23
[SYNPY-1295] Adding to the credentials.rst doc

BryanFauble:
Adding information about the `synapse config` command.

This is what the new section looks like:
image

I also added a link to the CLI `synapse config` in the more information section


thomasyu888 (approved):
🔥LGTM!

PR #998, opened 2023-10-30
[SYNPY-1316] Fix for cache item names matching

BryanFauble:
Problem:
When items in the cache on local machine had the same name for a file it would some-times assume it was the intended file to download and would not download the correct file from Synapse. Instead it would copy the incorrect file (With the correct name) to the requested location.

Solution:
I am doing a MD5 check of the cached item and the item from the server to verify they match

Testing:

1. I tested this is the script located in the comments on: https://sagebionetworks.jira.com/browse/SYNPY-1316
2. I also verified that with and without the changes that the introduced integration test passes/fails

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -1035,7 +1035,11 @@ class Synapse(object):
             if downloadFile:
                 if file_handle:
                     self._download_file_entity(
-                        downloadLocation, entity, ifcollision, submission
+                        downloadLocation,
+                        entity,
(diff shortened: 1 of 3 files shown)


thomasyu888 (approved):
🔥 LGTM! thanks for adding the extra type hinting.

In the future, we may want to look at performance if larger files are downloaded.  It doesn't generally take too long to calculate a md5, but something to look out for.

PR #999, opened 2023-10-30
[SYNPY-1298] flakey integration tests

BryanFauble:
Problem:
Several unit tests are very flaky, or timeout after 6 hours.

Solution:
Introduced 2 new packages:
```
    pytest-rerunfailures~=12.0
    func-timeout~=4.3
```

1. The `func-timeout` is to time bound some calls
2. The `rerunfailures` allows us to add a decorator to tests and conditionally re-run them a number of time (It's rare for them to fail multiple times in a row)

Testing:

1. See latest pipeline: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/6681317916 - Note: The pipeline in the PR might fail because of newly introduced dependencies that might not be in the cached dependencies


synapseclient/core/upload/multipart_upload.py
```diff
@@ -480,7 +479,7 @@ def part_fn(part_number):
         return _get_file_chunk(file_path, part_number, part_size)
 
     def md5_fn(part, _):
-        md5 = hashlib.md5()
+        md5 = hashlib.new("md5", usedforsecurity=False)
```
BryanFauble, line 482: These changes around MD5 changes for the same reason as: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/976#discussion_r1244461825

In order to support python 3.8 `usedforsecurity` can only be defined in the `.new` call - Newer python versions let you to do `hashlib.md5(data, usedforsecurity=False)`

.github/workflows/build.yml
```diff
@@ -133,8 +133,11 @@ jobs:
             export EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY="${{secrets.EXTERNAL_S3_BUCKET_AWS_SECRET_ACCESS_KEY}}"
 
              # use loadscope to avoid issues running tests concurrently that share scoped fixtures
-            pytest -sv tests/integration -n auto --dist loadscope
+            pytest -sv tests/integration -n auto --ignore=tests/integration/synapseclient/test_command_line_client.py --dist loadscope
+            # Execute the CLI tests in a non-dist way because they were causing some test instability when being run concurrently
+            pytest -sv tests/integration/synapseclient/test_command_line_client.py
```
thomasyu888, line 138: Would this still be an issue with the parallel OS executions?

synapseclient/core/upload/multipart_upload.py
thomasyu888: Should we add a TODO here to update when py 3.8 is at EOL?

synapseclient/core/upload/multipart_upload.py
```diff
@@ -531,9 +530,14 @@ def multipart_upload_string(
      https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.17
     """
 
+    def md5_fn(part, _):
```
thomasyu888, line 533: Should this be a utils function?

tests/integration/synapseclient/core/test_caching.py
```diff
@@ -145,22 +154,35 @@ def thread_keep_storing_one_File(syn, project, schedule_for_cleanup):
...
-        id = get_all_ids_from_Project(syn, project)
+        id = []
+        try:
+            id = get_all_ids_from_Project(syn, project)
+        except FunctionTimedOut:
+            syn.logger.warning(
+                f"thread_get_and_update_file_from_Project()::get_all_ids_from_Project timed out, [project: {project.id}]"
+            )
         if len(id) <= 0:
+            sleep_for_a_bit()
```
thomasyu888, line 185: Nit: I know this is old code but len of a list < 0?

tests/integration/synapseclient/integration_test_Entity.py
```diff
@@ -499,21 +499,22 @@ def test_create_or_update_project(syn, project, schedule_for_cleanup):
...
+    file = File(filepath, name=file_name, parent=project)
     file = syn.store(file)
 
     # Now hide the file from the cache and download with downloadFile=False
     os.rename(filepath, filepath + RENAME_SUFFIX)
     file = syn.get(file.id, downloadFile=False)
 
     # Change something and reupload the file's metadata
-    file.name = "Only change the name, not the file"
+    file.name = "Only change the name, not the file" + str(uuid.uuid4())
```
thomasyu888, line 517: Does this creates a new uuid4 string?

tests/integration/synapseclient/test_command_line_client.py
```diff
@@ -614,7 +614,11 @@ def test_command_get_recursive_and_query(test_state):
     cols = [Column(name="id", columnType="ENTITYID")]
 
     schema1 = test_state.syn.store(
-        Schema(name="Foo Table", columns=cols, parent=project_entity)
+        Schema(
+            name="Foo Table_test_command_get_recursive_and_query",
```
thomasyu888, line 618: Would uuid be more stable here?

tests/integration/synapseutils/test_synapseutils_copy.py
```diff
@@ -2,6 +2,7 @@
 import time
```
thomasyu888, line 2: Yes, the copy module tests could be fragile, especially if things are being copied into the same project across parallel tests that are running

tests/unit/synapseclient/core/upload/unit_test_multipart_upload.py
```diff
@@ -25,6 +25,12 @@
 )
 
 
+def md5_fn(part, _):
```
thomasyu888, line 28: See comment about adding this in utils.

thomasyu888 (commented):
Nice! I added some comments.

.github/workflows/build.yml
BryanFauble: I don't believe so - I think it's an issue with how the messages were getting parsed out of the logs to be used further within the CLI tests

synapseclient/core/upload/multipart_upload.py
BryanFauble: It'll work in either way - We don't get a real benefit from one or the other

tests/integration/synapseclient/integration_test_Entity.py
BryanFauble: Yes:
```
def uuid4():
    """Generate a random UUID."""
    return UUID(bytes=os.urandom(16), version=4)
```

synapseclient/core/upload/multipart_upload.py
BryanFauble: Moved over to a utils function: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/999/commits/4cd8886d95f2d23dc78406589aa1e83b3156ed2f

tests/integration/synapseclient/integration_test_Entity.py
thomasyu888: @BryanFauble Right, were we supposed to be printing out `filepath + RENAME_SUFFIX`

thomasyu888 (approved):
🤖 LGTM!  @BWMac if you wanted to review for for cross training, that'd be great, but it doesn't block the merging of this ticket.

tests/integration/synapseclient/integration_test_Entity.py
BryanFauble: In this test it's doing:
```
    # Change something and reupload the file's metadata
    file.name = "Only change the name, not the file" + str(uuid.uuid4())
    reupload = syn.store(file)
    assert reupload.path is None, "Path field should be null: %s" % reupload.path

    # This should still get the correct file
    reupload = syn.get(reupload.id)
    assert filecmp.cmp(filepath + RENAME_SUFFIX, reupload.path)
    assert reupload.name == file.name
```
My thought is that I want to avoid statically referencing anything - hence the addition of the UUID.

I'm not sure what you mean about `print out a random string`

thomasyu888: Ohhhh i understand now. Sorry about that - makes sense!

PR #1000, opened 2023-10-30
Fix for int test on windows for file rename

BryanFauble:
Problem:
Windows doesn't overwrite in an os.rename

`FileExistsError: [WinError 183] Cannot create a file when that file already exists: 'c:/users/runner~1/appdata/local/temp/tmplrhaqh7f.dat' -> 'c:/users/runner~1/appdata/local/temp/tmpj45y4tsc.dat'`

Solution:
Delete destination file before rename


BryanFauble:
Build pipeline passed for windows - merging into develop

PR #1001, opened 2023-10-31
[SYNPY-1274] pre commit in gh action

BryanFauble:
**Problem:**
We were only running the black linter during the GH action - which does not enforce everything the `pre-commit` check did.

**Solution:**
I set up  `pre-commit` job that runs this action: https://github.com/pre-commit/action

**Testing:**

1. Failure in pre-commit: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/6710081386/job/18234512611
2. Success in pre-commit: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/6710112140/job/18234608303

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,3 @@
 sphinx
 sphinx-argparse
 sphinx_rtd_theme
--e .
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -80,10 +80,8 @@ jobs:
             pip install numpy
           fi
 
-      # - name: lint
-      #   shell: bash
-      #   run: |
-      #     flake8
+      - name: Run static analysis and format checkers
+        run: pre-commit run --all-files
 
       - name: run-unit-tests
         shell: bash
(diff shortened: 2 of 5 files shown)


docs/requirements.txt
```diff
@@ -1,4 +1,3 @@
 sphinx
 sphinx-argparse
 sphinx_rtd_theme
--e .
```
BryanFauble, line 4: Having this included was giving me:
```
ERROR: file:///home/bfauble/BryansGreatWorkspace/synapsePythonClient/docs (from -r requirements.txt (line 4)) does not appear to be a Python project: neither 'setup.py' nor 'pyproject.toml' found.
```

Also - the pre-commit wanted to move this to line 1 - which is what prompted me to look at this.

thomasyu888: I think this was included only for a test read the docs integration.

Let's revisit the doc deployment strategy later on. It's an inconvenience that we have to build it ourselves when deploying - and our build doesn't allow for deployment of docs of past versions on our site.

thomasyu888 (approved):
🔥 Going to pre-approve.

PR #1002, opened 2023-10-31
[SYNPY-1316] Moved install_requires back to setup.py

BryanFauble:
Problem:
Installing the python client forces a downgrade of the `cryptography` package to `<3.4` - It is not optional as the comments of the file are implying

Solution:
Move back to setting the `install_requires` in the `setup.py` file instead of the `setup.cfg` file. The `setup.cfg` was not created with this kind of use-case in mind.

Testing:
I tested this locally and in some docker images when I had and had not installed `cruptography` before hand

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@
 import os
 from setuptools import setup
 
+import platform
 import json
 
 # make sure not to overwrite existing .synapseConfig with our example one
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -15,4 +16,33 @@
 with open("synapseclient/synapsePythonClient") as config:
     __version__ = json.load(config)["latestVersion"]
 
-setup(data_files=data_files, version=__version__)
+install_requires = [
```
thomasyu888, line 19: Will this impact the `pipenv` file?

PR #1003, opened 2023-10-31
[SYNPY-1316] Make cryptography an optional install

BryanFauble:
Problem:
Installing the python client forces a downgrade of the cryptography package to <3.4 - It is not optional as the comments of the file are implying

Solution:
Remove the 3.4 cryptography package as a required install. Instead let it be installed by anything that requires it.


thomasyu888 (approved):
🐛 lGTM!

PR #1007, opened 2023-11-02
[SYNPY-1304] Introduction of OpenTelemetry

BryanFauble:
**Background SPIKE:**

1. https://sagebionetworks.jira.com/browse/SYNPY-1290
2. Confluence Doc: https://sagebionetworks.jira.com/wiki/spaces/IBC/pages/3041099777/SYNPY-1289+Open+Telemetry+Metrics+OTEL+in+synapse+python+client

Solution:
This PR introduces 3 new dependencies to the Synapse Python project around OpenTelemetry. I have also manually instrumented the code in areas where it would be programmatically entered, and any areas where I/O is expected to occur for communication with our Rest APIs.

Please carefully review the changes to the README around this and let me know areas you'd like it improved.

JaegerUI to view traces:
image


.github/workflows/build.yml
```diff
@@ -63,7 +63,7 @@ jobs:
           path: |
             ${{ steps.get-dependencies.outputs.site_packages_loc }}
             ${{ steps.get-dependencies.outputs.site_bin_dir }}
-          key: ${{ runner.os }}-${{ matrix.python }}-build-${{ env.cache-name }}-${{ hashFiles('setup.py') }}-v7
+          key: ${{ runner.os }}-${{ matrix.python }}-build-${{ env.cache-name }}-${{ hashFiles('setup.py') }}-v7-otel
```
BryanFauble, line 66: To be reverted before merge - This is to allow the CICD pipeline to install the new dependencies and not used the cached version

README.md
```diff
@@ -153,6 +153,61 @@ The purpose of synapseutils is to create a space filled with convenience functio
         print(dirname)
         print(filename)
 
+OpenTelemetry (OTEL)
+--------------------------------
+[OpenTelemetry](https://opentelemetry.io/) is a tool that can be used to collect telemetry data. The synapseclient is ready to provide traces should you want them. The Synapse Python client supports OTLP Exports and can be configured via environment variables as defined [here](https://opentelemetry-python.readthedocs.io/en/stable/exporter/otlp/otlp.html).
```
thomasyu888, line 158: I think we can provide a little purpose around collecting telemetry data. Something like: "Analysis of traces and spans can provide insights into latency, errors, and other performance metrics." (pulled directly from google)

CONTRIBUTING.md
```diff
@@ -180,6 +180,9 @@ authEndpoint=https://repo-dev.dev.sagebase.org/auth/v1
 fileHandleEndpoint=https://repo-dev.dev.sagebase.org/file/v1
 ```
 
+#### Running OpenTelemetry in Integration Tests
+`tests/integration/conftest.py` is where we defining which trace provider to use. Set the `SYNAPSE_OTEL_INTEGRATION_TEST_PROVIDER` environment variable to `otlp` or `console` depending on your use case.
```
thomasyu888, line 184: Did we want to go into a bit of detail of what to add, or is that more on a per function basis?  For example, maybe we provide some direction about what attributes should be set.
```
        trace.get_current_span().set_attributes(
            {"synapse.file_handle_id": file_result["fileHandleId"]}
        )
```

synapseclient/core/upload/multipart_upload.py
```diff
@@ -315,6 +323,7 @@ def _upload_parts(self, part_count, remaining_part_numbers):
                     executor.submit(
                         self._handle_part,
                         part_number,
+                        context.get_current(),
```
thomasyu888, line 326: Ah, I see. a bit inconvenient with the way threading is in Python, but this makes sense

thomasyu888 (approved):
🔥 This looks great - id say since this is the intro to open telemetry, its ok for this PR to be so big.  We can take the sprint review to show everyone how this works.

README.md
BryanFauble: +1. I am swapping this paragraph to read as:
```suggestion
[OpenTelemetry](https://opentelemetry.io/) helps support the analysis of traces and spans which can provide insights into latency, errors, and other performance metrics. The synapseclient is ready to provide traces should you want them. The Synapse Python client supports OTLP Exports and can be configured via environment variables as defined [here](https://opentelemetry-python.readthedocs.io/en/stable/exporter/otlp/otlp.html).
```

CONTRIBUTING.md
BryanFauble: I added more about this in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1007/commits/0c91c295de5071ef5a01f3ef78264caefb293f29

PR #1010, opened 2023-11-08
[SYNPY-1319] Lower binary file size

BryanFauble:
**Problem:**
https://sagebionetworks.jira.com/browse/SYNPY-1319
One or more parts fails to upload

**Solution:**
To try to limit possible failures lower the size of the file to 1KB by default instead of 1MB

**Testing:**
Will rely on GH CICD actions


thomasyu888 (commented):
LGTM!

PR #1011, opened 2023-11-08
[SYNPY-816] Enable test_teams again

BryanFauble:
**Problem:**
https://sagebionetworks.jira.com/browse/SYNPY-816
Assertion error

**Solution:**
I enabled the test again

**Testing:**
I installed https://pypi.org/project/pytest-repeat/ and ran 100 times in parallel with 4 workers on my local machine

### DIFF
--- a/tests/integration/synapseclient/test_evaluations.py
+++ b/tests/integration/synapseclient/test_evaluations.py
@@ -5,9 +5,8 @@ import uuid
 import random
 
 import pytest
-import unittest
 
-from synapseclient import Evaluation, File, SubmissionViewSchema, Synapse, Team
+from synapseclient import Evaluation, File, SubmissionViewSchema, Synapse, Team, Project
 from synapseclient.core.exceptions import SynapseHTTPError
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!

thomasyu888 (commented):
Although - I will say, it doesn't look like we changed anything, I wonder if there is actually a possibility that we get duplicated team names with `"My Uniquely Named Team " + str(uuid.uuid4())` over time.

BryanFauble:
> Although - I will say, it doesn't look like we changed the logic within the tests, I wonder if there is actually a possibility that we get duplicated team names with `"My Uniquely Named Team " + str(uuid.uuid4())` over time.

@thomasyu888 
https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions
"the probability to find a duplicate within 103 trillion version-4 UUIDs is one in a billion. "

thomasyu888 (approved):

PR #1013, opened 2023-11-16
[SYNPY-1322] Object Orientated Programming Interfaces

BryanFauble:
**Background:**
Working with the Synapse python client is a mix of many competing technologies, frameworks, and documentation.

**Frameworks used:**

1. dataclasses: https://docs.python.org/3.8/library/dataclasses.html
2. asyncio: https://docs.python.org/3.8/library/asyncio-dev.html

**Performance insights:**
Some initial testing with the benchmark script herehttps://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/docs/scripts/benchmark.py - Using these new classes
| Test                      | Synapseutils Sync | os.walk + syn.store | S3 Sync | ASYNC/OOP   | Per file size |
|---------------------------|-------------------|---------------------|---------|---------|---------------|
| 25 Files 1MB total size   | 10.43s            | 8.99s               | 1.83s   | 5.45s   | 40KB          |
| 775 Files 10MB total size | 243.57s           | 257.27s             | 7.64s   | 123.99s | 12.9KB        |

**Examples of how this OOP approach works:**
See all of the scripts created in the `test_scripts/` folder. ie: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/SYNPY-1322-OOP-POC/docs/scripts/object_orientated_programming_poc/oop_poc_project.py

**Things left to do/investigate (Will be completed in follow-up epics):**

1. Would swapping away from the `requests` library (Like httpx: https://www.python-httpx.org/advanced/ ) lead to some better performance and async programming? Answer: Yeshttps://sagebionetworks.jira.com/browse/SYNPY-1411
2. All of the todos: https://sagebionetworks.jira.com/browse/SYNPY-1343
3. Any other domain models that we want to implement: https://sagebionetworks.jira.com/browse/SYNPY-1343


synapseclient/api/annotations.py
```diff
@@ -0,0 +1,39 @@
...
+    annotations: "Annotations", synapse_client: Optional[Synapse] = None
+):
+    """Call to synapse and set the annotations for the given input.
+
+    :param annotations: The annotations to set. This is expected to have the id, etag, and annotations filled in.
+    :return: _description_
+    """
+    annotations_dict = asdict(annotations)
+
+    # TODO: Is there a more elegant way to handle this - This is essentially being used
```
BWMac, line 26: There is another dataclass library that adds some more functionality. Its [`asdict`](https://dataclass-wizard.readthedocs.io/en/latest/dataclass_wizard.html#dataclass_wizard.dumpers.asdict) function has an `exclude` argument that you can pass to leave out class attributes you don't want in your dictionary:

```
from dataclasses import dataclass
from dataclass_wizard.dumpers import asdict

@dataclass
class Foo:
    bar: str
    baz: int

foo = Foo('bar', 42)
print(asdict(foo, exclude=['baz']))

> {'bar': 'bar'}
```

You'd still be hard-coding the excluded values though. Not sure of any other ways aside from implementing an `asdict` method specific to the class that excludes attributes not used in the API.

synapseclient/api/annotations.py
```diff
@@ -0,0 +1,39 @@
...
+    # TODO: to remove any fields that are not expected by the REST API.
+    filtered_dict = {k: v for k, v in annotations_dict.items() if k != "is_loaded"}
+
+    # TODO: This `restPUT` returns back a dict (or string) - Could we use:
+    # TODO: https://github.com/konradhalas/dacite to convert the dict to an object?
```
BWMac, line 31: I do like the look of `dacite`, but is that more efficient than just wrapping the response in the class object like we have elsewhere?

BWMac:
@BryanFauble Have you thought about having some sort of `BaseClass` that these new classes could extend? It seems like many have a lot of attributes in common and some methods (at least the names) too.

BryanFauble:
> @BryanFauble Have you thought about having some sort of BaseClass that these new classes could extend? It seems like many have a lot of attributes in common and some methods (at least the names) too.

@BWMac Yes, however inheritance is an easy way to have large refactoring down the line for simple changes. Especially in cases where a base class is a part of many parent classes. IMO The easiest way I've seen to deal with it is to just replicate things where they are needed. I do understand it adds some duplicated code, but the duplication far out-weighs the negatives.

synapseclient/api/annotations.py
BryanFauble: I'm not sure yet - I'll have to give it some more thought. 

One of the things we will need to do it at least have a thin translation layer between the REST api and the dataclasses because the names we are giving them in the python client are snake_case, vs the REST api is all in camelCase.

synapseclient/api/annotations.py
BryanFauble: Good idea - but i'm starting to think in the other direction now. What I mean by this is specify only those things I want to include and have a dataclass -> json/dict mapping process to format the input for the REST API as the API is expecting.

thomasyu888:
@BryanFauble , @BWMac 

I definitely agree that we should use inheritance carefully.  In the case of Entities, technically FileEntities do derive from Entities, so it _could_ make sense for that to use inheritance to avoid the duplication of code across 7 entities: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/Entity.html and having to update the code 7 separate times if something changes.  Each implemenation comes with its pros/cons.

It definitely makes sense to use MixIns/composition for different functionality for entities like 

* Versionable
* set ACL
* etc

BryanFauble:
@thomasyu888 @BWMac 

> In the case of Entities, technically FileEntities do derive from Entities, so it could make sense for that to use inheritance to avoid the duplication of code across 7 entities: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/Entity.html and having to update the code 7 separate times if something changes.

True - But also a recent example I came across is that a Project inherits an Entity, but `parentId` doesn't mean anything for a Project since it's treated as the highest level container of stuff. These are the kind of exceptions to the rule that make inheritance hard.

I do like composition though to contain common functionality.

Excellent discussion, let's continue this further when I get more models in place and we can combine as we see fit when we find overlapping functionality.

test_scripts/oop_poc_table.py
```diff
@@ -0,0 +1,160 @@
...
+    write_random_csv_with_data(path_to_csv)
+
+    csv_path = await copy_of_table.store_rows_from_csv(csv_path=path_to_csv)
+
+    print(csv_path)
+
+    # Querying for data from a table =====================================================
+    destination_csv_location = os.path.expanduser("~/temp/my_query_results")
+
+    await Table.query(
```
BWMac, line 140: Somthing about using a method from the `Table` class outside the context of an instance of that class, but then feeding an attribute of the instance `copy_of_table` rubs me the wrong way. 

Not sure how else it should be done though since you have to pass the query string as an argument to the `query` function somehow.

BWMac (commented):
Overall I found the example scripts to do a good job of demonstrating the power of the OOP changes. As a user I would much prefer this way when using the client within other python packages that I work on.

test_scripts/oop_poc_file.py
```diff
@@ -0,0 +1,144 @@
...
+    create_random_file(path_to_file)
+
+    # Creating and uploading a file to a project =========================================
+    file = File(
+        path=path_to_file,
+        name=name_of_file,
+        annotations=annotations_for_my_file,
+        parent_id=PROJECT_ID,
+        description="This is a file with random data.",
+    )
```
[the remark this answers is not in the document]
BryanFauble: Right - This was one thing I cut due to lack of time

test_scripts/oop_poc_table.py
BryanFauble: Yeah this is one of the things that I was struggling to figure out the best practice for. One thing to clarify is that we are not passing an instance of `copy_of_table`, in this case I am just grabbing the ID of the table. This is equivalent to these lines:

```
    table_id_to_query = copy_of_table.id
    await Table.query(
        query=f"SELECT * FROM {table_id_to_query}",
        result_format=CsvResultFormat(download_location=destination_csv_location),
    )
```

test_scripts/oop_poc_project.py
```diff
@@ -0,0 +1,140 @@
...
+    }
+
+    # Creating a project ==================================================================
+    project = Project(
+        name="bfauble_my_new_project_for_testing",
+        annotations=annotations_for_my_project,
+        description="This is a project with random data.",
+    )
+
+    project = await project.store()
```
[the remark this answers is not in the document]
BryanFauble: I agree - Likely people are making their projects in the UI and then grabbing the `syn` id afterword. Though I wanted to provide functionality for everything that you can do in the UI.

test_scripts/oop_poc_table.py
BWMac: Right, not an instance itself, but an attribute of that instance. 

It would feel better if you could do something like
```
copy_of_table.query(query_args)
```
So that it reads like you are querying the table you have already. But formatting the query could be weird and it's not like we don't want to let people do `Table.query` to query a table already in Synapse that they don't care to read into a Python object.

BWMac: This is very much a future work/off topic thing, but what if we didn't make people pass the whole query string themselves? 

We could use something like [jinjasql](https://github.com/sripathikrishnan/jinjasql) to allow people to pass parameters that the `Table` class could format into a query. Then someone could either pass the id of a table they want in an uninstantiated case or when there is an instance of the `Table` class we are already using, we just grab the `id` on the back end and plug it into the query.

Unfortunately, it looks like `jinjasql` has not been maintained for a while, but maybe there is something else out there that could provide similar functionality.

BryanFauble: Interesting library. My concerns when we get to solutioning on this are around continuing to provide the same if not more functionality for folks querying for data.

I had considered something like a "reference to my class instance" string that you could add into a query like:
`select * from :my-table-instance-id`, but that relying on those querying for the data to add in that string constant.

By not tying the query to a specific instance of a Table it makes things like joins easier to work with. SQL can be very complex and it'd be hard to implement everything we want into a Python class or semi-structured way to doing it.

synapseclient/models/annotations.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # TODO: Currently this is required - However, we do have the ability to make some assumptions based
+    # TODO: On the value types. For example, if it's a str, datetime, bool, etc...
+    type: AnnotationsValueType
+    # TODO: What are all the python types we are going to accept here
+    value: List[Union[str, bool]]
+
+
+@dataclass()
+class Annotations:
```
thomasyu888, line 37: Make sure we are using annotationsv2

synapseclient/models/file.py
```diff
@@ -0,0 +1,213 @@
...
+    desired annotations. The value is an object containing a list of values
+    (use empty list to represent no values for key) and the value type associated with
+    all values in the list."""
+
+    # TODO: We need to provide functionality for folks to store the file in Synapse, but
+    # TODO: Not upload the file.
+
+    is_loaded: bool = False
+
+    def convert_from_api_parameters(
```
thomasyu888, line 91: Is there a generalization camel case to snake case converter?

test_scripts/oop_poc_file.py
```diff
@@ -0,0 +1,144 @@
...
+        "my_key_double": AnnotationsValue(
+            type=AnnotationsValueType.DOUBLE, value=[1.2, 3.4, 5.6]
+        ),
+        "my_key_long": AnnotationsValue(
+            type=AnnotationsValueType.LONG, value=[1, 2, 3]
+        ),
+        "my_key_timestamp": AnnotationsValue(
+            type=AnnotationsValueType.TIMESTAMP_MS, value=[1701362964066, 1577862000000]
+        ),
+    }
```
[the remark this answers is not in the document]
thomasyu888: +1 here. Basically right now there's a convenience function to auto detect the type and add annotations.

That's probably the next step.

thomasyu888 (approved):
🔥 I'm going to pre-approve here.  Overall, I think we can refactor the syncToSynapse function which will help the upload speeds.

This will also make it as seamless of a transition as possible.

synapseclient/models/file.py
```diff
@@ -0,0 +1,213 @@
...
+
+if TYPE_CHECKING:
+    from synapseclient.models import Folder, Project
+
+
+tracer = trace.get_tracer("synapseclient")
+
+
+@dataclass()
+class File:
```
thomasyu888, line 23: This might look odd with the auto documentation. You may want to document the class with attributes.

test_scripts/oop_poc_file.py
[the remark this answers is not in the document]
BryanFauble: Will be implemented with https://sagebionetworks.jira.com/browse/SYNPY-1344

test_scripts/oop_poc_file.py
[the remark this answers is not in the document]
BryanFauble: I will implement this with https://sagebionetworks.jira.com/browse/SYNPY-1349

synapseclient/models/annotations.py
```diff
@@ -0,0 +1,103 @@
...
+        to the format used by this class.
+
+        Arguments:
+            synapse_annotations: The annotations from the synapse rest API.
+
+        Returns:
+            The annotations in python class format.
+        """
+        if synapse_annotations is None:
+            return None
```
BWMac, line 87: Should this method include type-hinting for the case where it returns `None`?

synapseclient/models/file.py
```diff
@@ -0,0 +1,271 @@
+import asyncio
+from dataclasses import dataclass
+from datetime import date, datetime
+from typing import Dict, List, Union
+from opentelemetry import trace, context
+from synapseclient.models import Annotations
+
+# import uuid
```
BWMac, line 8: remove commented out import (unless there is a reason not to)

synapseclient/models/folder.py
```diff
@@ -0,0 +1,294 @@
+import asyncio
+from dataclasses import dataclass, field
+from datetime import date, datetime
+from typing import Dict, List, Union
+from typing import Optional, TYPE_CHECKING
+from opentelemetry import trace, context
+
+# import uuid
```
BWMac, line 8: same as comment in `file`

BWMac (approved):
Just a couple minor comments but LGTM!

PR #1014, opened 2023-11-16
[SYNPY-1186] When a username is a number,  getUserProfile cannot retrieve the user

BWMac:
**Problem:**

When a user calls the `getUserProfile` method with a username that is a number, the appropriate `UserProfile` is not retrieved. This essentially boils down to the fact that we currently allow people to pass `userName`s and `ownerId`s to `getUserProfile`, and that people are allowed to create usernames on Synapse that are just a string of numbers.

**Solution:**

Split `getUserProfile` into two separate functions. One that accepts `userName` as a string only, and one that only will accept an `ownerId` in the form of an integer.  

A complication of this solution is that `getUserProfile` is used in a number of places, some of which also allow `userName`s or `ownerId`s. This refactoring and the eventual deprecation of `getUserProfile` will need to be done in a future PR.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -673,8 +673,8 @@ class Synapse(object):
         :returns: The user profile for the user of interest.
 
         Example::
-            my_profile = syn.getUserProfile()
-            freds_profile = syn.getUserProfile('fredcommo')
+            my_profile = syn.get_user_profile_by_id()
+            freds_profile = syn.get_user_profile_by_id(1234567)
         """
         if id:
             if not isinstance(id, int):
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -622,15 +622,15 @@ class Synapse(object):
             self.restDELETE("/secretKey", endpoint=self.authEndpoint)
 
     @functools.lru_cache()
-    def get_user_profile(
+    def get_user_profile_by_username(
         self,
-        id: Union[str, UserProfile, TeamMember] = None,
+        id: str = None,
         sessionToken: str = None,
     ) -> UserProfile:
         """
         Get the details about a Synapse user.
         Retrieves information on the current user if 'id' is omitted or is empty string.
-        :param id:           The userName, UserProfile, or TeamMember of a user
+        :param id:           The userName of a user
         :param sessionToken: The session token to use to find the user profile
         :returns: The user profile for the user of interest.
 
@@ -639,26 +639,16 @@ class Synapse(object):
             freds_profile = syn.get_user_name_profile('fredcommo')
         """
         if id:
-            # For UserProfile
-            if isinstance(id, collections.abc.Mapping) and "ownerId" in id:
-                id = id.ownerId
-            # For TeamMember
-            elif isinstance(id, TeamMember):
-                id = id.member.ownerId
-            # For userName
-            elif isinstance(id, str):
+            if isinstance(id, str):
                 principals = self._findPrincipals(id)
-                if len(principals) == 1:
-                    id = principals[0]["ownerId"]
+                for principal in principals:
+                    if principal.get("userName", None).lower() == id.lower():
+                        id = principal["ownerId"]
+                        break
                 else:
-                    for principal in principals:
-                        if principal.get("userName", None).lower() == id.lower():
-                            id = principal["ownerId"]
-                            break
-                    else:  # no break
-                        raise ValueError(f"Can't find user '{id}'")
+                    raise ValueError(f"Can't find user '{id}'")
             else:
-                raise TypeError("id must be a string, UserProfile, or TeamMember")
+                raise TypeError("id must be a 'userName' string")
         else:
             id = ""
         uri = f"/userProfile/{id}"
@@ -685,7 +675,10 @@ class Synapse(object):
             my_profile = syn.getUserProfile()
             freds_profile = syn.getUserProfile('fredcommo')
         """
-        if id is None:
+        if id:
+            if not isinstance(id, int):
+                raise TypeError("id must be an 'ownerId' integer")
+        else:
             id = ""
         uri = f"/userProfile/{id}"
         return UserProfile(
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -622,6 +622,76 @@ class Synapse(object):
             self.restDELETE("/secretKey", endpoint=self.authEndpoint)
 
     @functools.lru_cache()
+    def get_user_name_profile(
+        self,
+        id: Union[str, UserProfile, TeamMember] = None,
+        sessionToken: str = None,
+    ) -> UserProfile:
+        """
+        Get the details about a Synapse user.
+        Retrieves information on the current user if 'id' is omitted or is empty string.
+        :param id:           The userName, UserProfile, or TeamMember of a user
+        :param sessionToken: The session token to use to find the user profile
+        :returns: The user profile for the user of interest.
+
+        Example::
+            my_profile = syn.get_user_name_profile()
+            freds_profile = syn.get_user_name_profile('fredcommo')
+        """
+        if id:
+            if isinstance(id, collections.abc.Mapping) and "ownerId" in id:
+                id = id.ownerId
+            elif isinstance(id, TeamMember):
+                id = id.member.ownerId
+            elif isinstance(id, str):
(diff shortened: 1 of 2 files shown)


synapseclient/client.py
```diff
@@ -621,6 +621,76 @@ def invalidateAPIKey(self):
...
+        if not id:
+            id = ""
+        uri = f"/userProfile/{id}"
+        return UserProfile(
+            **self.restGET(
+                uri, headers={"sessionToken": sessionToken} if sessionToken else None
+            )
+        )
+
     @functools.lru_cache()
```
thomasyu888, line 694: One initial thought: do we put out a deprecation warning for this function?

synapseclient/client.py
```diff
@@ -621,6 +621,76 @@ def invalidateAPIKey(self):
         if self._is_logged_in():
             self.restDELETE("/secretKey", endpoint=self.authEndpoint)
 
+    @functools.lru_cache()
+    def get_user_name_profile(
```
BryanFauble, line 625: Thinking about naming of these functions:
`def get_user_profile_by_....`
`def get_user_profile_by_username`

might be a bit more consistent - I know when I'm searching for 'like' functions I usually expect the prefix to match.

BWMac: I agree. I'll do something like that

BryanFauble:
I like the approach you are taking on this - While looking for areas/dependent functions let's take the time to update the parameter section of those function(s) docstrings to be clear that typing of the data they're passing in matters and what we'll do with a string vs int

synapseclient/client.py
```diff
@@ -621,6 +621,76 @@ def invalidateAPIKey(self):
...
+            id = ""
+        uri = f"/userProfile/{id}"
+        return UserProfile(
+            **self.restGET(
+                uri, headers={"sessionToken": sessionToken} if sessionToken else None
+            )
+        )
+
+    @functools.lru_cache()
+    def get_user_profile_by_id(
```
BryanFauble, line 669: I might change this to be `get_user_profile_by_owner_id` to be the most specific . The reason is that `id` just means identifier, and username is an identifier - But we aren't intending for them to call this function with the username.

synapseclient/client.py
```diff
@@ -621,6 +621,76 @@ def invalidateAPIKey(self):
...
+        self,
+        id: int = None,
+        sessionToken: str = None,
+    ) -> UserProfile:
+        """
+        Get the details about a Synapse user.
+        Retrieves information on the current user if 'id' is omitted or is empty string.
+        :param id:           The ownerId of a user
```
BryanFauble, line 677: Would it make sense to swap this to `owner_id`?

synapseclient/client.py
thomasyu888: @BryanFauble one thing that we have to think about is that client users generally don't have an understanding of what an "owner_id" is. That said, I do like having being consistent with the API.

I'm slightly confused as why they named it `ownerId`.

BryanFauble: Yeah that makes sense. The changes here do work but I want to be clear for those that haven't worked in synapse before. Do you have an opinion on renaming or keeping? I'm ok with either.

BryanFauble: Given what we talked about in sprint review what this is should be fine since there is little understanding what ownerId is

BryanFauble (approved):
LGTM

synapseclient/client.py
```diff
@@ -621,6 +621,76 @@ def invalidateAPIKey(self):
...
+        :param sessionToken: The session token to use to find the user profile
+        :returns: The user profile for the user of interest.
+
+        Example::
+            my_profile = syn.getUserProfile()
+            freds_profile = syn.getUserProfile('fredcommo')
+        """
+        if not id:
```
thomasyu888, line 685: Nit: if Id is not None

thomasyu888 (approved):
Let's hold on the deprecation warning because we don't want it popping up every time someone uses the functions that leverages getUserProfile.

We will need to create separate tickets for the deprecation and also create platform tickets for not allowing integer usernames.

synapseclient/client.py
```diff
@@ -621,6 +621,72 @@ def invalidateAPIKey(self):
         if self._is_logged_in():
             self.restDELETE("/secretKey", endpoint=self.authEndpoint)
 
+    @functools.lru_cache()
+    def get_user_profile_by_username(
+        self,
+        id: str = None,
```
thomasyu888, line 627: Should this be "username"?

synapseclient/client.py
```diff
@@ -621,6 +621,72 @@ def invalidateAPIKey(self):
...
+                for principal in principals:
+                    if principal.get("userName", None).lower() == id.lower():
+                        id = principal["ownerId"]
+                        break
+                else:
+                    raise ValueError(f"Can't find user '{id}'")
+            else:
+                raise TypeError("id must be a 'userName' string")
+        else:
+            id = ""
```
thomasyu888, line 653: Can there be a way to un-nest this?  For example, did we want to fast fail if the username isn't a string or None?

thomasyu888 (changes requested):
Excellent - here's some feedback.

synapseclient/client.py
BWMac: changed in latest commit

synapseclient/client.py
BWMac: I moved the `TypeError` to the top of the function in the latest commit

synapseclient/client.py
```diff
@@ -621,6 +621,71 @@ def invalidateAPIKey(self):
...
+    @functools.lru_cache()
+    def get_user_profile_by_username(
+        self,
+        username: str = None,
+        sessionToken: str = None,
+    ) -> UserProfile:
+        """
+        Get the details about a Synapse user.
+        Retrieves information on the current user if 'id' is omitted or is empty string.
+        :param username:           The userName of a user
```
BryanFauble, line 633: nit: Line up the tabbing for the description of the :param with eachother

synapseclient/client.py
```diff
@@ -621,6 +621,71 @@ def invalidateAPIKey(self):
...
+        :param sessionToken: The session token to use to find the user profile
+        :returns: The user profile for the user of interest.
+
+        Example::
+            my_profile = syn.get_user_name_profile()
```
BryanFauble, line 638: These examples need to be update the reflect the new name

synapseclient/client.py
```diff
@@ -621,6 +621,71 @@ def invalidateAPIKey(self):
...
+            freds_profile = syn.get_user_name_profile('fredcommo')
+        """
+        if not isinstance(username, str) and username is not None:
```
BryanFauble, line 641: nit: It would read a little better to assign these conditions to variables like:

is_none = username is None
is_str = isinstance(username, str)
if not is_str and not is_none:
    # type error
if is_str:
    # do stuff

synapseclient/client.py
```diff
@@ -621,6 +621,71 @@ def invalidateAPIKey(self):
...
+    ) -> UserProfile:
+        """
+        Get the details about a Synapse user.
+        Retrieves information on the current user if 'id' is omitted.
+        :param id:           The ownerId of a user
+        :param sessionToken: The session token to use to find the user profile
+        :returns: The user profile for the user of interest.
+
+        Example::
+            my_profile = syn.getUserProfile()
```
BryanFauble, line 674: These examples need to be updated with the new function name

synapseclient/client.py
BWMac: added in latest commit

synapseclient/client.py
BWMac: updated in latest commit

synapseclient/client.py
```diff
@@ -621,6 +621,73 @@ def invalidateAPIKey(self):
...
+            freds_profile = syn.getUserProfile('fredcommo')
```
BryanFauble, line 677:
```suggestion
            my_profile = syn.get_user_profile_by_id()
            freds_profile = syn.get_user_profile_by_id(1234567)
```

BryanFauble (approved):
LGTM!

thomasyu888 (approved):
🚀 LGTM!

PR #1016, opened 2023-11-27
[SYNPY-1337] Adding note about unique .db file

BryanFauble:
**Problem:**

1. When running a migration for multiple projects/folders you cannot re-use the same `.db` file as there is a check that will prevent overwriting data for previous indexes.
2. During long-running migrations we want to be notified on failure or success

**Solution:**

1. Adding a best practice to specify a unique name per project/folder to migrate
2. Added a syn.sendMessage to the script to notify on success or failure
image

### DIFF
--- a/docs/articles/data_storage.rst
+++ b/docs/articles/data_storage.rst
@@ -198,61 +198,78 @@ Putting all the migration pieces together
 -----------------------------------------
   .. code-block::
 
-    import os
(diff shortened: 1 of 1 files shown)


BWMac (approved):
:thumbsup:

thomasyu888 (approved):
🔥 LGTM!

PR #1017, opened 2023-11-27
[SYNPY-1333] Print transfer progress for FTP

BryanFauble:
**Problem:**
When a file is being downloaded from an FTP link there is no indication that anything is occurring.

**Solution:**
Adding in a reporthook function that will print the transfer progress based on the block number, read size, and total size.

**Testing:**

1. I tested this out with a script similar to what tom added in the Jira:
```
import synapseclient

syn = synapseclient.Synapse(debug=True)
syn.login()


file_ent = synapseclient.File(
    "ftp://ftp.sra.ebi.ac.uk/vol1/fastq/SRR197/048/SRR19761348/SRR19761348_1.fastq.gz",
    name="SRR19761348_1.fastq.gz",
    # This parameter will store this file as a link
    synapseStore=False,
    parent="syn52569429",
)
file_ent = syn.store(obj=file_ent)
syn.get(entity=file_ent, downloadFile=True)

```

After running the script I get:
`Downloading  [--------------------]0.63%   9.0MB/1.4GB (3.1MB/s) SRR19761348_1.fastq.gz`

This is just like downloading synapse stored file which look like:
`Downloading  [####################]100.00%   12.0bytes/12.0bytes (4.5kB/s) my_super_cool_junk_file.txt`

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2731,6 +2731,7 @@ class Synapse(object):
         actual_md5 = None
         redirect_count = 0
         delete_on_md5_mismatch = True
+        self.logger.debug(f"Downloading from {url} to {destination}")
         while redirect_count < REDIRECT_LIMIT:
             redirect_count += 1
             scheme = urllib_urlparse.urlparse(url).scheme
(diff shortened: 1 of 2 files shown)


synapseclient/client.py
```diff
@@ -2747,7 +2748,24 @@ def _download_from_URL(
                 )
                 break
             elif scheme == "ftp":
-                urllib_request.urlretrieve(url, destination)
+                transfer_start_time = time.time()
+
+                def _ftp_report_hook(
```
thomasyu888, line 2753: Nit: does this hook already exist as a convenience function? we use this logging function in other places already.

thomasyu888 (approved):
🔥 nice! Thanks for pushing this through.  Just a small comment but it looks good.

synapseclient/client.py
thomasyu888: Makes sense to me.  Thanks!

PR #1018, opened 2023-11-30
Adding a label to the dockerfile to automatically labe it for this repo

BryanFauble:
Problem:
We need to manually add new containers to this GH repo

Solution:
Automatically set the label

Testing:
image


BWMac:
@BryanFauble This doesn't prevent us from needing to manually push container images, does it? From what I understand this is just adding metadata linking the container image to this repo.

BWMac (approved):
Cool. I created this [ticket](https://sagebionetworks.jira.com/browse/SYNPY-1341) to track automatically pushing new containers on version release.

PR #1019, opened 2023-11-30
Updates Dockerfile to Correctly Install Dependencies

BWMac:
**Problem:**

Newer dependencies are not correctly installed when you run `docker build .` with the current Dockerfile.

**Solution:**

Update the Dockerfile to correctly install the dependencies.


BryanFauble (approved):
LGTM! Thanks for making this change.

PR #1022, opened 2023-12-07
[SYNPY-1349] Detect annotation type

BryanFauble:
**Problem:**
Working with annotations was previously requiring a lot more verbose information. I was forcing a Type to be defined.

**Solution:**
Using the same logic that is in the current python client convert based off the class types into the appropriate type for the Synapse REST API. This allows us to simply define a list or singular annotation.

**Testing:**
See the test scripts.

### DIFF
--- a/test_scripts/oop_poc_file.py
+++ b/test_scripts/oop_poc_file.py
@@ -9,12 +9,12 @@ The following actions are shown in this script:
 """
 import asyncio
 import os
+
 from synapseclient.models import (
     File,
     Folder,
-    AnnotationsValueType,
-    AnnotationsValue,
 )
+from datetime import date, datetime, timedelta, timezone
 import synapseclient
 
 from opentelemetry import trace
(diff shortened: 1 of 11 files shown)


test_scripts/oop_poc_file.py
```diff
@@ -50,21 +50,19 @@ def create_random_file(
...
-        ),
-        "my_key_long": AnnotationsValue(
-            type=AnnotationsValueType.LONG, value=[1, 2, 3]
-        ),
-        "my_key_timestamp": AnnotationsValue(
-            type=AnnotationsValueType.TIMESTAMP_MS, value=[1701362964066, 1577862000000]
-        ),
+        "my_single_key_string": "a",
+        "my_key_string": ["b", "a", "c"],
+        "my_key_bool": [False, False, False],
```
thomasyu888, line 55: I forget, but what does that convert function do when there are mixed types?

thomasyu888 (approved):
🔥 LGTM!  Thanks for putting this together.

test_scripts/oop_poc_file.py
BryanFauble: It's returns an `object` which is than turned into a string:
`nested_annos[key] = {"type": "STRING", "value": [str(e) for e in elements]}`

PR #1024, opened 2023-12-13
[SYNPY-1321] Download benchmark results

BryanFauble:
**Problem:**
We had no stats on download times for varying file counts and sizes

**Solution:**
Create a benchmark script and ran on an EC2 instance created through the service catalog. This led to interesting information when starting to go into higher thread counts with > 1GB files. Something with the size of the files led to a lot of instability; As a result my recommendation was to not exceed the default thread count for these larger file sizes.


thomasyu888 (approved):
🔥 LGTM!

PR #1025, opened 2023-12-14
[SYNPY-1360] Migrating to mkdocstrings

BryanFauble:
Problem:
The Sphinx documentation was a eye-sore that was difficult to work with.

Solution:
Migrating to mkdocstrings/python along with updating many docstrings to google format

Testing:
- Stepped through the pages locally using both `mkdocs serve` and building the docs via `mkdocs build` and starting an HTTP server locally with `python3 -m http.server` and verifying that content was served correctly
- Auto deployment is setup and working. I set it to use this feature branch (Will update to on push to master) and verified that new changes are auto-deployed out to the `gh-pages` branch and than deployed to the live site:
image


docs/explanations/README
thomasyu888, line 1: I like this, we can build this out further as we continually improve the documentation

synapseclient/annotations.py
```diff
@@ -116,8 +121,15 @@ def _annotation_value_list_element_type(annotation_values: typing.List):
     return object
 
 
-def is_submission_status_annotations(annotations):
-    """Tests if the given dictionary is in the form of annotations to submission status"""
+def is_submission_status_annotations(annotations: collections.abc.Mapping) -> bool:
+    """Tests if the given dictionary is in the form of annotations to submission status
+
+    Arguments:
```
thomasyu888, line 127: Nit:  I think `Args` is also acceptable.  I think we either support both `Arguments` and `Args` or just one of them.  If one of them, I leave it to you what you think we should aim to support.

thomasyu888 (approved):
🔥 LGTM - I had some small notes here and there, but we can iterate on this.

synapseclient/annotations.py
BryanFauble: They are both supported by mkdocs. So I am ok with both being used interchangably

PR #1026, opened 2023-12-15
[SYNPY-1302] add method to implement get entity permission api

danlu1:
Problem:
The original `getPermission` method uses /acl endpoint only that shows [inconsistent behavior in the case of publicly downloadable content from Synapse.](https://sagebionetworks.jira.com/browse/PLFM-8079) 

Solution:
create a `get_permissions` method that implements [GET /entity/{id}/permissions API](https://rest-docs.synapse.org/rest/GET/entity/id/permissions.html). These API factors in the entity's ACL and the user's group membership when determining the caller's permission to the entity.  

Example:
Screenshot 2023-12-15 at 9 31 28 AM

Testing:
Unit tests and integration tests are included in the PR.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -120,6 +120,7 @@ from synapseclient.core.upload.upload_functions import (
 from synapseclient.core.dozer import doze
 from typing import Union
 from opentelemetry import trace
+from dataclasses import dataclass
 
 tracer = trace.get_tracer("synapseclient")
 
@@ -2312,6 +2313,28 @@ class Synapse(object):
                     permissions["accessType"]
                 )
         return list(effective_permission_set)
+
+    @tracer.start_as_current_span("Synapse::get_permissions")
+    def get_permissions(
+        self, entity: Union[Entity, Evaluation, str, collections.abc.Mapping]
+    ):
+        """Get the permissions that the caller has on an Entity.
+        https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/auth/UserEntityPermissions.html
+
+        :param entity:      An Entity or Synapse ID to lookup
+
+        :returns:           An Permission object
+        Example::
+            permissions = syn.get_permissions("syn12345")
+            permissions = syn.get_permissions(Entity Object)
+            permissions.access_types # to view the access types
+
+        """
+        entity_id = id_of(entity)
+        trace.get_current_span().set_attributes({"synapse.id": entity_id})
+        url = f"/entity/{entity_id}/permissions"
+        data = self.restGET(url)
+        return Permissions(data)
 
     @tracer.start_as_current_span("Synapse::setPermissions")
     def setPermissions(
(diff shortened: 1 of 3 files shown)


synapseclient/client.py
```diff
@@ -2313,6 +2314,28 @@ def getPermissions(
...
         return list(effective_permission_set)
 
+    @tracer.start_as_current_span("Synapse::get_permissions")
+    def get_permissions(
+        self, entity: Union[Entity, Evaluation, str, collections.abc.Mapping]
+    ):
+        """Get the permissions that the caller has on an Entity.
+        https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/auth/UserEntityPermissions.html
+
+        :param entity:      An Entity or Synapse ID to lookup
```
BryanFauble, line 2324: Let's swap to using google style docstrings as that's what we want to use moving forward.

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/a8bc6a899bfda682c998043bd5f813386dd9c1e6/synapseclient/client.py

For an example of what it looks like for the new doc pages.

In VSC my `Auto Docstring: Docstring Format` is set to:
`google-notypes`

synapseclient/client.py
```diff
@@ -2313,6 +2314,28 @@ def getPermissions(
                 )
         return list(effective_permission_set)
 
+    @tracer.start_as_current_span("Synapse::get_permissions")
+    def get_permissions(
+        self, entity: Union[Entity, Evaluation, str, collections.abc.Mapping]
+    ):
```
BryanFauble, line 2320: Set a return type on the function

synapseclient/client.py
```diff
@@ -5247,3 +5270,146 @@ def _return_rest_body(self, response):
         if is_json(response.headers.get("content-type", None)):
             return response.json()
         return response.text
+
+
+@dataclass
+class Permissions:
```
BryanFauble, line 5276: Let's move this out to a `models` directory to limit the amount of new content that we add to the `client.py`.

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/SYNPY-1322-OOP-POC/synapseclient/models/__init__.py is an example of what I had done for the Object-Oritentated-Programming PoC.

We could add this `Permissions` dataclass to both:
`synapseclient/models/__init__.py`
and
`synapseclient/__init__.py`

To allow either import.

synapseclient/client.py
```diff
@@ -5247,3 +5270,146 @@ def _return_rest_body(self, response):
...
+    """
+    The permission a User has for a given Entity.
+
+    Attributes:
+        can_view (Boolean): Can the user view this entity?
```
BryanFauble, line 5281: Let's plan to not include types in the docstring as `mkdocstrings` will automatically grab this information based off the typehint of the field.

synapseclient/client.py
```diff
@@ -5247,3 +5270,146 @@ def _return_rest_body(self, response):
...
+        can_enable_inheritance (Boolean): (Read Only) Can the user delete the entity's access control list (so it inherits settings from an ancestor)?
+        owner_principal_id (Integer): (Read Only) The principal ID of the entity's owner (i.e. the entity's 'createdBy')
+        can_public_read (Boolean): (Read Only) Is this entity considered public?
+        can_moderate (Boolean): Can the user moderate the forum associated with this entity? Note that only project entity has forum.
+        is_certification_required (Boolean): (Read Only) Is the certification requirement enabled for the project of the entity?
+        is_entity_open_data (Boolean): (Read Only) Returns true if the Entity's DateType equals 'OPEN_DATA', indicating that the data is safe to be released to the public.
+    """
+
+    can_view: bool = None
+    """Can the user view this entity?"""
```
BryanFauble, line 5302: Formatting might look a little better to add a new line inbetween the docstring and the new attribute below.

synapseclient/client.py
```diff
@@ -5247,3 +5270,146 @@ def _return_rest_body(self, response):
...
+    can_public_read: bool = None
+    """(Read Only) Is this entity considered public?"""
+    can_moderate: bool = None
+    """Can the user moderate the forum associated with this entity? Note that only project entity has forum."""
+    is_certification_required: bool = None
+    """(Read Only) Is the certification requirement enabled for the project of the entity?"""
+    is_entity_open_data: bool = None
+    """(Read Only) Returns true if the Entity's DateType equals 'OPEN_DATA', indicating that the data is safe to be released to the public."""
+
+    def __init__(self, data: dict):
```
BryanFauble, line 5338: I'm not sure this is how we should create an instance of this class. The reason is that if you are overriding the init dunder method than it prevent you from creating a Permissions instance like:
```
my_other_permissions = Permissions(can_view=True)
print(my_other_permissions)
```

This might make more sense:
```
    @classmethod
    def from_dict(cls, data: dict) -> "Permissions":
        """
            Convert a data dictory to an instance of this dataclass.
        """
        return cls(
            can_view=data["canView"],
            can_edit=data["canEdit"],
            can_move=data["canMove"],
            can_add_child=data["canAddChild"],
            can_certified_user_edit=data["canCertifiedUserEdit"],
            can_certified_user_add_child=data["canCertifiedUserAddChild"],
            is_certified_user=data["isCertifiedUser"],
            can_change_permissions=data["canChangePermissions"],
            can_change_settings=data["canChangeSettings"],
            can_delete=data["canDelete"],
            can_download=data["canDownload"],
            can_upload=data["canUpload"],
            can_enable_inheritance=data["canEnableInheritance"],
            owner_principal_id=data["ownerPrincipalId"],
            can_public_read=data["canPublicRead"],
            can_moderate=data["canModerate"],
            is_certification_required=data["isCertificationRequired"],
            is_entity_open_data=data["isEntityOpenData"],
        )
```

In the `get_permissions` function:
`return Permissions.from_dict(data)`


Thoughts?

tests/integration/synapseclient/integration_test.py
```diff
@@ -870,3 +870,293 @@ def test_get_permissions_for_project_with_public_and_registered_user(self):
...
+        "integration_test::TestPermissionsOnEntityForCaller::test_get_permissions_default"
+    )
+    def test_get_permissions_default(self):
+        # GIVEN a project created with default permissions of administrator
+        project_with_default_permissions: Entity = self.syn.store(
+            Project(name=str(uuid.uuid4()) + "test_get_permissions_default_permissions")
+        )
+        self.schedule_for_cleanup(project_with_default_permissions)
+
+        # AND the user that created the project
```
BryanFauble, line 891: Remove this commented out code

tests/integration/synapseclient/integration_test.py
```diff
@@ -870,3 +870,293 @@ def test_get_permissions_for_project_with_public_and_registered_user(self):
...
+            "CREATE",
+            "MODERATE",
+            "DOWNLOAD",
+        ]
+        assert set(expected_permissions) == set(permissions.access_types)
+
+    @tracer.start_as_current_span(
+        "integration_test::TestPermissionsOnEntityForCaller::test_get_permissions_for_project_with_public_and_registered_user"
+    )
+    def test_get_permissions_for_project_with_public_and_registered_user(self):
```
BryanFauble, line 1090: Is this test testing what we want out of it given that `get_permissions` only grabs the current logged in user permissions on an entity? Are we still able to get what the public user id: `273949` has as far as permissions?

synapseclient/client.py
```diff
@@ -2313,6 +2314,28 @@ def getPermissions(
                 )
         return list(effective_permission_set)
```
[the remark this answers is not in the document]
BryanFauble: Thinking about this:
The current `getPermissions` is doing what we would want a new `get_acl` function does. So perhaps we could create a new function named `get_acl`, cut the logic out of `getPermissions` and call `get_acl` from the current `getPermissions` function.


-----

When you add a deprecation warning on this you'll need to test what it looks like in the mkdocs.

There is no direct 1-to-1 replacement of:
```
    @deprecated.sphinx.deprecated(
        version="2.1.0",
        reason="deprecated and replaced with :py:meth:`xxxxx`",
    )
```

My suggestion is to add this to the top of the docstring:
```
"""**Deprecated** and replaced with [function_name][synapseclient.Synapse.function_name]"""
```

Here is the code I am thinking of:
```
    @tracer.start_as_current_span("Synapse::get_acl")
    def get_acl(
        self,
        entity: Union[Entity, Evaluation, str, collections.abc.Mapping],
        principalId: str = None,
    ) -> typing.List[str]:
        """Get the [ACL](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/ACCESS_TYPE.html) 
            that a user or group has on an Entity.

        Arguments:
            entity: An Entity or Synapse ID to lookup
            principalId: Identifier of a user or group (defaults to PUBLIC users)

        Returns:
            An array containing some combination of 
                ['READ', 'CREATE', 'UPDATE', 'DELETE', 'CHANGE_PERMISSIONS', 'DOWNLOAD'] 
                or an empty array
        """
        principal_id = self._getUserbyPrincipalIdOrName(principalId)

        trace.get_current_span().set_attributes(
            {"synapse.id": id_of(entity), "synapse.principal_id": principal_id}
        )

        acl = self._getACL(entity)

        team_list = self._find_teams_for_principal(principal_id)
        team_ids = [int(team.id) for team in team_list]
        effective_permission_set = set()

        # This user_profile_bundle is being used to verify that the principal_id is a registered user of the system
        user_profile_bundle = self._get_user_bundle(principal_id, 1)

        # Loop over all permissions in the returned ACL and add it to the effective_permission_set
        # if the principalId in the ACL matches
        # 1) the one we are looking for,
        # 2) a team the entity is a member of,
        # 3) PUBLIC
        # 4) A user_profile_bundle exists for the principal_id
        for permissions in acl["resourceAccess"]:
            if "principalId" in permissions and (
                permissions["principalId"] == principal_id
                or permissions["principalId"] in team_ids
                or permissions["principalId"] == PUBLIC
                or (
                    permissions["principalId"] == AUTHENTICATED_USERS
                    and user_profile_bundle is not None
                )
            ):
                effective_permission_set = effective_permission_set.union(
                    permissions["accessType"]
                )
        return list(effective_permission_set)

    @tracer.start_as_current_span("Synapse::getPermissions")
    def getPermissions(
        self,
        entity: Union[Entity, Evaluation, str, collections.abc.Mapping],
        principalId: str = None,
    ) -> typing.List[str]:
        """
        **Deprecated** and replaced with [get_acl][synapseclient.Synapse.get_acl].
        
        
        Get the permissions that a user or group has on an Entity.
        
        Arguments:
            entity: An Entity or Synapse ID to lookup
            principalId: Identifier of a user or group (defaults to PUBLIC users)

        Returns:
            An array containing some combination of
                ['READ', 'CREATE', 'UPDATE', 'DELETE', 'CHANGE_PERMISSIONS', 'DOWNLOAD']
                or an empty array
        """
        return self.get_acl(entity=entity, principalId=principalId)
```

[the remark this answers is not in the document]
BryanFauble: One thing I didn't consider in my above answer is that without the sphinx directive you lose the warning that is emitted when calling a deprecated method. We can directly replace the sphinx method though like:
Before:
```
    @deprecated.sphinx.deprecated(
        version="2.1.0",
        reason="deprecated and replaced with :py:meth:`get_annotations`",
    )
```

After:
```
@deprecated.deprecated(version="2.1.0", reason="deprecated and replaced with synapseclient.Synapse.get_annotations")
```

The `reason` in the deprecation warning is not added to the mkdocs, so we still need to include the information in the docstring pointing someone to the new method.

BryanFauble:
@danlu1 Something for when you are back in the office after the holidays.

I had missed this originally, but since this is running on the fork of the main `Sage-Bionetworks` repo you are missing a few key secrets for 1) Integration tests to run and 2) The newly added logic for SonarCloud scans to run in the CI pipeline.

Please raise a PR from within the main repo instead of a fork. I'll be working to update our documentation to mention these changes. This does not change external collaborators who are still required to work from a fork.

thomasyu888:
What do you all think about completing these changes here first due to all the comments that have been made?

Then create a feature branch in SYNPY to merge this branch into which can then be merged into develop.

BryanFauble:
> What do you all think about completing these changes here first due to all the comments that have been made?
Then create a feature branch in SYNPY to merge this branch into which can then be merged into develop.

Great plan! Let's keep the changes and discussion here and move it over when it's ready so we can see all the pre-merge checks.

danlu1:
Closed in favor of [new PR](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1037).

PR #1027, opened 2023-12-15
Draft: Testing some things that should fail sonar

BryanFauble:


synapseclient/client.py
```diff
@@ -5182,6 +5182,19 @@ def restPUT(
         )
         return self._return_rest_body(response)
 
+    
+
+    def a____sneaky_function_that_should_____Fail_sOnAr(self):
```
thomasyu888, line 5187: 👀👀👀👀

PR #1030, opened 2023-12-20
[SYNPY-1357] Allow multiple values in manifest TSV

BryanFauble:
**Problem:**

1. When using the `syncFromSynapse` function we were only storing the first indexed value for an annotation that had multiple values.
2. When using the `syncToSynapse` function we had no way to specify that multiple values are supposed to be added to an Annotation.

**Solution:**

1. During the `syncFromSynapse` function I am writing a comma delimited list wrapped in brackets for multiple annotation values. This is to match the expected format that works when syncing data to a FileView in Synapse.
2. During the `syncToSynapse` function I am splitting fields by the comma delimited list wrapped in brackets to allow for multiple annotation values to be applied to an individual annotation key.

**Testing:**

1. I updated unit tests and integration tests to test out this functionality.
2. I manually tested this and verified when using either function I could update annotations, and they were updated. Also that no-changes did not modify the annotations as well:

image

This manifest TSV generated the screenshot below:
```
path	parent	name	id	synapseStore	contentType	used	executed	activityName	activityDescription	my_string_with_a_backslash	multi_value_annotation_string	multi_value_annotation_date	multi_value_booleans	multi_value_longs	multi_value_ints	single_value_int
/home/bfauble/my_synapse_downloads/ParentFolder/another_file.txt	syn52570245	file_1.txt	syn52570249	True	text/plain					my\ first string	[hello,my,annotation,"with quotes","with,commas,in,quotes"]	[2000-01-01T00:00:00.000Z,2001-01-01T01:00:00.000Z,2002-01-01T02:00:00.000Z]	[false,FalSe,tRuE,TRUE]	[1.2,3.4,5.6]	[1,2,3,4]	[1]
```
image


docs/explanations/manifest_tsv.md
```diff
@@ -73,6 +75,14 @@ See:
...
-## See:
+### Dates in the manifest file
+Dates within the manifest file will always be written as [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) format in UTC without milliseconds. For example: `2023-12-20T16:55:08Z`.
+
+Dates can be written in other formats specified in ISO 8601 and it will be reconginzed, however, the [synapseutils.syncFromSynapse][] will always write this in the UTC format specified above. For example you may want to specify a datetime at a specific timezone like: `2023-12-20 23:55:08-07:00` and this will be recognized as a valid datetime.
+
+
+## Refernces:
 
+- [synapseutils.syncFromSynapse][]
```
thomasyu888, line 86: Do these references link to the python API docs?

synapseclient/core/utils.py
```diff
@@ -501,20 +551,41 @@ def from_unix_epoch_time(ms) -> datetime.datetime:
     return from_unix_epoch_time_secs(ms / 1000.0)
 
 
-def datetime_to_iso(dt, sep="T"):
-    # Round microseconds to milliseconds (as expected by older clients)
-    # and add back the "Z" at the end.
-    # see: http://stackoverflow.com/questions/30266188/how-to-convert-date-string-to-iso8601-standard
+def datetime_to_iso(
```
thomasyu888, line 554: Ah interesting, so there are places where we require the milliseconds, but just not for manifests?

synapseutils/sync.py
```diff
@@ -1021,9 +1130,50 @@ def syncToSynapse(
...
+    """
+    values_to_return = []
+
+    cell_values = cell.split(";")
+    for annotation_value in cell_values:
+        if possible_datetime := datetime_or_none(annotation_value):
+            values_to_return.append(possible_datetime)
+        elif (possible_int := int_or_none(annotation_value)) is not None:
+            values_to_return.append(possible_int)
+        elif (possible_float := float_or_none(annotation_value)) is not None:
```
thomasyu888, line 1155: You may want to flip these.  Technically you can do something like

```
int(3.4)
```
without an error

synapseclient/core/utils.py
```diff
@@ -326,6 +326,56 @@ def is_synapse_id_str(obj):
...
+    """
+    Attempts to convert a string to a bool. Returns None if it fails.
+
+    Args:
+        input_value: The string to convert to a bool
+
+    Returns:
+        The bool or None if the conversion fails
+    """
+    if input_value == "True" or input_value == "true":
```
thomasyu888, line 339: Will we need to worry about 1 and 0?

Nit: At this point in the `if else` we are assuming that `input_value` will most likely be a string, could we do `input_value.lower()` to simplify this just in case we get strings like `TRUE`

thomasyu888 (changes requested):
Awesome start, left some comments.

synapseutils/sync.py
```diff
@@ -796,14 +824,78 @@ def _get_file_entity_provenance_dict(syn, entity):
...
+
+    Returns:
+        The dict representation of the row.
+    """
+    data_to_write = {}
+    for key in keys:
+        data_for_key = row.get(key, "")
+        if isinstance(data_for_key, list):
+            items_to_write = _convert_manifest_data_items_to_string_list(data_for_key)
+            data_to_write[key] = ";".join(items_to_write)
```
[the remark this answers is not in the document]
BryanFauble: Yeah this is something I wasn't sure how we wanted to do it in the end, and I am open to whichever approach makes the most sense.

We already had an established pattern for the used and executed provenance fields: https://python-docs.synapse.org/explanations/manifest_tsv/#activityprovenance-fields

So that is the approach I took here.

Who would be a good source to make a decision on this direction?

docs/explanations/manifest_tsv.md
BryanFauble: Yes this is done through cross-references: https://mkdocstrings.github.io/usage/#cross-references

This will send you to the API Reference for this function.

tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -166,25 +167,78 @@ def test_syncToSynapse(test_state):
     # The dates in the manifest can accept a variety of formats, however we are always writing
     # them back in the same expected format. Verify they're converted correctly.
     assert new_anots.loc[:]["date_1"].tolist() == [
-        "2020-01-01 00:00:00+00:00",
+        "2020-01-01T00:00:00Z",
```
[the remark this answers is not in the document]
BryanFauble: All of the date formats will work for the syncToSynapse. ie:

- YYYY-MM-DD
- YYYY-MM-DDTHH:MM:SSZ
- YYYY-MM-DD HH:MM:SS
- YYYY-MM-DD HH:MM:SSZ-01:00

Basically anything defined in ISO 8601: https://en.wikipedia.org/wiki/ISO_8601

When we are writing the data back into the manifest TSV during the syncFromSynapse we will always write it as a UTC datetime in `YYYY-MM-DDTHH:MM:SSZ` format

synapseutils/sync.py
```diff
@@ -1021,9 +1130,50 @@ def syncToSynapse(
...
+
+    Returns:
+        The list of items to be used as annotations. Or a single instance if that is
+            all that is present.
+    """
+    values_to_return = []
+
+    cell_values = cell.split(";")
+    for annotation_value in cell_values:
+        if possible_datetime := datetime_or_none(annotation_value):
```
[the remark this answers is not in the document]
BryanFauble: Yeah I can add it here. Specifically I added it below because a bool `False` and number `0` would cause the if check to fail. In this datetime case we are getting back an Object or None.

synapseutils/sync.py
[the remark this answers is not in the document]
BryanFauble: Yes this is correct. I had considered adding in an escape `\` character in these cases.

If we keep the semi-colon delimited list I'll add escaping in, or, if we swap to `a string representation of an array` I'd need to make sure however we are storing the array also has any special characters properly escaped.

synapseclient/core/utils.py
BryanFauble: I'll double check on this logic. I know In the UI milliseconds are not enter-able. I'll check behavior through the `syn.set_annotations` functionality and align this logic to work the same way.

synapseutils/sync.py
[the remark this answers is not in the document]
thomasyu888: Let's float this on #synapse and tag milen and Andrew for feedback.

synapseclient/core/utils.py
BryanFauble: I think 1 and 0 should rightfully be parsed as an int as that's how pandas would be interpret the value.

I'll add in a case insensitive check here.

synapseutils/sync.py
BryanFauble: Great point. I will review this logic.

synapseutils/sync.py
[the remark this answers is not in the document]
BryanFauble: I updated this to store values comma delimited instead of semi-colon delimited in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1030/commits/f3c677daf9fac5ed6b50256b9a61e4c935effa34

In the end it'll look like:
image

`\` will be an escape character to allow a comma to be used in a string.

synapseutils/sync.py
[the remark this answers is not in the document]
BryanFauble: I added this escape character

synapseutils/sync.py
BryanFauble: I swapped to using https://docs.python.org/3/library/ast.html#ast.literal_eval here instead of doing my own parsing

synapseclient/core/utils.py
BryanFauble: I updated to allow case insensitivity here

synapseclient/core/utils.py
BryanFauble: ok so milliseconds are allowed. I updated the logic in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1030/commits/b73fd33bc4477996187f7fce209cf7dc1f5a8121 - This allows the date annotation to have milliseconds and when I am writing the manifest file only write the milliseconds if it's not 0.

tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -34,9 +35,9 @@ def __init__(self):
             self.f2 = utils.make_bogus_data_file(n=10)
             self.f3 = "https://www.synapse.org"
 
-            self.header = "path	parent	used	executed	activityName	synapseStore	foo	date_1	datetime_1	datetime_2	datetime_3\n"
+            self.header = "path	parent	used	executed	activityName	synapseStore	foo	date_1	datetime_1	datetime_2	datetime_3	multiple_strings	multiple_dates	multiple_bools	multiple_ints	multiple_floats\n"
             self.row1 = (
-                '%s	%s	%s	"%s;https://www.example.com"	provName		bar	2020-01-01	2023-12-04T07:00:00Z	2023-12-05 23:37:02.995000+00:00	2023-12-05 07:00:00+00:00\n'
+                '%s	%s	%s	"%s;https://www.example.com"	provName		bar	2020-01-01	2023-12-04T07:00:00Z	2023-12-05 23:37:02+00:00	2023-12-05 07:00:00+00:00	a,b,c,d	2020-01-01,2023-12-04T07:00:00.111Z,2023-12-05 23:37:02.333+00:00,2023-12-05 07:00:00+00:00	fAlSe,False,tRuE,True	1,2,3,4	1.2,3.4,5.6,7.8\n'
```
thomasyu888, line 40: Do we have a test case of a string with escaped commas?  What does that look like right now when theres a table value with 'my sentence, comma, has commas'?

synapseutils/sync.py
```diff
@@ -1021,9 +1157,55 @@ def syncToSynapse(
         _manifest_upload(syn, df)
 
 
-def _manifest_upload(syn, df):
+def _convert_cell_in_manifest_to_python_types(
+    cell: str,
+) -> typing.Union[typing.List, datetime.datetime, float, int, bool, str]:
+    """
+    Takes a possibly semi-colon delimited cell from the manifest TSV file into a list
```
thomasyu888, line 1164: Nit: update description

synapseutils/sync.py
```diff
@@ -1021,9 +1157,55 @@ def syncToSynapse(
...
+    Args:
+        cell: The cell item to convert.
+
+    Returns:
+        The list of items to be used as annotations. Or a single instance if that is
+            all that is present.
+    """
+    values_to_return = []
+
+    cell_values = re.split(pattern=COMMA_PATTERN, string=cell)
```
thomasyu888, line 1176: Is it confusing that setting multiple annotations via a fileview is different from setting multiple annotations via the entity itself?

tests/integration/synapseutils/test_synapseutils_sync.py
BryanFauble: I added this.

This is what this looks like in google sheets for this situation:
image

synapseutils/sync.py
BryanFauble: Updated

synapseutils/sync.py
BryanFauble: The biggest difference is around escaping the comma in this case.

Edit annotation in file view:
image

Edit annotation on the File:
image

Annotation on the Manifest TSV:
Which yes is different from setting it via the manifest TSV:
image

Annotation in Python:
image

BryanFauble:
Adding comments here around internal discussions:

This implementation is going to require a major version update to the client as those users who might currently be using commas or backslashes within their annotations managed through the Manifest TSV file will need to escape those characters.

We will be meeting internally again in January to settle on these changes and propose the path forward for this, possibly tying this with the release to support only the personal access token for authentication and other ground work for future releases.

thomasyu888 (approved):
🔥 nice! Going to approve here - we can create a ticket to revisit in January but this is a working solution.

From Milens comment in slack, it does seem like he wants to make it easier for users.

I think what this means is like you suggested, we should create a platform ticket to ask them how often "commas" are in annotations right now and then decide from there what is the least breaking change for all of our users. (Not just schematic)

thomasyu888 (approved):
🛸 LGTM!  I gave most of my feedback during the first round of PR.

PR #1032, opened 2023-12-21
[SYNPY-1370] Documentation Upgrade

jaymedina:
This PR will update the following documentation:

- [x] Small few left: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/annotations.py 
- [x] Evaluation - The section at the top of the file should be changed from RST into MD format. See [Table for an example](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/table.py): https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/evaluation.py 
- [x] Wiki The section at the top of the file should be changed from RST into MD format. See [Table for an example](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/table.py):: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/wiki.py 
- [x] https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/retry.py 
Should also add this to https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/docs/reference/core.md
- [x] https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/version_check.py 
- [x] https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/upload/multipart_upload.py 
- [x] https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseutils/sync.py

### DIFF
--- a/docs/reference/core.md
+++ b/docs/reference/core.md
@@ -7,7 +7,11 @@ functions.  Use at your own risk.
 ## Multipart Upload
 
 ::: synapseclient.core.upload.multipart_upload
-::: synapseclient.core.upload.multipart_upload._multipart_upload
+
+
+## Retry
+
+::: synapseclient.core.retry
 
 
 ## Utils
--- a/docs/reference/core.md
+++ b/docs/reference/core.md
@@ -7,6 +7,14 @@ functions.  Use at your own risk.
 ## Multipart Upload
 
 ::: synapseclient.core.upload.multipart_upload
+    options:
+        members:
+        - UploadAttempt
+        - executor
+        - multipart_upload_file
+        - multipart_upload_string
+        - multipart_copy
+        - _multipart_upload
 
 
 ## Retry
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -442,6 +442,12 @@ def convert_old_annotation_json(annotations):
     entity bundle JSON (Submissions). we don't need to support newer
     types here e.g. BOOLEAN because they did not exist at the time
     that annotation JSON was saved in this form.
+
+    Arguments:
+        annotations: A parsed JSON dictionary of old style annotations.
+
+    Returns:
+        A v2 Annotation-style dictionary.
     """
 
     meta_keys = ("id", "etag", "creationDate", "uri")
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -154,8 +154,10 @@ def to_submission_status_annotations(annotations, is_private=True):
             from synapseclient.annotations import to_submission_status_annotations
             from datetime import datetime as Datetime
 
-            ## create a submission and get its status
+            ## Initialize a Synapse object
             syn = Synapse()
+
+            ## create a submission and get its status
             submission = syn.submit(evaluation, 'syn11111111')
             submission_status = syn.getSubmissionStatus(submission)
 
@@ -219,6 +221,8 @@ def from_submission_status_annotations(annotations) -> dict:
 
     Example: Using this function
         Converting from submission status annotations
+
+            from synapseclient.annotations import from_submission_status_annotations
 
             submission_status.annotations = from_submission_status_annotations(submission_status.annotations)
     """
@@ -342,8 +346,9 @@ class Annotations(dict):
 
 
 def to_synapse_annotations(annotations: Annotations) -> typing.Dict[str, typing.Any]:
-    """Transforms a simple flat dictionary to a Synapse-style Annotation object.
-    https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/annotation/v2/Annotations.html
+    """Transforms a simple flat dictionary to a Synapse-style Annotation object. See
+    the [Synapse API](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/annotation/v2/Annotations.html)
+    documentation for more information on Synapse-style Annotation objects.
 
     Arguments:
         annotations: A simple flat dictionary of annotations.
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -149,11 +149,12 @@ def to_submission_status_annotations(annotations, is_private=True):
 
     Example: Using this function
         Adding and converting annotations
-
-            from synapseclient.annotations import to_submission_status_annotations, from_submission_status_annotations
+            from synapseclient.client import Synapse
+            from synapseclient.annotations import to_submission_status_annotations
             from datetime import datetime as Datetime
 
             ## create a submission and get its status
+            syn = Synapse()
             submission = syn.submit(evaluation, 'syn11111111')
             submission_status = syn.getSubmissionStatus(submission)
 
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -149,6 +149,7 @@ def to_submission_status_annotations(annotations, is_private=True):
 
     Example: Using this function
         Adding and converting annotations
+
             from synapseclient.client import Synapse
             from synapseclient.annotations import to_submission_status_annotations
             from datetime import datetime as Datetime
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -156,6 +156,7 @@ def to_submission_status_annotations(annotations, is_private=True):
 
             ## Initialize a Synapse object
             syn = Synapse()
+            syn.login()
 
             ## create a submission and get its status
             submission = syn.submit(evaluation, 'syn11111111')
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -150,12 +150,12 @@ def to_submission_status_annotations(annotations, is_private=True):
     Example: Using this function
         Adding and converting annotations
 
-            from synapseclient.client import Synapse
+            import synapseclient
             from synapseclient.annotations import to_submission_status_annotations
             from datetime import datetime as Datetime
 
             ## Initialize a Synapse object
-            syn = Synapse()
+            syn = synapseclient.Synapse()
             syn.login()
 
             ## create a submission and get its status
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -280,6 +280,8 @@ class Annotations(dict):
     Example: Creating a few instances
         Creating and setting annotations
 
+            from synapseclient import Annotations
+
             example1 = Annotations('syn123','40256475-6fb3-11ea-bb0a-9cb6d0d8d984', {'foo':'bar'})
             example2 = Annotations('syn123','40256475-6fb3-11ea-bb0a-9cb6d0d8d984', foo='bar')
             example3 = Annotations('syn123','40256475-6fb3-11ea-bb0a-9cb6d0d8d984')
@@ -308,6 +310,9 @@ class Annotations(dict):
 
         Example: Creating a few instances
             Creating and setting annotations
+
+                from synapseclient import Annotations
+
                 example1 = Annotations('syn123','40256475-6fb3-11ea-bb0a-9cb6d0d8d984', {'foo':'bar'})
                 example2 = Annotations('syn123','40256475-6fb3-11ea-bb0a-9cb6d0d8d984', foo='bar')
                 example3 = Annotations('syn123','40256475-6fb3-11ea-bb0a-9cb6d0d8d984')
--- a/synapseclient/annotations.py
+++ b/synapseclient/annotations.py
@@ -21,7 +21,8 @@ Annotate the entity with location data:
 entity.lat_long = [47.627477, -122.332154]
 ```
 
-Record when we collected the data. This will use the current timezone of the machine running the code.
+Record when we collected the data. This will use the current timezone of the machine
+running the code.
 
 ```python
 from datetime import datetime as Datetime
@@ -42,7 +43,8 @@ See:
 
 ## Annotating data sources
 
-Data sources are best recorded using Synapse's [Activity/Provenance][synapseclient.Activity] tools.
+Data sources are best recorded using Synapse's
+[Activity/Provenance][synapseclient.Activity] tools.
 
 ## Implementation details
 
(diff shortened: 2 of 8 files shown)


synapseclient/annotations.py
```diff
@@ -294,17 +294,18 @@ def __init__(
...
-            example3['foo'] = 'bar'
+        Attributes:
+            id:  A Synapse ID, a Synapse Entity object, a plain dictionary in which 'id' maps to a Synapse ID
+            etag: etag of the Synapse Entity
+            values: (Optional) dictionary of values to be copied into annotations
+            **kwargs: additional key-value pairs to be added as annotations
+
+        Example: Creating a few instances
+            Creating and setting annotations
+                example1 = Annotations('syn123','40256475-6fb3-11ea-bb0a-9cb6d0d8d984', {'foo':'bar'})
```
BryanFauble, line 305: While we are touching this let's add to this and include the import as well to the example:
```
from synapseclient import Annotations
```

synapseclient/evaluation.py
```diff
@@ -1,30 +1,28 @@
...
 
-Like entities, evaluations are access controlled via ACLs. The :py:func:`synapseclient.Synapse.getPermissions` and
-:py:func:`synapseclient.Synapse.setPermissions` methods work for evaluations:
+Like entities, evaluations are access controlled via ACLs. The [synapseclient.Synapse.getPermissions][] and
+[synapseclient.Synapse.setPermissions][] methods work for evaluations:
 
     access = syn.getPermissions(evaluation, user_id)
 
-The :py:func:`synapseclient.Synapse.submit` method returns a Submission_ object::
+The [synapseclient.Synapse.submit][] method returns a Submission_ object:
```
BryanFauble, line 20: For `Submission_` we can update that as well to look something like:
`[Submission][synapseclient.evaluation.Submission]`

Which should add an autoreference to the class definition further on the page.

synapseclient/evaluation.py
```diff
@@ -1,30 +1,28 @@
...
 
     entity = syn.get(synapse_id)
     submission = syn.submit(evaluation, entity, name='My Data', team='My Team')
 
-The Submission object can then be used to check the `status <#submission-status>`_ of the submission::
+The Submission object can then be used to check the `status <#submission-status>`_ of the submission:
```
BryanFauble, line 25: Similar to the comment above we can update this to look something like:
`[SubmissionStatus][synapseclient.evaluation.SubmissionStatus]`

BryanFauble:
Excellent start! When you get to another stopping point please use the `Re-request Review` icon next to my name in the Reviewers section and I will take a look.

synapseclient/annotations.py
```diff
@@ -150,9 +150,13 @@ def to_submission_status_annotations(annotations, is_private=True):
     Example: Using this function
         Adding and converting annotations
 
-            from synapseclient.annotations import to_submission_status_annotations, from_submission_status_annotations
+            from synapseclient.client import Synapse
```
jaymedina, line 153: There's more than one way I could have imported this but I chose to keep with the style of the import below it, rather than `from synapseclient import Synapse`, but let me know if there is a preference.

synapseclient/annotations.py
```diff
@@ -150,9 +150,13 @@ def to_submission_status_annotations(annotations, is_private=True):
...
+            from synapseclient.annotations import to_submission_status_annotations
             from datetime import datetime as Datetime
 
+            ## Initialize a Synapse object
+            syn = Synapse()
```
BryanFauble, line 158: Usually we've been adding a line with this as well:
```
syn.login()
```

Just to drill in with repetition that actions require you to intentionally log into the system. We have a small few things anonymous people can do, but this does not look like one of those things.

synapseclient/annotations.py
BryanFauble: Usually in the docs I've always seen:
```
import synapseclient
syn = synapseclient.Synapse()
syn.login()
```

Since it's used in so many places already lets try to keep this standard.

For other things it's been a bit more ambiguous, like:
```
import synapseclient
from synapseclient import Project, File, Column, Table, EntityViewSchema, EntityViewType
syn = synapseclient.Synapse()
syn.login()
```
Which yeah. Since we're already importing `synapseclient` we could do:
```
synapseclient.Project
synapseclient.File
synapseclient.Column
```

Let's at least keep the standard for now (The top most code block)

jaymedina: Noted ✅

jaymedina:
There are a few methods under the Wiki class that are marked as "For Internal use". Do we want to change their names to include `_` so they don't show up in the docs @BryanFauble? (e.g. `def _getURI(self)`)

BryanFauble:
> There are a few methods under the Wiki class that are marked as "For Internal use". Do we want to change their names to include _ so they don't show up in the docs @BryanFauble? (e.g. def _getURI(self))

@jaymedina 

No I don't think we should update these to be private since it's already out in the wild. What we can do instead is to add a filter to this markdown page specifically and filter these out from showing. To do something like that see how we are defining the filter option here:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/mkdocs.yml#L112

You should be able to take that an add an `options` section to the markdown page like we are defining the `members` section here:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/0bd479325edeedb369ebfd601ec350329b7c4672/docs/reference/client.md?plain=1#L6

jaymedina:
> add a filter to this markdown page specifically and filter these out from showing

Hey @BryanFauble - thanks for the example, I always appreciate those.

Before I do this, I think I need some buy-in: If we want to leave these methods public, wouldn't we want to leave their documentation public as well? So I'm either for: hiding the methods and their documentation, or exposing the methods and their documentation, but not hiding one and keeping the other.

BryanFauble:
> > add a filter to this markdown page specifically and filter these out from showing
> 
> Hey @BryanFauble - thanks for the example, I always appreciate those.
> 
> Before I do this, I think I need some buy-in: If we want to leave these methods public, wouldn't we want to leave their documentation public as well? So I'm either for: hiding the methods and their documentation, or exposing the methods and their documentation, but not hiding one and keeping the other.

The other option is to mark it as deprecated. Point them to a replacement if we have it or that it'll be removed in the future.
See my comment here: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1026#discussion_r1431778414

jaymedina:
> @jaymedina The other option is to mark it as deprecated. Point them to a replacement if we have it or that it'll be removed in the future. See my comment here: [#1026 (comment)](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1026#discussion_r1431778414)

Got it, I'll revisit this after a few more modules, thanks

docs/reference/core.md
```diff
@@ -7,7 +7,11 @@ functions.  Use at your own risk.
 ## Multipart Upload
 
 ::: synapseclient.core.upload.multipart_upload
-::: synapseclient.core.upload.multipart_upload._multipart_upload
+
```
jaymedina, line 10: I removed the explicit call to `::: synapseclient.core.upload.multipart_upload._multipart_upload` because it looked confusing in the documentation, and I wanted to have a conversation about it later. From a quick scan of the documentation, this looks to be the only hidden function that's explicitly documented. I see why that is. The returns of the other functions are calls to `_multipart_upload`, and so references to it are made in the Multipart section of Core. But this goes back to a previous conversation of whether or not we should expose the documentation of a hidden function and vice versa. To me if something is hidden with `_`, it's best not to add it to the documentation because it will be hidden from the user unless it's explicitly called, so it's best to avoid possible confusion. But I'm up for iterating on this before a final review.

thomasyu888: One thing we can explicitly state is that people should use functions within `core` at their own risk.   There's not really anything "hidden" within Python packages, but the functions within `core` are for developers only and aren't part of the "public" Python API.

jaymedina: What I mean by hidden is that the functions led by `_` are denoted to be for internal use only, and won't be available unless explicitly called, so it's not functionality that's immediately available to users (a user would have to look inside the code to know it's there).

image

But based on what you're saying, it sounds like we want to document everything in `core`, both internal and external functions since `core` is for users that are more advanced to begin with. If that's the case, I would add and update the docstring for `_multipart_upload` and also expose `retry.py`'s [_get_message](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/e80fa0d7816e6c97bb60c55f35d72a57d742d6c0/synapseclient/core/retry.py#L171)

thomasyu888: I think all functions in `core` and this package should have docstrings even if they aren't surfaced on the online python documentation site.

I don't think we actually need all of `core` to be available in the online python documentation site, but we should document that the functions in `core` are subject to change at any time and for python client users to use at their own risk.

jaymedina:
Hey @BryanFauble , happy new year :slightly_smiling_face: Just wanted quick clarification on the pep8-speaks status report on my PR: Did you want us to follow black's 88 character line-length rule you configured pep8-speaks to follow? Or should I just focus on resolving the SonarCloud issues (SonarCloud is following 120 char. limit)?

BryanFauble:
> Hey @BryanFauble , happy new year 🙂 Just wanted quick clarification on the pep8-speaks status report on my PR: Did you want us to follow black's 88 character line-length rule you configured pep8-speaks to follow? Or should I just focus on resolving the SonarCloud issues (SonarCloud is following 120 char. limit)?

@jaymedina Let's follow the 88 character limit. Sometimes we won't be able to always follow it. In some cases I have seen that having a URL that spans multiple lines breaks when it's rendered on the doc site, but we should strive to follow it as close as we can.

I went ahead and updated the SonarCloud rule to follow this 88 limit.

synapseclient/core/upload/multipart_upload.py
```diff
@@ -80,6 +78,8 @@ def _executor(max_threads, shutdown_wait):
 
 
 class UploadAttempt:
+    """ """
```
BryanFauble, line 81: I personally like having the docstring with a new line myself. But, I'm not sure we have a defined standard around. this. This style guide shows them without: https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings

I would be ok with either, or both.

@thomasyu888 @jaymedina Thoughts on which we should use?


```suggestion
    """
    Used to handle multi-threaded operations for uploading one or parts of a file.
    """
```

jaymedina: I have been following the convention below out of habit, but I can see how starting the first line under the first `"""` is visually cleaner, so I'm find with doing it this way from now on. Doesn't seem to go against pep257 rules.
```
def myfunc():
   """multi-line
   docstring

   """

File: synapseutils/sync.py [hunk 8]
```diff
@@ -719,7 +749,8 @@ def generateManifest(syn, allFiles, filename, provenance_cache=None) -> None:
         syn: A Synapse object with user's login, e.g. syn = synapseclient.login()
         allFiles: A list of File Entity objects on Synapse (can't be Synapse IDs)
         filename: file where manifest will be written
-        provenance_cache: an optional dict of known provenance dicts keyed by entity ids
+        provenance_cache: an optional dict of known provenance dicts keyed by entity
+        ids
```
BryanFauble, line 753: Can you check that this renders correctly on the docs page. This may need to be indented once is properly show up.

jaymedina:
> @jaymedina The other option is to mark it as deprecated. Point them to a replacement if we have it or that it'll be removed in the future. See my comment here: [#1026 (comment)](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1026#discussion_r1431778414)

@BryanFauble, there are no replacements to these so I won't mark them as deprecated, and they're called in other parts of the repo (`client.py`). I think I'll table this for now until the Python client refactoring happens, as I think it's just a design preference at this point. This PR is ready for review now. I allowed for a little bit of scope creep when it felt appropriate and improved the overall documentation (added docstrings to methods/functions without them, updated docstrings of internal functions). Per our conversation at the sprint planning yesterday I'm going to ignore the SonarCloud code smell messages related to lines beyond 88 chars. I can make a low priority Jira ticket to tackle formatting at a later time. Let me know your thoughts. Thanks!

synapseutils/sync.py
jaymedina: It does not. Fixed in the latest commit.

synapseclient/core/upload/multipart_upload.py
BryanFauble: This class will need a docstring on it

jaymedina: Lost this thread for a second. Made the commit, thanks

synapseutils/sync.py
```diff
@@ -837,11 +879,8 @@ def _checkProvenace(item, path):
 
         elif not utils.is_url(item) and (utils.is_synapse_id_str(item) is None):
             raise SynapseProvenanceError(
-                (
-                    "The provenance record for file: %s is incorrect.\n"
-                    "Specifically %s, is neither a valid URL or synapseId."
-                )
-                % (path, item)
+                "The provenance record for file: %s is incorrect.\n"
```
BryanFauble, line 882: nit: We are moving towards using f-strings over placeholders.

https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/

BryanFauble (approved):
Excellent works! Thanks for taking your time to think through all of these changes.

synapseutils/sync.py
jaymedina: Sounds good - instead of resolving this in this PR, what do you think of me creating a low-priority Jira ticket for formatting updates in the codebase (using f-strings & making sure lines do not exceed 88 chars), and marking it as "related to" the python client refactoring initiative in 2024? @BryanFauble @thomasyu888

BryanFauble: @jaymedina Sounds like a great plan. We'll be chipping away at these are we revisit portions of the code base, but it'll be nice to have this as something can focus on later.

jaymedina: I've made the [Jira ticket](https://sagebionetworks.jira.com/browse/SYNPY-1388) for this. Feel free to update as needed.

jaymedina:
Great, thanks reviewers for the help

PR #1035, opened 2023-12-22
omit the doc directory from code coverage

BryanFauble:
**Problem:**
Our docs directory is going to hold test scripts used during tutorials and other one-off processes. We are not going to test these through unit/integration tests and should not count against our code coverage.

**Solution:**
Exclude the docs directory.

**Testing:**
This PR will serve as the test to verify that the docs directory is excluded from the coverage report.

The code coverage % is going up it looks like by around the amount I'd be expecting. Once merged I'll give in the metrics again to make sure it is updated in the `develop` branch.
image

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[run]
+omit =
+    docs/*


jaymedina (approved):

thomasyu888 (approved):
💯

PR #1037, opened 2023-12-26
[SYNPY-1302] Replace getPermission with get_acl and add new get_permissions

danlu1:
Problem:
1. The original getPermission method uses /acl endpoint only that shows [inconsistent behavior in the case of publicly downloadable content from Synapse.](https://sagebionetworks.jira.com/browse/PLFM-8079)

Solution:
1. create a get_permissions method that implements [GET /entity/{id}/permissions API](https://rest-docs.synapse.org/rest/GET/entity/id/permissions.html). These API factors in the entity's ACL and the user's group membership when determining the **caller's** permission to the entity.
2. rename getPermission with get_acl indicating it extracts ACL information only
3. Deprecate getPermission

Example:
Screenshot 2023-12-15 at 9 31 28 AM
Screenshot 2023-12-26 at 2 56 46 PM

Testing:
Unit tests and integration tests are included in the PR.


synapseclient/client.py
```diff
@@ -2358,22 +2359,24 @@ def _getUserbyPrincipalIdOrName(self, principalId: str = None):
                 "Unknown Synapse user (%s).  %s." % (principalId, supplementalMessage)
             )
 
-    @tracer.start_as_current_span("Synapse::getPermissions")
-    def getPermissions(
+    @tracer.start_as_current_span("Synapse::get_acl")
+    def get_acl(
```
BryanFauble, line 2363: One additional change you'll need to make for documentation. On [the Client reference page](https://python-docs.synapse.org/reference/client/). I set it up so that we are specifying which functions are documented on the page so that we had control over the order without needing to move all functions in this `client.py`. 

Because of this the new changes you added here also need to be updated in this file: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/docs/reference/client.md?plain=1#L22

synapseclient/models/permission.py
```diff
@@ -0,0 +1,175 @@
+from dataclasses import dataclass
```
BryanFauble, line 1: For this `permission.py` let's create a new MD page and nav link so it'll show up in the documents pages.

Note on running the doc pages locally: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/CONTRIBUTING.md#python-doc-pages

Where you'll be adding the MD page to: https://github.com/Sage-Bionetworks/synapsePythonClient/tree/develop/docs/reference

Where you'll add it so that it shows up in the nav panel: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/mkdocs.yml#L53 - Let's add this above `Core` for now

synapseclient/models/permission.py
```diff
@@ -0,0 +1,175 @@
...
+from opentelemetry import trace
+
+tracer = trace.get_tracer("synapseclient")
```
BryanFauble, line 4: No OpenTelemetry tracing is occurring in this class so this and the import can be removed.

synapseclient/client.py
```diff
@@ -2358,22 +2359,24 @@ def _getUserbyPrincipalIdOrName(self, principalId: str = None):
...
         self,
         entity: Union[Entity, Evaluation, str, collections.abc.Mapping],
         principalId: str = None,
```
BryanFauble, line 2366: Since `get_acl` is a new function let's also swap out the arugments for snake_case as well.

synapseclient/models/permission.py
```diff
@@ -0,0 +1,175 @@
...
+        can_upload : (Read Only) Are there any access requirements precluding the user from uploading into this entity (folder or project)?
+        can_enable_inheritance : (Read Only) Can the user delete the entity's access control list (so it inherits settings from an ancestor)?
+        owner_principal_id : (Read Only) The principal ID of the entity's owner (i.e. the entity's 'createdBy')
+        can_public_read : (Read Only) Is this entity considered public?
+        can_moderate : Can the user moderate the forum associated with this entity? Note that only project entity has forum.
+        is_certification_required : (Read Only) Is the certification requirement enabled for the project of the entity?
+        is_entity_open_data : (Read Only) Returns true if the Entity's DateType equals 'OPEN_DATA', indicating that the data is safe to be released to the public.
+    """
+
+    can_view: bool = None
```
BryanFauble, line 33: When you get a chance take a peek at the SonarCloud results. One of the things it had flagged is this line:
https://sonarcloud.io/project/issues?id=Sage-Bionetworks_synapsePythonClient&pullRequest=1037&resolved=false&inNewCodePeriod=true

`Replace the type hint "bool" with "Optional[bool]" or don't assign "None" to "can_view"`

The import for this is:
`from typing import Optional`.

The code would end up like:
`can_view: Optional[bool] = None`

tests/integration/synapseclient/integration_test.py
```diff
@@ -870,3 +865,269 @@ def test_get_permissions_for_project_with_public_and_registered_user(self):
             "DOWNLOAD",
         ]
         assert set(expected_permissions) == set(permissions)
+
+
+class TestPermissionsOnEntityForCaller:
+    @pytest.fixture(autouse=True, scope="function")
+    def init(self, syn: Synapse, schedule_for_cleanup):
```
BryanFauble, line 872: SonarCloud is flagging these `schedule_for_cleanup` parameters for not having a function parameter. We should be able to use:
```
from typing import Callable
```

```suggestion
    def init(self, syn: Synapse, schedule_for_cleanup: Callable[..., None]):
```

In this case.

BryanFauble:
Excellent work! Thank you for incorporating the handful of suggested changes. There are a few more I added around documentation and some code smells that SonarCloud flagged.

synapseclient/models/permission.py
danlu1: Screenshot 2023-12-28 at 10 57 53 AM

docs/reference/core.md
```diff
@@ -17,3 +17,6 @@ functions.  Use at your own risk.
 
 ## Versions
 ::: synapseclient.core.version_check
+
+## Permissions
```
BryanFauble, line 21: Can we add this Permissions documentation to it's own page. I don't think this belongs in the `core.md` documentation.

So

1. Make a `permissions.md`
2. Move these lines over to that markdown
3. Add the permissions md reference in the `mkdocs.yml` nav section

danlu1: @BryanFauble Thanks for catching this. I must have missed the 'above core' part in your early comment. Will tweak this soon. Thanks

docs/reference/permissions.md
```diff
@@ -0,0 +1 @@
+::: synapseclient.models.permission
```
BryanFauble: I committed 1 additional change here - Please review and let me know your thoughts: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1037/commits/315eb98b66a896a6ae1d4dc28b512c0aa7bd2638

The reason I added it is because we have a docstring on each of the attributes, and we have them in the class docstring it ends up duplicating the information in the doc pages. See the screenshots below. Unfortunately because they serve 2 different purposes, [See this slack thread](https://sagebionetworks.slack.com/archives/C02M7QR8D46/p1702070318432449), they cannot be combined at the moment.

[In this settings](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/mkdocs.yml#L107) I have it setup to not show if no docstring, but again, since we are including it the attributes show up twice.

Before:
image

After:
image

BryanFauble (approved):
Thanks for completing this. 

Give a look over the few commits I added (I took care of some formatting issues I saw since it was an easy fix).

docs/reference/permissions.md
danlu1: I agree that we don't want to show duplicate information on the doc. I like your idea and found `show_docstring_attributes` might also be helpful. Let me test it and circle back.

danlu1: @BryanFauble Here is the output of using show_docstring_attributes as False. 
1. If we add show_docstring_attributes as false, nothing except the class description will be shown.
Screenshot 2023-12-29 at 11 12 11 AM
2. If we comment out the member section, we get the attribute and class property (access_types) as paragraph
Screenshot 2023-12-29 at 11 13 43 AM

Screenshot 2023-12-29 at 11 13 59 AM


I will go with the approach you commited since it will keep the attribute table. 


I will accept your commit si

PR #1040, opened 2024-01-04
[SYNPY-1371] Doc fixes

BryanFauble:
Problem:

1. Jenny had a great idea to move the initial setup to the home page instead of tutorials
2. Some links and formatting on a few pages was broken

Solution:
Implementing the changes

Testing:
Verified on docs site


BWMac (approved):
LGTM. Thanks for the updates.

PR #1045, opened 2024-01-16
[SYNPY-1357] Correction to strip spaces before and after values

BryanFauble:
Problem:

A Manifest TSV with the following values were behaving differently when using the `syncToSynapse` function.
1. `[by,"values, by"]`
2. `[by, "values, by"]`

Using the `syncFromSynapse` was adding some excessive quoting to the content written in the manifest TSV. For example the data above was being written as: `"[by,""values, by""]"`

Solution:

1. Updating the logic around splitting of comma values to use regex instead of the CSV library as it did not handle for this.
2. Updated the logic in `syncFromSynapse` to not use quote characters in the TSV as it is not required. The result is that the example above will be stored as `[by,"values, by"]` in the TSV.

Testing:

1. Manually tested with the affected content.
2. Updated some integration tests for the problem case.
3. Added unit tests around the functions for splitting and converting values


thomasyu888 (approved):
🔥 LGTM!

PR #1046, opened 2024-01-16
Move permissions from models folder

BryanFauble:
Problem:

1. With the OOP work in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1013 we are unable to pull in these models directly into the client.py. If you do it is causing a circular dependency because of the `__init.py__` has other models which are pulling in `synapseclient.Synapse`

Solution:

1. Moving the Permissions object into the core

Testing:

1. Integration/Unit testing
2. I also verified that the methods are still working:
```
perms = syn.get_permissions("syn52570249")

print(perms)
print(perms.access_types)

>> Permissions(can_view=True, can_edit=True, can_move=True, can_add_child=True, can_certified_user_edit=True, can_certified_user_add_child=True, is_certified_user=True, can_change_permissions=True, can_change_settings=True, can_delete=True, can_download=True, can_upload=True, can_enable_inheritance=True, owner_principal_id=3481671, can_public_read=False, can_moderate=True, is_certification_required=True, is_entity_open_data=False)
>> ['READ', 'UPDATE', 'CREATE', 'DELETE', 'DOWNLOAD', 'MODERATE', 'CHANGE_PERMISSIONS', 'CHANGE_SETTINGS']
```


thomasyu888:
🔥 LGTM

thomasyu888 (approved):

PR #1047, opened 2024-01-19
[SYNPY-1398] Support `.[version]` syntax for input SynIDs

jaymedina:
### problem

`syncToSynapse` fails for provenance pointing to a specific version of a synID using `syn[id].[version]` notation (e.g. `syn1234.5`). This revisits the discussion on the Python client adding support for synID`.[version]` notation.

image

---
### solution

The scope of this PR is constrained to allowing .version syntax in syncTo/FromSynapse `syn.get` and `synapse get` for the CLI

- [x] `syncToSynapse` supports `.[version]` notation
- [x] `syncFromSynapse` supports `.[version]` notation
- [x] `syn.get` supports `.[version]` notation
- [x] `synapse get` command supports `.[version]` notation
- [x] Corresponding documentation is updated to show new support
- [x] New unit/integration tests for feature coverage and error handling

---
### testing & preview

support for `syn.get(...)`  : 

image

support for `synapse get` for CLI : 

```
(synapsePythonClient) (synapseclient) bash-3.2$ synapse get syn53182713
Downloaded file: foo.txt
Creating /Users/jmedina/Documents/forks/synapsePythonClient/jout/foo.txt
(synapsePythonClient) (synapseclient) bash-3.2$ mv foo.txt foo_v2.txt
(synapsePythonClient) (synapseclient) bash-3.2$ synapse get syn53182713.1
Downloaded file: foo.txt
Creating /Users/jmedina/Documents/forks/synapsePythonClient/jout/foo.txt
(synapsePythonClient) (synapseclient) bash-3.2$ cat foo.txt
VERSION 1
(synapsePythonClient) (synapseclient) bash-3.2$ cat foo_v2.txt 
VERSION 2
(synapsePythonClient) (synapseclient) bash-3.2$ 
```

support for `syncFromSynapse` : 
image

support for `syncToSynapse` : 
image

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -957,7 +957,8 @@ class Synapse(object):
         """
 
         if utils.is_synapse_id_str(entity):
-            entity = self._getEntity(entity)
+            synid, version = utils.get_synid_and_version(entity)
+            entity = self._getEntity(synid, version)
         try:
             self.logger.info(
                 json.dumps(entity, sort_keys=True, indent=2, ensure_ascii=ensure_ascii)
@@ -1086,8 +1087,14 @@ class Synapse(object):
                     " Please use syn.store() to save your entity and try again."
                 )
             else:
-                version = kwargs.get("version", None)
-                bundle = self._getEntityBundle(entity, version)
+                input_version = kwargs.get("version", None)
+                synid_and_version = utils.get_synid_and_version(entity)
+                version = (
+                    input_version if input_version is not None else synid_and_version[1]
+                )
+                # If ``version`` is None, the arg will be ignored
+                bundle = self._getEntityBundle(synid_and_version[0], version)
+
             # Check and warn for unmet access requirements
             self._check_entity_restrictions(
                 bundle, entity, kwargs.get("downloadFile", True)
@@ -2396,9 +2403,12 @@ class Synapse(object):
 
         Returns:
             A dictionary of the Entity's benefactor
+
         """
         if utils.is_synapse_id_str(entity) or is_synapse_entity(entity):
-            return self.restGET("/entity/%s/benefactor" % id_of(entity))
+            synid, _ = utils.get_synid_and_version(entity)
+            return self.restGET("/entity/%s/benefactor" % synid)
+
         return entity
 
     def _getACL(self, entity: Union[Entity, str]) -> Dict[str, Union[str, list]]:
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -1024,10 +1024,11 @@ class Synapse(object):
         Gets a Synapse entity from the repository service.
 
         Arguments:
-            entity:           A Synapse ID, a Synapse Entity object, a plain dictionary in which 'id' maps to a
-                                Synapse ID or a local file that is stored in Synapse (found by the file MD5)
+            entity:           A Synapse ID (e.g. syn123 or syn123.1, with .1 denoting version), a Synapse Entity object,
+                              a plain dictionary in which 'id' maps to a Synapse ID or a local file that is stored in
+                              Synapse (found by the file MD5)
             version:          The specific version to get.
-                                Defaults to the most recent version.
+                                Defaults to the most recent version. If not denoted in the entity input.
             downloadFile:     Whether associated files(s) should be downloaded.
                                 Defaults to True.
             downloadLocation: Directory where to download the Synapse File Entity.
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -353,18 +353,30 @@ def get_synid_and_version(
 ) -> typing.Tuple[str, typing.Union[int, None]]:
     """Extract the Synapse ID and version number from input entity
 
+    Arguments:
+            obj: May be a string, Entity object, or dictionary.
+
     Returns:
         A tuple containing the synapse ID and version number,
         where the version number may be an integer or None if
         the input object does not contain a versonNumber or
         .version notation (if string).
+
+    Example: Get synID and version from string object
+        Extract the synID and version number of the entity string ID
+
+            from synapseclient.core import utils
+            utils.get_synid_and_version("syn123.4")
+
+        The call above will return the following tuple:
+
+            ('syn123', 4)
     """
 
     if isinstance(obj, str):
         synapse_id_and_version = is_synapse_id_str(obj)
-        assert (
-            synapse_id_and_version
-        ), "The input string was not determined to be a syn ID."
+        if not synapse_id_and_version:
+            raise ValueError("The input string was not determined to be a syn ID.")
         m = re.match(r"(syn\d+)(?:\.(\d+))?", synapse_id_and_version)
         id = m.group(1)
         version = int(m.group(2)) if m.group(2) is not None else m.group(2)
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -24,8 +24,6 @@ import uuid
 import warnings
 import zipfile
 
-import synapseclient
-
 
 UNIX_EPOCH = datetime.datetime(1970, 1, 1, 0, 0, tzinfo=datetime.timezone.utc)
 ISO_FORMAT = "%Y-%m-%dT%H:%M:%S.000Z"
@@ -341,7 +339,7 @@ def is_same_base_url(url1: str, url2: str) -> bool:
     return url1.scheme == url2.scheme and url1.hostname == url2.hostname
 
 
-def is_synapse_id_str(obj):
+def is_synapse_id_str(obj: str) -> typing.Union[str, None]:
     """If the input is a Synapse ID return it, otherwise return None"""
     if isinstance(obj, str):
         m = re.match(r"(syn\d+(\.\d+)?$)", obj)
@@ -364,14 +362,19 @@ def get_synid_and_version(
 
     if isinstance(obj, str):
         synapse_id_and_version = is_synapse_id_str(obj)
+        assert (
+            synapse_id_and_version
+        ), "The input string was not determined to be a syn ID."
         m = re.match(r"(syn\d+)(?:\.(\d+))?", synapse_id_and_version)
         id = m.group(1)
         version = int(m.group(2)) if m.group(2) is not None else m.group(2)
-    elif isinstance(obj, synapseclient.Entity):
-        id = id_of(obj)
-        version = None
-        if "versionNumber" in obj:
-            version = obj["versionNumber"]
+
+        return id, version
+
+    id = id_of(obj)
+    version = None
+    if "versionNumber" in obj:
+        version = obj["versionNumber"]
 
     return id, version
 
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -24,6 +24,8 @@ import uuid
 import warnings
 import zipfile
 
+import synapseclient
+
 
 UNIX_EPOCH = datetime.datetime(1970, 1, 1, 0, 0, tzinfo=datetime.timezone.utc)
 ISO_FORMAT = "%Y-%m-%dT%H:%M:%S.000Z"
@@ -342,10 +344,36 @@ def is_same_base_url(url1: str, url2: str) -> bool:
 def is_synapse_id_str(obj):
     """If the input is a Synapse ID return it, otherwise return None"""
     if isinstance(obj, str):
-        m = re.match(r"(syn\d+$)", obj)
+        m = re.match(r"(syn\d+(\.\d+)?$)", obj)
         if m:
             return m.group(1)
     return None
+
+
+def get_synid_and_version(
+    obj: typing.Union[str, collections.abc.Mapping]
+) -> typing.Tuple[str, typing.Union[int, None]]:
+    """Extract the Synapse ID and version number from input entity
+
+    Returns:
+        A tuple containing the synapse ID and version number,
+        where the version number may be an integer or None if
+        the input object does not contain a versonNumber or
+        .version notation (if string).
+    """
+
+    if isinstance(obj, str):
+        synapse_id_and_version = is_synapse_id_str(obj)
+        m = re.match(r"(syn\d+)(?:\.(\d+))?", synapse_id_and_version)
+        id = m.group(1)
+        version = int(m.group(2)) if m.group(2) is not None else m.group(2)
+    elif isinstance(obj, synapseclient.Entity):
+        id = id_of(obj)
+        version = None
+        if "versionNumber" in obj:
+            version = obj["versionNumber"]
+
+    return id, version
 
 
 def bool_or_none(input_value: str) -> typing.Union[bool, None]:
--- a/tests/integration/synapseclient/integration_test.py
+++ b/tests/integration/synapseclient/integration_test.py
@@ -115,6 +115,11 @@ def test_entity_version(syn, project, schedule_for_cleanup):
     assert returnEntity["description"] == "This is a test entity..."
     assert returnEntity["versionLabel"] == "Prada remix"
 
+    # Try retrieving the entity with synid.version notation
+    entity_string = f"{entity['id']}.{entity['versionNumber']}"
+    returnEntity_using_string = syn.get(entity_string)
+    assert returnEntity == returnEntity_using_string
+
     # Try the older Entity again
     returnEntity = syn.get(entity, version=1)
     assert returnEntity.versionNumber == 1
@@ -123,6 +128,9 @@ def test_entity_version(syn, project, schedule_for_cleanup):
 
     # Delete version 2
     syn.delete(entity, version=2)
+    # Setting versionNumber to None, so we can retrieve the latest version
+    # and confirm it is now version 1.
+    entity.versionNumber = None
     returnEntity = syn.get(entity)
     assert returnEntity.versionNumber == 1
 
--- a/tests/integration/synapseclient/integration_test.py
+++ b/tests/integration/synapseclient/integration_test.py
@@ -128,10 +128,10 @@ def test_entity_version(syn, project, schedule_for_cleanup):
 
     # Delete version 2
     syn.delete(entity, version=2)
-    # Setting versionNumber to None, so we can retrieve the latest version
-    # and confirm it is now version 1.
-    entity.versionNumber = None
-    returnEntity = syn.get(entity)
+    # Let's retrieve the entity again using its synId this time,
+    # since the old entity object had versionNumber = 2
+    entity_id = entity.id
+    returnEntity = syn.get(entity_id)
     assert returnEntity.versionNumber == 1
 
 
--- a/tests/integration/synapseutils/test_synapseutils_sync.py
+++ b/tests/integration/synapseutils/test_synapseutils_sync.py
@@ -49,7 +49,7 @@ def test_state(syn: Synapse, schedule_for_cleanup):
                 '%s	%s	"syn12"	" syn123 ;https://www.example.com"	provName2		bar\n'
                 % (self.f2, self.folder.id)
             )
-            self.row3 = '%s	%s	"syn12"		prov2	False	baz\n' % (self.f3, self.folder.id)
+            self.row3 = '%s	%s	"syn12.1"		prov2	False	baz\n' % (self.f3, self.folder.id)
             self.row4 = "%s	%s	%s		act		2\n" % (
                 self.f3,
                 self.project.id,
@@ -294,11 +294,24 @@ def test_syncFromSynapse(test_state):
         uploaded_paths.append(f)
         test_state.schedule_for_cleanup(f)
         test_state.syn.store(File(f, parent=folder_entity))
+
     # Add a file in the project level as well
     f = utils.make_bogus_data_file()
     uploaded_paths.append(f)
     test_state.schedule_for_cleanup(f)
-    test_state.syn.store(File(f, parent=project_entity))
+    entity = test_state.syn.store(File(f, parent=project_entity))
+
+    # Update the Entity and make sure the version is incremented
+    entity = test_state.syn.get(entity)
+    entity = test_state.syn.store(entity, forceVersion=True)
+    print(entity)
+    assert entity.versionNumber == 2
+
+    # Now get version 1 of the entity using .version syntax in the synid
+    synid_with_version_1 = f"{entity.id}.1"
+    entity_v1 = execute_sync_from_synapse(test_state.syn, synid_with_version_1)
+    # Confirm that the entity is version 1 and not 2
+    assert entity_v1[0].versionNumber == 1
 
     # syncFromSynapse() uses chunkedQuery() which will return results that are eventually consistent
     # but not always right after the entity is created.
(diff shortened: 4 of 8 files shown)


synapseclient/core/utils.py
```diff
@@ -342,12 +344,39 @@ def is_same_base_url(url1: str, url2: str) -> bool:
...
+    elif isinstance(obj, synapseclient.Entity):
+        id = id_of(obj)
+        version = None
+        if "versionNumber" in obj:
+            version = obj["versionNumber"]
+
+    return id, version
+
+
+def get_uri(method, entity_id, version=None):
```
BryanFauble, line 370: I don't think what you are trying to do is going to work here. Not every endpoint in the Synapse REST API can be accessed with a version.

Take the benefactor for example.

Works:
```
curl --location 'https://repo-prod.prod.sagebase.org/repo/v1/entity/syn53362132/benefactor' \
--header 'Authorization: Bearer '
```

Does not work because the endpoint does not exist:
```
curl --location 'https://repo-prod.prod.sagebase.org/repo/v1/entity/syn53362132/version/1/benefactor' \
--header 'Authorization: Bearer '
```

synapseclient/client.py
```diff
@@ -917,7 +917,8 @@ def printEntity(self, entity, ensure_ascii=True) -> None:
         """
 
         if utils.is_synapse_id_str(entity):
-            entity = self._getEntity(entity)
+            synid, _ = utils.get_synid_and_version(entity)
+            entity = self._getEntity(synid)
```
BryanFauble, line 921: You would need to pass in the version here as `_getEntity` supports version:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/SYNPY-1398-support-synid-version/synapseclient/client.py#L5487

synapseclient/core/utils.py
```diff
@@ -342,12 +344,39 @@ def is_same_base_url(url1: str, url2: str) -> bool:
...
     """If the input is a Synapse ID return it, otherwise return None"""
     if isinstance(obj, str):
-        m = re.match(r"(syn\d+$)", obj)
+        m = re.match(r"(syn\d+(\.\d+)?$)", obj)
         if m:
             return m.group(1)
     return None
 
 
+def get_synid_and_version(obj):
```
BryanFauble, line 353: nit: Typing on input and return values

synapseclient/core/utils.py
jaymedina: I see, yep I wrote this when I was planning for the larger `.version` support across the whole codebase. I'll retract this for now, but I think later maybe it's worth considering having a `get_uri` function in utils that will generate the URI for us on these calls, and have the endpoint be a kwarg the way it is now, and trigger building the URI with `version` if two conditions are met: the input entity (str or object) has a versionNumber, and the endpoint is one that accepts versionNumber, which can be pulled from a list.

This way any bugs caused by broken URIs can all be traced back to that one function.

Let me know your thoughts

BryanFauble: Sounds good. I might consider naming the function when you do implement it, something like `get_versionable_entity_uri`. `get_uri` is a bit too generic for the specific logic the function will hold.

jaymedina: Makes sense. I'll add this convo to the general `.version` notation ticket [here](https://sagebionetworks.jira.com/browse/SYNPY-699).

jaymedina:
I had a to-do item in the description for deprecating the `version` arg in `synapse.get` but I think it's best to leave that out until the full codebase supports `.version` syntax, what do you think @BryanFauble

thomasyu888:
> I had a to-do item in the description for deprecating the `version` arg in `synapse.get` but I think it's best to leave that out until the full codebase supports `.version` syntax, what do you think @BryanFauble

I think that would be a pretty big breaking change, and would require stakeholder feedback.  Let's hold off for now until I can gather some of that feedback

synapseclient/core/utils.py
```diff
@@ -342,12 +344,38 @@ def is_same_base_url(url1: str, url2: str) -> bool:
...
+        the input object does not contain a versonNumber or
+        .version notation (if string).
+    """
+
+    if isinstance(obj, str):
+        synapse_id_and_version = is_synapse_id_str(obj)
+        m = re.match(r"(syn\d+)(?:\.(\d+))?", synapse_id_and_version)
+        id = m.group(1)
+        version = int(m.group(2)) if m.group(2) is not None else m.group(2)
+    elif isinstance(obj, synapseclient.Entity):
```
thomasyu888, line 370: May need to think about table, view snapshotted versions, or add as a TODO.

synapseclient/core/utils.py
```diff
@@ -342,12 +344,38 @@ def is_same_base_url(url1: str, url2: str) -> bool:
...
     """If the input is a Synapse ID return it, otherwise return None"""
     if isinstance(obj, str):
-        m = re.match(r"(syn\d+$)", obj)
+        m = re.match(r"(syn\d+(\.\d+)?$)", obj)
         if m:
             return m.group(1)
     return None
 
 
+def get_synid_and_version(
```
thomasyu888, line 353: Nit: This would have been a good opportunity to practice test driven development: https://testdriven.io/blog/modern-tdd/ .  It's not the easiest to switch to, but it's more important to ensure that you've thought about the input / output of your function and you contribute unit tests along with your functions.

synapseclient/core/utils.py
```diff
@@ -342,12 +344,38 @@ def is_same_base_url(url1: str, url2: str) -> bool:
 def is_synapse_id_str(obj):
```
thomasyu888, line 344: I have not looked, but what functions are using this?  Would it make sense to update this function instead of adding another?

thomasyu888, line 344: Nit: I haven't looked at what other functions use this and I know you didn't add this function, but just wanted to point out that it is a bit misleading that the function doesn't return a boolean.  Do we know how other functions are using this?

When modifying other functions, can we add type hints?

jaymedina: It's used across `activity.py`, `client.py`, `sync.py` and `conftest.py` but mostly in `client.py`. In the examples I've seen, the `return m.group(1)` return is never used. It's just used for `if`/`if not` statements to continue a process. I'll look through all the references to confirm this, but I'm pretty sure I can return a boolean without breaking code.

Yep, I'll update with type hints.

jaymedina:
> One thought I had was to suggest the API itself to support the dot notation.

I think it's ideal for the Python client behavior to reflect the API and web client as closely as possible, so this makes sense. Do we have anyone from PLATFORM team coming to the demo on Friday? It might also be good to get more feedback from users on how valuable the `.version` support would be, as it's starting to look like a bigger undertaking than previously thought.

jaymedina:
This test seems to be failing because there is already a File entity called `foobarbat` in Project `syn12914772`. I was able to re-create the error in my local machine after running `syn.store` on my File entity `foobarhat` consecutively:

image


After deleting my `foobarhat` file from my Project that I'm testing on, the `syn.store` call works fine again:


image
image

It looks like the Projects don't get deleted after integration tests run. I'm wondering why that is? I notice when I run integration tests locally the Projects remain in my Synapse Projects page. Should we start an initiative to update tests so that we have complete teardowns after integration tests run/fail, or is there a reason for Entities to stay up after the fact? @BryanFauble @thomasyu888

tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -294,11 +294,24 @@ def test_syncFromSynapse(test_state):
         uploaded_paths.append(f)
         test_state.schedule_for_cleanup(f)
         test_state.syn.store(File(f, parent=folder_entity))
+
     # Add a file in the project level as well
     f = utils.make_bogus_data_file()
     uploaded_paths.append(f)
     test_state.schedule_for_cleanup(f)
-    test_state.syn.store(File(f, parent=project_entity))
```
thomasyu888, line 301: @jaymedina if you look in this class, you'll see `schedule_for_cleanup` functions, but it doesn't setup / teardown per test.

jaymedina: I see - It might be good to tear down per test because I turned off pytest `reruns` for this particular integration test, which uncovered the real error that was breaking this test. Or maybe consider turning off `reruns` entirely for integration tests where the Synapse data repositories get updated.

```
>           raise SynapseHTTPError(message, response=response)
E           synapseclient.core.exceptions.SynapseHTTPError: 404 Client Error: 
E           Cannot find a node with id syn53404745 and version 2
```

BryanFauble: > It might be good to tear down per test

If you have a suggestion to add feel free - Perhaps we could spy on the client and keep track of all of the things that need to be cleaned up and add a finalizer automatically.

Ideally everything created in an integration test is deleted.


Some background on the `reruns` - I added this in a few months ago as a band-aid fix. It got us into a "good enough" position to not spend the time to look further at the individual failures and patch each thing that randomly fails for us.



Also - As a heads up if you are running these integration tests make sure you are pointing to the dev endpoints (If you aren't already):
```
[endpoints]
repoEndpoint=https://repo-dev.dev.sagebase.org/repo/v1
authEndpoint=https://repo-dev.dev.sagebase.org/auth/v1
fileHandleEndpoint=https://repo-dev.dev.sagebase.org/file/v1
```

If you need help setting this up reach out.

tests/integration/synapseclient/integration_test.py
```diff
@@ -123,6 +123,9 @@ def test_entity_version(syn, project, schedule_for_cleanup):
 
     # Delete version 2
     syn.delete(entity, version=2)
+    # Setting versionNumber to None, so we can retrieve the latest version
+    # and confirm it is now version 1.
+    entity.versionNumber = None
```
jaymedina, line 128: After deleting version 2 of `entity`, this unit test was feeding back the original File Entity that had `versionNumber = 2` in its dictionary of properties.

This used to be fine, and the test used to return the latest version (in this case version 1 now) regardless, because `syn.get` was not getting version number from the `versionNumber` in the File Entity, but rather from the `version` kwarg if it was fed into `syn.get`: see [here for reference ](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/0ea22a414b25ca830e6e2bc5cb181727ccf4cf51/synapseclient/client.py#L1090)

But in this feature branch I updated `syn.get` to pull the version number from either: the `version` kwarg, the `.version` syntax in a synID string, or the `.versionNumber` in an entity, which caused this test to fail. The fix was to update `versionNumber` to None so that the latest version is pulled.

My current stance is that it makes sense for `syn.get` to pull the `.versionNumber` from the Entity object itself, so as to enforce correct Entity objects are fed into `syn.get`, and users are getting back exactly what they requested. Let me know your thoughts.

tests/integration/synapseutils/test_synapseutils_sync.py
jaymedina: Hi @BryanFauble , the `reruns` background makes sense, thanks for the context. In case that, it makes sense to leave it in for now. I think we would benefit more from cleanups after each test execution (pass or fail). That way each rerun starts on a blank canvas.

tests/integration/synapseclient/integration_test.py
BryanFauble: I'm thinking that the `delete` function should automatically do this for you. For example if any of our code is modifying what data is available we should keep the data we return from those functions to be consistent.

jaymedina: I see. Another way I can write this is call `syn.get()` on the synID string rather than the Entity object to avoid modifying data between steps.

synapseclient/core/utils.py
jaymedina: 3 unit tests fail when `is_synapse_id_str` returns `True` instead of the string. I'm going to leave this out for now since it's out of scope, and move forward updating the function with type hints, but me know if you think this should be in this PR

jaymedina:
> Another thing is this:
> 
> ```
> entity = syn.get("syn1234.3")
> # entity.id = "syn1234"
> ```
> 
> The `get` API endpoint returns this object: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/FileEntity.html.

I'm thinking of handling this in two ways:

1. Remove .version support for `syn.get` and the CLI `synapse get` for this PR. I'm pretty sure I can patch up `syncToSynapse` without needing to change `syn.get`. And I'll table the `syn.get` changes 
2. Update the documentation letting people know that their input `id` may not be reflected in the output Entity object, and that the `version`. But this isn't ideal since I think it defeats the purpose of supporting .version notation in the input.

Aside from this, I think this PR is in a good state for a final review.

Let me know if there are other ideas @thomasyu888 @BryanFauble

tests/integration/synapseutils/test_synapseutils_sync.py
BryanFauble: Agreeded

BryanFauble:
> > The `get` API endpoint returns this object: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/FileEntity.html.
> 
> I'm thinking of handling this in two ways:
> 
>     1. Start a new PR to fix `syncToSynapse` that excludes .version support for `syncFromSynapse`, `syn.get` and the CLI `synapse get`. This PR would be for the larger initiative to support .version across the whole codebase. I'm pretty sure I can patch up `syncToSynapse` without needing to change the others. The only thing I don't like about this is that I think `syncTo/FromSynapse` should always have very similar functionalities, since a user who uses one will probably use the other. And for .version to be supported in one and not the other is confusing.
>     2. Update the documentation letting people know that their input `id` may not be reflected in the output Entity object, and that the `version` is its own property. But this isn't ideal since I think it defeats the purpose of supporting .version notation in the input.



I agree that it'll be confusing, but IMO we should let folks use `.version` notation for inputs (As in support it), but don't return that format. I'm thinking that we should always return back 1 expected object format from our APIs. By this I mean we have an ID field for `syn123` and we have a `versionNumber` field denoting what version the entity is. I want those using the API to have confidence that all the information we're returning will always be consistent.


I also put together the following code to show explicit examples of what you are talking about with `syn.get` and what you are currently supporting. I am in agreement with how you implemented the changes.

```
import synapseclient
from synapseclient import Entity


syn = synapseclient.Synapse()
syn.login()

MY_ENTITY = "syn53369103"

# Works
by_id = syn.get(MY_ENTITY, downloadFile=False)
print(f"by_id: \n{by_id}")

# Works
by_entity = syn.get(Entity(id=MY_ENTITY), downloadFile=False)
print(f"by_entity: \n{by_entity}")

# Works
by_id_and_version = syn.get(f"{MY_ENTITY}.10", downloadFile=False)
print(f"by_id_and_version: \n{by_id_and_version}")

# Works - uses version 11 instead of 10
by_id_and_version_input_argument = syn.get(
    f"{MY_ENTITY}.10", version=11, downloadFile=False
)
print(f"by_id_and_version_input_argument: \n{by_id_and_version_input_argument}")

# Works
by_entity_and_version_field = syn.get(
    Entity(id=f"{MY_ENTITY}", versionNumber=10), downloadFile=False
)
print(f"by_entity_and_version_field: \n{by_entity_and_version_field}")

# Does not work
by_entity_and_version_in_id = syn.get(Entity(id=f"{MY_ENTITY}.10"), downloadFile=False)
print(f"by_entity_and_version_in_id: \n{by_entity_and_version_in_id}")
```

For a final version, should we support creating an Entity object like: `Entity(id=f"{MY_ENTITY}.10")`? Specifically I am thinking that we automatically split `.10` from the ID and put it into the `versionNumber` field. Handle conflicts in the same way you are in `.get()`.

synapseclient/core/utils.py
```diff
@@ -339,15 +339,46 @@ def is_same_base_url(url1: str, url2: str) -> bool:
...
         if m:
             return m.group(1)
     return None
 
 
+def get_synid_and_version(
+    obj: typing.Union[str, collections.abc.Mapping]
+) -> typing.Tuple[str, typing.Union[int, None]]:
+    """Extract the Synapse ID and version number from input entity
+
```
BryanFauble, line 355: nit:

Include
```

        Arguments:
            obj:
```
in the docstring

BryanFauble: Also include a few example inputs and what the output would be

synapseclient/core/utils.py
```diff
@@ -339,15 +339,46 @@ def is_same_base_url(url1: str, url2: str) -> bool:
...
+    Returns:
+        A tuple containing the synapse ID and version number,
+        where the version number may be an integer or None if
+        the input object does not contain a versonNumber or
+        .version notation (if string).
+    """
+
+    if isinstance(obj, str):
+        synapse_id_and_version = is_synapse_id_str(obj)
+        assert (
```
BryanFauble, line 365: In a few other functions in this utils class it is doing a `raise ValueError(msg)` rather than an assert. What's your opinion on one or the other?

BryanFauble: For example the closest in functionality is `id_of` which does a:
`raise ValueError("Invalid parameters: couldn't find id of " + str(obj))`

tests/unit/synapseclient/core/unit_test_utils.py
```diff
@@ -100,6 +100,37 @@ def __init__(self, id_attr_name, id):
         assert utils.id_of(foo) == "123"
 
 
+# TODO: Add a test for is_synapse_id_str(...)
```
BryanFauble, line 103: TODO

jaymedina:
> For a final version, should we support creating an Entity object like: `Entity(id=f"{MY_ENTITY}.10")`? Specifically I am thinking that we automatically split `.10` from the ID and put it into the `versionNumber` field. Handle conflicts in the same way you are in `.get()`.

I think that makes sense, and yes I can recycle the logic that splits up the synid from the version number. Are you suggesting folding this into the PR as well?

synapseclient/core/utils.py
jaymedina: Thanks for pointing this out - I think keeping with the `ValueError(msg) is better for consistency, and the error type is more specific. I'll make a commit to replace this `assert`.

jaymedina:
We've agreed on how I implemented the changes for `syn.get`, `synapse get` (CLI), `syncTo/FromSynapse` + tests pass, and I've completed the checklist, so I'll mark this as RFR and continue addressing comments as they come. Thank you Bryan and Tom for your comments so far.

tests/unit/synapseclient/core/unit_test_utils.py
jaymedina: Hi Bryan could you elaborate on this comment?

BryanFauble: You added a # TODO comment. Is this something you will plan to accomplish this this PR?

BryanFauble:
> > For a final version, should we support creating an Entity object like: `Entity(id=f"{MY_ENTITY}.10")`? Specifically I am thinking that we automatically split `.10` from the ID and put it into the `versionNumber` field. Handle conflicts in the same way you are in `.get()`.
> 
> I think that makes sense, and yes I can recycle the logic that splits up the synid from the version number. Are you suggesting folding this into the PR as well?

@jaymedina It's an option, but I don't think it's required for us to support at this exact moment.

tests/unit/synapseclient/core/unit_test_utils.py
jaymedina: No, I was going to make a ticket and add it to the Tech Debt Epic on Jira. I placed the TODO here so it's easier to see where the unit test should go.

tests/integration/synapseutils/test_synapseutils_sync.py
jaymedina: I've made a Jira ticket for this: https://sagebionetworks.jira.com/browse/SYNPY-1424

tests/unit/synapseclient/core/unit_test_utils.py
BryanFauble: Makes sense. When you create the Jira can you add it to the TODO here

jaymedina:
> > > For a final version, should we support creating an Entity object like: `Entity(id=f"{MY_ENTITY}.10")`? Specifically I am thinking that we automatically split `.10` from the ID and put it into the `versionNumber` field. Handle conflicts in the same way you are in `.get()`.
> > 
> > 
> > I think that makes sense, and yes I can recycle the logic that splits up the synid from the version number. Are you suggesting folding this into the PR as well?
> 
> @jaymedina It's an option, but I don't think it's required for us to support at this exact moment.

I've added this item to the Jira ticket for broader .version support across synpy: https://sagebionetworks.jira.com/browse/SYNPY-699

BryanFauble (approved):
LGTM!

jaymedina:
I need to leave this open for now because the `syncFromSynapse` call for synid.version input returns an empty etag for some reason. Need to investigate why. 

image

BryanFauble:
> I need to leave this open for now because the `syncFromSynapse` and `syn.get` calls for synid.version input returns an empty etag for some reason. Need to investigate why.
> image

@jaymedina I believe `00000s` is the intended eTag when you are grabbing the non-latest version of an Entity.

jaymedina:
That's right! Whew. Thanks!

image

synapseclient/client.py
```diff
@@ -2396,9 +2404,12 @@ def _getBenefactor(
 
         Returns:
             A dictionary of the Entity's benefactor
+
         """
         if utils.is_synapse_id_str(entity) or is_synapse_entity(entity):
-            return self.restGET("/entity/%s/benefactor" % id_of(entity))
+            synid, _ = utils.get_synid_and_version(entity)
+            return self.restGET("/entity/%s/benefactor" % synid)
```
thomasyu888, line 2411: Nit: in the future, when we see `%` lets switch to f-string interpolation.

synapseclient/core/utils.py
```diff
@@ -339,15 +339,58 @@ def is_same_base_url(url1: str, url2: str) -> bool:
...
+        version = int(m.group(2)) if m.group(2) is not None else m.group(2)
+
+        return id, version
+
+    id = id_of(obj)
+    version = None
+    if "versionNumber" in obj:
+        version = obj["versionNumber"]
+
+    return id, version
```
thomasyu888, line 391: Nit: Just a thought for the future, but tuples can be a bit risky because the order really matters.  Look into `namedtuples` for something more explicit.

tests/unit/synapseclient/core/unit_test_utils.py
```diff
@@ -100,6 +100,37 @@ def __init__(self, id_attr_name, id):
...
+# https://sagebionetworks.jira.com/browse/SYNPY-1425
+
+
+def test_get_synid_and_version():
```
thomasyu888, line 107: Nit: Not to over-engineer the solution, but in the future, these are technically different test cases.  For example, right now it would be harder to determine if Test 2 or Test 3 failed, but if you split them into separate tests OR use `pytest.parametrize` (it it makes sense to) then it'd be easier to navigate.

thomasyu888 (commented):
@jaymedina Just wanted to say this is great work and great reviews by @BryanFauble .  I left some nits just for future reference, but no action required.  Thanks everyone!

PR #1048, opened 2024-01-22
[SYNPY-1412] Fixing typing issue in copy_functions

jaymedina:
### problem

Importing `Entity` from `entity.py` was not working on Jupyter Notebook. This cell was executed after launching notebook off the latest `develop` branch on SynPy:

image

### solution

The solution was to have the type hints refer to the `Entity` class which is already imported at the top of `copy_functions.py` instead of calling it again from `synapseclient.entity.py`.

### testing & preview

With this change there are no import errors for `synapseutils`:

image


BryanFauble (approved):
LGTM!

PR #1049, opened 2024-01-23
Creating a decorator to be used in async class methods

BryanFauble:
Problem:

1. The default OTEL decorator does not work as expected in Async environments. It will only capture when the method is created but not capture the entire execution of the method.
2. We were required to use a context manager in the function leading to more complex code.

Solution:

1. I created a new decorator that will allow us to decorate these class methods. In the future I'll plan to create other decorators for async functions too.

Testing:

1. I verified that the name of the functions were correctly shown:
image
image
image

### DIFF
--- a/docs/reference/core.md
+++ b/docs/reference/core.md
@@ -42,6 +42,9 @@ functions.  Use at your own risk.
 
 ::: synapseclient.core.utils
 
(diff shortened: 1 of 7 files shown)


BWMac (approved):
LGTM

PR #1050, opened 2024-01-24
Remove return type hinting on decorator

BryanFauble:
Problem:
Type hinting on functions wrapped in this decorator were not showing correctly.

Solution:
Remove type hinting

Testing:
image


BWMac (approved):
Nice, I can see the type hints when I hover on my end.

PR #1051, opened 2024-01-24
Moving to pytest-socket for blocking socket connections in unit tests

BryanFauble:
Problem:
The following code was blanket stopping all sockets from being created, however, asyncIO uses sockets to function without sending data external to the machine it's executing.
```
@pytest.fixture(autouse=True, scope="session")
def block_remote_conections(request):
    """
    Block all remote calls during unit tests.
    """
    # inspired by https://stackoverflow.com/q/18601828

    def block_socket_side_effect(*args, **kwargs):
        raise ValueError(_BLOCKED_CONNECTION_MESSAGE)

    block_socket_patcher = mock.patch(
        "socket.socket", side_effect=block_socket_side_effect
    )
    block_socket_patcher.start()

    request.addfinalizer(block_socket_patcher.stop)
```
image

Solution:
Updating to use the pytest-socket library with `allow-unix-socket` turned on.

Testing:
- I was able to run a unit test locally that was ASYNC.
- Relying on current unit tests to pass as expected.
- I also updated the unit test that verifies connections cannot be started to assert that the new exception is raised.


BWMac (approved):
Awesome, thank you so much for your help!

PR #1052, opened 2024-01-24
[SYNPY-1344] Set client in conftest since login is not used in unit tests

BryanFauble:
**Problem:**

1. The `synapse_client` parameter in the Synapse class is only initialized when using `login` - Which is ok for integration tests and normal operations, however, because we are not using `login` for unit tests I need to manually set the parameter

**Solution:**
Set the client in the `conftest.py` for unit tests

**Testing:**
Unit tests Brad/Bryan are developing


BWMac (approved):
Looks good. I take it this removes the need to have the same line in each test?

BryanFauble:
> Looks good. I take it this removes the need to have the same line in each test?

yes

PR #1053, opened 2024-01-25
[SYNPY-1347] Adds OOP Team Model

BWMac:
**Description:**

This PR adds an OOP model for interacting with `Team`s in Synapse. It provides the ability to:
- Create a team
- Delete a team
- Retrieve information about a team by using its ID number or team name
- Retrieve information about a team's members
- Invite new members to a team
- View any unresolved invitations a team has sent to Synapse users

Notes:
- To support specifically creating and deleting teams, corresponding functions were added to `client.py`
- All new code has been tested with unit/integration tests where applicable
- `TeamMember` and `UserGroupHeader` models were also created to be used in the composition of `Team` objects
- A demo script `oop_poc_team.py` is included to showcase how a developer can interact with teams using the new model


docs/scripts/object_orientated_programming_poc/oop_poc_team.py
```diff
@@ -0,0 +1,57 @@
...
+async def new_team():
+    # Create a team
+    new_team = Team(
+        name="python-client-test-team",
+        description="testing OOP interface",
+        can_public_join=False,
+    )
+    my_synapse_team = await new_team.create()
+    print(my_synapse_team)
+    # sleep for a bit to allow the team to be created (Synapse API slow?)
```
BWMac, line 30: I was unable to get this to run without a sleep here. It seemed like the next lines were running before the team was actually created.

docs/reference/oop/file.md
```diff
@@ -0,0 +1,6 @@
+::: synapseclient.models.File
```
BWMac, line 1: Not sure whats up with these markdown files/why they are in this PR. I did not create them. Were they deleted from `develop` recently?

BWMac, line 1: Not sure why all of these md files are in my PR, I did not create them all. Were they deleted or moved in `develop` or something?

BryanFauble: I had them in the PoC originally - But removed them in favor of putting these into a single file:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/5f2e2b742d72c9aaf04ce3446d83744426d355dc/docs/reference/oop/models.md

BWMac: OK I will remove the individual ones here.

BryanFauble:
Please add some integration tests around the new methods.


Shoot me a re-review when you are ready for another round. Thanks for your help on these changes!

docs/scripts/object_orientated_programming_poc/oop_poc_team.py
thomasyu888: Weird... is 5 seconds arbitrary?  Would the same test would be to create a team and get the team right after without async?

synapseclient/client.py
```diff
@@ -3853,47 +3853,104 @@ def _find_teams_for_principal(self, principal_id: str) -> typing.Iterator[Team]:
...
+        description: str = None,
+        icon: str = None,
+        opentelemetry_context=None,
+    ) -> Team:
+        """
+        Creates a new team.
+
+        Arguments:
+            name: The name of the team to create.
+            description: A description of the team.
```
thomasyu888, line 3868: What was your rationale around these parameters?  A team object is: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/Team.html.

synapseclient/client.py
```diff
@@ -3853,47 +3853,104 @@ def _find_teams_for_principal(self, principal_id: str) -> typing.Iterator[Team]:
         for result in self._GET_paginated(f"/user/{principal_id}/team"):
             yield Team(**result)
 
-    @tracer.start_as_current_span("Synapse::getTeam")
-    def getTeam(self, id):
+    def create_team(
```
thomasyu888, line 3856: What was your rationale in creating this function?  Right now you can create a team by doing:

```
syn.store(Team(...))
```

synapseclient/client.py
```diff
@@ -3853,47 +3853,104 @@ def _find_teams_for_principal(self, principal_id: str) -> typing.Iterator[Team]:
...
+        ):
+            request_body = {"name": name, "description": description, "icon": icon}
+            return Team(
+                **self.restPOST(
+                    "/team",
+                    json.dumps(request_body),
+                )
+            )
+
+    def delete_team(self, id: int, opentelemetry_context=None) -> None:
```
thomasyu888, line 3887: Same thing as above, you can delete a team by doing 

```
syn.delete(Team(...))
```

synapseclient/models/team.py
```diff
@@ -1 +1,304 @@
...
+    team_id: Optional[int] = None
+    """The ID of the team"""
+
+    member: Optional[UserGroupHeader] = None
+    """An object of type [org.sagebionetworks.repo.model.UserGroupHeader](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/UserGroupHeader.html)"""
+
+    is_admin: Optional[bool] = None
+    """Whether the given member is an administrator of the team"""
+
+    def fill_from_dict(self, synapse_team_member: Synapse_TeamMember) -> "TeamMember":
```
thomasyu888, line 40: I noted in Bryan's PR too about the type hint for `synapse_team_member`.  Should it contain `dict` as well?

synapseclient/models/team.py
```diff
@@ -1 +1,304 @@
...
+            synapse_client: If not passed in or None this will use the last client from the `.login()` method.
+
+        Returns:
+            Team: The Team object.
+        """
+        loop = asyncio.get_event_loop()
+        current_context = context.get_current()
+        team = await loop.run_in_executor(
+            None,
+            lambda: Synapse.get_client(synapse_client=synapse_client).create_team(
```
thomasyu888, line 138: Ah, I understand why `create_team` now.  I'd say we could fill out the `create_team` function with more parameters, as it might get confusing to have `SynapseTeam` in here with all the arguments and a `syn.store` that calls it

synapseclient/models/team.py
```diff
@@ -1 +1,304 @@
...
+        team_id: The ID of the team
+        member: An object of type [org.sagebionetworks.repo.model.UserGroupHeader](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/UserGroupHeader.html)
+                describing the member
+        is_admin: Whether the given member is an administrator of the team
+    """
+
+    team_id: Optional[int] = None
+    """The ID of the team"""
+
+    member: Optional[UserGroupHeader] = None
```
thomasyu888, line 34: I've always been a bit confused as to why member doesn't just return UserProfiles: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/UserProfile.html, since that's really what they are.

thomasyu888 (commented):
🔥 Awesome - I left some comments that may require some changes, but it's great to see more OOP in the client.

docs/scripts/object_orientated_programming_poc/oop_poc_team.py
BWMac: 5 seconds was semi-arbitrary, 1 or 2 seconds did not work but 5 does.

BWMac:
Going to revert this back to draft mode while I address some issues.

synapseclient/client.py
BWMac: The reason was this on the Synapse API [docs](https://rest-docs.synapse.org/rest/POST/team.html).

> The passed request body may contain the following fields:
> 
> name - Give your new Team a name. The name must be unique, not used by an existing Team (required).
> description - a short text description of the Team's purpose (optional).
> icon - a fileHandle ID for an icon image file previously uploaded to Synapse (optional).

synapseclient/client.py
BWMac: As it stands, `syn.delete` does not work by simply passing it the new model, and you cannot delete a team by just passing its `id` to `syn.delete` as you can with entities that have a `syn123` identifier. So then to leverage the existing function I would have to convert the new model back to the old one which seems counterintuitive if our goal is to eventually transition to only using the new ones.

synapseclient/client.py
BWMac: The rest of the fields are filled in by the API for the old model, and then filled in by `fill_from_dict` in the new model

synapseclient/models/team.py
BWMac: Should we wait to make this change until the `UserProfile` model is added?

BWMac:
@BryanFauble I added integration tests as requested. I still need to look more into [this](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1053#discussion_r1467230623) (any suggestions would be appreciated), but I think this is ready for review.

BryanFauble:
> @BryanFauble I added integration tests as requested. I still need to look more into [this](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1053#discussion_r1467230623) (any suggestions would be appreciated), but I think this is ready for review.

@BWMac 

I'm not sure where my comment ran off to - because I responded to Tom's comment but I cannot find it.

Likely this is not a client side issue, but a server side issue. I'd expect that if we called this from both a sync and async environment we are going to get the same behavior.

My comment earlier to this was:

> If I understand this is only applicable when using your from_name method. Using from_id worked perfectly for me.

> Could you write up a PLFM jira ticket with what you found around this REST API: https://rest-docs.synapse.org/rest/GET/teams.html - Is [eventually consistent](https://en.wikipedia.org/wiki/Eventual_consistency) and if that's expected?

synapseclient/models/team.py
thomasyu888: No because the API for TeamMembers does actually return UserGroupHeaders

tests/integration/models/test_team.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @pytest.mark.asyncio
+    async def test_delete(self) -> None:
+        # GIVEN a team object self.team
+        # WHEN I create the team on Synapse
+        test_team = await self.team.create()
+        # AND I delete the team
+        await test_team.delete()
+        # THEN I expect the team to no longer exist
+        with pytest.raises(SynapseHTTPError):
```
BryanFauble, line 55: nit: We could verify the return message. For example I had done this in Activity:
```
        with pytest.raises(
            client.SynapseHTTPError, match="404 Client Error: \nNo activity"
        ):
```

synapseclient/models/team.py
```diff
@@ -1 +1,304 @@
...
+
+    created_by: Optional[str] = None
+    """The ID of the user that created this team"""
+
+    modified_by: Optional[str] = None
+    """The ID of the user that last modified this team"""
+
+    def fill_from_dict(self, synapse_team: Synapse_Team) -> "Team":
+        self.id = (
+            int(synapse_team.get("id", None)) if synapse_team.get("id", None) else None
```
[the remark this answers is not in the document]
BWMac: It doesn't because it evaluates the `if` part first. It won't try to do the `int` part unless `synapse_team.get("id", None)` is not `None`:
```
team = Team()
team.fill_from_dict(synapse_team={})

Team(id=None, name=None, description=None, icon=None, can_public_join=False, can_request_membership=True, etag=None, created_on=None, modified_on=None, created_by=None, modified_by=None)
```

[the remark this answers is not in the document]
BryanFauble: Yeah, you're right. My mistake!

synapseclient/models/team.py
```diff
@@ -1 +1,304 @@
...
+                id=id, opentelemetry_context=current_context
+            ),
+        )
+        self.fill_from_dict(team)
+        return self
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Team_From_Name: {self.name}"
+    )
+    async def from_name(
```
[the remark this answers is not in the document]
BWMac: I converted them both to classmethods in my latest commit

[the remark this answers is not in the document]
BryanFauble: I was thinking a bit about this - What are your thoughts around providing a classmethod AND a get method? My thought is around providing the most convenience for someone in these cases:

1. I have an instance of the class and I want to load it's current state from Synapse (ie: Call `get`)
2. I don't have an instance of the class but I want to create a loaded instance from Synapse (ie: Call `from_name` or `from_id`)

If we go down this idea path we could update the other models to follow this same idea - Where either:

1. `get` is a wrapper for these classmethods
2. OR the classmethods are wrappers for the `get`

Basically the 2 ideas being:
1.
```
@classmethod
  async def from_id(
      cls, id: int, synapse_client: Optional[Synapse] = None
  ) -> "Team":
    # Implementation

@classmethod
async def from_name(
    cls, name: str, synapse_client: Optional[Synapse] = None
) -> "Team":
    # Implementation

async def get(self, synapse_client: Optional[Synapse] = None) -> "Team":
    # Retrieve by ID `if self.id:` by calling `from_id`
    # Retrieve by Name `if self.name: by calling `from_name`
```
2.
```
@classmethod
  async def from_id(
      cls, id: int, synapse_client: Optional[Synapse] = None
  ) -> "Team":
    # Call GET

@classmethod
async def from_name(
    cls, name: str, synapse_client: Optional[Synapse] = None
) -> "Team":
    # Call GET

async def get(self, synapse_client: Optional[Synapse] = None) -> "Team":
    # Retrieve by ID `if self.id:` 
        # IMPLEMENTATION
    # Retrieve by Name `if self.name: 
        # IMPLEMENTATION
```

Whichever we want to follow we could also apply it to the other Models where we have similar rules to follow:
```
Project:
- Name
or
- ID

Folder:
- ID
or
- Name and Parent

File:
- ID
or
- Name and Parent
or
- MD5 (List of matching Files)

Table:
- ID
or
- Name and Project

Team:
- ID
or
- Name
```

cc @BWMac @thomasyu888 @jaymedina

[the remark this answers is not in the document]
BWMac: I was thinking about this the way that I thought about this [ticket](https://sagebionetworks.jira.com/browse/SYNPY-1186?jql=project%20in%20(SYNPY)%20AND%20assignee%3D%22Brad%20Macdonald%22%0A), where we determined that the best solution was to separate the concerns of using a username and using an ID number. I'm not strongly opposed to including a `get` method of some sort, but I think that both cases

> I have an instance of the class and I want to load it's current state from Synapse (ie: Call get)
and
> I don't have an instance of the class but I want to create a loaded instance from Synapse (ie: Call from_name or > 
> from_id)

are already covered by the `from_name` and `from_id` class methods:

```
# Load current state of my team
my_team = await Team.from_id(id=3492636)
print(my_team)

Team(id=3492636, name='python-client-test-team-1', description='testing OOP interface', icon=None, can_public_join=False, can_request_membership=True, etag='6639cfde-c5bb-4984-b51d-cd4cd2982469', created_on='2024-01-26T18:08:55.071Z', modified_on='2024-01-29T16:13:01.741Z', created_by='3458117', modified_by='3458117')

# Then, I go into the UI and change the ability for the public to join
# Reload current state of my team

await my_team.from_id(id=3492636)
print(my_team)

Team(id=3492636, name='python-client-test-team-1', description='testing OOP interface', icon=None, can_public_join=True, can_request_membership=True, etag='6639cfde-c5bb-4984-b51d-cd4cd2982469', created_on='2024-01-26T18:08:55.071Z', modified_on='2024-01-29T16:13:01.741Z', created_by='3458117', modified_by='3458117')
```

So I think the convenience is already there, and we currently don't need if/else logic to route the user's query to the appropriate class method.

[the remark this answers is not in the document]
BryanFauble: What you describing in this case case wouldn't work if I understand correctly:

```
# Load current state of my team
my_team = await Team.from_id(id=3492636)
print(my_team)

Team(id=3492636, name='python-client-test-team-1', description='testing OOP interface', icon=None, can_public_join=False, can_request_membership=True, etag='6639cfde-c5bb-4984-b51d-cd4cd2982469', created_on='2024-01-26T18:08:55.071Z', modified_on='2024-01-29T16:13:01.741Z', created_by='3458117', modified_by='3458117')

# Then, I go into the UI and change the ability for the public to join
# Reload current state of my team

await my_team.from_id(id=3492636)
print(my_team)

```

What would not work:
```
# This is the first instance - We create `my_team` from our classmethod
my_team = await Team.from_id(id=3492636)
print(my_team)

# A new instance of `my_team` is created, but thrown away because we are not assigning it to anything. The `my_team` class instance is NOT updated because we are not updating `self` in `from_id`
await my_team.from_id(id=3492636)
print(my_team)
```

Please let me know if my assumption here is incorrect (Because it'll save us a lot of headache if I am)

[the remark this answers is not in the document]
BWMac: Ah right, you are correct. It would have worked that way before I changed them to classmethods though.

BryanFauble:
@BWMac Do we have an 'update' method available? Suppose I'd like to update the name of the team, can I do that through this interface?

synapseclient/models/team.py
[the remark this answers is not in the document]
BryanFauble: > was thinking about this the way that I thought about this [ticket](https://sagebionetworks.jira.com/browse/SYNPY-1186?jql=project%20in%20(SYNPY)%20AND%20assignee%3D%22Brad%20Macdonald%22%0A), where we determined that the best solution was to separate the concerns of using a username and using an ID number.

Something to mention about this ticket - In this ticket both the username and an ID number were being passed into the method in the same parameter. Meaning we didn't know which one they actually meant (ID or a username that happens to be a numerical number).

However - In this case they are separate fields on the Team (`id` and `name`). If we specify the order which we would use them in `get` I think that covers our use-case. We could say "We will retrieve the Team from Synapse by `id` or `name` in that order depending on which is filled in."

One other thing I was thinking about is whichever route we want to take (If we make these changes) that we should make sure that the object we call `get` with is also updated. That would mean the `get` could look like:
 ```
    async def get(self, synapse_client: Optional[Synapse] = None) -> "Team":
        if self.id:
            new_instance = await Team.from_id(id=self.id, synapse_client=synapse_client)
            self.__dict__.update(new_instance.__dict__)
            return self
        elif self.name:
            new_instance = await Team.from_name(
                name=self.name, synapse_client=synapse_client
            )
            self.__dict__.update(new_instance.__dict__)
            return new_instance
        raise ValueError("Team must have either an id or a name")
```

OR if we go the other direction it would look something like:
```
    async def get(self, synapse_client: Optional[Synapse] = None) -> "Team":
        if self.id:
            loop = asyncio.get_event_loop()
            current_context = context.get_current()
            api_team = await loop.run_in_executor(
                None,
                lambda: Synapse.get_client(synapse_client=synapse_client).getTeam(
                    id=self.id, opentelemetry_context=current_context
                ),
            )
            return self.fill_from_dict(api_team)
        elif self.name:
            loop = asyncio.get_event_loop()
            current_context = context.get_current()
            api_team = await loop.run_in_executor(
                None,
                lambda: Synapse.get_client(synapse_client=synapse_client).getTeam(
                    id=self.name, opentelemetry_context=current_context
                ),
            )
            return self.fill_from_dict(api_team)
        raise ValueError("Team must have either an id or a name")

    @classmethod
    @otel_trace_method(
        method_to_trace_name=lambda cls, id, **kwargs: f"Team_From_Id: {id}"
    )
    async def from_id(cls, id: int, synapse_client: Optional[Synapse] = None) -> "Team":
        """Gets Team object using its integer id.

        Arguments:
            id: The id of the team.
            synapse_client: If not passed in or None this will use the last client from the `.login()` method.

        Returns:
            Team: The Team object.
        """

        return await cls(id=id).get(synapse_client=synapse_client)

    @classmethod
    @otel_trace_method(
        method_to_trace_name=lambda cls, name, **kwargs: f"Team_From_Name: {name}"
    )
    async def from_name(
        cls, name: str, synapse_client: Optional[Synapse] = None
    ) -> "Team":
        """Gets Team object using its string name.

        Arguments:
            name: The name of the team.
            synapse_client: If not passed in or None this will use the last client from the `.login()` method.

        Returns:
            Team: The Team object.
        """
        return await cls(name=name).get(synapse_client=synapse_client)
```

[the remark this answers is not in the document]
Comment (BWMac) [reply synapseclient/models/team.py:207]: I guess my gut feeling is to go with option 1, though I'm not sure I could make a strong case that it is better than 2. Do you have a preference @BryanFauble?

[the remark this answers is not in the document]
Comment (BryanFauble) [reply synapseclient/models/team.py:207]: I like the feeling of 2 better. Mainly because of this line:
`self.__dict__.update(new_instance.__dict__)`

That is required in the first implementation feels more like a hack than anything.

**Comment by BWMac** [comment]
@BryanFauble No, I don't believe you can do that with this interface is it stands. I am also unaware of if there is such a method already implemented in `client.py`.

File: docs/scripts/object_orientated_programming_poc/oop_poc_team.py [hunk 12]
```diff
@@ -0,0 +1,57 @@
...
+    await asyncio.sleep(5)
+    # Instantiate a Team object from a Synapse team
+    my_team = await Team().from_id(id=my_synapse_team.id)
```
BryanFauble, line 33:
```suggestion
    my_team = await Team.from_id(id=my_synapse_team.id)
```

docs/scripts/object_orientated_programming_poc/oop_poc_team.py
```diff
@@ -0,0 +1,57 @@
...
+    print(my_team)
+
+    my_team = await Team().from_name(name=my_synapse_team.name)
```
BryanFauble, line 36:
```suggestion
    my_team = await Team.from_name(name=my_synapse_team.name)
```

BryanFauble (approved):
🔥 🔥 

These changes look good to me!

synapseclient/client.py
```diff
@@ -3853,47 +3853,104 @@ def _find_teams_for_principal(self, principal_id: str) -> typing.Iterator[Team]:
...
+            id: The ID of the team to delete.
+            opentelemetry_context: OpenTelemetry context to propogate to this function to use for tracing.
+                                    Used in cases where concurrent operations need to be linked to parent spans.
+        """
+        with tracer.start_as_current_span(
+            "Synapse::delete_team", context=opentelemetry_context
+        ):
+            return self.restDELETE(f"/team/{id}")
+
+    def getTeam(self, id: Union[int, str], opentelemetry_context=None) -> Team:
```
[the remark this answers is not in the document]
BWMac: ticket: https://sagebionetworks.jira.com/browse/PLFM-8240

synapseclient/models/team.py
```diff
@@ -1 +1,357 @@
...
+        """
+        loop = asyncio.get_event_loop()
+        current_context = context.get_current()
+        team = await loop.run_in_executor(
+            None,
+            lambda: run_and_attach_otel_context(
+                lambda: Synapse.get_client(synapse_client=synapse_client).create_team(
+                    name=self.name,
+                    description=self.description if self.description else None,
+                    icon=self.icon if self.icon else None,
```
thomasyu888, line 151: Here you aren't passing in the "canPublicJoin" parameter, so it wouldn't be configuring the Team correctly on Synapse.

synapseclient/client.py
BryanFauble: Great point @thomasyu888 . @BWMac Could you make a modification here to pass along the possible creation parameters for the team for the REST API: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/Team.html

As Tom mentions I think it's just `canPublicJoin`, and `canRequestMembership`. The other fields look to be read only, not applicable to a create, or you're already passing it along. These two bools can be arguments to this method.

docs/scripts/object_orientated_programming_poc/oop_poc_team.py
BWMac: @BryanFauble @thomasyu888 This is the [response](https://sagebionetworks.jira.com/browse/PLFM-8240?focusedCommentId=196075) from platform about this delay with using `from_name` to retrieve a team. It appears as though this behavior is intended. Should I just document this in the docstring?

BryanFauble: Yes, that works for me! Let's document that it is the expected behavior for that method.

synapseclient/client.py
BWMac: added in my latest commit

synapseclient/client.py
```diff
@@ -3852,6 +3852,8 @@ def create_team(
         name: str,
         description: str = None,
         icon: str = None,
+        canPublicJoin: bool = False,
```
BryanFauble, line 3855: nit: Since these are new methods we should follow snake_case

thomasyu888 (approved):
🔥 LGTM!

BryanFauble (approved):

PR #1054, opened 2024-01-25
[SYNPY-1344] Allows Socket Connections for Windows Systems in Unit Tests

BWMac:
**Problem:**

Asynchronous function unit testing does not work in Windows GH Actions runners due to a [lack of support](https://github.com/python/cpython/issues/77589) for UNIX socket for Windows in Python. 

**Solution:**

Enable socket connections in unit tests on Windows systems.

### DIFF
--- a/tests/unit/conftest.py
+++ b/tests/unit/conftest.py
@@ -22,6 +22,7 @@ def pytest_runtest_setup():
 
     allow_unix_socket=True is required for async to work.
     """
+    # This is a work-around because of https://github.com/python/cpython/issues/77589
     if platform.system() != "Windows":
         disable_socket(allow_unix_socket=True)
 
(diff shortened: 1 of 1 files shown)


tests/unit/conftest.py
```diff
@@ -22,14 +22,16 @@ def pytest_runtest_setup():
 
     allow_unix_socket=True is required for async to work.
     """
-    disable_socket(allow_unix_socket=True)
+    if platform.system() != "Windows":
```
BryanFauble, line 25:
```suggestion
    # This is a work-around because of https://github.com/python/cpython/issues/77589
    if platform.system() != "Windows":
```

BryanFauble (approved):
LGTM!

PR #1056, opened 2024-01-26
[SYNPY-1344] Remove need to manually propogate otel context

BryanFauble:
@BWMac Which would you prefer: The current way we are passing the context along (No change), or wrapping the calling side to handle for this (This change)?

**Problem:**
Because we are using `run_in_executor` to execute blocking code in an async context otel context is not propogating to the new code that is executing. As a result we were required to pass the context along and modify the functions to accept the parameter to make it work.

**Solution:**
Creating a wrapper function that we can add to `run_in_executor` that attaches the otel context for us; prevents us from needing to modifying all of the sync code function signatures.

**Testing:**
I verified I could run our OOP test scripts and all otel context remained:
image

image
image

image
image


synapseclient/api/annotations.py
thomasyu888, line 1: @BryanFauble On a high level, I prefer this over passing around the context manager.  I haven't had the time to review closer, but this is great.

Can the team @Sage-Bionetworks/dpe want pitch in on this review?

synapseclient/core/utils.py
```diff
@@ -1262,52 +1265,35 @@ def print_tick(self):
...
-            """Wrapper for the function to be traced."""
-            trace_name = (
-                method_to_trace_name(self, *arg, **kwargs)
-                if method_to_trace_name
-                else None
+    This is a generic function that will run a callable function and attach the passed in
+    OpenTelemetry context to the thread or context that the function is running on.
+
+    This is a hack to get around AsyncIO `run_in_executor` not propagating the context
+    to the code it's executing. When we are directly calling async functions after
```
BWMac, line 1276: If we can remove this after SYNPY-1411 is complete, what does the difference between making this change now vs. later look like?

BWMac (approved):
I like this change. LGTM but just one question.

synapseclient/core/utils.py
BryanFauble: Making the change to use this wrapper now prevents us from needing to modify the code in `client.py` to manually propagate the OTEL context.

Making the change later means we will have had to propagate the OTEL to the rest of the sync methods we want to call in `client.py` that haven't been called yet.

BWMac (commented):
Looks good, I really like the mixin for permissions. Only thing is permissions issues for the demo script. I'll come back and approve once I can run it.

BWMac (commented):
Looks good, I really like the mixin for permissions. Only thing is permissions issues for the demo script. I'll come back and approve once I can run it.

PR #1058, opened 2024-01-30
[SYNPY-1348] Add User profile and Permission models

BryanFauble:
**Problem:**

1. User profile and permissions need to be added to the OOP model

**Solution:**

1. Adding models to the client

**Testing:**

1. The sample script
2. Integration/unit tests

### DIFF
--- a/docs/scripts/object_orientated_programming_poc/oop_poc_user_and_permission.py
+++ b/docs/scripts/object_orientated_programming_poc/oop_poc_user_and_permission.py
@@ -1,0 +1,97 @@
+"""The purpose of this script is to demonstrate how to use the OOP interface for
+UserProfile and permissions on some entities.
+The following actions are shown in this script:
+1. Get a UserProfile by name and id
+2. Check if a UserProfile is certified
+3. Set permissions on a File
+4. Get permissions on a File
+"""
+import asyncio
+import os
+
+from synapseclient.models import (
+    File,
+    UserProfile,
+)
+import synapseclient
+
+PROJECT_ID = "syn52948289"
+DPE_TEAM_SERVICE_ACCOUNT_NAME = "synapse-service-dpe-team"
+DPE_TEAM_SERVICE_ACCOUNT_ID = "3485485"
+
+syn = synapseclient.Synapse(debug=True)
+syn.login()
+
+
+def create_random_file(
+    path: str,
+) -> None:
+    """Create a random file with random data.
+
+    :param path: The path to create the file at.
+    """
+    with open(path, "wb") as f:
+        f.write(os.urandom(1))
+
+
+async def user_profile():
+    dpe_team_profile = await UserProfile(username=DPE_TEAM_SERVICE_ACCOUNT_NAME).get()
+    print(f"Getting account by name: {dpe_team_profile}")
+
+    dpe_team_profile = await UserProfile(id=DPE_TEAM_SERVICE_ACCOUNT_ID).get()
+    print(f"Getting account by id: {dpe_team_profile}")
+
+    dpe_team_profile = await UserProfile.from_username(
+        username=DPE_TEAM_SERVICE_ACCOUNT_NAME
+    )
+    print(f"Getting account by name: {dpe_team_profile}")
+
+    dpe_team_profile = await UserProfile.from_id(user_id=DPE_TEAM_SERVICE_ACCOUNT_ID)
+    print(f"Getting account by id: {dpe_team_profile}")
+
+    print(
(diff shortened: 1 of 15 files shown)


docs/scripts/object_orientated_programming_poc/oop_poc_user_and_permission.py
```diff
@@ -0,0 +1,97 @@
...
+        f"is certified - By name?: {await UserProfile(username=DPE_TEAM_SERVICE_ACCOUNT_NAME).is_certified()}"
+    )
+
+
+async def set_permissions():
+    name_of_file = "my_file_with_read_permissions_for_service_account.txt"
+    path_to_file = os.path.join(os.path.expanduser("~/temp"), name_of_file)
+    create_random_file(path_to_file)
+
+    file_with_read = await File(
```
BWMac, line 65: @BryanFauble I am running into permissions issues here
```
synapseclient.core.exceptions.SynapseHTTPError: 403 Client Error: 
Lack of READ permission on the requested entity.
```

BWMac (commented):
looks great. I really like the mixin implementation. Only thing is permission issues in the demo script. I'll come back to approve once I can run it to completion

synapseclient/client.py
```diff
@@ -2619,7 +2619,7 @@ def setPermissions(
         Arguments:
             entity: An Entity or Synapse ID to modify
             principalId: Identifier of a user or group. '273948' is for all registered Synapse users
-                            and '273949' is for public access.
+                            and '273949' is for public access. None implies public access.
```
thomasyu888, line 2622: Does `None` imply both public and all registered Synapse users OR just public access?

docs/scripts/object_orientated_programming_poc/oop_poc_user_and_permission.py
BryanFauble: My mistake - During testing I removed the DPE team from sharing settings on the File in order to verify that only the user permissions were being applied (Instead of BOTH the user and team permissions). 

What this means is because the DPE team service account is a member of the DPE team when you create the local sharing settings and print out the effective ACL it's going to be the union of both ACL sets.

Instead of using the DPE team service account I swapped this to use my other prod testing account. You should be able to run the script now @BWMac

synapseclient/client.py
BryanFauble: I copied the comment over from one of the methods this calls. But after setting it to None and running the code it sets `Anyone on the web` to those permissions:
image

"Only registered synapse users" permissions are not touched

BWMac (approved):
@BryanFauble It works now, thanks for updating the permissions.

synapseclient/models/mixins/access_control.py
thomasyu888, line 1: Just a note, Evaluation objects in Synapse have a different access control list.

BryanFauble: Good to know. When that gets created we may need to make a mixin for that specifically than

synapseclient/models/user.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+@dataclass()
+class UserPreference:
+    """
+    A UserPreference represents a user preference in the system.
+
+    Attributes:
+        name: The name of the user preference.
+        concrete_type: The concrete type of the user preference.
```
thomasyu888, line 23: @BryanFauble are you missing concrete_type here?

BryanFauble: No. I just removed it from the docstring

synapseclient/models/user.py
```diff
@@ -0,0 +1,311 @@
...
+    """This person's default team name"""
+
+    send_email_notifications: Optional[bool] = True
+    """Should the user receive email notifications? Default true."""
+
+    mark_emailed_messages_as_read: Optional[bool] = False
+    """Should messages that are emailed to the user be marked as
+    READ in Synapse? Default false."""
+
+    preferences: Optional[List[UserPreference]] = field(default_factory=list)
```
thomasyu888, line 121: Nit: Not related, but I don't understand this `preferences` parameter.  I took a look at the API and it doesn't tell you much about it either...

thomasyu888 (approved):
🔥 LGTM!

synapseclient/models/user.py
BryanFauble: I was messing around with this as well since it isn't explained anywhere. It let's you set arbitrary booleans on your account. For example if you get the userprofile for my account i set an example value. `bfauble`

PR #1059, opened 2024-01-31
[SYNPY-1411] Prevent Repeated `syn.get` calls in `_help_walk`

BWMac:
**Problem:**

It was reported recently that running `synapseutils.walk` on large projects in Synapse with many folders and subfolders was taking a very long time due to `syn.get` being executed on every recursive `_helpWalk` call. Further, as pointed out by @linglp, this is unnecessary since we are also using `syn.getChildren` which provides the same information about entity types.

**Solution:**

Update `_helpWalk` to use the entities returned by `syn.getChildren` in the recursive function call. Specifically, repeated calls to `syn.get` are prevented by adding a new parameter that carries the starting entity into each recursive iteration. `syn.get` is called once per `walk` call to initiate the process from the parent folder or project.

Notes:
- A brief exploration was done to determine if we could further optimize this function by using [functools.cache](https://docs.python.org/3/library/functools.html), but this was not implemented due to the way that `cache` [interacts with generators](https://stackoverflow.com/questions/36803626/why-does-functools-lru-cache-break-this-function).
- `_helpWalk` and its parameters were converted to using snake case
- Unit tests for the walk functions were updated as needed.

### DIFF
--- a/synapseutils/walk_functions.py
+++ b/synapseutils/walk_functions.py
@@ -8,7 +8,7 @@ from synapseclient.entity import is_container
 def walk(
     syn: synapseclient.Synapse,
     synId: str,
-    includeTypes=[
+    includeTypes: typing.List[str] = [
         "folder",
         "file",
         "table",
@@ -67,14 +67,15 @@ def walk(
     # Ensure that "folder" is included so the hierarchy can be traversed
     if "folder" not in includeTypes:
         includeTypes.append("folder")
-    return _helpWalk(syn, synId, includeTypes)
+    return _help_walk(syn=syn, syn_id=synId, include_types=includeTypes)
 
 
 # Helper function to hide the newpath parameter
-def _helpWalk(
+def _help_walk(
     syn: synapseclient.Synapse,
-    synId: str,
-    includeTypes: typing.List[str],
+    syn_id: str,
+    include_types: typing.List[str],
+    start_entity: typing.Union[synapseclient.Entity, dict] = None,
     newpath: str = None,
 ):
     """Helper function that helps build the directory path per result by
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):
Changes lgtm!

Ran this sample script to verify everything in the project printed:
```
for directory_path, directory_names, file_name in synapseutils.walk(
    syn=syn, synId="syn49637038"
):
    for directory_name in directory_names:
        print(
            f"Directory ({directory_name[1]}): {directory_path[0]}/{directory_name[0]}"
        )

    for file in file_name:
        print(f"({file[1]}): {directory_path[0]}/{file[0]}")
```

And no errors were received.

thomasyu888 (approved):
🔥 Excellent - this is another great example of iterative improvement to the codebase.

synapseutils/walk_functions.py
```diff
@@ -67,14 +67,15 @@ def walk(
     # Ensure that "folder" is included so the hierarchy can be traversed
     if "folder" not in includeTypes:
         includeTypes.append("folder")
-    return _helpWalk(syn, synId, includeTypes)
+    return _help_walk(syn=syn, syn_id=synId, include_types=includeTypes)
 
 
 # Helper function to hide the newpath parameter
-def _helpWalk(
+def _help_walk(
```
thomasyu888, line 74: Nit: Not sure what I was thinking when I was writing this, but the `_` is almost redundant with `help`.  I can imagine this could be `_walk`, but... this is good.  no action required.  Just a thought in the future for naming functions.

BWMac:
@BryanFauble Thanks for testing, next time I will include a script/snippet to save reviewers the trouble.

PR #1061, opened 2024-02-07
[SYNPY-1415] Folder model finishing touches

BryanFauble:
Problem:

1. The folder model needed some TLC and bug fixing
2. We needed to add some more functionality to the Folder model to make it more usable

Solution:

1. Updating and bug squashing
2. Writing a sync_from_synapse function in the Folder model
3. Including a copy method

Testing:

1. Integration and unit testing
2. Some initial testing (Obviously more is needed) compared to the `synapseutils.syncFromSynapse` on a project with 10 files/9 directories:

async: 1.9427805209998041
syncFromSynapse: 3.578380188002484

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,6 +12,7 @@ repos:
   - id: check-merge-conflict
   - id: check-xml
   - id: check-yaml
+    exclude: ^mkdocs\.yml$
   - id: debug-statements
   - id: end-of-file-fixer
   - id: requirements-txt-fixer
--- a/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
+++ b/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
@@ -158,7 +158,7 @@ async def store_folder():
         name="destination_for_copy", parent_id=PROJECT_ID
     ).store()
     coped_folder = await root_folder_for_my_project.copy(
-        destination_id=destination_folder_to_copy_to.id
+        parent_id=destination_folder_to_copy_to.id
     )
 
     print(
--- a/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
+++ b/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
@@ -41,10 +41,11 @@ def create_random_file(
         f.write(os.urandom(1))
 
 
-async def try_delete_folder(folder_name: str, parent_id: str):
+async def try_delete_folder(folder_name: str, parent_id: str) -> None:
+    """Simple try catch to delete a folder."""
     try:
         await (await Folder(name=folder_name, parent_id=parent_id).get()).delete()
-    except Exception as e:
+    except Exception:
         pass
 
 
--- a/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
+++ b/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
@@ -1,4 +1,7 @@
-"""The purpose of this script is to demonstrate how to use the new OOP interface for folders.
+"""
+Expects that ~/temp exists and is a directory.
+
+The purpose of this script is to demonstrate how to use the new OOP interface for folders.
 The following actions are shown in this script:
 1. Creating a folder
 2. Storing a folder to a project
@@ -114,6 +117,8 @@ async def store_folder():
     }
 
     for file in folder_copy.files:
+        file.download_file = False
+        await file.get()
         file.annotations = new_annotations
 
     for folder in folder_copy.folders:
--- a/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
+++ b/docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
@@ -7,10 +7,13 @@ The following actions are shown in this script:
 2. Storing a folder to a project
 3. Storing several files to a folder
 4. Storing several folders in a folder
-5. Getting metadata about a folder
+5. Getting metadata about a folder and it's immediate children
 6. Updating the annotations in bulk for a number of folders and files
 7. Deleting a folder
+8. Copying a folder
+9. Moving a folder
 """
+
 import asyncio
 import os
 from synapseclient.models import (
@@ -37,7 +40,21 @@ def create_random_file(
         f.write(os.urandom(1))
 
 
+async def try_delete_folder(folder_name: str, parent_id: str):
+    try:
+        await (await Folder(name=folder_name, parent_id=parent_id).get()).delete()
+    except Exception as e:
+        pass
+
+
 async def store_folder():
+    # Clean up synapse for previous runs:
+    await try_delete_folder("my_new_folder_for_this_project", PROJECT_ID)
+    await try_delete_folder("destination_for_copy", PROJECT_ID)
+    await try_delete_folder(
+        "my_new_folder_for_this_project_I_want_to_delete", PROJECT_ID
+    )
+
     # Creating annotations for my folder ==================================================
     annotations_for_my_folder = {
         "my_single_key_string": "a",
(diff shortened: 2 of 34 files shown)


.pre-commit-config.yaml
```diff
@@ -12,6 +12,7 @@ repos:
   - id: check-merge-conflict
   - id: check-xml
   - id: check-yaml
+    exclude: ^mkdocs\.yml$
```
BryanFauble, line 15: `format: !!python/name:pymdownx.superfences.fence_code_format` is not valid syntax according this this check-yml check - However, this is the format required for it to work: https://squidfunk.github.io/mkdocs-material/reference/diagrams/#configuration

synapseclient/client.py
```diff
@@ -2342,27 +2341,28 @@ def getChildren(
 
         - [synapseutils.walk][]
         """
-        parentId = id_of(parent) if parent is not None else None
+        with tracer.start_as_current_span("Synapse::getChildren"):
```
thomasyu888, line 2344: I thought we were using the decorator?

synapseclient/core/exceptions.py
```diff
@@ -15,6 +15,10 @@ class SynapseFileNotFoundError(SynapseError):
     """Error thrown when a local file is not found in Synapse."""
 
 
+class SynapseNotFoundError(SynapseError):
```
thomasyu888, line 18: Nit: does python base exceptions have something for Not found errors?

synapseclient/models/file.py
```diff
@@ -816,14 +816,14 @@ async def copy(
...
-            Copy the file but do not persist annotations of activity:
+            Copy the file but do not persist annotations or activity:
 
                 new_file_instance = await File(id="syn123").copy(destination_id="syn456", copy_annotations=False, copy_activity=None)
 
         Raises:
             ValueError: If the file does not have an ID and destination_id to copy.
         """
-        if not self.id:
+        if not self.id or not destination_id:
```
thomasyu888, line 826: I would love to pivot away from destination id and just say parent id

synapseclient/models/folder.py
```diff
@@ -266,6 +312,78 @@ async def delete(self, synapse_client: Optional[Synapse] = None) -> None:
...
+                current_context=current_context,
+            ),
+        )
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Folder_Copy: {self.id}"
+    )
+    async def copy(
+        self,
+        destination_id: str,
```
thomasyu888, line 324: Same note above about wanting to be consistent in the future and not have destination id

synapseclient/models/mixins/__init__.py
```diff
@@ -1,5 +1,9 @@
+"""References to the mixins that are used in the Synapse models."""
+
```
thomasyu888, line 2: Out of scope for this ticket but I think there is also a versionable mixin that currently exists in the client.

tests/integration/synapseclient/test_json_schema_services.py
```diff
@@ -63,13 +63,13 @@ def test_json_schema_organization(js):
 
 
 class TestJsonSchemaSchemas:
-    @pytest.fixture(autouse=True)
+    @pytest.fixture(autouse=True, scope="function")
```
BryanFauble, line 66: This is an effort to resolve these integration test errors:

> Semantic version: '0.48786.1' already exists for this JSON schema

That have been popping up. These passed on my local machine - Will be verifying they also pass in the CI pipeline

synapseclient/core/exceptions.py
BryanFauble: I did not find one - The closet from: https://docs.python.org/3/library/exceptions.html

were:

1. FileNotFoundError - Derived from OSError and not applicable in all cases
2. ModuleNotFoundError - Not really the same thing

synapseclient/models/folder.py
BryanFauble: Updated in latest commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1061/commits/33a1f4721bd7b6c4168d107cfcc0fb257304520b

tests/integration/synapseclient/test_json_schema_services.py
jaymedina: Can confirm these tests also passed on my machine.

docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
```diff
@@ -87,52 +107,88 @@ async def store_folder():
...
+    ).store()
+    current_parent_id = folder_i_am_going_to_move.parent_id
+    folder_i_am_going_to_move.parent_id = destination_folder_to_copy_to.id
+    await folder_i_am_going_to_move.store()
+    print(
+        f"Moved folder from {current_parent_id} to {folder_i_am_going_to_move.parent_id}"
+    )
+
+    # 10) Using sync_from_synapse to download the files and folders ======================
+    # This will download all the files and folders in the folder to the local file system
```
jaymedina, line 187: This looks good to me! Do you think it would be useful to have a `Folder().getChildren()` functionality to see what the child entities are for the Folder before syncing from Synapse?

synapseclient/models/services/storable_entity_components.py
```diff
@@ -51,59 +79,86 @@ async def store_entity_components(
...
+            raise ex
 
-    re_read_required = await _store_activity(
-        root_resource, synapse_client=synapse_client
-    )
+    return re_read_required
+
+
+def _resolve_store_task(
+    result: Union[bool, "Folder", "File", BaseException],
```
jaymedina, line 108: Are the "Folder" / "File" optional values correct ? This `result` arg later gets called to get the `.__class__.__name__` attribute which would just be `'str'` for these. Unless there is something I'm missing about type hints here

jaymedina (approved):
This looks good. Integration tests pass and I can confirm the poc script for the Folder model works. I had a few comments, but if anything comes out of them, I think it can be for another PR. Don't see anything getting in the way of having this one merged for validation.

Just a note: The provenance of copied entities gets updated with where they were copied from and the version appears to reset, which I believe is expected behavior(?) but want to post here in case not
image

docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
```diff
@@ -87,52 +107,88 @@ async def store_folder():
...
-    folder = await folder.store()
+    root_folder_for_my_project.folders = folders_to_store
+    root_folder_for_my_project = await root_folder_for_my_project.store()
 
-    # Getting metadata about a folder =====================================================
-    folder_copy = await Folder(id=folder.id).get(include_children=True)
+    # 5) Getting metadata about a folder and it's immediate children =====================
+    new_folder_instance = await Folder(
+        id=root_folder_for_my_project.id
+    ).sync_from_synapse(download_file=False, recursive=False)
```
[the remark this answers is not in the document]
BryanFauble: Yes - As doing this from a File entity would be the same as calling `.get()`

docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
BryanFauble: This functionality is available by:

```
                from synapseclient.models import Folder
                syn = Synapse()
                syn.login()
                my_folder = Folder(id="syn12345")
                await my_folder.sync_from_synapse(download_file=False, recursive=False)
                for folder in my_folder.folders:
                    print(folder.name)
                for file in my_folder.files:
                    print(file.name)
```

synapseclient/models/services/storable_entity_components.py
BryanFauble: It is correct here - The problem that this is solving is an issue of circular dependencies. `isinstance` only works if you are passing an initialized reference to the class - But I was unable to import it at the top of the file (circular dependency) or under `if TYPE_CHECKING` (Not initialized).

I could have done a lazy import (ie: Importing in the method) - But this also works.

docs/scripts/object_orientated_programming_poc/oop_poc_folder.py
jaymedina: Not for this PR but I think it would be good to have an example like this before example 10 in your Folder poc script later down the line

BryanFauble: This example is available in the docstring of the function. IMO I don't think it's necessary to give all possible ways the function can be used in these example scripts - Although I'm more than happy to make the change.

jaymedina: If it seems redundant, then no need to add it. You mentioned writing up some tutorials for synpy support at the team retro - I think this type of example is better suited for that type of thing, especially if we go with Jupyter Notebooks for the tutorials.

BryanFauble: This is the start of those tutorials - though geared towards the current interfaces
http://python-docs.synapse.org/tutorials/python_client/

PR #1063, opened 2024-02-08
[SYNPY-1253] Check MD5 before upload to verify change in content

BryanFauble:
**Problem:**

1. When there is a cache miss we never checked if the content of the file to be uploaded actually changed. As a result we would always upload the file to Synapse.

**Solution:**

1. If we need to upload a file to Synapse and the file already exists in Synapse and we got back an MD5 from Synapse compare the MD5 in Synapse to what the local file is - Is they are the same do not upload the file.

**Testing:**

1. Integration tests around this logic

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -1620,35 +1620,35 @@ class Synapse(object):
                         fileHandle["externalURL"] != entity["externalURL"]
                     )
                 else:
+                    synapse_store_flag = entity["synapseStore"] or local_state.get(
+                        "synapseStore"
+                    )
                     # Check if we need to upload a new version of an existing
                     # file. If the file referred to by entity['path'] has been
                     # modified, we want to upload the new version.
                     # If synapeStore is false then we must upload a ExternalFileHandle
-                    needs_upload = not entity[
-                        "synapseStore"
-                    ] or not self.cache.contains(
+                    needs_upload = not synapse_store_flag or not self.cache.contains(
                         bundle["entity"]["dataFileHandleId"], entity["path"]
                     )
 
+                    md5_stored_in_synapse = local_state.get("_file_handle", {}).get(
+                        "contentMd5", None
+                    ) or (fileHandle or {}).get("contentMd5", None)
+
                     # Check if we got an MD5 checksum from Synapse and compare it to the local file
                     if (
-                        (entity["synapseStore"] or local_state.get("synapseStore"))
-                        and os.path.isfile(entity["path"])
+                        synapse_store_flag
                         and needs_upload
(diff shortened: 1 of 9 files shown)


synapseclient/client.py
```diff
@@ -1628,6 +1629,26 @@ def store(
...
+                        and needs_upload
+                        and (
+                            md5_stored_in_synapse := (
+                                local_state.get("_file_handle", {}).get(
+                                    "contentMd5", None
+                                )
+                                or (fileHandle or {}).get("contentMd5", None)
+                            )
+                        )
+                    ):
```
BWMac, line 1646: The logic in this function is pretty complicated/difficult to read. Would it be possible to break up some of these logical statements or pull some out of this `if` block?

synapseclient/client.py
```diff
@@ -1657,7 +1678,7 @@ def store(
                     if (synapseStore or local_state_fh.get("externalURL") is None)
                     else local_state_fh.get("externalURL"),
                     synapseStore=synapseStore,
-                    md5=local_state_fh.get("contentMd5"),
+                    md5=local_file_md5_hex or local_state_fh.get("contentMd5"),
```
BWMac, line 1681: What happens if both these things are `None`?

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -1033,7 +1033,7 @@ def abort_side_effect(futures):
                 uploader.upload(items)
 
             # it would be aborted with Futures
-            mock_abort.assert_called_once_with([ANY])
+            mock_abort.assert_called_once()
             isinstance(mock_abort.call_args_list[0][0], Future)
```
BWMac, line 1037: I know this wasn't part of this PR, but should there be an `assert` before this `isinstance`?

BWMac (approved):
Looks good! Just a few comments/questions

synapseclient/client.py
BryanFauble: Further down in the code it will get the md5 from the file on disk. Both are none during the first upload of a file - since the code above is only checked for a file that already exists in Synapse. And the file handle hasn't been created yet.

synapseclient/client.py
BryanFauble: I modified this if check a bit to make it a bit easier to read - I also verified that `md5_for_file` utility function is still only getting called once and `local_file_md5_hex` is getting pass along to the `upload_file_handle` function so it doesn't get re-calculated again later on

thomasyu888:
🔥

PR #1065, opened 2024-02-10
[SYNPY-448] allow changing Synapse name in changeFileMetaData

danlu1:
Problem:

Python client doesn't allow changing Synapse name for file metadata 

Solution:

Add a name argument to change_metadata in file.py and changeFileMetaData function in synapseutils

Testing:

Integration and unit testing are included in this PR

### DIFF
--- a/synapseutils/copy_functions.py
+++ b/synapseutils/copy_functions.py
@@ -264,7 +264,7 @@ def changeFileMetaData(
 
             file_entity = syn.get(synid)
             print(os.path.basename(file_entity.path))  ## prints, e.g., "my_file.txt"
-            file_entity = synapseutils.changeFileMetaData(syn, file_entity, "my_new_name_file.txt", "my_new_downloadAs_name_file.txt")
+            file_entity = synapseutils.changeFileMetaData(syn=syn, entity=file_entity, downloadAs="my_new_downloadAs_name_file.txt", name="my_new_name_file.txt")
             print(os.path.basename(file_entity.path))  ## prints, "my_new_downloadAs_name_file.txt"
             print(file_entity.name) ## prints, "my_new_name_file.txt"
     """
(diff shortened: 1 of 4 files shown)


synapseutils/copy_functions.py
```diff
@@ -239,6 +239,7 @@ def _copy_cached_file_handles(cache: Cache, copiedFileHandles: dict) -> None:
 def changeFileMetaData(
     syn: synapseclient.Synapse,
     entity: typing.Union[str, Entity],
+    name: str = None,
```
BryanFauble, line 242: I am going to make 1 change here to put name at the end of the parameters being passed in - This is because if someone were using this function without explicit kwarguments then this would break their code.

synapseutils/copy_functions.py
```diff
@@ -242,27 +242,31 @@ def changeFileMetaData(
...
     Example: Using this function
-        Can be used to change the filename or the file content-type without downloading:
+        Can be used to change the filename, the filename when the file is downloaded, or the file content-type without downloading:
 
-        file_entity = syn.get(synid)
-        print(os.path.basename(file_entity.path))  ## prints, e.g., "my_file.txt"
-        file_entity = synapseutils.changeFileMetaData(syn, file_entity, "my_new_name_file.txt")
+            file_entity = syn.get(synid)
+            print(os.path.basename(file_entity.path))  ## prints, e.g., "my_file.txt"
+            file_entity = synapseutils.changeFileMetaData(syn, file_entity, "my_new_name_file.txt", "my_new_downloadAs_name_file.txt")
```
BWMac, line 267: This looks like it needs to be updated to reflect the change in argument order

BryanFauble: Great catch! I updated the order, but I also changed this to use kwargs: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1065/commits/65dd776245bfe10eeb3f46cf714ea5071ffb941e

BWMac (approved):

BryanFauble (approved):
Thanks for the changed Dan!

synapseutils/copy_functions.py
danlu1: Good catch! Thanks @BryanFauble

PR #1066, opened 2024-02-12
[SYNPY-1326] Expand range for pandas dependency to <3.0

BryanFauble:
Problem:

1. We were locking the pandas dependency to <2.1

Solution:

1. Expanding the pandas range to allow any 2.x version - From a look at the change log there are no breaking changes here
2. Updating pipfile


thomasyu888 (approved):
🔥 LGTM!

PR #1067, opened 2024-02-13
[SYNPY-1414] Project model finishing touches

BryanFauble:
**Problem:**

1. Not all fields were being saved on the Project
2. I missed the `is_restricted` flag at the folder level that gets passed to syn.store()
3. We are not able to store projects or folders by ID only in a non-destructive way. When we did an upsert action (Called `createOrUpdate`) we did so only after we got an HTTP 409 conflict, but only if we did not specify the ID of the resource. This led to inconsistent behavior if you supplied an ID, instead of a name when saving resources like the Project.
4. You could not specify project alias through the client before
5. copy needed to be added to project
6. Copy on the project/folder level could take in parameters that affect how files are copied
7. Annotations were not copied over during a project copy

**Solution:**

1. Adding missing fields to be saved on project
2. Adding an `is_restricted` flag to the folder level
3. Allow storing projects/folders by ID with an upsert by reading for the stored synapse instance if we didn't already communicate with Synapse for that instance before. Updating when an 'upsert' logic occurs to be before we attempt to persist anything to Synapse. First we do a check to find the ID of the resource being updated, and if found, retrieve the current state from Synapse. We then merge the object in Synapse to the requested changed - Only if there is an update will we then store the data via an HTTP PUT. If there are no change to be made we do not make the call. If the resource is not found in Synapse we perform a create via HTTP POST.
4. Adding project alias
5. Adding copy to Project
6. Adding in project/folder level parameters for file copying
7. Copy project annotations

**Testing:**

1. Added the POC script for working with a project object
2. Added unit/integration testing


docs/scripts/object_orientated_programming_poc/oop_poc_project.py
```diff
@@ -53,35 +53,51 @@ async def store_project():
...
-    # Creating a project =============================================================
+    # 1) Creating a project ==============================================================
     project = Project(
         name="my_new_project_for_testing",
         annotations=my_annotations,
         description="This is a project with random data.",
+        alias="my_project_alias_" + str(uuid.uuid4()).replace("-", "_"),
     )
 
     project = await project.store()
```
BWMac, line 67: I am getting a `Lack of READ permission on the requested entity` error on this line.

synapseclient/models/project.py
```diff
@@ -146,23 +153,51 @@ class Project(AccessControllable, StorableContainer):
...
+        )
+
     def _set_last_persistent_instance(self) -> None:
         """Stash the last time this object interacted with Synapse. This is used to
         determine if the object has been changed and needs to be updated in Synapse."""
         del self._last_persistent_instance
         self._last_persistent_instance = replace(self)
+        self._last_persistent_instance.annotations = (
+            deepcopy(self.annotations) if self.annotations else None
+        )
```
BWMac, line 200: These also are in the folder model, is there a way to have them as part of a mixin or another way to not have the code repeated?

docs/scripts/object_orientated_programming_poc/oop_poc_project.py
BryanFauble: I gave the DPE team permissions on the project

synapseclient/client.py
```diff
@@ -5823,6 +5823,7 @@ def _rest_call(
         Returns:
             JSON encoding of response
         """
+        self.logger.debug(f"Sending {method} request to {uri}")
```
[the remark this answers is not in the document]
BryanFauble: I hadn't intended to remove it - But I am curious to know your thoughts. Personally I think it's useful if you turn on debug level to know what HTTP requests you are sending out. It's also super useful in cases where folks may complain that the client is "stuck" and doing nothing - But really may be doing a lot of HTTP calls that don't have any other log messages.

synapseclient/models/project.py
BryanFauble: Yes - But I am interested to get other models in place before I commit to a mixin for this. In particular table is likely to work the same/similar to this, but I wanted to get to it before making this change.

File is also similar to this - But slightly different in that it also have an `Activity` model which also needs the `dataclasses.replace` function to be called on it.

synapseclient/client.py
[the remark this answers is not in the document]
BWMac: Yeah OK I can see this being useful in that sense. I forgot that the POC script had `debug` enabled so I thought it was doing this by default at first. If this isn't documented somewhere we should include it in some sort of "help yourself with debugging the client before filing an issue" document.

synapseclient/models/project.py
BWMac: Makes sense to me!

synapseclient/client.py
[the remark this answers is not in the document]
BryanFauble: I added: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1067/commits/0a0f9773b0ce7d5c601af5b70e8c63ed1db115e6 - Is this what you had in mind?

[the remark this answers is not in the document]
BWMac: I think that is helpful for developers, but I was thinking more from an individual user's perspective. So something in the doc site probably.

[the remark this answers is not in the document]
BryanFauble: I added this section to the home page of the tutorials:
image

How does this look? https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1067/commits/b0a740cc4bb1f11bd5e4a1427bfa4e02367e4280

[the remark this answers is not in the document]
BWMac: Awesome, thanks!

BWMac (approved):

docs/scripts/object_orientated_programming_poc/oop_poc_project.py
```diff
@@ -53,35 +53,51 @@ async def store_project():
...
+        "annotation_i_want_to_delete": "I want to delete this annotation",
     }
 
-    # Creating a project =============================================================
+    # 1) Creating a project ==============================================================
     project = Project(
         name="my_new_project_for_testing",
         annotations=my_annotations,
         description="This is a project with random data.",
+        alias="my_project_alias_" + str(uuid.uuid4()).replace("-", "_"),
```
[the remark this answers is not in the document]
BryanFauble: When you set a project alias you may access it via:
https://www.synapse.org/alias_bfauble_oop_refactor_testing_d51be
or
https://www.synapse.org/#!Synapse:syn52948289

synapseclient/core/utils.py
```diff
@@ -1331,3 +1335,39 @@ def delete_none_keys(incoming_object: typing.Dict) -> None:
         for key in list(incoming_object.keys()):
             if incoming_object[key] is None:
                 del incoming_object[key]
+
+
+def merge_dataclass_entities(
+    source: typing.Union["Project", "Folder"],
```
thomasyu888, line 1341: I can see this function being helpful for any Entities that support annotations. (File, Table, Dataset, etc). Not sure if we need to add it in right here, but just a note

docs/scripts/object_orientated_programming_poc/oop_poc_project.py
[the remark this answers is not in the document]
thomasyu888: Edit: Nevermind, i think this is just "shorthand". so you can do something like

synapse.org/genie

synapseclient/core/utils.py
thomasyu888: Do you think that there is an "annotation" mixIn here where the classes that have it can do something like sets in python and have `.update` functionality?

BryanFauble: Can you elaborate a bit more on your thoughts on this? I see what sets is doing, but I'm unclear what you mean about applying it in this use-case.

synapseclient/models/folder.py
```diff
@@ -324,6 +370,8 @@ async def copy(
         parent_id: str,
         copy_annotations: bool = True,
         exclude_types: Optional[List[str]] = None,
+        file_update_existing: bool = False,
```
thomasyu888, line 373: This makes sense in a Folder context, but you can also update existing projects, tables, and other entities.

synapseclient/models/project.py
```diff
@@ -113,25 +120,25 @@ class Project(AccessControllable, StorableContainer):
     concurrent updates. Since the E-Tag changes every time an entity is updated it
     is used to detect when a client's current representation of an entity is out-of-date."""
 
-    created_on: Optional[str] = None
-    """The date this entity was created."""
+    created_on: Optional[str] = field(default=None, compare=False)
```
thomasyu888, line 123: TIL: `compare=False` - very neat!

synapseclient/models/project.py
```diff
@@ -286,15 +384,119 @@ async def delete(self, synapse_client: Optional[Synapse] = None) -> None:
...
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Project_Copy: {self.id}"
+    )
+    async def copy(
+        self,
+        destination_id: str,
+        copy_annotations: bool = True,
+        copy_wiki: bool = True,
+        exclude_types: Optional[List[str]] = None,
+        file_update_existing: bool = False,
```
thomasyu888, line 424: Same comment as above, now that this is on a project level, there's other entities that can be updated beyond files

synapseclient/models/services/search.py
```diff
@@ -0,0 +1,84 @@
...
+        The ID of the entity.
+
+    Raises:
+        ValueError: If the entity ID or Name and parent is not set.
+        SynapseNotFoundError: If the entity is not found in Synapse.
+    """
+    can_search = (
+        entity.id
+        or (
+            entity.name and (entity.__class__.__name__ == "Project" or entity.parent_id)
```
thomasyu888, line 55: Nit: Is this where you can use `isinstance(entity, Project)`?

thomasyu888 (approved):
🔥 Nice work here!  I'm going to pre-approve, just some minor comments.

synapseclient/core/utils.py
BryanFauble: > `MyFileEntity.update(MySecondFileEntity) and it would just update the annotations.`

I don't think so in this case because we are setting annotations to be a dict. This function was built out of the need that dataclasses didn't have any nifty mechanism I could find that would do a merge.

synapseclient/models/folder.py
BryanFauble: Per this [documentation here](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseutils/copy_functions.py#L325) - It is explicit that updateExisting is only applicable in the case of a File copy. Agreed on the Table front - However, [the boolean isn't being used](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseutils/copy_functions.py#L645) during a table copy.

thomasyu888: Shows how much I remember for a function I wrote. 🙃

PR #1068, opened 2024-02-16
[SYNPY-1433] Synchronous wrappers for all external facing async methods

BryanFauble:
**Problem:**

1. When calling code in a synchronous context, you must wrap your code in `asyncio. run`, which was a poor experience.
2. Some docstrings needed fixing

**Solution:**

1. Creating an `async_to_sync` decorator that can be applied at the class level that will at runtime create a non-async version of a method. The problem with this is that because it is at runtime static type checkers do not know of its existence. To get around this issue I created a protocol class for all of models that creates the missing method definitions. Again - The problem here is that docstrings need to be duplicated for both the mkdocs site AND, hoverdocs to work correctly.
2. Updating some docstrings.

**Testing:**

1. I copied all of the existing integration and unit tests into an async and sync directory. I am running both sets of tests for both interfaces. I was able to duplicate the tests without changes - Just updating some mocks, and remove asyncio from the non-async tests.
2. I verified calling sync code is still running concurrently as expected:
image


synapseclient/core/async_utils.py
```diff
@@ -48,3 +51,73 @@ async def otel_trace_method_wrapper(self, *arg, **kwargs) -> None:
...
+                return self.fn(obj, *args, **kwds)
+            else:
+                return self.fn(cls, *args, **kwds)
+
+        functools.update_wrapper(f, self.fn)
+        return f
+
+
+# Adapted from https://github.com/keflavich/astroquery/blob/30deafc3aa057916bcdca70733cba748f1b36b64/astroquery/utils/process_asyncs.py#L11
+def async_to_sync(cls):
```
BryanFauble, line 74: The is the defining change that allows us to create the wrappers for the async functions at run-time.

synapseclient/models/protocols/access_control_protocol.py
```diff
@@ -0,0 +1,104 @@
...
+        Example: Using this function:
+            Getting permissions for a Synapse Entity
+
+                permissions = File(id="syn123").get_permissions()
+
+            Getting access types list from the Permissions object
+
+                permissions.access_types
+        """
+        return self
```
BryanFauble, line 42: I would have liked to keep a return off of this - However, whatever I tried would cause an error to pop up if it was not included.

image

docs/reference/oop/models_async.md
```diff
@@ -0,0 +1,82 @@
...
+        members:
+        - get_async
+        - store_async
+        - delete_async
+        - sync_from_synapse_async
+        - get_permissions_async
+        - get_acl_async
+        - set_permissions_async
+---
+::: synapseclient.models.File
```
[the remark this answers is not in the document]
BryanFauble: Good catch - Adding this in the latest commit

mkdocs.yml
```diff
@@ -29,19 +29,19 @@ nav:
...
-      - Versions: tutorials/python/versions.md
-      - Activity/Provenance: tutorials/python/activity.md
-      - File View: tutorials/python/file_view.md
-      - Table: tutorials/python/table.md
-      - Using a Table: tutorials/python/table_crud.md
-      - Dataset: tutorials/python/dataset.md
-      - Sharing Settings: tutorials/python/sharing_settings.md
-      - Wiki: tutorials/python/wiki.md
-      - Team: tutorials/python/team.md
+      # - Versions: tutorials/python/versions.md
```
jaymedina, line 32: Are these commented-out sections meant to be uncommented or deleted?

synapseclient/core/async_utils.py
```diff
@@ -48,3 +51,77 @@ async def otel_trace_method_wrapper(self, *arg, **kwargs) -> None:
...
+                loop = asyncio.get_running_loop()
+            except RuntimeError:
+                return asyncio.run(wrapper(*args, **kwargs))
+            else:
+                nest_asyncio.apply(loop=loop)
+                return loop.run_until_complete(wrapper(*args, **kwargs))
+
+        return newmethod
+
+    methods = cls.__dict__.keys()
```
jaymedina, line 105: Does this call only pull in the defined methods within a class or does it also pull in the methods that were inherited through something like class inheritance? Is that something that could lead to trouble down the line?

BryanFauble: I found that in the inherited mixins I needed to add the decorator to those classes as well.

synapseclient/models/file.py
```diff
@@ -68,8 +68,10 @@ class FileHandle:
         Synapse will generate this ID when the FileHandle is created."""
 
     etag: Optional[str] = None
-    """FileHandles are immutable from the perspective of the API. The only field that can
-        be change is the previewId. When a new previewId is set, the etag will change."""
+    """
+    FileHandles are immutable from the perspective of the API. The only field that can
+    be change is the previewId. When a new previewId is set, the etag will change.
```
jaymedina, line 73: _"The only field that can be changed is the `previewId`"_

Are we sure this is correct? I thought `downloadAs` was part of the file handle.

synapseclient/models/file.py
```diff
@@ -476,27 +482,27 @@ async def store(
         Example: Using this function
             File with the ID `syn123` at path `path/to/file.txt`:
 
-                file_instance = await File(id="syn123", path="path/to/file.txt").store()
+                file_instance = await File(id="syn123", path="path/to/file.txt").store_async()
```
jaymedina, line 485: Do we want to update these examples in the docstrings without the async syntax ( `await ...` and `_async()` ) since it's no longer necessary?

synapseclient/models/file.py
```diff
@@ -697,7 +705,7 @@ async def get(
         self.fill_from_dict(synapse_file=entity, set_annotations=True)
 
         if include_activity:
-            self.activity = await Activity.from_parent(
+            self.activity = await Activity.from_parent_async(
```
jaymedina, line 708: I'm confused as to why the synchronous `from_parent` isn't being used here

synapseclient/models/file.py
BryanFauble: This was taken directly from:
https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/file/FileHandle.html

synapseclient/models/file.py
BryanFauble: When we are already in an async context we don't need to run the synchronous version of the code. Only the entry point into the code should ever be calling the non-async version (What the person using the client is calling).

synapseclient/models/file.py
BryanFauble: The non-async method definitions will have the non-async examples: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f2830f87f72c9468cef8ddccbda569f06e3d012d/synapseclient/models/protocols/file_protocol.py#L51

synapseclient/models/protocols/access_control_protocol.py
```diff
@@ -0,0 +1,104 @@
...
+class AccessControllableSynchronousProtocol(Protocol):
+    """
+    The protocol for methods that are asynchronous but also
+    have a synchronous counterpart that may also be called.
+    """
+
+    def get_permissions(
+        self,
+        synapse_client: Optional[Synapse] = None,
+    ) -> "Permissions":
```
jaymedina, line 20: why is this not the Self type hint

synapseclient/models/protocols/access_control_protocol.py
```diff
@@ -0,0 +1,104 @@
...
+                Synapse users and `273949` is for public access. None implies
+                public access.
+            access_type: Type of permission to be granted. One or more of CREATE,
+                READ, DOWNLOAD, UPDATE, DELETE, CHANGE_PERMISSIONS.
+
+                **Defaults to ['READ', 'DOWNLOAD']**
+            modify_benefactor: Set as True when modifying a benefactor's ACL
+            warn_if_inherits: Set as False, when creating a new ACL. Trying to modify
+                the ACL of an Entity that inherits its ACL will result in a warning
+            overwrite: By default this function overwrites existing permissions for the
```
jaymedina, line 88: synapseclient arg is missing

synapseclient/models/protocols/activity_protocol.py
```diff
@@ -0,0 +1,90 @@
...
+
+class ActivitySynchronousProtocol(Protocol):
+    """
+    The protocol for methods that are asynchronous but also
+    have a synchronous counterpart that may also be called.
+    """
+
+    def store(
+        self,
+        parent: Optional[Union["Table", "File"]] = None,
```
jaymedina, line 20: should we not have it default to None here? have it a required arg? would storing the activity without a parent lead to an error

synapseclient/models/protocols/file_protocol.py
```diff
@@ -0,0 +1,274 @@
+"""Protocol for the specific methods of this class that have synchronous counterparts
+generatred at runtime."""
```
jaymedina, line 2:
```suggestion
generated at runtime."""
```

synapseclient/models/protocols/access_control_protocol.py
BryanFauble: Permissions is the correct type hint in this case. This is because [this is a mixin](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/a079191c7e515f741472c69f82017991d50f80f7/synapseclient/models/mixins/access_control.py) that is used in other classes (Like Project for example) where the `Self` type hint would mean `Project` instead of `Permissions`.

synapseclient/models/protocols/activity_protocol.py
BryanFauble: Not in this case. I still need to add some input validation (SYNPY-1400), but Activity can exist without a parent. Activity is it's concept where other Synapse entities (File, Table, ect...) have inbound edges pointing at the activity:

image


This means that the Activity has its own ID that can be used to store the activity with and update information.

synapseclient/models/protocols/file_protocol.py
```diff
@@ -0,0 +1,274 @@
...
+        download_as: Optional[str] = None,
+        content_type: Optional[str] = None,
+        synapse_client: Optional[Synapse] = None,
+    ) -> "File":
+        """
+        Change File Entity metadata for properties that are immutable after creation
+        through the store method.
+
+        Arguments:
+            name: Specify filename to change the filename of a file.
```
jaymedina, line 88: nit: "...change the filename of a file as seen on Synapse."

synapseclient/models/protocols/file_protocol.py
```diff
@@ -0,0 +1,274 @@
...
+            synapse_client: If not passed in or None this will use the last client from
+                the `.login()` method.
+
+        Returns:
+            The file object.
+
+        Example: Using this function
+            Can be used to change the filename, the filename when the file is downloaded, or the file content-type without downloading:
+
+                file_entity = File(id="syn123").get()
```
jaymedina, line 100: the example is described as no downloading required but `download_file` isn't set to `False`

BryanFauble: Corrected: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1068/commits/a40bfdbd525b4b35ec462c2974a91f1f508e849f

synapseclient/models/protocols/file_protocol.py
BryanFauble: Great suggestion - Added: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1068/commits/b0227f9330b9b310b9e7c9bfc55fd6a9a181ca27

synapseclient/models/protocols/activity_protocol.py
jaymedina: Interesting, okay. Thanks for the diagram!

synapseclient/models/protocols/file_protocol.py
```diff
@@ -0,0 +1,274 @@
...
+            Assuming you have a file at the path "path/to/file.txt":
+
+                file_instance = File.from_path(path="path/to/file.txt")
+        """
+        from synapseclient.models import File
+
+        return File()
+
+    def delete(self, synapse_client: Optional[Synapse] = None) -> None:
+        """Delete the file from Synapse.
```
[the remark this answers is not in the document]
BryanFauble: Great point. I updated the docstring and I added support to delete a specific file version if you pass a boolean flag into this method. See the implementation here: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1068/commits/e8105045dd27c0335760576a9b783fc9858cdd2d

synapseclient/models/protocols/folder_protocol.py
```diff
@@ -0,0 +1,132 @@
...
+    have a synchronous counterpart that may also be called.
+    """
+
+    def store(
+        self,
+        parent: Optional[Union["Folder", "Project"]] = None,
+        failure_strategy: FailureStrategy = FailureStrategy.LOG_EXCEPTION,
+        synapse_client: Optional[Synapse] = None,
+    ) -> "Folder":
+        """Storing folder and files to synapse.
```
jaymedina, line 26: The File model has its own `store` procedure in `file_protocol`, so can this be reworded to mention only Folders

BryanFauble: True - However, if you have files in the `files` attribute or folders in the `folders` it will store those as well:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/685c3641d8e1b31a6fe99da67cd1086c17c2faf1/synapseclient/models/services/storable_entity_components.py#L66

BryanFauble: I added a more explicit comment here: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1068/commits/d8ca50963ce3839d83982f30696ff69a652b8e69

tests/integration/synapseclient/models/async/test_file_async.py
```diff
@@ -791,13 +805,39 @@ async def test_delete(self, file: File) -> None:
...
+        # GIVEN a file stored in synapse
+        assert file.id is not None
+        assert file.version_number == 1
+
+        # AND I update the file
+        file.description = "new description"
+        await file.store_async()
+        assert file.version_number == 2
+
+        # WHEN I delete the file for a specific version
```
jaymedina, line 826: Suggestion to add unit tests to verify the error handling for `version_number` with no `version_only` and vice verrsa

synapseclient/models/protocols/folder_protocol.py
```diff
@@ -0,0 +1,134 @@
...
+        self,
+        parent_id: str,
+        copy_annotations: bool = True,
+        exclude_types: Optional[List[str]] = None,
+        file_update_existing: bool = False,
+        file_copy_activity: Union[str, None] = "traceback",
+        synapse_client: Optional[Synapse] = None,
+    ) -> "Folder":
+        """
+        Copy the folder to another Synpase location. This will recursively copy all
```
jaymedina, line 99:
```suggestion
        Copy the folder to another Synapse location. This will recursively copy all
```

synapseclient/models/protocols/table_protocol.py
```diff
@@ -0,0 +1,128 @@
...
+class ColumnSynchronousProtocol(Protocol):
+    """
+    The protocol for methods that are asynchronous but also
+    have a synchronous counterpart that may also be called.
+    """
+
+    def store(self, synapse_client: Optional[Synapse] = None) -> Self:
+        """Persist the column to Synapse.
+
+        :param synapse_client: If not passed in or None this will use the last client
```
jaymedina, line 31: nit: docstring needs to be updated with Google docs style

synapseclient/models/protocols/team_protocol.py
```diff
@@ -0,0 +1,139 @@
...
+            synapse_client: If not passed in or None this will use the last client
+                from the `.login()` method.
+
+        Returns:
+            None
+        """
+        return None
+
+    def get(self, synapse_client: Optional[Synapse] = None) -> "Team":
+        """Gets a Team from Synapse.
```
jaymedina, line 42: Need to specify how a Team is retrieved with this call (id and/or name) - I know it's implied under `Raises`, but just to make it a little more clear. Also, what happens with `id` and `name` are fed but one does not match the other?

synapseclient/models/protocols/user_protocol.py
```diff
@@ -0,0 +1,91 @@
...
+            synapse_client: If not passed in or None this will use the last client
+                from the `.login()` method.
+
+        Returns:
+            True if the user is certified, False otherwise.
+
+        Raises:
+            ValueError: If id nor username is specified.
+        """
+        return True
```
jaymedina, line 91: should this be `return bool()`?

tests/integration/synapseclient/models/synchronous/test_permissions.py
```diff
@@ -0,0 +1,1298 @@
...
+        project_with_permissions_through_single_team.set_permissions(
+            principal_id=p1.id, access_type=[]
+        )
+
+        # WHEN I get the permissions for the user on the entity
+        permissions = project_with_permissions_through_single_team.get_acl(
+            principal_id=p1.id
+        )
+
+        # THEN I expect to see the permissions of the team
```
jaymedina, line 138: If it's not too much trouble (and maybe this can be for another PR), I'd like to see some tests where permissions are revoked to confirm the ACL gets updated accordingly. So for this example, can we confirm that re-setting permissions such that READ and DOWNLOAD are removed from the access types reflects on the ACL?

tests/integration/synapseclient/models/synchronous/test_permissions.py
```diff
@@ -0,0 +1,1298 @@
...
+            "UPDATE",
+            "CHANGE_PERMISSIONS",
+            "CREATE",
+            "MODERATE",
+            "DOWNLOAD",
+        ]
+        assert set(expected_permissions) == set(permissions)
+
+
+class TestAclOnFolder:
```
jaymedina, line 287: Would like to see a test where the local share settings on a subfolder is changed for a user and/or team, without affecting the ACL of the parent folder. Apologies if this was already written and I've missed it

jaymedina (approved):
Nice job - I'm pre-approving since you've addressed my main comments. Thanks!

Also ran all the poc scripts with the new sync versions of the methods and can confirm they all work.

synapseclient/models/protocols/annotations_protocol.py
thomasyu888, line 1: Will we have to keep adding functions to these `Protocols`?  We will have to add good documentation when people are contributing / editing functions right?

synapseclient/models/protocols/file_protocol.py
```diff
@@ -0,0 +1,284 @@
...
+                file_entity = file_entity.change_metadata(name="my_new_name_file.txt", download_as="my_new_downloadAs_name_file.txt", content_type="text/plain")
+                print(os.path.basename(file_entity.path))  ## prints, "my_new_downloadAs_name_file.txt"
+                print(file_entity.name) ## prints, "my_new_name_file.txt"
+
+        Raises:
+            ValueError: If the file does not have an ID to change metadata.
+        """
+        return self
+
+    def get(
```
thomasyu888, line 112: So I'm assuming that the functions here are all copied from the original models?

synapseclient/models/protocols/user_protocol.py
thomasyu888, line 1: I wonder if it's worthwhile to have code that auto-creates these protocol classes or if that's even more risky than just adding them.  What I worry about is not having one source of truth for adding functions to models.

thomasyu888 (approved):
🔥 LGTM! Going to pre-approve, just some comments.

synapseclient/models/file.py
thomasyu888, line 1: So just to clarify, you can do...

```
await File(id="syn123", path="path/to/file.txt").store_async()
File(id="syn123", path="path/to/file.txt").store()
```

BryanFauble: Yes that's correct

synapseclient/models/protocols/user_protocol.py
BryanFauble: IMO The async methods are the source of truth as in the protocol classes they're generated from the async methods. The only thing we're doing is to provide a static interface so folks know how to call the method.

I agree though if we let them get out of sync we're going to have problems where the protocol says "X", but we're really expecting "Y".

synapseclient/models/protocols/annotations_protocol.py
BryanFauble: [I added notes here](https://github.com/Sage-Bionetworks/synapsePythonClient/commit/13b9d41a7076a859c8e8654618e72296f9c7769a) - Let me know what you think

synapseclient/core/async_utils.py
```diff
@@ -1,7 +1,10 @@
 """This utility class is to hold any utilities that are needed for async operations."""
 
+import asyncio
 from typing import Callable, Union
 from opentelemetry import trace
+import functools
+import nest_asyncio
```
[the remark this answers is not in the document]
BryanFauble: I added this extension and ran it on all these files: https://marketplace.visualstudio.com/items?itemName=ms-python.isort

https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1068/commits/2c8c1fba6b288b681421d794f99c6b94e2b9b3f9

synapseclient/models/protocols/user_protocol.py
BryanFauble: Both work - But I swapped over to your suggestion.

synapseclient/models/protocols/table_protocol.py
BryanFauble: I will be handling this in https://sagebionetworks.jira.com/browse/SYNPY-1418

tests/integration/synapseclient/models/async/test_file_async.py
BryanFauble: This is tested here:
version_number with no version_only:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/2c8c1fba6b288b681421d794f99c6b94e2b9b3f9/tests/integration/synapseclient/models/synchronous/test_file.py#L764

no version_number with version_only:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/2c8c1fba6b288b681421d794f99c6b94e2b9b3f9/tests/unit/synapseclient/models/synchronous/unit_test_file.py#L846

synapseclient/models/protocols/folder_protocol.py
BryanFauble: This was an issue in several spots. I fixed all of them.

tests/integration/synapseclient/models/synchronous/test_permissions.py
BryanFauble: Added more tests here: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1068/commits/32f0778dd1fc097d0fcabc7cd8c108a3ec9aab5f

PR #1069, opened 2024-02-21
Regen pipfile.lock for dependabot alert

BryanFauble:
**Problem:**

1. https://github.com/Sage-Bionetworks/synapsePythonClient/security/dependabot/10

**Solution:**

1. Generate a new pipfile.lock

**Testing:**

1. Verified that the new version is outside of the affected versions. `=42.0.4`

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -434,18 +434,18 @@
         },
         "boto3": {
             "hashes": [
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #1071, opened 2024-02-23
[SYNPY-1442] Log update to capture exception when download-list fails

BryanFauble:
**Problem:**

1. If an exception was raised during get-download-list we did not print it out anywhere (Exception Swallowing)

**Solution:**

1. use `.exception` on the logger which will capture the exception being raised

**Testing:**

1. Verified functionality in a unit test
2. Also verified by manually placing temporary code in `.get()` that the exception is logged to console:
```
Downloading  [####################]100.00%   418.0bytes/418.0bytes (227.4kB/s) manifest.csv Done...
    [ERROR] Unable to download file
  File "/home/bfauble/BryansGreatWorkspace/synapsePythonClient/synapseclient/client.py", line 2195, in get_download_list
    entity = self.get(row["ID"], downloadLocation=downloadLocation)
  File "/home/bfauble/.pyenv/versions/3.8.18/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/home/bfauble/BryansGreatWorkspace/synapsePythonClient/synapseclient/client.py", line 1098, in get
    raise SynapseError("This is a test error")
synapseclient.core.exceptions.SynapseError: This is a test error
[WARNING] A manifest was created, but no files were downloaded
Manifest file: manifest_1708712431265684147.csv
```

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2204,7 +2204,7 @@ class Synapse(object):
                 except Exception:
                     row["path"] = ""
                     row["error"] = "DOWNLOAD FAILED"
-                    self.logger.error("Unable to download file")
+                    self.logger.exception("Unable to download file")
                 writer.writerow(row)
 
         # Don't want to clear all the download list because you can add things
--- a/synapseclient/__main__.py
+++ b/synapseclient/__main__.py
@@ -4,6 +4,7 @@ The Synapse command line client.
 For a description of its usage and parameters, see its documentation:
 https://python-docs.synapse.org/build/html/CommandLineClient.html
 """
+
 import argparse
 import collections.abc
 import logging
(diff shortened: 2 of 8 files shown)


synapseclient/client.py
```diff
@@ -2204,7 +2204,7 @@ def get_download_list(self, downloadLocation: str = None) -> str:
                 except Exception:
                     row["path"] = ""
                     row["error"] = "DOWNLOAD FAILED"
-                    self.logger.error("Unable to download file")
+                    self.logger.exception("Unable to download file")
```
thomasyu888, line 2207: TIL : I did not know you could do this!

thomasyu888 (approved):
🔥 LGTM!

PR #1072, opened 2024-02-23
[SYNPY-1443] Correct which fields are merged if there is a change

BryanFauble:
**Problem:**

1. The usage of `asdict` was converting the folder/file dataclasses to a dict on the Project class - This pevented further code from calling methods on the Folder/File class.
```
  File "/home/bfauble/BryansGreatWorkspace/synapsePythonClient/synapseclient/models/project.py", line 308, in store_async
    await store_entity_components(
  File "/home/bfauble/BryansGreatWorkspace/synapsePythonClient/synapseclient/models/services/storable_entity_components.py", line 80, in store_entity_components
    folder.store_async(
AttributeError: 'dict' object has no attribute 'store_async'
```

**Solution:**

1. Only apply changes for modified items. We are not retrieving any dataclasses during the `get_async` calls [during this line](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/SYNPY-1443-store-before-get-for-containers/synapseclient/models/project.py#L288) besides `self` - So we won't run into issues where a dataclass is getting replaced by a dict.
2. If we are to retrieve dataclasses during the `get_async` calls we'll need to make a small modification to this code to ensure the merge occurs in that dataclass (recursively) and is converted back into the correct class instance. I will take ownership of this problem when we need to cross that bridge.

**Testing:**

1. Verified the broken script works as expected:
```
import synapseclient
from synapseclient.models import Project, Folder, File

syn = synapseclient.login()
file_for_demo = "/home/bfauble/temp/my_file_with_random_data.txt"
my_project_name = "My Super Cool Project"


my_project = Project(name=my_project_name)

my_folder = Folder(name="my folder")

my_file = File(path=file_for_demo)

my_folder.files.append(my_file)

my_project.folders.append(my_folder)

my_project.store()
```

1. Integration test around this logic

### DIFF
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -1355,19 +1355,20 @@ def merge_dataclass_entities(
     # Convert dataclasses to dictionaries
     destination_dict = asdict(destination)
     source_dict = asdict(source)
+    modified_items = {}
 
     # Update destination_dict with source_dict, keeping destination's values in case of conflicts
     for key, value in source_dict.items():
         if key not in destination_dict or destination_dict[key] is None:
-            destination_dict[key] = value
+            modified_items[key] = value
         elif key == "annotations":
-            destination_dict[key] = {
+            modified_items[key] = {
                 **(value or {}),
                 **destination_dict[key],
             }
 
     # Update destination's fields with the merged dictionary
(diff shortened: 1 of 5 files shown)


danlu1 (approved):
LGTM

PR #1078, opened 2024-03-14
DRAFT: [SYNPY-1420] Re-write uploads to mix async, multi-threading, and multi-processing

BryanFauble:
Problem:

1. The current upload process was slow as pieces that could be executed concurrently were executed sequentially.
2. Items like MD5 calculation are CPU bound which blocks the Python GIL from doing anything else.

Solution:

1. Part uploads are pushed to other threads
2. MD5 checksum calculations are pushed to other processes (To prevent blocking the main AsyncIO Thread)
3. All HTTP calls are using AsyncIO async/await syntax to prevent any blocking while we are waiting for network I/O

Testing:
Some initial testing figures:

10 files, 100MB/file, 1GB total:
Before: 33s
After: 16s

1 file, 10GB:
Before: 221s - Average ~73.5 MB/s
After: 164s - Average ~124 MB/s

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,6 +53,7 @@ install_requires =
     opentelemetry-sdk~=1.21.0
     opentelemetry-exporter-otlp-proto-http~=1.21.0
     nest-asyncio~=1.6.0
+    asyncio-atexit~=1.0.1
     httpx~=0.27.0
 tests_require =
     pytest>=6.0.0,<7.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -55,6 +55,7 @@ install_requires =
     nest-asyncio~=1.6.0
     asyncio-atexit~=1.0.1
     httpx~=0.27.0
+    tqdm>=4.66.2,<5.0
 tests_require =
     pytest>=6.0.0,<7.0
     pytest-mock>=3.0,<4.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,6 +53,7 @@ install_requires =
     opentelemetry-sdk~=1.21.0
     opentelemetry-exporter-otlp-proto-http~=1.21.0
     nest-asyncio~=1.6.0
+    httpx~=0.27.0
 tests_require =
     pytest>=6.0.0,<7.0
     pytest-mock>=3.0,<4.0
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -23,6 +23,7 @@ from .file_services import (
     put_file_multipart_add,
     put_file_multipart_complete,
     post_file_multipart_presigned_urls_async,
+    put_file_multipart_add_async,
 )
 
 __all__ = [
@@ -43,6 +44,7 @@ __all__ = [
     "post_external_filehandle",
     "post_file_multipart_presigned_urls",
     "post_file_multipart_presigned_urls_async",
+    "put_file_multipart_add_async",
     # entity_services
     "get_entity",
     "put_entity",
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -12,6 +12,7 @@ from .entity_services import (
     get_upload_destination_location,
     post_entity,
     put_entity,
+    create_access_requirements_if_none,
 )
 from .file_services import (
     get_file_handle,
@@ -51,4 +52,5 @@ __all__ = [
     "post_entity",
     "get_upload_destination",
     "get_upload_destination_location",
+    "create_access_requirements_if_none",
 ]
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -22,6 +22,7 @@ from .file_services import (
     post_file_multipart_presigned_urls,
     put_file_multipart_add,
     put_file_multipart_complete,
+    post_file_multipart_presigned_urls_async,
 )
 
 __all__ = [
@@ -41,6 +42,7 @@ __all__ = [
     "get_file_handle",
     "post_external_filehandle",
     "post_file_multipart_presigned_urls",
+    "post_file_multipart_presigned_urls_async",
     # entity_services
     "get_entity",
     "put_entity",
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -1,7 +1,17 @@
 # These are all of the models that are used by the Synapse client.
 from .annotations import set_annotations
+from .entity_bundle_services_v2 import (
+    get_entity_id_bundle2,
+    get_entity_id_version_bundle2,
+    post_entity_bundle2_create,
+    put_entity_id_bundle2,
+)
 
 
 __all__ = [
     "set_annotations",
+    "get_entity_id_bundle2",
+    "get_entity_id_version_bundle2",
+    "post_entity_bundle2_create",
+    "put_entity_id_bundle2",
 ]
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -6,12 +6,45 @@ from .entity_bundle_services_v2 import (
     post_entity_bundle2_create,
     put_entity_id_bundle2,
 )
+from .file_services import (
+    post_file_multipart,
+    put_file_multipart_add,
+    put_file_multipart_complete,
+    post_file_multipart_presigned_urls,
+    post_external_object_store_filehandle,
+    post_external_s3_file_handle,
+    get_file_handle,
+    post_external_filehandle,
+)
+from .entity_services import (
+    get_upload_destination,
+    get_entity,
+    put_entity,
+    post_entity,
+    get_upload_destination_location,
+)
 
 
 __all__ = [
+    # annotations
     "set_annotations",
     "get_entity_id_bundle2",
     "get_entity_id_version_bundle2",
     "post_entity_bundle2_create",
     "put_entity_id_bundle2",
+    # file_services
+    "post_file_multipart",
+    "put_file_multipart_add",
+    "put_file_multipart_complete",
+    "post_file_multipart_presigned_urls",
+    "post_external_object_store_filehandle",
+    "post_external_s3_file_handle",
+    "get_file_handle",
+    "post_external_filehandle",
+    # entity_services
+    "get_entity",
+    "put_entity",
+    "post_entity",
+    "get_upload_destination",
+    "get_upload_destination_location",
 ]
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -6,24 +6,23 @@ from .entity_bundle_services_v2 import (
     post_entity_bundle2_create,
     put_entity_id_bundle2,
 )
+from .entity_services import (
+    get_entity,
+    get_upload_destination,
+    get_upload_destination_location,
+    post_entity,
+    put_entity,
+)
 from .file_services import (
-    post_file_multipart,
-    put_file_multipart_add,
-    put_file_multipart_complete,
-    post_file_multipart_presigned_urls,
-    post_external_object_store_filehandle,
-    post_external_s3_file_handle,
     get_file_handle,
     post_external_filehandle,
+    post_external_object_store_filehandle,
+    post_external_s3_file_handle,
+    post_file_multipart,
+    post_file_multipart_presigned_urls,
+    put_file_multipart_add,
+    put_file_multipart_complete,
 )
-from .entity_services import (
-    get_upload_destination,
-    get_entity,
-    put_entity,
-    post_entity,
-    get_upload_destination_location,
-)
-
 
 __all__ = [
     # annotations
@@ -41,6 +40,7 @@ __all__ = [
     "post_external_s3_file_handle",
     "get_file_handle",
     "post_external_filehandle",
+    "post_file_multipart_presigned_urls",
     # entity_services
     "get_entity",
     "put_entity",
--- a/synapseclient/api/entity_bundle_services_v2.py
+++ b/synapseclient/api/entity_bundle_services_v2.py
@@ -1,0 +1,136 @@
+"""This module is responsible for exposing the services defined at:
+<https://rest-docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.EntityBundleV2Controller>
+"""
+
+import json
+from typing import Any, Dict, Optional
+
+from synapseclient import Synapse
+
+
+async def get_entity_id_bundle2(
+    entity_id: str,
+    request: Optional[Dict[str, bool]] = None,
+    synapse_client: Optional[Synapse] = None,
+) -> Dict[str, Any]:
+    """
+    Arguments:
+        entity_id: The ID of the entity to which the bundle belongs
+        request: The request for the bundle matching
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/entitybundle/v2/EntityBundleRequest.html>.
+            If not passed in or None, the default request will be used:
+
+            - includeEntity: True
+            - includeAnnotations: True
+            - includeFileHandles: True
+            - includeRestrictionInformation: True
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+    Returns:
+        The requested entity bundle matching
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/entitybundle/v2/EntityBundle.html>
+    """
+    if not request:
+        request = {
+            "includeEntity": True,
+            "includeAnnotations": True,
+            "includeFileHandles": True,
+            "includeRestrictionInformation": True,
+        }
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_post_async(
+        uri=f"/entity/{entity_id}/bundle2",
+        body=json.dumps(request),
+    )
+
+
+async def get_entity_id_version_bundle2(
+    entity_id: str,
+    version: int,
+    request: Optional[Dict[str, bool]] = None,
+    synapse_client: Optional[Synapse] = None,
+) -> Dict[str, Any]:
+    """
+    Arguments:
+        entity_id: The ID of the entity to which the bundle belongs
+        version: The version of the entity to which the bundle belongs
+        request: The request for the bundle matching
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/entitybundle/v2/EntityBundleRequest.html>.
+            If not passed in or None, the default request will be used:
+
+            - includeEntity: True
+            - includeAnnotations: True
+            - includeFileHandles: True
+            - includeRestrictionInformation: True
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+    Returns:
+        The requested entity bundle matching
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/entitybundle/v2/EntityBundle.html>
+    """
+    if not request:
+        request = {
+            "includeEntity": True,
+            "includeAnnotations": True,
+            "includeFileHandles": True,
+            "includeRestrictionInformation": True,
+        }
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_post_async(
+        uri=f"/entity/{entity_id}/version/{version}/bundle2",
+        body=json.dumps(request),
+    )
+
+
+async def post_entity_bundle2_create(
+    request: Dict[str, Any],
+    generated_by: Optional[str] = None,
+    synapse_client: Optional[Synapse] = None,
+) -> Dict[str, Any]:
+    """
+    Arguments:
+        request: The request for the bundle matching
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/entitybundle/v2/EntityBundleCreate.html>
+        generated_by: The ID of the activity to associate with the entity.
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+    Returns:
+        The requested entity bundle matching
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/entitybundle/v2/EntityBundle.html>
+    """
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_post_async(
+        uri="/entity/bundle2/create"
+        + (f"?generatedBy={generated_by}" if generated_by else ""),
+        body=json.dumps(request),
+    )
+
+
+async def put_entity_id_bundle2(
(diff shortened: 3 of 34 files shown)


synapseclient/api/__init__.py
```diff
@@ -6,12 +6,50 @@
     post_entity_bundle2_create,
     put_entity_id_bundle2,
 )
-
+from .entity_services import (
```
BryanFauble, line 9: These API services are just splitting out functionality in the `client.py` around the various REST APIs. Since this is going to be replaced with the OpenAPI Generated client I am going to focus on integration testing higher level pieces than these individual level functions.

synapseclient/core/async_utils.py
```diff
@@ -93,14 +93,19 @@ async def wrapper(*args, **kwargs):
                 """Wrapper for the function to be called in an async context."""
                 return await getattr(self, async_method_name)(*args, **kwargs)
 
+            loop = None
```
BryanFauble, line 96: This change cleans up the stack trace logs a little bit.

synapseclient/core/retry.py
```diff
@@ -340,7 +341,130 @@ async def foo(a, b, c): return [a, b, c]
...
-                    else "Rasing the exception: %s"
+                    else "Raising the exception: %s"
+                ),
+                str(caught_exception_info[0]),
+            )
+            raise caught_exception
+        return response
+
+
+def with_retry_time_based(
```
BryanFauble, line 352: Nearly a 1-to-1 copy from above. But without async/await. And using the time library to wait.

synapseclient/core/upload/multipart_upload.py
```diff
@@ -412,28 +420,6 @@ def __call__(self):
         return upload_status_response
 
 
-def _get_file_chunk(file_path, part_number, chunk_size):
```
BryanFauble, line 415: I pulled out common functionality out of the previous upload scripts that could be re-used. I placed them into `upload_utils.py`

synapseclient/models/file.py
```diff
@@ -516,40 +625,82 @@ async def store_async(
                 "The file must have an (ID with a (path or `data_file_handle_id`)), or a "
                 "(path with a (`parent_id` or parent with an id)) to store."
             )
+        self.parent_id = parent.id if parent else self.parent_id
```
BryanFauble, line 628: The changes in this `store_async` function serve a few purposes:

1. Retrieve the current entity data from Synapse and determine if we need to PUT new data onto Synapse
2. Determine based on the file Synapse current has if an upload needs to occur
3. Swapping over to an async/HTTPX interface for storing all of this data

synapseclient/models/services/storable_entity.py
```diff
@@ -0,0 +1,103 @@
...
+)
+from synapseclient.core.utils import get_properties
+
+if TYPE_CHECKING:
+    from synapseclient.models import File, Folder, Project
+
+tracer = trace.get_tracer("synapseclient")
+
+
+async def store_entity(
```
BryanFauble, line 21: I intend for this to be replaced when https://sagebionetworks.jira.com/browse/SYNPY-1437 - can be worked on (Currently blocked by a few PLFM tickets)

synapseclient/api/entity_services.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(
+        uri=f"/entity/{entity_id}",
+    )
+
+
+@alru_cache(ttl=60)
```
BryanFauble, line 97: This was a nice performance gain when looking at cases where multiple files are getting stored in a container. This package works in such a way that if multiple things call this co-routine it is only executed the FIRST time it is called - The rest of the calls await the result.

In the 1000 Files, 1 GiB total size test case this had around a 25% time improvement.

https://github.com/aio-libs/async-lru

thomasyu888: Just out of curiosity, what's the issue when using the built in lru_cache?

synapseclient/api/entity_services.py
```diff
@@ -0,0 +1,173 @@
...
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    existing_restrictions = await client.rest_get_async(
+        f"/entity/{entity_id}/accessRequirement?offset=0&limit=1"
+    )
+    if len(existing_restrictions["results"]) <= 0:
```
thomasyu888, line 164: Can this ever be negative?

Nit: you can do this - `if not existing_restrictions["results"]`, but it's fine to be explicit as well.

synapseclient/api/entity_services.py
```diff
@@ -0,0 +1,173 @@
...
+        access_requirements = await client.rest_post_async(
```
thomasyu888, line 165: It would be nice if we were given the ability to use these endpoints but not get all the emails the ACL team gets.

synapseclient/core/async_utils.py
```diff
@@ -71,6 +73,28 @@ def f(*args, **kwds):
         return f
 
 
+def wrap_async_to_sync(coroutine, syn: "Synapse"):
```
thomasyu888, line 76: How difficult is it to type the coroutine?

synapseclient/core/remote_file_storage_wrappers.py
```diff
@@ -276,14 +292,27 @@ def upload_file(
         Returns:
             The URL of the uploaded file.
         """
+        progress_bar = tqdm(
```
thomasyu888, line 295: This is awesome - will we be able to remove the old maintained function?

synapseclient/core/retry.py
```diff
@@ -340,7 +374,130 @@ async def foo(a, b, c): return [a, b, c]
...
+    function,
```
thomasyu888, line 386: Nit: Have you looked at the backoff package?

BryanFauble: I had not - Reading over the docs it does handle for most if not all the use-cases here. I won't make changes since this is working, but it may be a candidate to swap to if this does not work out.

thomasyu888: This might be a good future project!

synapseclient/core/async_utils.py
BryanFauble: Not difficult - I'll add a Type hint:
`: Coroutine[Any, Any, Any]`

Since these are generic functions that take in any coroutine the type hints are less useful, but at least typed to a `Coroutine`

synapseclient/api/entity_services.py
BryanFauble: Can you clarify what you mean? Are you suggesting a change to this [REST API](https://rest-docs.synapse.org/rest/POST/entity/id/lockAccessRequirement.html)?

synapseclient/api/entity_services.py
BryanFauble: I hadn't stopped to consider this line. I took this from the `client.py` method:
```
    @tracer.start_as_current_span("Synapse::_createAccessRequirementIfNone")
    def _createAccessRequirementIfNone(self, entity: Union[Entity, str]) -> None:
        """
        Checks to see if the given entity has access requirements. If not, then one is added

        Arguments:
            entity: A Synapse ID or a Synapse Entity object
        """
        existingRestrictions = self.restGET(
            "/entity/%s/accessRequirement?offset=0&limit=1" % id_of(entity)
        )
        if len(existingRestrictions["results"]) <= 0:
            self.restPOST("/entity/%s/lockAccessRequirement" % id_of(entity), body="")
```

Looking at it the logic I can probably change this to:
`if existing_restrictions is None or not hasattr(existing_restrictions, "results") or len(existing_restrictions["results"]) == 0:`

synapseclient/api/entity_services.py
BryanFauble: The problem with lru_cache is the way that it works doesn't work for asyncio. What I mean by this:

Suppose we have 10 coroutines running to save 10 files. Each of those 10 files needs to determine where the file is going to be uploaded and calls this `get_upload_destination()` method. lru_cache works in this way:

1. If a result is already present in the cache return the result to the caller
2. If a result is not present in the cache enter the function and cache the result
In the case of asyncio - all 10 coroutines are going to enter this `get_upload_destination` because the result hasn't yet returned and been cached.

ALRU cache works by:

1. If a result is already present in the cache return the result to the caller
2. If a result is not present in the cache and a future already exists that promises the result - return that future to the caller (This means all 10 coroutines will wait on a single Future to get the result)
3. If a result is not present and no future is already retrieving the result - enter the function, create a future that promises the result, execute, and cache the result.

synapseclient/api/entity_bundle_services_v2.py
```diff
@@ -3,15 +3,16 @@
...
 import json
-from typing import Any, Dict, Optional
+from typing import Any, Dict, Optional, TYPE_CHECKING
 
-from synapseclient import Synapse
+if TYPE_CHECKING:
+    from synapseclient import Synapse
 
 
 async def get_entity_id_bundle2(
```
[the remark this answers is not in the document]
BryanFauble: Agreeded

synapseclient/api/annotations.py
```diff
@@ -2,21 +2,23 @@
 The purpose of this module is to provide any functions that are needed to interact with
 annotations that are not cleanly provided by the synapseclient library.
 """
+
 import json
 
 from dataclasses import asdict
 
 from typing import TYPE_CHECKING, Optional
-from synapseclient import Synapse
```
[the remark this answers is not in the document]
BryanFauble: Yes - This worked right away, but if you have other suggestions to make this cleaner I am all for making a change. This didn't "feel" like the best solution but it wasn't that important to spend more time on atm.

setup.cfg
```diff
@@ -53,7 +53,12 @@ install_requires =
     opentelemetry-sdk~=1.21.0
     opentelemetry-exporter-otlp-proto-http~=1.21.0
     nest-asyncio~=1.6.0
+    asyncio-atexit~=1.0.1
     httpx~=0.27.0
+    tqdm>=4.66.2,<5.0
+    loky~=3.0.0
```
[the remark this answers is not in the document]
BryanFauble: https://github.com/minrk/asyncio-atexit is the only one that hasn't been worked on in a while - But it's small enough (a single 122 line file) that even if I were to need to apply a bug fix it wouldn't be a problem to do so.

docs/scripts/uploadBenchmark.py
```diff
@@ -60,34 +76,48 @@ def create_folder_structure(
...
+        The total number of directories, total number of files, and the size of each
+        file in bytes
     """
     # Calculate total number of files and size of each file
     total_dirs = sum(
         [num_sub_directories**i for i in range(1, depth_of_directory_tree + 1)]
     )
     total_files = total_dirs * num_files_per_directory
-    total_size_of_files_bytes = total_size_of_files_mbytes * 1024 * 1024
+    total_size_of_files_bytes = total_size_of_files_mib * MiB
```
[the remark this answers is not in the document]
BryanFauble: Good catch - I missed this when I reverted some lines:
https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1078/commits/a0118b7dcda2724986aa9457f02ab8479e94d47e

docs/explanations/benchmarking.md
```diff
@@ -1,9 +1,50 @@
...
+| 100 File/100 MiB ea | 10GiB               | 124.14s              | 248.57s                 | 150.46s     | 320.50s                 | 227.75s      | 170.82s                   | 294.69s                              |
+| 10 File/100 MiB ea  | 1GiB                | 15.13s               | 28.38s                  | 18.14s      | 33.74s                  | 26.64s       | 17.69s                    | 32.80s                               |
+| 100 File/10 MiB ea  | 1GiB                | 19.24s               | 141.23s                 | 19.59s      | 139.31s                 | 48.50s       | 18.34s                    | 138.14s                              |
+| 1000 File/1 MiB ea  | 1GiB                | 152.65s              | 1044.42s                | 25.03s      | 1101.90s                | 340.07s      | 100.94s                   | 1106.40s                             |
+
+#### A high level overview of the differences between each of the upload methods:
+- **OOP Models Interface:** Uploads all files and 8MB chunks of each file in parallel using a new upload algorithm
+- **Synapseutils:** Uploads all files in parallel and 8MB chunks of each file in parallel using the old upload algorithm
```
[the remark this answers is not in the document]
BryanFauble: At the moment - No. When it is replaced I'll be bench marking this again at that point.

docs/explanations/benchmarking.md
```diff
@@ -1,9 +1,50 @@
...
+data driven decisions on where to spend time optimizing the client. Additionally, it
+will give us a way to measure the impact of changes to the client.
 
 ## Results
 
+### 04/01/2024: Uploading files to Synapse
+These benchmarking results bring together some important updates to the Upload logic. It
+has been re-written to bring a focus to concurrent file uploads and more effecient use
+of available threads. As a result of this change it is not reccommended to
+increase `max_threads` manually. Based on the available CPU cores this python package
```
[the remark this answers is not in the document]
BryanFauble: For downloads it'll continue to follow the previous recommendations:
```
During this download test I tried various thread counts to see what performance looked like at different levels. What I found was that going over the default count of threads during download of large files (10GB and over) led to significantly unstable performance. The client would often crash or hang during execution. As a result the general recommendation is as follows:

- For files over 1GB use the default number of threads: `multiprocessing.cpu_count() + 4`
- For a large number of files 1GB and under 40-50 threads worked best
```

docs/explanations/benchmarking.md
```diff
@@ -1,9 +1,50 @@
...
+
+The results were created on a `t3a.micro` EC2 instance with a 200GB disk size running
+in us-east-1. The script that was run can be found in `docs/scripts/uploadBenchmark.py`.
+
+
+#### Some insights:
+- The use of the
+[Object-Orientated interfaces/Models Interface](../reference/oop/models.md) results in
+the best performance out of the box and will scale based on the hardware it's run on.
+- Increasing the number of threads did increase performance for the previous upload
```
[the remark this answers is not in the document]
BryanFauble: I felt like the full piece gave a decent disclaimer:
```
- Increasing the number of threads did increase performance for the previous upload
logic and Synapseutils functionality. It may also increase performance for new uploads,
however, it was not tested again. Increasing the number of threads in use also has
inconsistent stability as found in previous benchmarks.
```

In particular:
```
Increasing the number of threads in use also has inconsistent stability as found in previous benchmarks.
```

This is a "change at your own risk" kind of thing.

synapseclient/api/entity_services.py
thomasyu888: So currently only members of a Synapse ACL Team can execute these endpoints.  At least to my knowledge, our user accounts don't have permission to execute these.

thomasyu888: Interesting!  I do know that there are specific endpoints that only ACT team members can execute - maybe that's not one of them.  Sorry about that!

synapseclient/client.py
thomasyu888, line 1: I'm starting to like all the "client" functionality being built into this class.  I can slowly start to see the other functions within this class transfer out of here.

synapseclient/client.py
```diff
@@ -349,8 +362,133 @@ def __init__(
 
         transfer_config = self._get_transfer_config()
         self.max_threads = transfer_config["max_threads"]
+        self._thread_executor = {}
+        self._process_executor = {}
+        self._md5_semaphore = {}
```
thomasyu888, line 367: Do we imagine the need for other semaphores?

synapseclient/client.py
```diff
@@ -1543,6 +1691,11 @@ def store(
                             You will be contacted with regards to the specific data being restricted and the
                             requirements of access.
             set_annotations: If True, set the annotations on the entity. If False, do not set the annotations.
+            async_file_handle_upload: Temporary feature flag that will be removed at an
```
thomasyu888, line 1694: Nit: not to complicate things, but is it worth doing a `_async...` parameter?

synapseclient/core/remote_file_storage_wrappers.py
```diff
@@ -35,7 +37,7 @@ def _attempt_import_boto3():
 
     @staticmethod
     def _create_progress_callback_func(
-        file_size: int, filename: str, prefix: str = None
+        progress_bar: tqdm,
```
thomasyu888, line 40: Amazing!  I can't wait to start seeing the cutting down of code

synapseclient/core/upload/multipart_upload_async.py
```diff
@@ -0,0 +1,884 @@
...
+            return refreshed_url
+
+    async def _handle_part_wrapper(self, part_number: int) -> HandlePartResult:
+        loop = asyncio.get_running_loop()
+        otel_context = context.get_current()
+
+        mem_info = psutil.virtual_memory()
+
+        if mem_info.available <= self._part_size * 2:
+            gc.collect()
```
thomasyu888, line 306: I'm surprised garbage collection wasn't part of the description in the PR.  Can you add a description as to why you did this?

synapseclient/core/utils.py
```diff
@@ -65,19 +67,77 @@ def md5_for_file(
...
         while True:
+            loop_iteration += 1
             if callback:
                 callback()
             data = f.read(block_size)
             if not data:
                 break
             md5.update(data)
+            del data
+            # Garbage collect every 100 iterations
```
thomasyu888, line 82: What was the rationale for 100 here?

synapseclient/core/upload/multipart_upload_async.py
BryanFauble: I'm not sure if I really like having this logic in the code and I might get rid of it or change it to only look at the lowest generation for GC.

The problem I am solving:
Each part (8MB) of a file is loaded into memory before it's sent to the storage provider. In testing on a 1GB EC2 instance, I kept killing the machine for out-of-memory issues. What I believe the issue was is that there was not enough items (700 items, 25 items, 10 items IIRC for each generation level) to cause a the scheduled garbage collection.

When I had tested this logic it seemed to have worked - though it was all very inconsistent.

thomasyu888 (approved):
🔥 Excellent work here!  I just had some comments here and there, and I didn't take a super hard look at the tests.  Could someone else @Sage-Bionetworks/dpe take a look through these?

synapseclient/client.py
BryanFauble: Not for this. I intend that this is removed with https://sagebionetworks.jira.com/browse/SYNPY-1356 - Which is not that far away.

synapseclient/client.py
BryanFauble: Perhaps around things like querying for table data later on, though I'm not certain.

BWMac (approved):
I took a close look at the tests and they look great to me! I still love the "GIVEN/WHEN/THEN" format. Makes it so easy to read through the testing logic.

synapseclient/api/file_services.py
```diff
@@ -0,0 +1,335 @@
...
+    part_number: int,
+    md5_hex: str,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """
+    <https://rest-docs.synapse.org/rest/PUT/file/multipart/uploadId/add/partNumber.html>
+
+    Arguments:
+        upload_id: The unique identifier of the file upload.
+        part_number: The part number to add. Must be a number between 1 and 10,000.
```
jaymedina, line 62: Do you think we should enforce this limit (1-10,000) in the code itself?

synapseclient/api/file_services.py
```diff
@@ -0,0 +1,335 @@
...
+            the `.login()` method.
+
+    Returns:
+        A FileHandle for objects that are stored externally.
+        <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/file/ExternalFileHandleInterface.html>
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    file_name = external_url.split("/")[-1]
```
jaymedina, line 217: You can use `os.path.basename(external_url)` for this.

image

synapseclient/api/file_services.py
```diff
@@ -0,0 +1,335 @@
...
+    Raises:
+        ValueError: If neither parent nor storage_location_id is specified, or if
+            both are specified.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    if storage_location_id:
+        if parent:
```
jaymedina, line 278: why not `if storage_location_id & parent`?

synapseclient/api/file_services.py
```diff
@@ -0,0 +1,335 @@
...
+            raise ValueError("Pass parent or storage_location_id, not both")
+    elif not parent:
+        raise ValueError("One of parent or storage_location_id is required")
+    else:
+        upload_destination = await get_upload_destination(
```
jaymedina, line 283: I guess not related to this PR, but was this func meant to be named `get_upload_destination_async`?

synapseclient/core/remote_file_storage_wrappers.py
```diff
@@ -127,21 +118,30 @@ def download_file(
...
 
             s3_obj.download_file(
                 download_file_path,
                 Callback=progress_callback,
                 Config=transfer_config,
             )
+            if progress_bar:
+                progress_bar.close()
 
             # why return what we were passed...?
```
jaymedina, line 146: Has this been addressed?

synapseclient/api/file_services.py
BryanFauble: We do not want to fall into the else block. `if storage_location_id & parent` means if they supplied a storage ID, but not a parent (What they should do) we'll overwrite the location they told us to upload to.

synapseclient/api/file_services.py
BryanFauble: I am not planning to include this for all the api functions. They're geared towards internal use and I am not providing both a sync and async call for them.

jaymedina (approved):

synapseclient/core/remote_file_storage_wrappers.py
BryanFauble: I removed the comment

synapseclient/api/file_services.py
BryanFauble: At this point I do not think it's needed - But I do think a larger-scale input validation is going to be important in the Open API generated client.

PR #1079, opened 2024-04-02
[SYNPY-1420] Clean up otel spans

BryanFauble:
**Problem:**

1. OTEL span were a bit all over the place with varying depth and information

**Solution:**

1. Focus OTEL spans on the I/O portions of the application to know when requests are start/stopping or long running MD5 calculations are taking place

**Testing:**

1. Verified by running the upload benchmark script and checking a Jaeger container deployed on my local:
image

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -134,6 +134,8 @@ from synapseclient.core.async_utils import wrap_async_to_sync
 from typing import Any, Union, Dict, List, Optional, Tuple
 from synapseclient.core.models.permission import Permissions
 from opentelemetry import trace
+from opentelemetry.trace import SpanKind
+
 
 tracer = trace.get_tracer("synapseclient")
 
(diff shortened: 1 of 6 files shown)


synapseclient/client.py
```diff
@@ -310,9 +312,40 @@ def __init__(
...
+            """
+            span = span_dict.pop(response.request)
+            span.set_attribute("http.response.status_code", response.status_code)
+            span.end()
+
+        event_hooks = {"request": [log_request], "response": [log_response]}
         httpx_timeout = httpx.Timeout(70)
         self._requests_session_storage = requests_session_storage or httpx.Client(
-            timeout=httpx_timeout
+            timeout=httpx_timeout, event_hooks=event_hooks
```
thomasyu888, line 348: Well, this is convenient.  Are we no longer using `requests`?

synapseclient/core/utils.py
```diff
@@ -126,8 +126,8 @@ async def md5_for_file_multiprocessing(
     Returns:
         The MD5 Checksum
     """
-    with tracer.start_as_current_span("Utils::md5_for_file_multiprocessing"):
-        async with md5_semaphore:
+    async with md5_semaphore:
+        with tracer.start_as_current_span("Utils::md5_for_file_multiprocessing"):
```
thomasyu888, line 130: Did you flip these intentionally?

synapseclient/client.py
BryanFauble: > Are we no longer using requests?

In all new code we're going to be using the httpx client. I cannot replace the requests library yet (For the previous code) because it would be a breaking change if someone were passing in their own client into the `Synapse()` class initializer.

> Basically, how I'm viewing this is that tracers are only provided by the httpx calls correct?

That's what I am moving towards - Yes. The caveat however is that I want a span for the overall Synapse entity doing actions. So a span is started for a Project, Folder, File, Table, Team ect.... that is being saved. Then inside of that span all network calls will have their span (So we can see start/end times) and I/O work like MD5 calculation will have a span.

image

synapseclient/core/utils.py
BryanFauble: Yes, this is intentional. For MD5 calculation I only have 1 additional Python process that gets started up. I found some instability when I let the multi-processing library handle waiting on and scheduling all of these files - Which is why you see the `md5_semaphore` here. By placing the tracer within the `async with md5_semaphore` it means that the trace is only started when this file has gotten the semaphore. I am not capturing the time spent waiting to acquire the semaphore.

thomasyu888 (approved):
🔥 LGTM!  If someone else from @Sage-Bionetworks/dpe wants to take a look before merge, lets do so before EOD Friday :)

PR #1080, opened 2024-04-02
[SYNPY-1453] Guard around modified time item in cache

BryanFauble:
**Problem:**

1. In the cache if a cache value was a dictionary but it did not contain the exact key of `modified_time` the code would return back the parent dict - causing further code to fail that was expecting a string to be returned.

**Solution:**

1. Adding a guard statement in `_get_cache_modified_time` that verified that the data returned is going to be a string or the value for `modified_time`

**Testing:**

Added unit tests around the logic that verify the behavior if the cache map:

1. Is a dict but does not contain `modified_time` key
2. Is a dict but contains neither of the expected keys
3. The entry is an empty dict
4. The entry is None

### DIFF
--- a/synapseclient/core/cache.py
+++ b/synapseclient/core/cache.py
@@ -153,6 +153,9 @@ class Cache:
         then it is a new entry and we can return the value. If it does not, then it
         is an old entry and we should return the `cache_map_entry` itself.
 
+        The caveat is that `cache_map_entry` needs to be a string to return the value
+        otherwise it will return None.
+
         Arguments:
             cache_map_entry: The entry from the cache map
 
@@ -161,8 +164,9 @@ class Cache:
         """
         if cache_map_entry is not None and "modified_time" in cache_map_entry:
             return cache_map_entry.get("modified_time", None)
-        else:
+        elif cache_map_entry is not None and isinstance(cache_map_entry, str):
             return cache_map_entry
+        return None
 
     def _get_cache_content_md5(
         self, cache_map_entry: typing.Union[str, dict, None]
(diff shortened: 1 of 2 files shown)


synapseclient/core/cache.py
```diff
@@ -161,8 +164,9 @@ def _get_cache_modified_time(
         """
         if cache_map_entry is not None and "modified_time" in cache_map_entry:
             return cache_map_entry.get("modified_time", None)
-        else:
+        elif cache_map_entry is not None and isinstance(cache_map_entry, str):
```
thomasyu888, line 167: I understand how this scenario occurs, but I'm confused as to how the cache map would get into that state. I know the old cache map looked something like this, but I thought we had tested downloading files between different cache map states.

```
{"/Users/tyu/.synapseCache/115/134016115/SYNAPSE_TABLE_QUERY_134016115.csv": "2024-03-07T06:29:19.000Z"}
```

thomasyu888 (commented):
Thanks for the quick bug fix, just a comment

synapseclient/core/cache.py
BryanFauble: I look around the old and new code and I couldn't really find an explanation. The only case I'm thinking is if the cache map entry looked like:
```
{"/Users/tyu/.synapseCache/115/134016115/SYNAPSE_TABLE_QUERY_134016115.csv": {}}
```
or
```
{"/Users/tyu/.synapseCache/115/134016115/SYNAPSE_TABLE_QUERY_134016115.csv": {"someOtherValue": 123}}
```
I compared the previous code from [3.1.0](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/v3.1.0/synapseclient/core/cache.py) to the code in the file now and I do not see anywhere we are intentionally setting the value to an empty dict, or a dict that does not contain the expected keys.

As for testing I had verified that going forward on client versions handled this gracefully.

thomasyu888 (approved):
🔥 LGTM!

PR #1083, opened 2024-04-10
[SYNPY-1356] Refactor sync to synapse

BryanFauble:
**Problem:**

1. The syncToSynapse function provided in the utility package had it's own multi-threaded implementation to allow the upload of files in parallel. Using this it had a shared thread pool executor with the file part upload as well.

**Solution:**

1. This PR contains the changes that were independently reviewed in these 2 PRs:
2. https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1084
3. https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1089
4. Conversion of the multi-threaded portion of the syncToSynapse code to instead rely on asyncIO tasks/futures - Meaning all of the code is going to be executed in a single thread and only down the line will parts of files be uploaded in parallel on different threads.

**Testing:**

1. Planning to use the benchmark upload script in addition to integration testing around this logic.


synapseutils/sync.py
```diff
@@ -589,188 +594,138 @@ def _order_items(items):
...
-                                or abort_event.is_set()
-                            )
-                        )
-
-                pending_provenance.reset_count()
-
-            ordered_items = skipped_items
+                initial_items_to_pop.append(key)
+        for key in initial_items_to_pop:
+            ordered_items.pop(key)
```
BWMac, line 634: Why not just remove the dictionary items directly at the end of the `if not value:` block?

synapseutils/sync.py
```diff
@@ -589,188 +594,138 @@ def _order_items(items):
...
+                    stored_file = completed_task.result()
+                    finished_items[stored_file.path] = stored_file
+                    ordered_items.pop(stored_file.path, None)
+
+                    for key, activity_dependencies in ordered_items.items():
+                        if activity_dependencies and all(
+                            activity_dependency in finished_items
+                            if os.path.isfile(activity_dependency)
+                            else True
+                            for activity_dependency in activity_dependencies
```
BWMac, line 652: It might be beneficial to break up this block if possible for readability

synapseutils/sync.py
```diff
@@ -589,188 +594,138 @@ def _order_items(items):
...
+                try:
+                    stored_file = completed_task.result()
+                    finished_items[stored_file.path] = stored_file
+                    ordered_items.pop(stored_file.path, None)
+
+                    for key, activity_dependencies in ordered_items.items():
+                        if activity_dependencies and all(
+                            activity_dependency in finished_items
+                            if os.path.isfile(activity_dependency)
+                            else True
```
BWMac, line 651: Am I reading this wrong or would this evaluate to `True` no matter what?

synapseutils/sync.py
```diff
@@ -589,188 +594,138 @@ def _order_items(items):
...
+                if is_url(used_item):
+                    if used_item in used:
+                        used_activity.append(UsedURL(url=used_item))
+                    else:
+                        executed_activity.append(UsedURL(url=used_item))
+
+                # -- Synapse Entity ID (assuming the string is an ID)
+                elif isinstance(used_item, str):
+                    if not is_synapse_id_str(used_item):
+                        raise ValueError("%s is not a valid Synapse id" % used_item)
```
BWMac, line 694:
```suggestion
                        raise ValueError(f"{used_item} is not a valid Synapse id")
```

synapseutils/sync.py
```diff
@@ -589,188 +594,138 @@ def _order_items(items):
...
+                    synid, version = get_synid_and_version(
+                        used_item
+                    )  # Handle synapseIds of from syn234.4
+                    targetVersion = None
```
BWMac, line 698: NIT: snake case

synapseutils/sync.py
BryanFauble: 100% agree

synapseutils/sync.py
BryanFauble: Not quite. In an activity you have either a `URL`, a `syn123` ID, or a reference to a file on disk.

In this case the check is attempting to verify:

For every dependencies that you have:

If it's a file -> Make sure that it has already been saved
If it is not a file -> Pass over the dependency as we don't have anything it could be waiting on

This is the important piece how I am determining when I can store a node from the Directed Acyclic Graph:
https://www.geeksforgeeks.org/topological-sorting/

I want to modify the code in `_order_items` to instead give back an order that is a bit easier than this logic here. The difficult part the code here solved was to make sure that all outbound edge nodes were saved before the starting node is saved.
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/SYNPY-1356-refactor-sync-to-synapse/synapseutils/sync.py#L571

synapseutils/sync.py
BryanFauble: Updated in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1083/commits/699c015fd8eef415e13b95a39e434f1ef8d6a323

synapseutils/sync.py
```diff
@@ -558,14 +560,17 @@ def __init__(
         """
         self._syn = syn
 
-        max_concurrent_file_transfers = max(
-            int(max_concurrent_file_transfers or self._syn.max_threads / 2), 1
+        # TODO: Extract this to a shared location that is further up the stack
+        self._max_concurrent_file_transfers = max(
+            int(max_concurrent_file_transfers or self._syn.max_threads), 1
```
jaymedina, line 565: why not `max_threads * 2` as it is [here](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1084)

BryanFauble: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1084

Will be getting merged into this draft

synapseclient/models/file.py
```diff
@@ -814,6 +816,7 @@ async def change_metadata_async(
         """
         if not self.id:
             raise ValueError("The file must have an ID to change metadata.")
+        from synapseutils.copy_functions import changeFileMetaData
```
thomasyu888, line 819: Should these just be internal client functions instead of synapseutils?

synapseclient/core/upload/multipart_upload_async.py
```diff
@@ -126,6 +127,22 @@
 DEFAULT_PART_SIZE = 8 * MB
 MAX_RETRIES = 7
 
+_thread_local = threading.local()
+
+
+@contextmanager
+def shared_progress_bar(progress_bar):
```
thomasyu888, line 134: Will this only be used in the multi part upload?

thomasyu888 (approved):
🔥 Nice!

synapseclient/core/upload/multipart_upload_async.py
BryanFauble: For now - Yes. I may extract this with changing how downloads are working.

PR #1084, opened 2024-04-10
[SYNPY-1356] Moving file semaphore under file store

BryanFauble:
**Problem:**

1. The sync util function had a mechanism to limit the number of files that can enter the store process.

**Solution:**

1. Adding a semaphore to the client.
2. Moving code from the sync function into the File.store_async() function that wraps the upload method. Now only a limited amount of files can enter into the "uploading" state.

**Testing:**
I tested out storing a few different data sizes.
1000/1GiB: 125.61s
1/10GiB: 170.75s
100/10GiB: 156.92s

These numbers are in-line with the bench marked numbers collected on the table here: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/620937a91398a2446b0483619ae88a15adede3f8/docs/explanations/benchmarking.md#some-insights

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -402,6 +402,7 @@ class Synapse(object):
         self.max_threads = transfer_config["max_threads"]
         self._thread_executor = {}
         self._process_executor = {}
+        self._parallel_file_transfer_semaphore = {}
         self._md5_semaphore = {}
         self.use_boto_sts_transfers = transfer_config["use_boto_sts"]
 
@@ -561,6 +562,29 @@ class Synapse(object):
         self._md5_semaphore.update({asyncio_event_loop: asyncio.Semaphore(1)})
 
         return self._md5_semaphore[asyncio_event_loop]
+
+    def _get_parallel_file_transfer_semaphore(
+        self, asyncio_event_loop: asyncio.AbstractEventLoop
+    ) -> asyncio.Semaphore:
+        """
+        Retrieve the semaphore for the Synapse client. Or create a new one if it does
+        not exist. This semaphore is used to limit the number of files that can actively
+        enter the uploading/downloading process.
+
+        This is expected to be called from within an AsyncIO loop.
+        """
+        if (
+            hasattr(self, "_parallel_file_transfer_semaphore")
+            and asyncio_event_loop in self._parallel_file_transfer_semaphore
+            and self._parallel_file_transfer_semaphore[asyncio_event_loop] is not None
+        ):
+            return self._parallel_file_transfer_semaphore[asyncio_event_loop]
+
+        self._parallel_file_transfer_semaphore.update(
+            {asyncio_event_loop: asyncio.Semaphore(max(self.max_threads * 2, 1))}
+        )
+
+        return self._parallel_file_transfer_semaphore[asyncio_event_loop]
 
     # initialize logging
     def _init_logger(self):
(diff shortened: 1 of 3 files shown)


synapseutils/sync.py
```diff
@@ -671,58 +660,53 @@ async def _upload_item_async(
...
+                target_version = None
+                if version:
+                    target_version = int(version)
+                if used_item in used:
+                    used_activity.append(
+                        UsedEntity(
+                            target_id=synid, target_version_number=target_version
                         )
+                    )
                 else:
```
jaymedina, line 691: might it be safer to make this an elif? i.e. `elif used_item in executed:`

BryanFauble: Great point, I will be planning to clean all of this code up in the next PR. I wanted to keep this focused on the semaphore change.

synapseclient/client.py
```diff
@@ -562,6 +563,29 @@ def _get_md5_semaphore(
...
+        """
+        if (
+            hasattr(self, "_parallel_file_transfer_semaphore")
+            and asyncio_event_loop in self._parallel_file_transfer_semaphore
+            and self._parallel_file_transfer_semaphore[asyncio_event_loop] is not None
+        ):
+            return self._parallel_file_transfer_semaphore[asyncio_event_loop]
+
+        self._parallel_file_transfer_semaphore.update(
+            {asyncio_event_loop: asyncio.Semaphore(max(self.max_threads * 2, 1))}
```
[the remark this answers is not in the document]
BryanFauble: I am updating the docstring to read:
```
        """
        Retrieve the semaphore for the Synapse client. Or create a new one if it does
        not exist. This semaphore is used to limit the number of files that can actively
        enter the uploading/downloading process.

        This is expected to be called from within an AsyncIO loop.
        
        By default the number of files that can enter the "uploading" state will be
        limited to 2 * max_threads. This is to ensure that the files that are entering
        into the "uploading" state will have priority to finish. Additionally, it means
        that there should be a good spread of files getting up to the "uploading"
        state, entering the "uploading" state, and finishing the "uploading" state.
        
        If we break these states down into large components they would look like:
        - Before "uploading" state: HTTP rest calls to retrieve what data Synapse has
        - Entering "uploading" state: MD5 calculation and HTTP rest calls to determine
          how/where to upload a file to.
        - "uploading" state: Uploading the file to a storage provider.
        - After "uploading" state: HTTP rest calls to finalize the upload.
        """
```

synapseutils/sync.py
jaymedina: Oh, excellent, thanks for limiting the PR scope! Feel free to add me as a reviewer for the other one.

synapseutils/sync.py
```diff
@@ -671,58 +660,53 @@ async def _upload_item_async(
         store_args,
```
thomasyu888, line 660: nit: can there be type hints here?

thomasyu888 (approved):
🔥 LGTM!  This looks great minus the failing tests.

BryanFauble:
> 🔥 LGTM! This looks great minus the failing tests.
> 
> Also - are we not executing sonarcloud currently - or am I just missing it somehow...

@thomasyu888 
Sonar is setup to run after the tests successfully execute since only then do we have the coverage report. However, you bring up a good point - we should let sonar run even on failure. Would you mind creating a tech debt synpy ticket for me for this?

PR #1090, opened 2024-05-02
[SYNPY-1466] update user agent for command line 

danlu1:
Problem:

In the datawarehouse, the client version isn’t being parsed for user_agent of the synapse command line client

Solution:

Update the user agent of the  synapse command line client by removing the `synapseclient` substring with empty string so the client version can be pulld by Synapse-ETL-Jobs

Testing:

Passed both unit test and integration test

### DIFF
--- a/synapseclient/__main__.py
+++ b/synapseclient/__main__.py
@@ -1871,8 +1871,10 @@ def _authenticate_login(
 
 def main():
     args = build_parser().parse_args()
-    synapseclient.USER_AGENT["User-Agent"] = (
-        "synapsecommandlineclient " + synapseclient.USER_AGENT["User-Agent"]
+    synapseclient.USER_AGENT[
+        "User-Agent"
+    ] = "synapsecommandlineclient" + synapseclient.USER_AGENT["User-Agent"].replace(
+        "synapseclient", ""
     )
     if args.otel:
         trace.set_tracer_provider(
--- a/synapseclient/__main__.py
+++ b/synapseclient/__main__.py
@@ -1871,11 +1871,7 @@ def _authenticate_login(
 
 def main():
     args = build_parser().parse_args()
-    synapseclient.USER_AGENT[
-        "User-Agent"
-    ] = "synapsecommandlineclient" + synapseclient.USER_AGENT["User-Agent"].replace(
-        "synapseclient", ""
-    )
+    synapseclient.USER_AGENT = synapseclient.USER_AGENT_COMMAND_LINE
     if args.otel:
         trace.set_tracer_provider(
             TracerProvider(
--- a/synapseclient/__init__.py
+++ b/synapseclient/__init__.py
@@ -88,6 +88,11 @@ USER_AGENT = {
     % (__version__, requests.utils.default_user_agent())
 }
 
+USER_AGENT_COMMAND_LINE = {
+    "User-Agent": "synapsecommandlineclient/%s %s"
+    % (__version__, requests.utils.default_user_agent())
+}
+
 # patch json
 from .core.models import custom_json  # noqa
 


synapseclient/__main__.py
```diff
@@ -1871,8 +1871,10 @@ def _authenticate_login(
 
 def main():
     args = build_parser().parse_args()
-    synapseclient.USER_AGENT["User-Agent"] = (
-        "synapsecommandlineclient " + synapseclient.USER_AGENT["User-Agent"]
+    synapseclient.USER_AGENT[
+        "User-Agent"
+    ] = "synapsecommandlineclient" + synapseclient.USER_AGENT["User-Agent"].replace(
```
thomasyu888, line 1876: Nice clever trick! Could we just get rid of the + and replace synapseclient with synapsecommandlineclient?

synapseclient/__main__.py
```diff
@@ -1871,8 +1871,10 @@ def _authenticate_login(
 
 def main():
     args = build_parser().parse_args()
-    synapseclient.USER_AGENT["User-Agent"] = (
-        "synapsecommandlineclient " + synapseclient.USER_AGENT["User-Agent"]
+    synapseclient.USER_AGENT[
```
BryanFauble, line 1874: I might suggest a slightly different solution to create a second variable in Synapseclient instead of doing a string replacement. 
```
USER_AGENT_COMMAND_LINE = {
    "User-Agent": "synapsecommandlineclient/%s %s"
    % (__version__, requests.utils.default_user_agent())
}

```

Then use that here. The reason is that it us more clear what the agents are expected to be used by this library.

thomasyu888: This is the better solution. 👌

danlu1: Thanks for sharing.  Before adopting this solution,  I would like to confirm that this won't change the work of chain in the client. From what I've learned, the command line client wraps synapseclient functions but adds arg parser to it. There are other scripts utilizing synapseclient.USER_AGENT: 

1. [synapseclient/core/version_check.py](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/0b903e3797206650e0d9137516ce46eb1f454e77/synapseclient/core/version_check.py#L177). In this script, USER_AGENT is added to the headers of a API call to GitHub site and I checked the GitHub returns the content of the target site with or without the USER_AGENT. 

2. [synapseclient/client.py](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/0b903e3797206650e0d9137516ce46eb1f454e77/synapseclient/client.py#L694). In line 694, the USER_AGENT is used to construct a request session. If I understand it correctly, this would trigger the downstream log of the USER_AGENT on Synapse server side.

BryanFauble: What the code is currently doing is redefining an attribute on the `USER_AGENT` constant defined in the client with a new value. This isn't the best, but it's fine because this code is only ever invoked from the CLI and temporary reassignments are just that - Temporary.

Ideally this would instead be a property on the Synapse class instance. But it would involve a larger number of changes.

What you would be doing with this suggestion is something like:

`synapseclient.USER_AGENT = synapseclient.USER_AGENT_COMMAND_LINE`

danlu1: Gotcha. On it.

BryanFauble (approved):
The changes look good to me. @danlu1 until I finish the flaky integration tests ticket can you retry the build cicd pipeline until you have at least 1 full success. We would like to at least have this level of verification in the changes are not breaking anything. 

In addition - if you've already done testing, are you able to verify that the correct results show up the day after in snowflake?

danlu1:
Thanks. @BryanFauble Do I wait for the `Flaky Integration tests` ticket solved to merge the PR or merge it now and rebuild the CICD later? I just tested this on my local and haven't checked the snowflake database. But I'm on it. So tomorrow we might be able to know.

BryanFauble:
> Thanks. @BryanFauble Do I wait for the `Flaky Integration tests` ticket solved to merge the PR or merge it now and rebuild the CICD later? I just tested this on my local and haven't checked the snowflake database. But I'm on it. So tomorrow we might be able to know.

We can wait for `Flaky Integration tests` to be merged in. I am waiting on the pipelines to build before I update https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1085 to non-draft status.

BryanFauble:
@danlu1 https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1085 is ready to be reviewed. When that is approved and merged please pull develop back into your feature branch and we should see a successful CI build afterword.

PR #1092, opened 2024-05-07
[SYNPY-1456] Revert to old ubuntu

BryanFauble:
Problem:
****
1. Running integration tests in ubuntu-22.04 takes between 3-4 hours to complete.

**Solution:**

1. Revert ubuntu versions for the GH runner

**Testing:**

1. Verified after this change that a pipeline run only took ~30 minutes to complete

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -72,7 +72,7 @@ jobs:
           path: |
             ${{ steps.get-dependencies.outputs.site_packages_loc }}
             ${{ steps.get-dependencies.outputs.site_bin_dir }}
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM - do you know why there might be such a big difference?

BryanFauble:
@thomasyu888 @jaymedina I'm not sure why we are seeing such a slowdown here. When I looked back at the runs it seemed like the majority of time was spent doing "nothing" where the next test was waiting to be run. Perhaps there is a bug with the new ubuntu version and with xdist that causes this.

PR #1093, opened 2024-05-09
Update README.md

allaway:
I discussed this with @BryanFauble a while back but only remembered to do this now...


BryanFauble (approved):
Excellent changes. Thanks much!

PR #1094, opened 2024-05-09
[SYNPY-1470]Remove SonarCloud's dependency on test job

danlu1:
Problem:

Currently, SonarCloud only executes when tests pass, but even if tests fail, its valuable for sonarcloud to execute.

Solution:

Remove `needs:[test]` so it doesn't require the test job complete successfully before kicking off SonarCloud job. Add `if: ${{ always() }}` in the  SonarCloud step so the it can be run at all time no matter the status of previous steps.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -206,7 +206,6 @@ jobs:
           path: coverage.xml
 
   sonarcloud:
-    needs: [test]
     name: SonarCloud
     runs-on: ubuntu-20.04
     steps:
@@ -222,6 +221,7 @@ jobs:
         run: sed -i "s/<source>\/home\/runner\/work\/synapsePythonClient<\/source>/<source>\/github\/workspace<\/source>/g" coverage.xml
       - name: SonarCloud Scan
         uses: SonarSource/sonarcloud-github-action@master
+        if: ${{ always() }}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -206,7 +206,6 @@ jobs:
           path: coverage.xml
 
   sonarcloud:
-    needs: [test]
     name: SonarCloud
     runs-on: ubuntu-20.04
     steps:
```
BryanFauble, line 211: Will need to make a change to the `download-artifact` job to make sure a fail here doesn't prevent a red Error to show up in the pipeline:
https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/9022604848/job/24792505985?pr=1094
image
You'll probably need a step similar to:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/.github/workflows/build.yml#L113-L121

That runs a bash script to check if the file is present. If not present don't run `Download coverage report` or `Override Coverage Source Path for Sonar`

danlu1: Yea, I forgot to add this and will work on it soon. Thanks~

.github/workflows/build.yml
```diff
@@ -206,7 +206,6 @@ jobs:
           path: coverage.xml
 
   sonarcloud:
-    needs: [test]
```
BryanFauble, line 209: We will want to wait for `test` to execute before running this Job. Perhaps doing this might work?

```
    needs: [test]
    if: always()
```

danlu1: As the description "Use jobs.<job_id>.needs to identify any jobs that must complete successfully before this job will run." I think needs: [test] will require the test run through before starting.

BryanFauble: Yes - However if you include `always` this will be run even if `test` fails, however, it should wait until `test` fails to run this `sonarcloud` job.

BryanFauble: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

> If you would like a job to run even if a job it is dependent on did not succeed, use the always() conditional expression in jobs.<job_id>.if.

PR #1095, opened 2024-05-10
[SYNPY-1470] Remove SonarCloud's dependency on test job

danlu1:
Problem:
Currently, sonarcloud only executes when tests pass, but even if tests fail, it's valuable for sonarcloud to execute.

Solution:
Let SonarCloud scan always run no matter what status of the previous steps. Add a separate step to check if coverage report exist, if not, skip downloading coverage report. Run `Override Coverage Source Path for Sonar`, only if the coverage report download step has succeeded. 

Testing:
1. Skip downloading the coverage report when the coverage report does not exist
2. The SonarCloud scan runs even though `Override Coverage Source Path for Sonar` fails.
Screenshot 2024-05-09 at 8 45 42 PM
3. Skip Override Coverage Source Path for Sonar when the coverage report does not exist.
Screenshot 2024-05-10 at 9 08 14 AM

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -206,7 +206,6 @@ jobs:
           path: coverage.xml
 
   sonarcloud:
-    needs: [test]
     name: SonarCloud
     runs-on: ubuntu-20.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -213,15 +214,23 @@ jobs:
...
+        with:
+          files: "coverage.xml"
       - name: Download coverage report
         uses: actions/download-artifact@v2
+        if: steps.check_coverage_report.outputs.files_exists == 'true'
         with:
           name: coverage-report
         # This is a workaround described in https://community.sonarsource.com/t/sonar-on-github-actions-with-python-coverage-source-issue/36057
       - name: Override Coverage Source Path for Sonar
+        if: ${{ success() }}
```
BryanFauble, line 229:
```suggestion
        if: steps.check_coverage_report.outputs.files_exists == 'true'
```

BryanFauble (approved):
LGTM!

PR #1096, opened 2024-05-11
[SYNPY-1470] Always run SonarCloud as long as previous jobs are not cancelled

danlu1:
A quick fix for the previous change.


BryanFauble (approved):
LGTM!

PR #1098, opened 2024-05-15
[SYNPY-1356] Logic around disassociating Activity from File Entity

BryanFauble:
**Problem:**

1. When you remove an activity from a file within a manifest (Or through the OOP model interface) it does not delete the Activity off the entity. It will only remove the activity relationship to new versions of that entity, since activity relationships are not brought forward onto new entity versions automatically. The old syncToSynapse relied on this behavior as every sync (Even if a file within the manifest did not change) caused a version update to occur.

**Solution:**

1. Note: A large portion of these changes were already reviewed via this PR: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1099
2. Added in some (unfortunately hacky) backwards compatible code that will handle disassociating an Activity from an entity when the used/executed fields are present in a manifest upload, but no values for those cells.
3. Updated the logic around when a version of a File entity is updated to match the OOP model. The OOP model works by only causing a version update on an Entity if there is metadata about the Entity that changed.

**Testing:**

1. Integration testing around this logic
4. I also verified that I am able to go onto the master branch (Before the syncToSynapse changes) and run these integration tests (With a minor update to `storable_container` to retrieve the activity for the loaded file). This means that behavior wise according these integration tests before/after is working the same*, with 1 small cavet around entity versioning detailed above in the Solution.


tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -391,6 +368,311 @@ async def test_sync_to_synapse_activities_pointing_to_files(
...
+
+        # AND the file is the one we uploaded
+        assert folder.files[0].version_number == 1
+        assert folder.files[0].path == temp_file
+        assert folder.files[0].activity.name == BOGUS_ACTIVITY
+        assert folder.files[0].activity.description == BOGUS_DESCRIPTION
+        assert len(folder.files[0].activity.used) == 1
+        assert folder.files[0].activity.used[0].url == SYNAPSE_URL
+
+        # WHEN I update the manifest file to remove the activities
```
thomasyu888, line 413: Just wanted to confirm that this is existing behavior for the provenance to be removed when there are blank values in used / executed

tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -391,6 +368,311 @@ async def test_sync_to_synapse_activities_pointing_to_files(
...
+        df = pd.DataFrame(
+            {
+                "path": [temp_file],
+                "parent": folder.id,
+                "used": "",
+                "executed": "",
+                "activityName": "",
```
thomasyu888, line 420: I know this is existing behavior but what would happen if an annotation was "activityName" ?

tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -391,6 +368,311 @@ async def test_sync_to_synapse_activities_pointing_to_files(
...
+        for file in temp_files:
+            schedule_for_cleanup(file)
+
+        # AND A manifest file with the paths to the temp files exists
+        annotations = ["foo", "bar", "baz", "qux", "quux"]
+        df = pd.DataFrame(
+            {
+                "path": temp_files,
+                "parent": folder.id,
+                "my_file_annotations": annotations,
```
thomasyu888, line 623: Is this existing behavior as well? This makes partial updates more difficult if you only want to update the annotations in one column.

thomasyu888 (commented):
Thanks Bryan - the code looks great, but just wanted to have some discussions around some of the comments I made.

tests/integration/synapseutils/test_synapseutils_sync.py
BryanFauble: Excellent point. Perhaps the way I should make this work is you may specify optional booleans in the function:

1. Clear annotations from Synapse that are not defined in my file
2. Remove activity/provenance if not defined in my file
3. I could even allow something like: Apply existing provenance to new file versions

tests/integration/synapseutils/test_synapseutils_sync.py
BryanFauble: It's slightly different. If used/executed it not specified it just won't apply a new Activity to the new version of your file. 

In this case I'm removing the Activity from your current Entity version if you don't specify it. If you have a new version 2, the original Activity on version 1 still exists, and no Activity exists on 2.

Since this is slightly different, See the other thread about my boolean idea to extend the things you can do with this function.

tests/integration/synapseutils/test_synapseutils_sync.py
thomasyu888: That is a great idea.  For no. 3, we may want to hold off on that because technically when you create a new version of a file, you may have a new activity.  That said, maybe the new annotations added wouldn't have a new activity since the content in the file is the same.

synapseclient/core/utils.py
```diff
@@ -1400,8 +1400,9 @@ def delete_none_keys(incoming_object: typing.Dict) -> None:
 
 def merge_dataclass_entities(
     source: typing.Union["Project", "Folder", "File"],
-    destination: typing.Union["Project", "Folder"],
-) -> typing.Union["Project", "Folder"]:
+    destination: typing.Union["Project", "Folder", "File"],
+    fields_to_ignore: typing.List[str] = None,
```
thomasyu888, line 1404: Nit: can we add a parameter description?

thomasyu888 (approved):
🔥 LGTM! Going to pre-approve, there's some comments but they are minor.

PR #1099, opened 2024-05-16
[SYNPY-1356] Updating logic around activity/annotations to be optional

BryanFauble:
**Problem:**

1. Some changes I introduced in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1098 would change default behavior in ways we did not want that would cause some potential data deletion.

**Solution:**

1. Updating changed logic to be behind optional booleans that allow folks to have more fine tuned control over how changes can be synced to synapse.

**Testing:**

1. Relying on the integration test cases


thomasyu888 (commented):
I see this PR, but wanted to give a chance for @Sage-Bionetworks/dpe to review

synapseclient/core/utils.py
```diff
@@ -1443,49 +1443,3 @@ def merge_dataclass_entities(
         setattr(destination, key, value)
 
     return destination
-
-
-def merge_non_modifiable_manifest_fields(
```
rxu17, line 1448: Why was this function removed in favor of `_determine_fields_to_ignore_in_merge` added to             `merge_dataclass_entities` in the file's store function?

synapseclient/models/services/storable_entity_components.py
```diff
@@ -203,23 +203,24 @@ async def _store_activity_and_annotations(
...
     ):
-        if root_resource.activity:
-            result = await root_resource.activity.store_async(
-                parent=root_resource, synapse_client=synapse_client
-            )
-            root_resource.activity = result
-        else:
-            from synapseclient.models import Activity
-
-            await Activity.disassociate_from_entity_async(
```
rxu17, line 220: Why do we remove this part that disassociates the activity from the entity if it didn't meet the requirements for it to be stored to synapse?

synapseclient/models/services/storable_entity_components.py
```diff
@@ -203,23 +203,24 @@ async def _store_activity_and_annotations(
 
     if (
         hasattr(root_resource, "activity")
-        and (root_resource.activity or last_persistent_instance)
+        and root_resource.activity is not None
         and (
             last_persistent_instance is None
-            or last_persistent_instance.activity != root_resource.activity
+            or (
```
rxu17, line 209: Nit: there is a lot of nested logic going on here, is it possible to break it out into several variables that give insight as to what it is checking for and how it fulfills what the function states that it should be doing?

rxu17 (approved):
LGTM! Mainly questions about what the code is doing

synapseclient/core/utils.py
BryanFauble: I wrote it in this way to keep the utility class here with the least amount of responsibility. If we determine the fields we are not going to merge before calling this utility function it'll mean that this code can be generically re-used for other Entities. File was the first entity to have drastically different rules determining how data in merged and it made sense to leave it within the context of the File

synapseclient/models/services/storable_entity_components.py
BryanFauble: I made a change here that stop the store process from automatically removing an Activity during the store operation. I am currently setting this up such that you need to directly interact with the Activity (`disassociate_from_entity_async` or `delete_async`) to modify these relationships.

I think more discussion needs to happen if we want to automatically modify the activity/provenance relationship for folks like this code was doing.

synapseclient/models/services/storable_entity_components.py
BryanFauble: Take a look at this code change and let me know what you think: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1099/commits/8b353053d04619ac5d3039f84244059aece15562

tests/integration/synapseutils/test_synapseutils_sync.py
```diff
@@ -322,7 +322,7 @@ async def test_sync_to_synapse_activities_pointing_to_files(
         schedule_for_cleanup(folder.id)
 
         # AND 5 temporary files on disk:
```
thomasyu888, line 324:
```suggestion
        # AND 3 temporary files on disk:
```

thomasyu888 (approved):
🔥 LGTM!  i took a look at your test cases, and the conditions are sound to me!  Thanks for adding these fields - going to pre-approve pending Rixing's final review!

synapseclient/models/services/storable_entity_components.py
rxu17: Looks good to me! thanks!

PR #1100, opened 2024-05-22
[SYNPY-1341] Add CI step to containerize and upload synpy to the GHCR

jaymedina:
### problem

_We recently added a GHCR [image](https://github.com/orgs/sage-bionetworks/packages/container/package/synapsepythonclient) for the python client because there was no up-to-date image hosted anywhere. We should add a step to the CI workflow to publish new versions of the client as docker images automatically when a new version is released._

### solution

Two new jobs (`ghcr-build-and-push-on-merge`, `ghcr-build-and-push-on-develop`) are created to build a multi-platform image of the repository and push it to the GitHub Container Registry.

`ghcr-build-and-push-on-release`
- [x] The job should be triggered only for an official version release
~~- [ ] The docker image should be multi-platform (AMD64, ARM64)~~
- [x] The docker image tag should be named after the release version number 

`ghcr-build-and-push-on-develop`
- [x] The job should be triggered only for pushes to `develop` (i.e. PR merges)
~~- [ ] The docker image should be multi-platform (AMD64, ARM64)~~
- [x] The docker image tag should be named after the latest commit SHA on `develop`
### testing & preview

`ghcr-build-and-push-on-release`

I manually fed a version `v1.2.3test` that will be fed into the `Build and push Docker image` step to test that this communication happens between `Extract Release Version` and `Build and push Docker image`.
image

Triggering the job results in a new release with the expected version name
image

I pulled the docker image and ran a container to confirm that the latest version `v4.2.0` of the python client is running in the container:
image

`ghcr-build-and-push-on-develop`

`develop` image tag name reflects the latest commit it points to. Below is the example for commit https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1100/commits/830399960ebf890e045220b8c68a08a9e953e8b5 of this branch: 

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -467,7 +467,7 @@ jobs:
           push: true
           tags: ghcr.io/sage-bionetworks/synapsepythonclient:latest,ghcr.io/sage-bionetworks/synapsepythonclient:${{ github.run_number }}
           file: ./Dockerfile
-          platforms: linux/arm64
+          platforms: linux/amd64, linux/arm64
           cache-from: type=registry,ref=ghcr.io/sage-bionetworks/synapsepythonclient:latest
           cache-to: type=inline
       - name: Output image digest
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -433,29 +433,33 @@ jobs:
           exit 1
 
   # finally, containerize the package and upload to the GHCR
-  ghcr-upload:
+  build-and-push-to-ghcr:
     needs: package
-    runs-on: ubuntu-20.04
+    if: github.event_name == 'release' && github.event.release.prerelease == 'false'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+
     steps:
-      - name: Checkout repository
-        uses: actions/checkout@v2
-
+      - name: Check out the repo
+        uses: actions/checkout@v4
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v1
-        if: startsWith(github.ref, 'refs/tags/')
-
-      - name: Login to GitHub Container Registry
+        uses: docker/setup-buildx-action@v2
+      - name: Log in to GitHub Container Registry
         uses: docker/login-action@v2
-        if: startsWith(github.ref, 'refs/tags/')
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-
       - name: Build and push Docker image
-        uses: docker/build-push-action@v2
-        if: startsWith(github.ref, 'refs/tags/')
+        uses: docker/build-push-action@v3
         with:
-          context: .
           push: true
-          tags: ghcr.io/${{ github.repository_owner }}/your-image-name:latest
+          tags: ghcr.io/${{ github.repository }}/synapseclient:latest,ghcr.io/${{ github.repository }}/synapseclient:${{ github.run_number }}
+          file: ./Dockerfile
+          platforms: linux/amd64, linux/arm64
+          cache-from: type=registry,ref=ghcr.io/${{ github.repository }}/synapseclient:latest
+          cache-to: type=inline
+      - name: Output image digest
+        run: echo "The image digest is ${{ steps.docker_build.outputs.digest }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -439,3 +439,39 @@ jobs:
...
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        id: docker_build
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          tags: ghcr.io/sage-bionetworks/synapsepythonclient:latest,ghcr.io/sage-bionetworks/synapsepythonclient:${{ env.RELEASE_VERSION }}
```
rxu17, line 471: What currently uses the synapseclient docker container (the older non-ghcr versions)? I'm guessing it won't be needed to be tagged for `develop` or feature branches?

rxu17 (approved):
LGTM! Just a question

.github/workflows/build.yml
BryanFauble: > What currently uses the synapseclient docker container (the older non-ghcr versions)?

I'm not sure myself, but we are only going to be publishing on GHCR from now on as previous solutions were manual.

> I'm guessing it won't be needed to be tagged for develop or feature branches?

This job is only run during the package step and when `[github.event.release.prerelease == 'false'`](if: github.event_name == 'release' && github.event.release.prerelease == 'false')`. These 2 things make sure that we're only releasing tags when we create them. Based on what I am seeing in the code Jenny wrote is that it'll automatically pull the version of the tag and use that. Which is what we want.

BryanFauble (approved):
Excellent work!

.github/workflows/build.yml
```diff
@@ -439,3 +439,39 @@ jobs:
 
           echo "Failed to install expected version $VERSION"
           exit 1
+
+  # finally, containerize the package and upload to the GHCR
+  build-and-push-to-ghcr:
+    needs: package
```
thomasyu888, line 445: Should we have a "latest" one too when code is pushed to dev?

.github/workflows/build.yml
rxu17: Gotcha, yeah that makes sense if this is only needed during package release step

.github/workflows/build.yml
jaymedina: The job currently tags the latest container with `latest` and the version of the release. Are you saying it would add value to push 2 versions of a container: one reflecting the bleeding edge of the repository e.g. `latest-dev` and one reflecting the latest version on `PyPI` (`latest`)?

BryanFauble: We could have `latest` and the tagged versions be the same, ie: `4.2.0` and `latest`.

I like the idea of pushing one out when changes are committed to develop, but that one should be tagged something like `develop`, or `dev`. Thoughts?

jaymedina: I can see this adding value to users who work exclusively in docker container environments for developing. Can do 👍

BryanFauble: No worries since you're already making the changes - But another possible way to stage these new changes could be in a new PR. Branch off this feature branch and then point it back at this feature branch with this specific change.

jaymedina: Good point! Since I've already implemented the change, I won't do it this time but will do it in the future. I agree, I don't like big PRs :) I'm finishing up some troubleshooting work having to do with multiple untagged images being pushed alongside my tagged multi-platform images, after which I'll post new screenshots with more test results and re-request review. Thanks!

jaymedina:
This PR is now ready for a final review.

Latest changes:
* Each push to `develop` will trigger the job to generate a new image tagged with `develop-[commit SHA]` so that all images can be traced back to a point in the commit history. Examples:

image

* **TLDR: The scope of this PR is adjusted to work with single platform images. The support of multi-platform images will be an enhancement ticket after this one is complete.** The original intention was to support multi-platform images, however the way this has been working so far has led to untagged images being pushed alongside the official tagged image. These untagged images seem to be individual layers corresponding to one of the designated platforms (e.g. `linux/arm64` created an untagged image, and `linux/amd64` creates another). These extra untagged images being pushed to the registry add extra clutter, which I would like to minimize since the `develop` images will be filling up the registry on top of this, so I'll be opening a new ticket for tackling this behavior in a separate PR, and this feature will introduce automating the creation and push of SINGLE platform images onto the registry.

Feedback is welcome. Thanks!

BryanFauble (approved):
This all looks good to me! We can test this out and adjust based on feedback.

jaymedina:
Great thanks @BryanFauble - Once all tests pass I'll squash and merge this.

rxu17 (approved):

PR #1103, opened 2024-06-04
[SYNPY-1476] Splitting logic for downloading content from client.py

BryanFauble:
**Problem:**

1. There is a significant amount of business logic within client.py related to configuration and downloading of files.
2. Many warnings were popping up when running unit/integration tests.
3. As items are being migrated out we will need to mark them as deprecated.
4. Lot's of items are flagged by sonarcloud that need to be addressed.

**Solution:**

1. Laying the ground work for migrating logic out of client.py. I started this process by splitting functions out into `synapseclient/api/configuration_services.py`, `synapseclient/api/file_services.py`, and `synapseclient/core/download/download_functions.py`
2. Updating unit/integration tests for items I saw that were spitting out warnings or other code smells.
3. Marking items as deprecated and updating all internal usage of those deprecated methods to their replacements.
4. Starting to address some sonarcloud flagged issues.

**Testing:**

1. Light manual testing but mostly focusing on unit/integration testing around these changes. Further PRs will include heavier testing of downloading logic that is split out of client.py

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -1,0 +1,646 @@
+[MAIN]
+
+# Analyse import fallback blocks. This can be used to support both Python 2 and
+# 3 compatible code, which means that the block might have code that exists
+# only in one or another interpreter, leading to false positives when analysed.
+analyse-fallback-blocks=no
+
+# Clear in-memory caches upon conclusion of linting. Useful if running pylint
+# in a server-like mode.
+clear-cache-post-run=no
+
+# Load and enable all available extensions. Use --list-extensions to see a list
+# all available extensions.
+#enable-all-extensions=
+
+# In error mode, messages with a category besides ERROR or FATAL are
+# suppressed, and no reports are done by default. Error mode is compatible with
+# disabling specific errors.
+#errors-only=
+
+# Always return a 0 (non-error) status code, even if lint errors are found.
+# This is primarily useful in continuous integration scripts.
+#exit-zero=
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-allow-list=
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code. (This is an alternative name to extension-pkg-allow-list
+# for backward compatibility.)
+extension-pkg-whitelist=
+
+# Return non-zero exit code if any of these messages/categories are detected,
+# even if score is above --fail-under value. Syntax same as enable. Messages
+# specified are enabled, while categories only check already-enabled messages.
+fail-on=
+
+# Specify a score threshold under which the program will exit with error.
+fail-under=10
+
+# Interpret the stdin as a python script, whose filename needs to be passed as
+# the module_or_package argument.
+#from-stdin=
+
+# Files or directories to be skipped. They should be base names, not paths.
+ignore=CVS
+
+# Add files or directories matching the regular expressions patterns to the
+# ignore-list. The regex matches against paths and can be in Posix or Windows
+# format. Because '\\' represents the directory delimiter on Windows systems,
+# it can't be used as an escape character.
+ignore-paths=
+
+# Files or directories matching the regular expression patterns are skipped.
+# The regex matches against base names, not paths. The default value ignores
+# Emacs file locks
+ignore-patterns=^\.#
+
+# List of module names for which member attributes should not be checked and
+# will not be imported (useful for modules/projects where namespaces are
+# manipulated during runtime and thus existing member attributes cannot be
+# deduced by static analysis). It supports qualified module names, as well as
+# Unix pattern matching.
+ignored-modules=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use, and will cap the count on Windows to
+# avoid hangs.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# List of plugins (as comma separated values of python module names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Resolve imports to .pyi stubs if available. May reduce no-member messages and
+# increase not-an-iterable messages.
+prefer-stubs=no
+
+# Minimum Python version to use for version dependent checks. Will default to
+# the version used to run pylint.
+py-version=3.8
+
+# Discover python modules and packages in the file system subtree.
+recursive=no
+
+# Add paths to the list of the source roots. Supports globbing patterns. The
+# source root is an absolute path or a path relative to the current working
+# directory used to determine a package namespace for modules located under the
+# source root.
+source-roots=
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+# In verbose mode, extra non-checker-related info will be displayed.
+#verbose=
+
+
+[BASIC]
+
+# Naming style matching correct argument names.
+argument-naming-style=snake_case
+
+# Regular expression matching correct argument names. Overrides argument-
+# naming-style. If left empty, argument names will be checked with the set
+# naming style.
+#argument-rgx=
+
+# Naming style matching correct attribute names.
+attr-naming-style=snake_case
+
+# Regular expression matching correct attribute names. Overrides attr-naming-
+# style. If left empty, attribute names will be checked with the set naming
+# style.
+#attr-rgx=
+
+# Bad variable names which should always be refused, separated by a comma.
+bad-names=foo,
+          bar,
+          baz,
+          toto,
+          tutu,
+          tata
+
+# Bad variable names regexes, separated by a comma. If names match any regex,
+# they will always be refused
+bad-names-rgxs=
+
+# Naming style matching correct class attribute names.
+class-attribute-naming-style=any
+
+# Regular expression matching correct class attribute names. Overrides class-
+# attribute-naming-style. If left empty, class attribute names will be checked
+# with the set naming style.
+#class-attribute-rgx=
+
+# Naming style matching correct class constant names.
+class-const-naming-style=UPPER_CASE
+
+# Regular expression matching correct class constant names. Overrides class-
+# const-naming-style. If left empty, class constant names will be checked with
+# the set naming style.
+#class-const-rgx=
+
+# Naming style matching correct class names.
+class-naming-style=PascalCase
+
+# Regular expression matching correct class names. Overrides class-naming-
+# style. If left empty, class names will be checked with the set naming style.
+#class-rgx=
+
+# Naming style matching correct constant names.
+const-naming-style=UPPER_CASE
+
+# Regular expression matching correct constant names. Overrides const-naming-
+# style. If left empty, constant names will be checked with the set naming
+# style.
+#const-rgx=
+
+# Minimum line length for functions/classes that require docstrings, shorter
+# ones are exempt.
+docstring-min-length=-1
+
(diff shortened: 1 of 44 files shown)


pylintrc
```diff
@@ -0,0 +1,646 @@
+[MAIN]
```
BryanFauble, line 1: I generated this with `pylint --generate-rcfile > pylintrc` and add private member access and non-top level imports to the ignored messages. The rest of the file is auto-generated defaults.

synapseclient/api/configuration_services.py
```diff
@@ -0,0 +1,136 @@
...
+    return get_config_section_dict(
+        section_name=config_file_constants.AUTHENTICATION_SECTION_NAME,
+        config_path=config_path,
+    )
+
+
+def get_transfer_config(
+    config_path: str,
+) -> Dict[str, str]:
+    """section_name=
```
jaymedina, line 99: Is this supposed to be here? `section_name=`?

synapseclient/api/configuration_services.py
```diff
@@ -0,0 +1,136 @@
...
+    Arguments:
+        config_path:  Path to configuration file on local file system
+
+    Returns:
+        A RawConfigParser populated with properties from the user's configuration file.
+    """
+
+    try:
+        config = configparser.RawConfigParser()
+        config.read(config_path)  # Does not fail if the file does not exist
```
jaymedina, line 27: Do we want to add error-handling to address the comment here?

synapseclient/api/configuration_services.py
```diff
@@ -0,0 +1,136 @@
...
+) -> Dict[str, str]:
+    """
+    Get a profile section in the configuration file with the section name.
+
+    Arguments:
+        section_name: The name of the profile section in the configuration file
+        config_path:  Path to configuration file on local file system
+
+    Returns:
+        A dictionary containing the configuration profile section content
```
jaymedina, line 45: This should mention that it will return an empty dictionary if it runs into a `NoSectionError`

synapseclient/api/configuration_services.py
```diff
@@ -0,0 +1,136 @@
...
+        config.read(config_path)  # Does not fail if the file does not exist
+        return config
+    except configparser.Error as ex:
+        raise ValueError(f"Error parsing Synapse config file: {config_path}") from ex
+
+
+def get_config_section_dict(
+    section_name: str,
+    config_path: str,
+) -> Dict[str, str]:
```
jaymedina, line 36: I think return type should be `Dict[str, Any]`

synapseclient/api/configuration_services.py
```diff
@@ -0,0 +1,136 @@
...
+    Arguments:
+        endpoint: The location of the target service
+        bucket:   AWS S3 bucket name
+        config_path:  Path to configuration file on local file system
+
+    Returns:
+        The authenticated S3 profile
+    """
+
+    config_section = endpoint + "/" + bucket
```
jaymedina, line 72: Do we want to use `os.path.join(endpoint, bucket)` instead? Manually adding forward slashes like this makes the code a bit more flimsy because you're relying on `endpoint` not already having the forward slash.

Here is an example of why `os.path.join` is more robust:

image

synapseclient/api/configuration_services.py
```diff
@@ -0,0 +1,136 @@
...
+            if k == "max_threads":
+                try:
+                    transfer_config["max_threads"] = int(v)
+                except ValueError as cause:
+                    raise ValueError(
+                        f"Invalid transfer.max_threads config setting {v}"
+                    ) from cause
+
+            elif k == "use_boto_sts":
+                lower_v = v.lower()
```
jaymedina, line 128: The default `use_boto_sts` is a boolean, but the key value is treated as a string here

synapseclient/api/configuration_services.py
BryanFauble: Each of the configuration sections are simple key,value pairs. The config parser library that is extracting these values is always returning back a string for the value even if it's not exactly a string:

image


image

synapseclient/api/configuration_services.py
BryanFauble: No. I think leaving this as is is the appropriate choice. The reason is that there are multiple authentication methods and the config file is only one of them. In the case the file is not found it should fall through to the next auth method. Then only at the end once no other methods can authenticate you are we raising an exception.

synapseclient/api/configuration_services.py
BryanFauble: See https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1103#discussion_r1629864731 - This is how the config parser library behaves.

synapseclient/api/configuration_services.py
thomasyu888: Nit: urllib?

synapseclient/client.py
```diff
@@ -3377,6 +3461,11 @@ def _download_from_url_multi_threaded(
 
         return destination
 
+    @deprecated(
```
thomasyu888, line 3464: Excited to see all these deprecation warnings :)

synapseclient/api/configuration_services.py
BryanFauble: > Manually adding forward slashes like this makes the code a bit more flimsy because you're relying on endpoint not already having the forward slash.

I agree - However this relies on the contract that the platform team set in the rest API when we are looking at specific files `fileHandle["endpointUrl"]`

BryanFauble: Swapped over to urllib to handle this. Verified I could still upload to an external object store I set up before:

image

`Uploading to endpoint: [https://s3.amazonaws.com] bucket: [sc-237179673806-pp-3ywcz7qgqp4tk-s3bucket-s7wyruiz8llz]: 100%|█|`

synapseclient/client.py
```diff
@@ -357,7 +368,7 @@ def log_response(response: httpx.Response) -> None:
         # Check for a config file
         self.configPath = configPath
         if os.path.isfile(configPath):
-            config = self.getConfigFile(configPath)
+            config = get_config_file(configPath)
```
jaymedina, line 371: If we want to keep this consistent with other function calls, do we want this to be `self.configPath`?

synapseclient/core/download/download_functions.py
```diff
@@ -0,0 +1,727 @@
...
+        else os.path.basename(cached_file_path)
+    )
+
+    # Decide the best download location for the file
+    if download_location is not None:
+        # Make sure the specified download location is a fully resolved directory
+        download_location = ensure_download_location_is_directory(download_location)
+    elif cached_file_path is not None:
+        # file already cached so use that as the download location
+        download_location = os.path.dirname(cached_file_path)
```
jaymedina, line 115: Would this lead to overwriting the file without warning?

synapseclient/core/download/download_functions.py
```diff
@@ -0,0 +1,727 @@
...
+) -> str:
+    """
+    Download a file from the given URL to the local file system.
+
+    Arguments:
+        file_handle_id: The id of the FileHandle to download
+        synapse_id:     The id of the Synapse object that uses the FileHandle e.g. "syn123"
+        entity_type:   The type of the Synapse object that uses the FileHandle e.g. "FileEntity"
+        destination:  The destination on local file system
+        retries:      The Number of download retries attempted before throwing an exception.
```
jaymedina, line 179: nit: Add `Default is 5.`

synapseclient/core/download/download_functions.py
```diff
@@ -0,0 +1,727 @@
...
+                        prefix="Downloading ",
+                        postfix=os.path.basename(destination),
+                        dt=time.time() - transfer_start_time,
+                    )
+
+            urllib_request.urlretrieve(
+                url=url, filename=destination, reporthook=_ftp_report_hook
+            )
+            break
+        elif scheme == "http" or scheme == "https":
```
jaymedina, line 454: nit: easier to read: `elif scheme in ["http", "https"]`

tests/unit/synapseclient/core/download/unit_test_download_functions.py
```diff
@@ -0,0 +1,42 @@
+"""Unit tests for synapseclient.core.download.download_functions"""
```
jaymedina, line 1: Might be a good idea to test out `resolve_download_path_collisions`

synapseclient/client.py
BryanFauble: There is no difference in this case as just above we're doing `self.configPath = configPath`

synapseclient/core/download/download_functions.py
BryanFauble: > Would this lead to overwriting the file without warning?

Not in this case - This first check is "If I were to download the file, where _would_ I download that file to". Later on the `resolve_download_path_collisions` handles for this.

> Also I remember you mentioning something about os.path not working for Windows because of the \ delimiter, is this a concern in this context?

It's not a problem in this case because the script running on a windows machine is going to have drives mapped with the correct delimiter with `os.path`.

synapseclient/core/download/download_functions.py
```diff
@@ -0,0 +1,727 @@
...
+        # _downloadFileHandle may also return None to indicate that the download failed
+        download_path = await download_by_file_handle(
+            file_handle_id=entity.dataFileHandleId,
+            synapse_id=object_id,
+            entity_type=object_type,
+            destination=download_path,
+        )
+
+        if download_path is None or not os.path.exists(download_path):
+            return
```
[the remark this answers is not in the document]
BryanFauble: I'll think more about how we want this to work as I refactor more. The way I'd like this to work is that the methods to download the files (Think FTP, Multi-threaded, ect..) should be responsible for writing out any log messages when it failed to download the files.

synapseclient/core/download/download_functions.py
BryanFauble: I'll be removing this shortly to convert this to be a time based retry system.

synapseclient/core/download/download_functions.py
```diff
@@ -0,0 +1,727 @@
...
+    expected_md5: str = None,
+    synapse_client: Optional["Synapse"] = None,
+) -> str:
+    """
+    Download a file from the given URL using multiple threads.
+
+    Arguments:
+        file_handle_id: The id of the FileHandle to download
+        object_id:      The id of the Synapse object that uses the FileHandle e.g. "syn123"
+        object_type:    The type of the Synapse object that uses the FileHandle e.g. "FileEntity"
```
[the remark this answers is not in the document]
BryanFauble: It can be any of these: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/file/FileHandleAssociateType.html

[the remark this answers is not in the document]
BryanFauble: I'll be sure to update the doc strings to far up to include the link to these types

synapseclient/core/download/download_functions.py
```diff
@@ -0,0 +1,727 @@
...
+    """
+    Download a file from the given URL to the local file system.
+
+    Arguments:
+        url:           The source of download
+        destination:   The destination on local file system
+        file_handle_id:  Optional. If given, the file will be given a temporary name that includes the file
+                                handle id which allows resuming partial downloads of the same file from previous
+                                sessions
+        expected_md5:  Optional. If given, check that the MD5 of the downloaded file matches the expected MD5
```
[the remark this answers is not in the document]
BryanFauble: This was fixed in this PR: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1106

tests/unit/synapseclient/core/download/unit_test_download_functions.py
BryanFauble: I added integration tests around these different file collision types in: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1106/files#diff-e1692a03c6aa3f388426318480c8b7405e995d62eec5a19098876ebf3a5953c5

synapseclient/core/download/download_functions.py
BryanFauble: Great suggestion!

BryanFauble: Added in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1106/commits/9e8a1226cafca81d3fb0956ca552322e2842c71d

synapseclient/core/download/download_functions.py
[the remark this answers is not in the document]
BryanFauble: Added in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1106/commits/9e8a1226cafca81d3fb0956ca552322e2842c71d

PR #1104, opened 2024-06-07
[SYNPY-1482] Updating to remove loop executor during annotation store

BryanFauble:
**Problem:**

1. Storage of annotation was taking place in a loop executor

**Solution:**

1. Updating logic to be stored without the executor/other threads

**Testing:**

1. Unit/Integration tests

### DIFF
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -1,5 +1,5 @@
 # These are all of the models that are used by the Synapse client.
-from .annotations import set_annotations
+from .annotations import set_annotations, set_annotations_async
 from .entity_bundle_services_v2 import (
     get_entity_id_bundle2,
     get_entity_id_version_bundle2,
@@ -30,6 +30,7 @@ from .file_services import (
 __all__ = [
     # annotations
     "set_annotations",
+    "set_annotations_async",
     "get_entity_id_bundle2",
     "get_entity_id_version_bundle2",
     "post_entity_bundle2_create",
--- a/synapseclient/api/annotations.py
+++ b/synapseclient/api/annotations.py
@@ -47,3 +47,36 @@ def set_annotations(
             }
         ),
     )
+
+
+async def set_annotations_async(
+    annotations: "Annotations",
+    synapse_client: Optional["Synapse"] = None,
+):
+    """Call to synapse and set the annotations for the given input.
+
+    Arguments:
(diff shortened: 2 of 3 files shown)


synapseclient/api/annotations.py
```diff
@@ -47,3 +47,36 @@ def set_annotations(
...
+            the `.login()` method.
+
+    Returns: The annotations set in Synapse.
+    """
+    annotations_dict = asdict(annotations)
+
+    synapse_annotations = _convert_to_annotations_list(
+        annotations_dict["annotations"] or {}
+    )
+    from synapseclient import Synapse
```
linglp, line 71: why here the import statement is in the function call?

BryanFauble: This is to prevent circular import issues. This is a "lazy" import. It was the easiest way to fix the problem, but I'd be open to suggestions.

synapseclient/api/__init__.py
```diff
@@ -30,6 +30,7 @@
 __all__ = [
     # annotations
     "set_annotations",
+    "set_annotations_async",
```
linglp, line 33: I am also curious what is the purpose of "__all__" here and why did you add set_annotations_async?

BryanFauble: This is defining what can be imported from `synapseclient.api`:

https://www.geeksforgeeks.org/python-__all__/

It allows you to do:
`from synapseclient.api import set_annotations_async` instead of `from synapseclient.api.annotations import set_annotations_async`

synapseclient/api/annotations.py
linglp: Thanks. I actually have never heard of lazy import before. I looked it up and it seems like the benefit of lazy import is to load a module when the module is being used. But`Synapse` is so commonly used which makes me wonder how big the benefit is.. But thanks for teaching me lazy import, I learned something new today!

synapseclient/api/__init__.py
linglp: wow thank you for the explanation Bryan!

synapseclient/api/annotations.py
thomasyu888: @linglp to circle back around to this.  It's important to note that we use lazy importing to avoid circular imports.

thomasyu888 (approved):
🔥 LGTM!

PR #1105, opened 2024-06-07
[SYNPY-1483] Update default order for credential provider

BryanFauble:
**Problem:**

1. There have been 2 service desk tickets from folks unable to use `synapse login -p $AUTH_TOKEN` due to having a synapse config that might not have the appropriate data.

**Solution:**

1. Updating the credential chain to use CLI arguments first

**Testing:**

1. Verified that setting `-p` is used if supplied, but not if not supplied:
image


rxu17 (approved):
LGTM!

PR #1106, opened 2024-06-07
[SYNPY-1476] Increase code coverage and stage files

BryanFauble:
Problem:

1. Code coverage around download functions in terms of integration tests is severely lacking.
2. Some dependency imports need to be updated to start pointing at locations where new code is going to exist.

**Solution:**

1. Adding integration tests to cover more scenarios.
2. Staging files to be ready for future changes. In particular, the `download_async.py` file is close to a copy of `download_threads.py` as it is going to replace that python script.

**Testing:**

1. Integration testing

### DIFF
--- a/synapseclient/core/download/download_async.py
+++ b/synapseclient/core/download/download_async.py
@@ -1,0 +1,477 @@
+"""Logic required for the actual transferring of files."""
+
+try:
+    import threading as _threading
+except ImportError:
+    import dummy_threading as _threading
+
+import concurrent.futures
+import datetime
+import os
+import time
+from contextlib import contextmanager
+from http import HTTPStatus
+from typing import Generator, NamedTuple, TYPE_CHECKING
+from urllib.parse import parse_qs, urlparse
+
+from requests import Response, Session
+from requests.adapters import HTTPAdapter
+from urllib3.util.retry import Retry
+
+from synapseclient.api import get_file_handle_for_download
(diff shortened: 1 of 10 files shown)


synapseclient/core/download/download_async.py
```diff
@@ -0,0 +1,477 @@
+"""Logic required for the actual transferring of files."""
```
BryanFauble, line 1: `download_async.py` file is close to a copy of `download_threads.py`. No significant changes here, but I wanted to start to point files to the correct locations they're going to eventually live at.

synapseclient/core/upload/upload_functions_async.py
```diff
@@ -306,6 +306,11 @@ async def upload_synapse_s3(
     return await get_file_handle(file_handle_id=file_handle_id, synapse_client=syn)
 
 
+def _get_aws_credentials() -> None:
```
thomasyu888, line 309: I know it's used only in testing purposes, but what was happening before?

BryanFauble: You could only specify the aws account to use by profile name. Then that means you have to have the correct .aws config available on disk.

Doing it this way allows me to mock which credentials are being used easier.

I don't think there was a test case testing this before, so it was not an issue.

thomasyu888 (approved):
🤖 LGTM!  Took some time for me to go through and read the test cases, but then comments really make it easy to follow

PR #1107, opened 2024-06-11
[SYNPY-1476] Refactoring multithreaded downloads to be more fault tolerant

BryanFauble:
**As a heads up for this review:** There are a large number of issues present in this code that was fixed later on in this second PR: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1111 - This Pull Request is scoped towards the re-write of the core download logic, while 1111 is geared towards the re-write of the outer logic that uses this re-written download logic.

**Problem:**

1. The download logic that was applicable to the requests library no longer works the same way when swapping to HTTPX - In particular around how streams are handled.
2. The download logic for multi-threaded downloads is not extremely fault tolerant.
3. The new download logic did not work for the way that synapseutils.syncFromSynapse works.

**Solution:**

1. Removing the custom solution built to submit jobs to a thread executor and instead swapping to use asyncio to wrap any thread execution.
2. Moving several APIs to async & organizing code to handle all of this.
4. Updating the retry logic here to be time based instead of number of retry based.
5. Updating the logic within synapseutils.sync to use the new models interfaces, updating code within the model interfaces to allow backwards compatibility with this function.
6. Updating the new models sync_from_synapse to support traversing links.

**Testing:**

1. (Not currently included) For cursory testing I wrote an integration test that would fail the download of individual file chunks at a 90% rate. Old logic would require that the test itself be ran 3-5 times before it would be successful. New logic works 100% of the time without needing to restart the test. This means that downloads should be significantly more fault tolerant.

**Items left to complete:**

- [x] Fix bugs with the tqdm progress bar for all types of downloads and the shared bar from synapseutils.sync
- [x] Determine the best way to handle displaying when a file is being downloaded or an MD5 check is running on it (Will be handled in: https://sagebionetworks.jira.com/browse/SYNPY-1486)
- [x] Update the code in Project, Folder, File `.get_async()` to no longer be a simple wrapper for the `SynapseClient.get()` function. https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1111
- [x] Benchmark downloads and update the related scripts
- [x] In download_functions.py determine the best method to handle using `with_retry_time_based` and HTTPX streams
- [x] Update code in `download_from_url` to use the HTTP client
- [x] Update any functions within `download_functions.py` and `download_async.py` to use the HTTP client/async interfaces to make HTTP calls
- [x] Test download of local file
- [x] Test download of SFTP
- [x] Test download of FTP
- [x] Test downloading of non-multithreaded downloads
- [x] Test download with external object store using AWS profile on machine
- [x] Test download of STS enabled storage with S3

Mermaid diagram describing how the multi-threaded logic is called:

```mermaid
sequenceDiagram
    title Multi-Threaded Download Process with Retry Mechanism

    actor Client as Client
    participant download_functions as download_functions
    participant download_async as download_async
    participant download_execution as download_execution
    participant multi_threaded_download as multi_threaded_download
    participant remote_storage_server as remote_storage_server
    participant file as file

    activate Client
    Client ->> download_functions: download_by_file_handle
    activate download_functions

    loop retryable

        alt Download type = multi_threaded
            note over download_functions: download_from_url_multi_threaded

            download_functions ->> download_async: download_file
            activate download_async

            download_async ->> download_async: _generate_stream_and_write_chunk_tasks


            loop for each download task
                download_async ->> download_execution: _execute_download_tasks
                activate download_execution

                par MULTI-THREADED: Run in thread executor
                    download_execution ->> multi_threaded_download: _stream_and_write_chunk
                    activate multi_threaded_download

                    loop stream chunk into memory
                        multi_threaded_download ->> remote_storage_server: stream chunk from remote server
                        remote_storage_server -->> multi_threaded_download: Return partial range
                    end

                    note over multi_threaded_download: Chunk loaded into memory

                    alt obtain thread lock [Failed]
                        note over multi_threaded_download: Wait to obtain lock
                    else obtain thread lock [Success]
                        multi_threaded_download ->> file: write chunk to file
                        file -->> multi_threaded_download: .
                        note over multi_threaded_download: Update progress bar
                        note over multi_threaded_download: Release lock
                    end
                    multi_threaded_download -->> download_execution: .
                end
                download_execution -->> download_async: .
                note over download_async: Run garbage collection every 100 iterations
                deactivate multi_threaded_download
                deactivate download_execution
            end

            download_async -->> download_functions: .
            deactivate download_async

            download_functions ->> download_functions: md5_for_file
            download_functions -->> Client: File downloaded
            deactivate download_functions
        else Download type = non multi_threaded
            note over download_functions: Execute `download_from_url`
        else Download type = external s3 object store
            note over download_functions: Execute `S3ClientWrapper.download_file`
        else Download type = aws s3 sts storage
            note over download_functions: Execute `S3ClientWrapper.download_file` with with_boto_sts_credentials
        end
    end

    deactivate Client
```

### DIFF
--- a/synapseclient/api/entity_services.py
+++ b/synapseclient/api/entity_services.py
@@ -3,7 +3,7 @@
 """
 
 import json
-from typing import TYPE_CHECKING, Any, Dict, Optional, Union
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 
 from async_lru import alru_cache
 
@@ -14,6 +14,7 @@ if TYPE_CHECKING:
 async def post_entity(
     request: Dict[str, Any],
     generated_by: Optional[str] = None,
+    *,
     synapse_client: Optional["Synapse"] = None,
 ) -> Dict[str, Any]:
     """
@@ -44,6 +45,7 @@ async def put_entity(
     request: Dict[str, Any],
     new_version: bool = False,
     generated_by: Optional[str] = None,
+    *,
     synapse_client: Optional["Synapse"] = None,
 ) -> Dict[str, Any]:
     """
@@ -104,7 +106,7 @@ async def get_entity(
 
 @alru_cache(ttl=60)
 async def get_upload_destination(
-    entity_id: str, synapse_client: Optional["Synapse"] = None
+    entity_id: str, *, synapse_client: Optional["Synapse"] = None
 ) -> Dict[str, Union[str, int]]:
     """
     <https://rest-docs.synapse.org/rest/GET/entity/id/uploadDestination.html>
@@ -128,7 +130,7 @@ async def get_upload_destination(
 
 
 async def get_upload_destination_location(
-    entity_id: str, location: str, synapse_client: Optional["Synapse"] = None
+    entity_id: str, location: str, *, synapse_client: Optional["Synapse"] = None
 ) -> Dict[str, Union[str, int]]:
     """
     <https://rest-docs.synapse.org/rest/GET/entity/id/uploadDestination/storageLocationId.html>
@@ -153,7 +155,7 @@ async def get_upload_destination_location(
 
 
 async def create_access_requirements_if_none(
-    entity_id: str, synapse_client: Optional["Synapse"] = None
+    entity_id: str, *, synapse_client: Optional["Synapse"] = None
 ) -> None:
     """
     Checks to see if the given entity has access requirements. If not, then one is added
@@ -187,6 +189,7 @@ async def create_access_requirements_if_none(
 
 async def delete_entity_generated_by(
     entity_id: str,
+    *,
     synapse_client: Optional["Synapse"] = None,
 ) -> None:
     """
@@ -202,4 +205,57 @@ async def delete_entity_generated_by(
     client = Synapse.get_client(synapse_client=synapse_client)
     return await client.rest_delete_async(
         uri=f"/entity/{entity_id}/generatedBy",
+    )
+
+
+async def get_entity_path(
+    entity_id: str,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, List[Dict[str, Union[str, int, bool]]]]:
+    """
+    Implements:
+    <https://rest-docs.synapse.org/rest/GET/entity/id/path.html>
+
+    Arguments:
+        entity_id: The ID of the entity.
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+    Returns:
+        Entity paths matching:
+        <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityPath.html>
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(
+        uri=f"/entity/{entity_id}/path",
+    )
+
+
+async def get_entities_by_md5(
+    md5: str,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Union[int, List[Dict[str, Any]]]]:
+    """
+    Implements:
+    <https://rest-docs.synapse.org/rest/GET/entity/md5/md5.html>
+
+    Arguments:
+        md5: The MD5 of the entity.
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+    Returns:
+        Paginated results of:
+        <https://rest-docs.synapse.org/rest/org/sagebionetworks/reflection/model/PaginatedResults.html>
+        <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityHeader.html>
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(
+        uri=f"/entity/md5/{md5}",
     )
--- a/synapseclient/api/entity_services.py
+++ b/synapseclient/api/entity_services.py
@@ -3,7 +3,8 @@
 """
 
 import json
-from typing import Any, Dict, Optional, Union, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, Dict, Optional, Union
+
 from async_lru import alru_cache
 
 if TYPE_CHECKING:
--- a/synapseclient/api/entity_services.py
+++ b/synapseclient/api/entity_services.py
@@ -75,6 +75,8 @@ async def put_entity(
 
 async def get_entity(
     entity_id: str,
+    version_number: int = None,
+    *,
     synapse_client: Optional["Synapse"] = None,
 ) -> Dict[str, Any]:
     """
@@ -90,9 +92,14 @@ async def get_entity(
     from synapseclient import Synapse
 
     client = Synapse.get_client(synapse_client=synapse_client)
-    return await client.rest_get_async(
-        uri=f"/entity/{entity_id}",
-    )
+    if version_number:
+        return await client.rest_get_async(
+            uri=f"/entity/{entity_id}/version/{version_number}",
+        )
+    else:
+        return await client.rest_get_async(
+            uri=f"/entity/{entity_id}",
+        )
 
 
 @alru_cache(ttl=60)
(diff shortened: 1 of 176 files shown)


synapseclient/core/download/download_async.py
```diff
@@ -220,104 +205,59 @@ def _pre_signed_url_expiration_time(url: str) -> datetime:
...
-        time_made, ISO_AWS_STR_FORMAT
-    )
-    expires: str = parsed_query["X-Amz-Expires"][0]
-    return time_made_datetime + datetime.timedelta(seconds=int(expires))
+    parsed_query = parse_qs(urlparse(url).query)
+    time_made = parsed_query["X-Amz-Date"][0]
+    time_made_datetime = datetime.datetime.strptime(time_made, ISO_AWS_STR_FORMAT)
+    expires = parsed_query["X-Amz-Expires"][0]
+    return_data = time_made_datetime + datetime.timedelta(seconds=int(expires))
+    if return_data.tzinfo is None:
```
thomasyu888, line 213: When would this be None?

synapseclient/core/download/download_async.py
thomasyu888, line 1: I think it would be valuable to walk through this code as something educational for the team!  In particular, I like how you use async to submit jobs to the executor.  That said, is that that connected to the `threading.lock` which is a separate from an `async` function?

BryanFauble: Do you think the code spaces meeting is a good location to cover this? I am going to put together a sequence diagram regardless, I think it'll help to show how asyncio is kicking off the multi-threaded process of downloading.

synapseclient/core/download/download_async.py
BryanFauble: I updated this to be:
```
    return_data = (
        time_made_datetime + datetime.timedelta(seconds=int(expires))
    ).replace(tzinfo=datetime.timezone.utc)
```

By default all datetime objects in Python are not TZ aware (Bad decision IMO), but this lets everything down stream know for certain what TZ this is expected to be.

synapseclient/core/download/download_async.py
```diff
@@ -326,169 +266,286 @@ class _MultithreadedDownloader:
...
+            getattr(_thread_local, "progress_bar", None) is None
+        )
 
-    def download_file(self, request: DownloadRequest) -> None:
+    async def download_file(self) -> None:
         """
         Splits up and downloads a file in chunks from a URL.
 
         Arguments:
             request: A DownloadRequest object specifying what Synapse file to download.
```
jaymedina, line 296: Request is no longer an arg and should be removed

synapseclient/core/download/download_async.py
```diff
@@ -326,169 +266,286 @@ class _MultithreadedDownloader:
...
+            )
+            download_tasks = pending_tasks
+            for completed_task in done_tasks:
+                try:
+                    start_bytes, end_bytes = completed_task.result()
+                    del completed_task
+                    self._syn._parts_transfered_counter += 1
+
+                    # Garbage collect every 100 iterations
+                    if self._syn._parts_transfered_counter % 100 == 0:
```
jaymedina, line 337: Would there ever be a case where the total # of parts the file has been split into never exceeds 100 and therefore garbage collection never happens?

synapseclient/core/download/download_async.py
```diff
@@ -326,169 +266,286 @@ class _MultithreadedDownloader:
...
+
+    def _create_progress_bar(self, file_size: int) -> None:
+        """Handle creating the progress bar.
 
-            try:
-                os.remove(request.path)
-            except FileNotFoundError:
-                pass
+        Arguments:
+            file_size: The size of the file to download
```
jaymedina, line 376: To clarify: This `file_size` is the size of a file _or_ part, correct?

synapseclient/core/download/download_async.py
```diff
@@ -326,169 +266,286 @@ class _MultithreadedDownloader:
...
+                )
 
-        return start, response
+    def _stream_and_write_chunk(
+        self,
+        session: httpx.Client,
+        presigned_url_provider: PresignedUrlProvider,
+        start: int,
+        end: int,
+        otel_context: Union[Context, None],
```
jaymedina, line 461: do we want to set the default to `otel_context` as `None` here?

synapseclient/models/activity.py
```diff
@@ -55,6 +67,20 @@ class UsedURL:
     """The external URL of the file that was used such as a link to a GitHub commit
     or a link to a specific version of a software tool."""
 
+    def format_for_manifest(self) -> str:
```
jaymedina, line 70: There are 2 functions with this same name and docstring, is one meant to be replaced?

synapseutils/sync.py
```diff
@@ -177,334 +167,170 @@ def syncFromSynapse(
...
+            entity=entity_bundle["entity"],
+            path=path,
+            if_collision=if_collision,
+            follow_link=follow_link,
+            download_file=download_file,
+            manifest=manifest,
+        )
+    else:
+        raise ValueError(
+            "Cannot initiate a sync from an entity that is not a File, Folder, or Link to a File/Folder."
```
jaymedina, line 330: `...that is not a File, Folder, Project, or Link`

synapseclient/api/entity_services.py
```diff
@@ -75,6 +75,8 @@ async def put_entity(
 
 async def get_entity(
     entity_id: str,
```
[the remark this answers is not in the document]
BryanFauble: It does - however I would like to keep this specific function working like this so it's a consistent and documented experience. Where this function is being called is handling for this "." notation:
```
    if is_synapse_id_str(entity):
        # ensure that we seed with an actual entity
        synid, version = get_synid_and_version(obj=entity)
        entity = await get_entity(
            entity_id=synid, version_number=version, synapse_client=syn
        )
```

synapseclient/core/download/download_async.py
BryanFauble: Yes it would occur if the file is not 800MIB or larger. But in this case it'll get cleaned up during normal GC runs.

It isn't an issue in this case as other parts of the code have enough references that get created/released that'll cause GC to kick off it's normal garbage culling

synapseclient/core/download/download_async.py
BryanFauble: Removed in the second PR
https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1111

synapseclient/models/file.py
```diff
@@ -167,6 +168,30 @@ def fill_from_dict(
 
         return self
 
+    def _convert_into_legacy_file_handle(self) -> Dict[str, Union[str, bool, int]]:
+        """Convert the file handle object into a legacy File Handle object."""
```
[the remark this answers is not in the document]
BryanFauble: The current syncFromSynapse function is returning the "legacy" FIle class object. This was a way to be able to change the internal implementation, but keep the same interface.

synapseclient/core/download/download_async.py
BryanFauble: This was the total size of the file. I've changed this implementation a bit after getting a context managed TQDM bar for the sync function. See https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1111/files#diff-29894805101b443f78e9548b65256eaff42ea1fd15290de9ab09857dac4765a1 for the new implementation

synapseclient/core/download/download_async.py
BryanFauble: Since this is a "private" function it doesn't matter much as I am always passing in `Union[Context, None]` in this case. I'll add a doc string to this though

synapseclient/models/activity.py
BryanFauble: No. They're on 2 different dataclasses:
`UsedEntity` and `UsedURL`. The way the data is going to be loaded into the manifest is different so they have different implementations

jaymedina:
Looks good so far. Are you planning to merge #1111 into this one for a final review?

BryanFauble:
> Looks good so far. Are you planning to merge #1111 into this one for a final review?

@jaymedina What's your opinion on this? #1111 and #1107 could be reviewed/approved separately, or go through a final review process like you mention.

jaymedina:
@BryanFauble I like the idea of merging the chunked PRs into a final one, after they've been reviewed and approved, to make sure the changes jive well together. Let's try that out for this case and go from there!

jaymedina (approved):
LGTM! 🥇

PR #1111, opened 2024-06-20
[SYNPY-1476] Remove thread executor usage for Project, Folder, File

BryanFauble:
**Problem:**

1. Within Project, Folder, File, in order to wrap the existing Synapse Python Client `.get()` function we were running the code within the asyncio `run_in_executor`. This worked fine up until now. The problem at this point is that utility functions are now going to be directly using these new model functions and it leads to a problem where we are changing from a non-async context (Before SynapseUtils) -> Async context (Inside the model) -> Non-Async context (Running in a thread executor) -> Async context (To download a file) -> Multi-threaded download.
2. The progress bars for both individual file downloads, and syncFromSynapse was broken and needed to be mended.

**Solution:**

1. Breaking out the logic that is used to retrieve entities. This new logic is now following a factory-type pattern allowing us to easily add more entities to it.
3. Updating the Project, Folder, File, functions to use the new factory-type pattern to retrieve the entity data.
4. Updating the logic around shared and individual download progress bars

**Testing:**

1. Integration test runs
2. Manual verification via steps laid out in jira comment of: https://sagebionetworks.jira.com/browse/SYNPY-1476

### DIFF
--- a/synapseclient/api/entity_factory.py
+++ b/synapseclient/api/entity_factory.py
@@ -178,9 +178,10 @@ async def _search_for_file_by_md5(
     from synapseclient import Synapse
 
     syn = Synapse.get_client(synapse_client=synapse_client)
+    md5 = md5 or utils.md5_for_file_hex(filename=filepath)
     results = (
         await get_entities_by_md5(
-            md5=md5 or utils.md5_for_file_hex(filename=filepath),
+            md5=md5,
             synapse_client=synapse_client,
         )
     )["results"]
@@ -215,7 +216,9 @@ async def _search_for_file_by_md5(
         version=entity_version,
         synapse_client=synapse_client,
     )
-    syn.cache.add(file_handle_id=bundle["entity"]["dataFileHandleId"], path=filepath)
+    syn.cache.add(
+        file_handle_id=bundle["entity"]["dataFileHandleId"], path=filepath, md5=md5
+    )
 
     return bundle
 
--- a/synapseclient/api/entity_factory.py
+++ b/synapseclient/api/entity_factory.py
@@ -1,0 +1,386 @@
+"""Factory type functions to create and retrieve entities from Synapse"""
+
+import os
+from typing import TYPE_CHECKING, Any, Dict, Optional, Union
+
+from opentelemetry import trace
+
+from synapseclient.api import (
+    get_entities_by_md5,
+    get_entity_id_bundle2,
+    get_entity_id_version_bundle2,
+    get_entity_path,
+)
+from synapseclient.core import utils
+from synapseclient.core.constants import concrete_types
+from synapseclient.core.download import download_file_entity_model
+from synapseclient.core.exceptions import (
+    SynapseFileNotFoundError,
+    SynapseUnmetAccessRestrictions,
+)
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+    from synapseclient.models import File, Folder, Project
+
+
+async def get_from_entity_factory(
+    synapse_id_or_path: str,
+    version: int = None,
+    if_collision: str = "keep.both",
+    limit_search: str = None,
+    md5: str = None,
+    download_file: bool = True,
+    download_location: str = None,
+    follow_link: bool = False,
+    entity_to_update: Union["Project", "File", "Folder"] = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Union["Project", "File", "Folder"]:
+    """
+    Factory type function to retrieve an entity from Synapse. Optionally you may also
+    pass in `entity_to_update` if you want to update the fields on the existing entity
+    instead of creating a new instance.
+
+    Arguments:
+        synapse_id_or_path: The Synapse ID or file path of the entity to retrieve.
+        version:            The version number of the entity to retrieve.
+        if_collision: Determines how to handle file collisions. May be:
+
+                - `overwrite.local`
+                - `keep.local`
+                - `keep.both`
+        limit_search:       Limit the search to a specific project or folder. Only used
+            if `synapse_id_or_path` is a path.
+        md5: The MD5 of the file to retrieve. If not passed in, the MD5 will be
+            calculated. Only used if `synapse_id_or_path` is a path.
+        download_file: Whether associated files should be downloaded.
+        download_location: The directory to download the file to.
+        follow_link: Whether to follow a link to its target. This will only do a single
+            hop to the target of the link.
+        entity_to_update: An existing entity class instance to update with data from
+            Synapse.
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+        Example: Using this function
+            Download file into cache
+
+                from synapseclient import Synapse
+                from synapseclient.api import get_from_entity_factory
+
+                syn = Synapse()
+                syn.login()
+
+                entity = await get_from_entity_factory(synapse_id_or_path='syn1906479')
+                print(entity.name)
+                print(entity.path)
+
+            Download file into current working directory
+
+                from synapseclient import Synapse
+                from synapseclient.api import get_from_entity_factory
+
+                syn = Synapse()
+                syn.login()
+
+                entity = await get_from_entity_factory(synapse_id_or_path='syn1906479', download_location='.')
+                print(entity.name)
+                print(entity.path)
+
+    Raises:
+        SynapseFileNotFoundError: If the id is not a synapse ID and it is not a valid
+            file path.
+    """
+
+    # If entity is a local file determine the corresponding synapse entity
+    if isinstance(synapse_id_or_path, str) and os.path.isfile(synapse_id_or_path):
+        bundle = await _search_for_file_by_md5(
+            filepath=synapse_id_or_path,
+            limit_search=limit_search,
+            md5=md5,
+            synapse_client=synapse_client,
+        )
+        download_file = False
+
+    elif isinstance(synapse_id_or_path, str) and not utils.is_synapse_id_str(
+        obj=synapse_id_or_path
+    ):
+        raise SynapseFileNotFoundError(
+            (
+                f"The parameter {synapse_id_or_path} is neither a local file path "
+                " or a valid entity id"
+            )
+        )
+    else:
+        synid_and_version = utils.get_synid_and_version(obj=synapse_id_or_path)
+        version = version if version is not None else synid_and_version[1]
+        if version:
+            bundle = await get_entity_id_version_bundle2(
+                entity_id=synid_and_version[0],
+                version=version,
+                synapse_client=synapse_client,
+            )
+        else:
+            bundle = await get_entity_id_bundle2(
+                entity_id=synid_and_version[0], synapse_client=synapse_client
+            )
+
+    # Check and warn for unmet access requirements
+    _check_entity_restrictions(
+        bundle=bundle, synapse_id=bundle["entity"]["id"], download_file=download_file
+    )
+
+    return_data = await _cast_into_class_type(
+        entity_to_update=entity_to_update,
+        entity_bundle=bundle,
+        download_file=download_file,
+        download_location=download_location,
+        if_collision=if_collision,
+        follow_link=follow_link,
+        synapse_client=synapse_client,
+    )
+    trace.get_current_span().set_attributes(
+        {
+            "synapse.id": return_data.id,
+            "synapse.type": return_data.__class__.__name__,
+        }
+    )
+    return return_data
+
+
+async def _search_for_file_by_md5(
+    filepath: str,
+    md5: str,
+    limit_search: str = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """
+    Handle using md5 for a local file to search through Synapse to find a match. By
+    default this will search through every entity that you as a user have access to.
+    However, you can limit the search to a specific project or folder by passing in the
+    Synapse ID of the project or folder into the `limit_search` field.
+
+    Arguments:
+        filepath: The path to the file to search for.
+        md5: The MD5 of the file to retrieve. If not passed in, the MD5 will be
+            calculated.
+        limit_search: Limit the search to a specific project or folder.
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+    Returns:
+        A dictionary containing the entity bundle of the file found.
+
+    Raises:
+        SynapseFileNotFoundError: If the file is not found in Synapse.
+    """
+    from synapseclient import Synapse
+
+    syn = Synapse.get_client(synapse_client=synapse_client)
+    results = (
+        await get_entities_by_md5(
+            md5=md5 or utils.md5_for_file_hex(filename=filepath),
+            synapse_client=synapse_client,
+        )
+    )["results"]
+    if limit_search is not None:
+        # Go through and find the path of every entity found
+        paths = [
+            await get_entity_path(entity_id=ent["id"], synapse_client=synapse_client)
+            for ent in results
+        ]
+        # Filter out all entities whose path does not contain limitSearch
+        results = [
+            ent
+            for ent, path in zip(results, paths)
+            if utils.is_in_path(id=limit_search, path=path)
+        ]
+    if len(results) == 0:  # None found
+        raise SynapseFileNotFoundError(f"File {filepath} not found in Synapse")
+    elif len(results) > 1:
+        id_txts = "\n".join([f"{r['id']}.{r['versionNumber']}" for r in results])
+        syn.logger.warning(
+            f"\nThe file {filepath} is associated with many files in Synapse:\n{id_txts}\n"
+            "You can limit to files in specific project or folder by setting the limitSearch to the"
+            " synapse Id of the project or folder.\n"
+            "Will use the first one returned: \n"
+            f"{results[0]['id']} version {results[0]['versionNumber']}\n"
+        )
+    entity_id = results[0]["id"]
+    entity_version = results[0]["versionNumber"]
+
+    bundle = await get_entity_id_version_bundle2(
+        entity_id=entity_id,
+        version=entity_version,
+        synapse_client=synapse_client,
+    )
+    syn.cache.add(file_handle_id=bundle["entity"]["dataFileHandleId"], path=filepath)
+
+    return bundle
+
+
+async def _cast_into_class_type(
+    entity_bundle: Dict[str, Any],
+    download_file: bool = True,
+    download_location: str = None,
+    if_collision: str = None,
+    submission: str = None,
+    follow_link: bool = False,
+    entity_to_update: Union["Project", "File", "Folder"] = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Union["Project", "File", "Folder"]:
+    """
+    Take an entity_bundle returned from the Synapse API and cast it into the appropriate
+    class type. This will also download the file if `download_file` is set to True.
+    Additionally, if `entity_to_update` is passed in, the fields of the existing entity
+    will be updated instead of creating a new instance. If the entity is a link and
+    `follow_link` is set to True, the target entity will be retrieved.
+
+    Arguments:
+        entity_bundle: The entity bundle to cast into a class type.
+        download_file: Whether associated files should be downloaded.
+        download_location: The directory to download the file to.
+        if_collision: Determines how to handle file collisions. May be:
+
+                    - `overwrite.local`
+                    - `keep.local`
+                    - `keep.both`
+        submission: The ID of the submission to which the entity belongs.
+        follow_link: Whether to follow a link to its target. This will only do a single
+            hop to the target of the link.
+        entity_to_update: An existing entity class instance to update with data from
+            Synapse.
+        synapse_client: If not passed in or None this will use the last client from
+            the `.login()` method.
+
+    Returns:
+        A Synapse entity object.
+
+    Raises:
+        ValueError: If the entity type is not supported.
+    """
+    from synapseclient import Synapse
+
+    syn = Synapse.get_client(synapse_client=synapse_client)
+
+    # If Link, get target ID entity bundle
+    if (
+        entity_bundle["entity"]["concreteType"] == concrete_types.LINK_ENTITY
+        and follow_link
+    ):
+        target_id = entity_bundle["entity"]["linksTo"]["targetId"]
+        target_version = entity_bundle["entity"]["linksTo"].get("targetVersionNumber")
+        entity_bundle = await get_entity_id_version_bundle2(
+            entity_id=target_id, version=target_version, synapse_client=synapse_client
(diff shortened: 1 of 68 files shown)


tests/integration/synapseclient/models/async/test_file_async.py
```diff
@@ -1225,7 +1225,7 @@ async def test_delete(
         # THEN I expect the file to be deleted
         with pytest.raises(SynapseHTTPError) as e:
             await file.get_async()
-        assert f"404 Client Error: \nEntity {file.id} is in trash can." in str(e.value)
+        assert f"404 Client Error: Entity {file.id} is in trash can." in str(e.value)
```
BryanFauble, line 1228: Not exactly sure why the newline character is removed in the real exception being raised. The rest of the exception still has the expected new lines in it.

synapseclient/models/file.py
```diff
@@ -660,15 +661,16 @@ async def _load_local_md5(self, syn: "Synapse") -> None:
...
+            # self.content_md5 = await utils.md5_for_file_multiprocessing(
+            #     filename=self.path,
+            #     process_pool_executor=syn._get_process_pool_executor(
+            #         asyncio_event_loop=asyncio.get_running_loop()
+            #     ),
+            #     md5_semaphore=syn._get_md5_semaphore(
+            #         asyncio_event_loop=asyncio.get_running_loop()
+            #     ),
+            # )
+            self.content_md5 = utils.md5_for_file_hex(filename=self.path)
```
BryanFauble, line 673: I want to re-visit the usage of using multi-processing for this. There is a tipping point somewhere that the time to start up a new process actually takes longer than just running this in the entrant python process.

Also, I was using `top` to look at resource usage and overall system usage is significant lower when not running this in a second process.

synapseclient/core/download/download_async.py
```diff
@@ -156,14 +173,32 @@ def _get_pre_signed_info(self) -> PresignedUrlInfo:
         Returns:
             Information about a retrieved presigned-url from a new request.
         """
-        response = wrap_async_to_sync(
```
BryanFauble, line 159: I needed to take away this `wrap_async_to_sync`. There are some issues it appears when running this many times. From what I've found out we should only be using this async -> sync conversion once.

There are 2 issues otherwise:

1. stack overflows can occur
2. `nest_asyncio` library doesn't work and raises exceptions in some of the libraries that it is patching

synapseclient/api/entity_factory.py
```diff
@@ -0,0 +1,391 @@
...
+    synapse_id_or_path: str,
+    version: int = None,
+    if_collision: str = "keep.both",
+    limit_search: str = None,
+    md5: str = None,
+    download_file: bool = True,
+    download_location: str = None,
+    follow_link: bool = False,
+    entity_to_update: Union["Project", "File", "Folder"] = None,
+    *,
```
thomasyu888, line 36: I like this factory class!  It's a lot more readable than the old code was as it has a clear series of steps that it does.

Just curious, what was the rationale behind the `*`?

synapseclient/api/entity_factory.py
```diff
@@ -0,0 +1,391 @@
...
+        )
+    entity = entity_bundle["entity"]
+
+    from synapseclient.models import Annotations
+
+    annotations = Annotations.from_dict(
+        synapse_annotations=entity_bundle.get("annotations", None)
+    )
+
+    from synapseclient.models import File, Folder, Project
```
thomasyu888, line 289: I think you already import this up top?

synapseclient/api/entity_factory.py
BryanFauble: @Sage-Bionetworks/dpe This would be a great spot to get some feedback from the team.

What the asterisk in the parameters means:

> You use a bare asterisk to define a boundary between arguments that you can pass by either position or keyword from those that you must pass by keyword.
image


Here my my thought process:

1. By requiring that `synapse_client` is always a KW only argument it keeps a consistent experience across every function.
2. Passing `synapse_client` into a function is not meant to be the norm, but rather the exception to allow for you to use multiple logged in synapse client sessions during a single script execution run. Providing this functionality is to allow for a use case where a single running server instance might be providing the interface for multiple user accounts.
3. In the future when we want to add more arguments to a function then we will never have a case where `synapse_client` ends up being in the "middle" of several other positional arguments.

What is everyone thought on using `*` across the board for this purpose? Is this something you support, think would be confusing, or is a bad decision?

synapseclient/api/entity_factory.py
BryanFauble: I am - however it's behind a `TYPE_CHECKING` if statement:
```

if TYPE_CHECKING:
    from synapseclient.models import File, Folder, Project
```

What it means in this context is that I can reference the type in an argument of a function like:
```
def my_function(file: "File") -> None:
```

However I cannot create an instance of a File like this, the following is NOT valid:
```
def my_function(file: "File") -> None:
    instance_of_file = File()
```
image


This is another instance of lazy importing to prevent circular dependency trees. Because `File, Folder, Project` themselves are importing this factory function to be called from within their respective `get_async()` function calls.

synapseclient/api/entity_factory.py
jaymedina: @BryanFauble I think the reasoning is sound and I'm for turning it into a kw-only arg, so long as the documentation examples using the `synapse_client` keyword are updated to explain this. I imagine it would be a bit of a clumsy transition for users who are not as experienced with keyword-only args and already have pre-existing scripts where they use `synapse_client` in the old way.

thomasyu888 (commented):
Awesome work here!

tests/integration/synapseclient/models/async/test_file_async.py
BryanFauble: I might have also slightly altered the format in https://github.com/Sage-Bionetworks/synapsePythonClient/blob/SYNPY-1476-remove-thread-executor-usage/synapseclient/core/exceptions.py#L189-L279 when I wrote it a few months ago, and now when we've swapped to HTTPX

synapseclient/models/file.py
BryanFauble: I created https://sagebionetworks.jira.com/browse/SYNPY-1489 to look back at this

synapseclient/api/entity_factory.py
```diff
@@ -0,0 +1,391 @@
...
+        synapse_client=synapse_client,
+    )
+    syn.cache.add(
+        file_handle_id=bundle["entity"]["dataFileHandleId"], path=filepath, md5=md5
+    )
+
+    return bundle
+
+
+async def _cast_into_class_type(
```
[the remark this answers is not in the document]
BryanFauble: Yes - I am planning for this to support anything with a SYNid here. It'll take some time to support them all though

synapseclient/api/file_services.py
```diff
@@ -422,3 +431,64 @@ async def get_file_handle_for_download(
...
+    )
+
+    result = response["requestedFiles"][0]
+    failure = result.get("failureCode")
+    if failure == "NOT_FOUND":
+        raise SynapseFileNotFoundError(
+            f"The fileHandleId {file_handle_id} could not be found"
+        )
+    elif failure == "UNAUTHORIZED":
+        raise SynapseError(
```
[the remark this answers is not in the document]
BryanFauble: I am going to create `SynapseAuthorizationError` and use it here

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
     entity.cacheDir = os.path.dirname(download_path)
 
 
+async def download_file_entity_model(
+    *,
```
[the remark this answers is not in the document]
BryanFauble: Yes. The original function I wrote this after did this so I had just kept that in.

synapseclient/core/multithread_download/download_threads.py
```diff
@@ -170,21 +168,16 @@ def _get_pre_signed_info(self) -> PresignedUrlInfo:
...
+            self.request.object_id,
+            objectType=self.request.object_type,
         )
         file_name = response["fileHandle"]["fileName"]
         pre_signed_url = response["preSignedURL"]
         return PresignedUrlInfo(
-            file_name=file_name,
-            url=pre_signed_url,
-            expiration_utc=_pre_signed_url_expiration_time(pre_signed_url),
+            file_name, pre_signed_url, _pre_signed_url_expiration_time(pre_signed_url)
```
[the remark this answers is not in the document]
BryanFauble: For this I reverted the logic back to it was before I started touching this. Rather than patching the logic I just went back to what it was; Will be removing this in v5 when we remove the deprecated code (This class is deprecated)

synapseclient/api/entity_factory.py
```diff
@@ -0,0 +1,391 @@
...
+    if_collision: str = "keep.both",
+    limit_search: str = None,
+    md5: str = None,
+    download_file: bool = True,
+    download_location: str = None,
+    follow_link: bool = False,
+    entity_to_update: Union["Project", "File", "Folder"] = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Union["Project", "File", "Folder"]:
```
jaymedina, line 38: Is there a plan to support Links in the future?

BryanFauble: That's the plan. I'll want to eventually support everything that has a synapseID here. 


We can do more to extract some functionality in this class to lower the cognitive complexity, but, we'll worry about doing that refactor when we add more supported entities here.

synapseclient/api/entity_factory.py
```diff
@@ -0,0 +1,391 @@
...
+        raise SynapseFileNotFoundError(
+            (
+                f"The parameter {synapse_id_or_path} is neither a local file path "
+                " or a valid entity id"
+            )
+        )
+    else:
+        synid_and_version = utils.get_synid_and_version(obj=synapse_id_or_path)
+        version = version if version is not None else synid_and_version[1]
+        if version:
```
jaymedina, line 117: Why do we need this `if version` conditional here? It looks like `version` gets assigned something in the previous line no matter what.

BryanFauble: It does get assigned, but it can be assigned to `None`. The return type of `get_synid_and_version` is:
`typing.Tuple[str, typing.Union[int, None]]`

jaymedina: Thanks for clarifying!

synapseclient/api/entity_factory.py
```diff
@@ -0,0 +1,391 @@
...
+    follow_link: bool = False,
+    entity_to_update: Union["Project", "File", "Folder"] = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Union["Project", "File", "Folder"]:
+    """
+    Take an entity_bundle returned from the Synapse API and cast it into the appropriate
+    class type. This will also download the file if `download_file` is set to True.
+    Additionally, if `entity_to_update` is passed in, the fields of the existing entity
+    will be updated instead of creating a new instance. If the entity is a link and
```
jaymedina, line 241: nit: "If the entity is a link..." but links are not supported yet. I worry this will lead to confusion, unless you are planning on including Links pretty soon

BryanFauble: Links are supported enough to be included here. We are not returning a Link entity yet, however, we are supporting following a link to find a target entity.

For example links that look like the following are supported:

1. Link -> Project
2. Link -> Folder
3. Link -> File

If link is not pointing to one of those resources a `ValueError` is raised.

jaymedina: Makes sense! Thanks

synapseclient/core/remote_file_storage_wrappers.py
```diff
@@ -72,7 +75,7 @@ def download_file(
         *,
         profile_name: str = None,
         credentials: typing.Dict[str, str] = None,
-        show_progress: bool = True,
+        progress_bar: tqdm = None,
```
jaymedina, line 78: nit: shouldn't `progress_bar` type-hint be `Union[tqdm, None]`

synapseclient/models/project.py
```diff
@@ -365,27 +368,18 @@ async def get_async(
...
-        entity = await loop.run_in_executor(
-            None,
-            lambda: run_and_attach_otel_context(
-                lambda: Synapse.get_client(synapse_client=synapse_client).get(
-                    entity=entity_id,
-                ),
-                current_context,
-            ),
+        await get_from_entity_factory(
+            entity_to_update=self,
```
jaymedina, line 372: is this an incomplete attribute?

BryanFauble: no, it's intentional. This argument is `entity_to_update: Union["Project", "File", "Folder"] = None`. So in this case since it used within one of those classes I pass `self` into this so updates occur on the class instance passed in.

synapseclient/core/remote_file_storage_wrappers.py
BryanFauble: Updated

synapseclient/core/download/download_functions.py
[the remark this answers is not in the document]
jaymedina: Is this something we want to add later, though? I don't know how I feel about some funcs having all kw-only args, and some not having that requirement. It's a bit confusing from the user pov.

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+    synapse_client: Optional["Synapse"] = None,
+) -> None:
+    """
+    Download file entity
+
+    Arguments:
+        download_location: The download location
+        entity:           The File object
+        if_collision:      Determines how to handle file collisions.
+                            May be
```
jaymedina, line 188: This docstring is incomplete. "May be ..." - also I would suggest a more detailed description of the other definitions

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+
+            - `overwrite.local`
+            - `keep.local`
+            - `keep.both`
+
+        submission:       Access associated files through a submission rather than through an entity.
```
jaymedina, line 194: `synapse_client` arg is missing. Was this added in 1107?

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+    file: "File",
+    if_collision: str,
+    submission: str,
+    synapse_client: Optional["Synapse"] = None,
+) -> None:
+    """
+    Download file entity
+
+    Arguments:
+        download_location: The download location
```
jaymedina, line 185: Would suggest making the docstrings more detailed since they also serve as documentation for users. e.g "The location on your local machine where the entity will be downloaded. Note that this location will be searched for any previous versions of the entity that were downloaded already, and will be handled with the `if_collision` argument"

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+
+    # location in .synapseCache where the file would be corresponding to its FileHandleId
+    synapse_cache_location = client.cache.get_cache_dir(
+        file_handle_id=file.data_file_handle_id
+    )
+
+    file_name = (
+        file.file_handle.file_name
+        if cached_file_path is None
+        else os.path.basename(cached_file_path)
```
jaymedina, line 217: Ah nice. This `else` statement is meant to catch for cases where a user re-named the file right?

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+
+    # Decide the best download location for the file
+    if download_location is not None:
+        # Make sure the specified download location is a fully resolved directory
+        download_location = ensure_download_location_is_directory(download_location)
+    elif cached_file_path is not None:
+        # file already cached so use that as the download location
+        download_location = os.path.dirname(cached_file_path)
+    else:
+        # file not cached and no user-specified location so default to .synapseCache
```
jaymedina, line 228: nit: would be nice to have a warning output message notifying the user of this behavior

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+        with logging_redirect_tqdm(loggers=[client.logger]):
+            download_path = await download_by_file_handle(
+                file_handle_id=file.data_file_handle_id,
+                synapse_id=object_id,
+                entity_type=object_type,
+                destination=download_path,
+            )
+
+        if download_path is None or not os.path.exists(download_path):
+            return
```
jaymedina, line 266: Same comment here re: no silent returns

synapseclient/core/download/download_functions.py
BryanFauble: Nope - I'll add this now

synapseclient/core/download/download_functions.py
BryanFauble: So for these functions within `core` they are not meant to be used directly by users of the python client. They're functions to be used by higher-level code. There is a disclaimer on the core python docs page that reads:

https://python-docs.synapse.org/reference/core/

> This section is for super users / developers only. These functions are subject to change as they are internal development functions. Use at your own risk.

For example `entity_factory.py` uses this, and that function has more in-depth docstrings.

With that being said: I'll update the docstring with your example since it's great!

BryanFauble:
```
        download_location: The location on disk where the entity will be downloaded. If
            there is a matching file at the location, the download collision will be
            handled according to the `if_collision` parameter.
```

synapseclient/core/download/download_functions.py
BryanFauble: The `else` in this case is when we have a "cache hit" and the file was already downloaded and present within the cache. 

As for exactly why this was put in place, I'm not certain. It was one of the items in place where this code was taken from, and I didn't want to change this behavior.

synapseclient/core/download/download_functions.py
BryanFauble: There is actually an info level message that is printed in `download_by_file_handle` that will specify the directory where the file was downloaded to:
`syn.logger.info(f"Downloaded {synapse_id} to {downloaded_path}")`

So it'll point you to the `.synapseCache` directory in this case

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+    # ifcollision value
+    download_path = resolve_download_path_collisions(
+        download_location=download_location,
+        file_name=file_name,
+        if_collision=if_collision,
+        synapse_cache_location=synapse_cache_location,
+        cached_file_path=cached_file_path,
+    )
+    if download_path is None:
+        return
```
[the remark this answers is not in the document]
BryanFauble: I would put a message inside of where `download_path` is resolved: `resolve_download_path_collisions`

How does this look to you:
image

Code snippet:
```
        elif if_collision == COLLISION_KEEP_LOCAL:
            client.logger.info(f"Found existing file at {download_path}, skipping download.")

            # Don't want to overwrite the local file.
            download_path = None
```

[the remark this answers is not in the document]
jaymedina: Looks great - thanks!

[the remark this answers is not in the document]
BryanFauble: I suppose I might also want to add in a message when we're copying the file from cache as well. So we would have 3 messages popping up:

1. When we already found the file in the location (keep.local collision is set)
2. We have an existing file already present in cache so we perform a copy
3. The file needs to be downloaded:

image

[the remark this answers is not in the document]
BryanFauble: Added in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1111/commits/c269dc7d830d437d15d2c382f6cc10d251ba3bb9

synapseclient/core/download/download_functions.py
BryanFauble: Added in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1111/commits/c269dc7d830d437d15d2c382f6cc10d251ba3bb9

synapseclient/core/download/download_functions.py
BryanFauble: Added in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1111/commits/c269dc7d830d437d15d2c382f6cc10d251ba3bb9

synapseclient/core/download/download_functions.py
jaymedina: Excellent thanks!

synapseclient/core/download/download_functions.py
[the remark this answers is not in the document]
jaymedina: Perfect

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
     entity.cacheDir = os.path.dirname(download_path)
 
 
+async def download_file_entity_model(
+    *,
+    download_location: str,
```
jaymedina, line 181: There is a conditional further down that is `if download_location is not None:` - should we have this arg default to `None` to allow for this to happen?

synapseclient/core/download/download_functions.py
```diff
@@ -161,6 +170,105 @@ async def download_file_entity(
...
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    # set the initial local state
+    file.path = None
+
+    # check to see if an UNMODIFIED version of the file (since it was last downloaded) already exists
+    # this location could be either in .synapseCache or a user specified location to which the user previously
+    # downloaded the file
+    cached_file_path = client.cache.get(
+        file_handle_id=file.data_file_handle_id, path=download_location
```
jaymedina, line 217: Let's say I download a file `myfile.csv` and then rename it something later called `myfile_original.csv` before downloading the entity again (the entity has not changed). Because this is searching for the file via the `file_handle_id` and not the name of the file on disk, this would not lead to a download that overwrites the `myfile_original.csv` correct?

BryanFauble: So it's finding the path by `file_handle_id` and also the `download_location` passed in. In this case it would be a cache miss and it would download the file again. Example:
image

synapseclient/core/download/download_functions.py
BryanFauble: I'll update the type to `Union[str, None]`

synapseclient/core/download/download_functions.py
jaymedina: Awesome, thanks for testing this :)

synapseclient/core/download/download_functions.py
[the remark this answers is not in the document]
BryanFauble: This was swapped to just the `synapse_client` being a kwarg

PR #1112, opened 2024-06-21
[SYNPY-1457] Fix deprecated Node.js build warnings

jaymedina:
## problem

We are using versions of GitHub actions that still use`Node.js 16`, which is now deprecated and leads to deprecation warnings in the builds. This clutters build output messages which is unhelpful when troubleshooting/monitoring the status of your builds.

## solution

- [x] Update GitHub actions used in our workflows to versions that use `Node.js 20`
- [x] Update one of our integration tests in `tests/integration/synapseclient/test_tables.py` that uses `np.string_(s)` which is no longer supported with the updated actions that uses a newer `numPy`

## testing & preview

Latest run: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/9618137457

UPDATED:

Latest run: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/9620252513

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,10 +33,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
-    - uses: actions/setup-python@v4
+    - uses: actions/setup-python@v5
       with:
         python-version: '3.11'
-    - uses: pre-commit/action@v3.0.0
+    - uses: pre-commit/action@v3.0.1
 
 
   # run unit (and integration tests if account secrets available) on our build matrix
@@ -55,7 +55,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python }}
 
@@ -70,7 +70,7 @@ jobs:
 
       - name: Cache py-dependencies
         id: cache-dependencies
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         env:
           cache-name: cache-py-dependencies
         with:
@@ -197,7 +197,7 @@ jobs:
           # Execute the CLI tests in a non-dist way because they were causing some test instability when being run concurrently
           # pytest -sv --reruns 3 --cov-append --cov=. --cov-report xml tests/integration/synapseclient/test_command_line_client.py
       - name: Upload otel spans
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         if: always()
         with:
           name: otel_spans_integration_testing_${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -197,15 +197,15 @@ jobs:
...
           name: otel_spans_integration_testing_${{ matrix.os }}
           path: tests/integration/otel_spans_integration_testing_*.ndjson
           if-no-files-found: ignore
       - name: Upload coverage report
         id: upload_coverage_report
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v4
         if: ${{ contains(fromJSON('["3.9"]'), matrix.python) && contains(fromJSON('["ubuntu-20.04"]'), matrix.os)}}
         with:
           name: coverage-report
```
BryanFauble, line 211: @jaymedina Looks like there is 1 last item to update during the `sonarcloud` job:
https://github.com/actions/checkout

```

SonarCloud
Node.js 16 actions are deprecated. Please update the following actions to use Node.js 20: actions/checkout@v3. For more information see: https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/.
```

.github/workflows/build.yml
```diff
@@ -217,7 +217,7 @@ jobs:
     name: SonarCloud
     runs-on: ubuntu-20.04
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
       - name: Check coverage report existence
```
BryanFauble, line 223: I think there is a breaking change around the code coverage report that'll need to be debugged a bit further:
image

The coverage.xml does exist, but is packaged into a zip. I don't recall the exact specifics of how this was working before:
image

I believe that this was broken as a result of some changes that @danlu1 had made here:
https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1095

The code coverage graph in sonarcloud coincides with this timeline where `Covered Lines` goes to 0:
image


Would you want to resolve this issue here, or do you think another ticket is appropriate?

BryanFauble (approved):
There is an issue with code coverage that I noted, but was not introduced with these changes.

If we could also address it (Could be another ticket, or an extension of this ticket) that would be amazing!

Otherwise these changes look good to me!

.github/workflows/build.yml
jaymedina: @BryanFauble I'll make a separate Jira ticket and PR for this, for book-keeping purposes, but can definitely investigate it this week. Would you mind sharing where I can find the contents in the last two screenshots?

thomasyu888 (approved):
🔥 ! Thanks for doing this update!

.github/workflows/build.yml
BryanFauble: @jaymedina 
The coverage-report is here: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/9620252513?pr=1112

The historical code coverage graph is available in sonarcloud: https://sonarcloud.io/project/overview?id=Sage-Bionetworks_synapsePythonClient

jaymedina: Thanks! I'll be merging this then and starting this separately.

PR #1115, opened 2024-07-02
[SYNPY-1476] Benchmarking results

BryanFauble:
**Problem:**

1. We did not have benchmarking results of the download changes.
2. The download/upload test file scripts needed to be patched for the test cases we wanted

**Solution:**

1. Collecting benchmark results for downloads at varying sizes
2. Updating the download/upload test file scripts

**Testing:**

1. The benchmark results were collected on a `t3a.micro` instance. No issues during downloads, no crashes, and no hang-ups.

The mkdocs page:
image

### DIFF
--- a/docs/scripts/uploadTestFiles.py
+++ b/docs/scripts/uploadTestFiles.py
@@ -84,9 +84,9 @@ def create_folder_structure(
 
     def create_files_in_current_dir(path_to_create_files: str) -> None:
         for i in range(1, num_files_per_directory + 1):
-            chunk_size = 1048576  # size of each chunk in bytes
+            chunk_size = MiB  # size of each chunk in bytes
             num_chunks = size_of_each_file_bytes // chunk_size
-            filename = f"{path_to_create_files}/file{i}.txt"
+            filename = os.path.join(path_to_create_files, f"file{i}.txt")
             if (
                 os.path.isfile(filename)
                 and os.path.getsize(filename) == size_of_each_file_bytes
--- a/docs/scripts/uploadTestFiles.py
+++ b/docs/scripts/uploadTestFiles.py
@@ -2,22 +2,31 @@
 Create some test files and upload them to Synapse and S3. This is used as the first step
 for benchmarking downloads.
 """
+
+import asyncio
 import os
 import shutil
 import subprocess  # nosec
 
 import synapseclient
 import synapseutils
-from synapseclient.entity import Project
+from synapseclient.models import Folder, Project
 
 PARENT_PROJECT = "syn$FILL_ME_IN"
 S3_BUCKET = "s3://$FILL_ME_IN"
 S3_PROFILE = "$FILL_ME_IN"
 
-PROJECT_25_FILES_1MB = "download_benchmarking_25_files_1mb"
-PROJECT_775_FILES_10MB = "download_benchmarking_775_files_10mb"
-PROJECT_10_FILES_1GB = "download_benchmarking_10_files_1gb"
-PROJECT_10_FILES_100GB = "download_benchmarking_10_files_100gb"
+
+# Create a bunch of folders with known names to be used during the benchmarking
+FOLDER_10_FILES_10GIB = "download_benchmarking_10_files_10gib"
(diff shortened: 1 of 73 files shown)


docs/scripts/uploadTestFiles.py
```diff
@@ -49,36 +58,52 @@ def create_folder_structure(
...
+        The total number of directories, total number of files, and the size of each
+        file in bytes
     """
     # Calculate total number of files and size of each file
     total_dirs = sum(
         [num_sub_directories**i for i in range(1, depth_of_directory_tree + 1)]
     )
     total_files = total_dirs * num_files_per_directory
-    total_size_of_files_bytes = total_size_of_files_mbytes * 1024 * 1024
+    total_size_of_files_bytes = total_size_of_files_mib * MiB
```
jaymedina, line 77: Why is this being converted to mib, if the argument should already be fed in mib?

docs/scripts/uploadTestFiles.py
```diff
@@ -49,36 +58,52 @@ def create_folder_structure(
...
     print(f"total_files: {total_files}")
     print(f"total_size_of_files_bytes: {total_size_of_files_bytes}")
-    print(f"size_of_each_file_bits: {size_of_each_file_bytes}")
+    print(f"size_of_each_file_bytes: {size_of_each_file_bytes}")
 
-    def create_files_in_current_dir(path_to_create_files):
+    def create_files_in_current_dir(path_to_create_files: str) -> None:
         for i in range(1, num_files_per_directory + 1):
-            chunk_size = 1024  # size of each chunk in bytes
+            chunk_size = 1048576  # size of each chunk in bytes
```
jaymedina, line 87: nit: This can just be `MiB`

docs/scripts/uploadTestFiles.py
```diff
@@ -49,36 +58,52 @@ def create_folder_structure(
...
             num_chunks = size_of_each_file_bytes // chunk_size
-
-            with open(f"{path_to_create_files}/file{i}.txt", "wb") as f:
-                for _ in range(num_chunks):
+            filename = f"{path_to_create_files}/file{i}.txt"
```
jaymedina, line 89: I would suggest `os.path.join` here in case `path_to_create_files` is already fed like `/my/path/` which creates e.g. `/my/path//file1.txt` with how this is currently written.

docs/scripts/uploadTestFiles.py
```diff
@@ -148,83 +149,150 @@ def sync_to_synapse(path: str, project_id: str, syn: synapseclient.Synapse) -> N
...
-    os.remove(f"{path}/benchmarking_manifest.tsv")
-    execute_sync_to_s3(path=path, key_in_bucket=PROJECT_25_FILES_1MB)
-    shutil.rmtree(path)
+async def set_up_folders_one_time(path: str, syn: synapseclient.Synapse) -> None:
+    project = await Project(id=PARENT_PROJECT).get_async()
 
+    depth = 1
+    sub_directories = 1
+    files_per_directory = 1000
+    size_mib = 1024
```
jaymedina, line 165: isn't this supposed to be 1? `size_mib` is fed to `total_size_of_files_mib` which is used to calculate `total_size_of_file_bytes` by multiplying by `MiB` which is already 1 mib i.e. `1024 * 1024`

jaymedina (approved):
Great results!! 🔥 We'll get good use out of these changes on `nf-synapse-challenge`, as the [SYNAPSE_STAGE](https://github.com/Sage-Bionetworks-Workflows/nf-synapse-challenge/blob/05a0cb0913913ac511e33dfd4c936a9f218f8bf3/modules/synapse_stage.nf#L18) step takes a little long for challenges with large data vols. Thanks for these changes @BryanFauble

Approved, pending resolution to my comments/questions

docs/scripts/uploadTestFiles.py
BryanFauble: Yeah, I had written this so that I passed in the total size I wanted to test the transfer with. Then the code will distribute that size evenly over the requested number of files.

PR #1117, opened 2024-07-10
[SYNPY-1494] Remove process pool executor

BryanFauble:
**Problem:**

1. loky crash when used from a uwsgi process: https://github.com/joblib/loky/issues/207

**Solution:**

1. Removing the use of multi-processing from the client

**Testing:**

1. Lingling was able to verify that the changes are working as expected
2. Integration tests are still all passing here

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -18,11 +18,11 @@
     "default": {
         "anyio": {
             "hashes": [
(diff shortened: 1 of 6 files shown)


linglp (approved):
Thanks again Bryan!

thomasyu888 (approved):
🔥 LGTM!

PR #1118, opened 2024-07-11
Update os matrix for check-deploy

BryanFauble:
**Problem:**

The workflow runs for check-deploy have been failing with:

check-deploy (macos-11, 3.10)
This request was automatically failed because there were no enabled runners online to process the request for more than 1 days.

**Solution:**

1. Updating image versions

**Testing:**

1. Will be testing on next check deploy run


thomasyu888 (approved):
🔥 LGTM

PR #1120, opened 2024-07-12
[SYNPY-1504] Remove `download_location` from File and rely only on Path

BryanFauble:
**Problem:**

1. When working with the File model you needed to specify a `download_location` parameter when using the `.get()` function, but `path` when using the `store()` function. This was a bit confusing.

**Solution:**

1. Updating logic such that only the path parameter is needed on the file for either operation.

**Testing:**

1. Some manual testing to verify getting a file already on disk, not on disk, and on disk - but downloading to another directory

### DIFF
--- a/synapseclient/models/file.py
+++ b/synapseclient/models/file.py
@@ -206,7 +206,15 @@ class File(FileSynchronousProtocol, AccessControllable):
             contain: letters, numbers, spaces, underscores, hyphens, periods, plus
             signs, apostrophes, and parentheses. If not specified, the name will be
             derived from the file name.
-        path: The path to the file on disk.
+        path: The path to the file on disk. Using shorthand `~` will be expanded to the
+            user's home directory.
+
+            This is used during a `get` operation to specify where to download the file
+            to. It should be pointing to a directory.
+
+            This is also used during a `store` operation to specify the file to upload.
+            It should be pointing to a file.
+
         description: The description of this file. Must be 1000 characters or less.
         parent_id: The ID of the Entity that is the parent of this Entity. Setting this
             to a new value and storing it will move this File under the new parent.
@@ -297,7 +305,6 @@ class File(FileSynchronousProtocol, AccessControllable):
 
     Attributes:
         download_file: (Get only) If True the file will be downloaded.
-        download_location: (Get only) The location to download the file to.
         if_collision: (Get only)
             Determines how to handle file collisions. Defaults to "keep.both". May be:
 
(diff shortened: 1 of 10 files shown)


synapseclient/models/file.py
```diff
@@ -1013,13 +1021,15 @@ async def get_async(
             if_collision=self.if_collision,
             limit_search=self.synapse_container_limit or self.parent_id,
             download_file=self.download_file,
-            download_location=self.download_location,
+            download_location=os.path.dirname(self.path)
+            if self.path and os.path.isfile(self.path)
+            else self.path,
```
jaymedina, line 1026: Nice! I'm thinking about this a little bit more:

What happens if you store a file with `path=/user/bfauble/folder/file.txt` for example, and I try to download it. Does my machine automatically make a folder called `bfauble` and a subfolder called `folder` to store my file?

Another thought, but related: This looks to me that if a folder you're downloading has `path` in the filehandle, it will replace synapse cache as the new default, which is new behavior and should be heavily documented everywhere

BryanFauble: > What happens if you store a file with path=/user/bfauble/folder/file.txt for example, and I try to download it. Does my machine automatically make a folder called bfauble and a subfolder called folder to store my file?

No. The file will be downloded based on the `path` that you specify, or the `.synapseCache` if no path is specified.

> Another thought, but related: This looks to me that if a file you're downloading has path in the filehandle, path will replace synapse cache as the new default, which is new behavior and should be heavily documented everywhere


The is the current behavior of the client through the current Synapse class.

> Another thought: Have you tested this for a case where you're downloading a Folder, and 2 Files within the folder have different paths?

This isn't a scenario that would happen I don't think. You can use the Folder.sync_from_synapse function for sure, but that function takes in a `path` argument, and all files/folders under it will be downloaded to the respective location(s) based on how they're organized in Synapse.

jaymedina: > No. The file will be downloded based on the path that you specify, or the .synapseCache if no path is specified.

You're right, I see that now. `self.path` doesn't get assigned `path` from an entity, or anything like that, if it's not declared.

Thanks!

tests/integration/synapseclient/core/test_download.py
```diff
@@ -191,12 +191,11 @@ async def test_download_cached_file_to_new_directory(
 
             # WHEN I download the file to another location
             updated_location = os.path.join(
-                os.path.dirname(original_file_path), "subdirectory"
+                os.path.dirname(original_file_path),
+                "subdirectory/asd/asdas/dasd/asdas/dasdasd",
```
jaymedina, line 195: I remember you had written this as part of your demo today and it should be reverted back to `"subdirectory"` haha

PR #1121, opened 2024-07-15
[SYNSD-1127] Retry on several more exceptions for transfers

BryanFauble:
**Problem:**

1. In https://sagebionetworks.jira.com/browse/SYNSD-1127 - Some exceptions were raised in the HTTPX library that the retry function was not accounting for.

**Solution:**

1. Adding the class name of the exceptions we're now looking for.

**Testing:**

1. I added some temporary debug code to verify that an individual part is the only thing retried, not an entire file
2. Added an integration test around the logic change

### DIFF
--- a/synapseclient/core/retry.py
+++ b/synapseclient/core/retry.py
@@ -64,6 +64,11 @@ RETRYABLE_CONNECTION_EXCEPTIONS = [
     "timeout",
     "ReadError",
     "ReadTimeout",
+    # HTTPX Specific connection exceptions:
+    "RemoteProtocolError",
+    "TimeoutException",
+    "ConnectError",
+    "ConnectTimeout",
 ]
 
 DEBUG_EXCEPTION = "calling %s resulted in an Exception"
--- a/tests/integration/synapseclient/core/test_download.py
+++ b/tests/integration/synapseclient/core/test_download.py
@@ -475,6 +475,80 @@ class TestDownloadFromUrlMultiThreaded:
         assert file.file_handle.content_md5 == file_md5
         assert os.path.exists(file_path)
 
+    async def test_download_from_url_multi_threaded_random_protocol_exceptions(
+        self,
+        syn: Synapse,
+        project_model: Project,
+        schedule_for_cleanup: Callable[..., None],
+    ) -> None:
+        """Test download of a file if downloaded in multiple parts. In this case I am
+        dropping the download part size to 500 bytes to force multiple parts download.
+
+        This function will randomly fail the download of a part with protocol exceptions
+
+        """
+        # Set the failure rate to 90%
+        failure_rate = 0.9
+
+        # GIVEN a file stored in synapse
+        file_path = utils.make_bogus_data_file()
+        file = await File(path=file_path, parent_id=project_model.id).store_async()
+        schedule_for_cleanup(file.id)
+        schedule_for_cleanup(file_path)
+        file_md5 = file.file_handle.content_md5
+        assert file_md5 is not None
+        assert os.path.exists(file_path)
+
+        # AND the file is not in the cache
+        syn.cache.remove(file_handle_id=file.file_handle.id)
+        os.remove(file_path)
+        assert not os.path.exists(file_path)
+
+        # AND an httpx client that is not mocked
+        httpx_timeout = httpx.Timeout(70, pool=None)
+        client = httpx.Client(timeout=httpx_timeout)
+
+        # AND the mock httpx send function to simulate a failure
+        def mock_httpx_send(**kwargs):
+            """Conditionally mock the HTTPX send function to simulate a failure. The
+            HTTPX .stream function internally calls a non-contexted managed send
+            function. This allows us to simulate a failure in the send function."""
+            is_part_stream = False
+            for header in kwargs.get("request").headers.raw:
+                if header[0].lower() == b"range":
+                    is_part_stream = True
+                    break
+            if is_part_stream and random.random() <= failure_rate:
+                raise httpx.RemoteProtocolError(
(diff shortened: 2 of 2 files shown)


synapseclient/core/retry.py
```diff
@@ -64,6 +64,11 @@
     "timeout",
     "ReadError",
     "ReadTimeout",
+    # HTTPX Specific connection exceptions:
```
BryanFauble, line 67: I picked the others out of the list from all the exceptions that the package raises as they look equivalent to the ones we were retrying from the requests library:
```
from httpx._exceptions import (
    CloseError,
    ConnectError,
    ConnectTimeout,
    CookieConflict,
    DecodingError,
    HTTPError,
    HTTPStatusError,
    InvalidURL,
    LocalProtocolError,
    NetworkError,
    PoolTimeout,
    ProtocolError,
    ProxyError,
    ReadError,
    ReadTimeout,
    RemoteProtocolError,
    RequestError,
    RequestNotRead,
    ResponseNotRead,
    StreamClosed,
    StreamConsumed,
    StreamError,
    TimeoutException,
    TooManyRedirects,
    TransportError,
    UnsupportedProtocol,
    WriteError,
    WriteTimeout,
)
```

tests/integration/synapseclient/core/test_download.py
```diff
@@ -475,6 +475,80 @@ def mock_httpx_send(**kwargs):
...
+        assert not os.path.exists(file_path)
+
+        # AND an httpx client that is not mocked
+        httpx_timeout = httpx.Timeout(70, pool=None)
+        client = httpx.Client(timeout=httpx_timeout)
+
+        # AND the mock httpx send function to simulate a failure
+        def mock_httpx_send(**kwargs):
+            """Conditionally mock the HTTPX send function to simulate a failure. The
+            HTTPX .stream function internally calls a non-contexted managed send
```
philerooski, line 514: non-contexted managed?

tests/integration/synapseclient/core/test_download.py
```diff
@@ -475,6 +475,80 @@ def mock_httpx_send(**kwargs):
...
+            function. This allows us to simulate a failure in the send function."""
+            is_part_stream = False
+            for header in kwargs.get("request").headers.raw:
+                if header[0].lower() == b"range":
+                    is_part_stream = True
+                    break
+            if is_part_stream and random.random() <= failure_rate:
```
philerooski, line 521: Why is it okay to have a random component to this test?

tests/integration/synapseclient/core/test_download.py
BryanFauble: In the HTTPX library they use a contextmanager to handle the `stream` function that we directly call. It makes mocking the call really difficult. To get around this I am mocking the `.send()` function directly to replace the call:

```
    @contextmanager
    def stream(
        self,
        method: str,
        url: URLTypes,
        *,
        content: RequestContent | None = None,
        data: RequestData | None = None,
        files: RequestFiles | None = None,
        json: typing.Any | None = None,
        params: QueryParamTypes | None = None,
        headers: HeaderTypes | None = None,
        cookies: CookieTypes | None = None,
        auth: AuthTypes | UseClientDefault | None = USE_CLIENT_DEFAULT,
        follow_redirects: bool | UseClientDefault = USE_CLIENT_DEFAULT,
        timeout: TimeoutTypes | UseClientDefault = USE_CLIENT_DEFAULT,
        extensions: RequestExtensions | None = None,
    ) -> typing.Iterator[Response]:
        """
        Alternative to `httpx.request()` that streams the response body
        instead of loading it into memory at once.

        **Parameters**: See `httpx.request`.

        See also: [Streaming Responses][0]

        [0]: /quickstart#streaming-responses
        """
        request = self.build_request(
            method=method,
            url=url,
            content=content,
            data=data,
            files=files,
            json=json,
            params=params,
            headers=headers,
            cookies=cookies,
            timeout=timeout,
            extensions=extensions,
        )
        response = self.send(
            request=request,
            auth=auth,
            follow_redirects=follow_redirects,
            stream=True,
        )
        try:
            yield response
        finally:
            response.close()
```

tests/integration/synapseclient/core/test_download.py
BryanFauble: This was to prove that in adverse conditions we don't have direct control over that the code still behaves as expected. The behavior is that it would retry up to 5 minutes to complete the transfer, after which it would fail.

Since the exception is being raised without making any HTTP call the fail -> retry loop is very quick. In this test I set a maximum backoff of 0.2 seconds between retries.


I used this same kind of testing to verify what would occur before the download re-write changes, to post re-write. I found the new code to work 100% of the time, whereas the previous code worked ~20% of the time.

philerooski: With the failure rate what it is, you're practically guaranteed a failure. But wouldn't it be better to make the test entirely deterministic? Or would that require jumping through too many hoops?

BryanFauble: > With the failure rate what it is, you're practically guaranteed a failure

Yes, that is the purpose of having the failure rate what it is. It is meant to verify that the retry logic is sufficient to handle 90% of the parts of a file being dropped.

> But wouldn't it be better to make the test entirely deterministic? Or would that require jumping through too many hoops?

Integration tests are rarely 100% deterministic since it relies on so many outside factors. This is to help simulate if one of those outside factors (The remote file server) is one of those flaky items.

Since this gives the retry logic more testing I am keen to keep this as is.

thomasyu888 (approved):
🔥 LGTM!  Thanks for the great work and reviews!

PR #1122, opened 2024-07-30
[SYNPY-1436] avoid duplicate user profile call during login

danlu1:
Problem:

When calling syn.login() method 2 HTTP calls are sent to the Synapse server to /userProfile

Solution:

Add displayName info from profile to SynapseAuthTokenCredentials

Test:

Relevant unit tests have been updated. Unit test and integration test passed.


synapseclient/core/credentials/cred_data.py
```diff
@@ -61,11 +61,12 @@ def _validate_token(cls, token):
             # ValueError if the split string is not base64 encoded or if the decoded base64 is not json
             pass
 
-    def __init__(self, token, username=None):
+    def __init__(self, token, username=None, displayname=None):
```
BryanFauble, line 64: nit: Could you update the type for the arguments here and a return type of None

synapseclient/core/credentials/credential_provider.py
```diff
@@ -77,6 +77,7 @@ def _create_synapse_credential(
             profile = syn.restGET("/userProfile", auth=credentials)
             profile_username = profile.get("userName")
             profile_emails = profile.get("emails", [])
+            profile_displayname = profile.get("displayName")
```
BryanFauble, line 80: From the Rest API docs:

> displayName 	STRING 	This field is deprecated and will always be null.

https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/UserProfile.html

Is this still being set?

BryanFauble: This would need to follow the same check that existed in the client.py script before:

```
            display_name = (
                profile["displayName"]
                if "displayName" in profile
                else self.credentials.username
            )
```

I think:
`profile_displayname = profile.get("displayName", profile_username)`

That is assuming that `displayName` has cases where it is not None. If it is not None then I think we can revert most of these changes and instead just use username in the welcome message:
`self.logger.info(f"Welcome, {self.credentials.username}!\n")`

BryanFauble: I did test this out and from that API I did get back a `displayName` for my account:

`"displayName": "Bryan Fauble",`

So I think the documentation for that API is incorrect. Could you create a PLFM jira ticket and mention this? The docs should be corrected, or the field should indeed return null.

BryanFauble (changes requested):
1 change needed when setting display name to just use username if display name is not set.

Otherwise if the user does not have a display name the welcome message would show up as:
`Welcome, None!`

synapseclient/core/credentials/credential_provider.py
danlu1: Thanks for catching this. I got the displayName from my profile as well. I will add the check back.

synapseclient/core/credentials/cred_data.py
```diff
@@ -96,10 +96,10 @@ def __call__(self, r):
         return r
 
     def __repr__(self):
-        return  f"SynapseAuthTokenCredentials("\
-            f"username='{self.username}', "\
-            f"displayname='{self.displayname}', "\
-            f"token='{self.secret}')"
+        return (f"SynapseAuthTokenCredentials("
+            f"username='{self.username}', "
+            f"displayname='{self.displayname}', "
```
thomasyu888, line 101: @danlu1 / @BryanFauble Thoughts on just removing displayname?

danlu1: I'm open to either discard displayname or check displayname first otherwise use username.

BryanFauble: If synapse is giving us the data I am in favor of leaving it in.

thomasyu888: Lets find out if it is actually deprecated? Then go from there.

BryanFauble: Works for me. @danlu1 Do you mind following up with Platform on this item?

danlu1: Sure. I tested it with my personal account and DPE test account. With my personal account, `syn.restGET("/userProfile", auth=credentials)` gives 

```{'ownerId': '3434599', 'etag': '01f8652a-55b8-4816-b8b5-cd8d3e479feb', 'firstName': 'Dan', 'lastName': 'Lu', 'emails': ['dan.lu@sagebase.org'], 'openIds': [], 'userName': 'danlu', 'displayName': 'Dan Lu', 'summary': '', 'position': 'Research Software Engineer', 'location': 'Seattle, WA, USA', 'industry': '', 'company': 'Sage Bionetworks', 'profilePicureFileHandleId': '81598436', 'url': '', 'notificationSettings': {'sendEmailNotifications': True}, 'createdOn': '2021-09-15T16:12:59.000Z', 'twoFactorAuthEnabled': False}```

And with the DPE test account,
```
{'ownerId': '3336428', 'etag': '9b4a8f96-ffe7-41bc-b4c9-82d198f8b176', 'emails': ['test@sagebase.org'], 'openIds': [], 'userName': 'test', 'notificationSettings': {'sendEmailNotifications': True}, 'createdOn': '2021-09-11T06:53:31.000Z', 'twoFactorAuthEnabled': False}
```

It seems that the displayName will be given back if it exists for the account.

BryanFauble (approved):
Changes look good so I'll go ahead and approve

synapseclient/core/credentials/cred_data.py
danlu1: Tracked [here](https://sagebionetworks.jira.com/browse/PLFM-8522).

PR #1125, opened 2024-08-16
[SYNPY-1511] Correct missing synapse client propogation and caching

BryanFauble:
**Problem:**

1. Around 8 months ago I had added in logic to cache the logged in synapse client instance as a class variable to prevent the need from passing your instance into all methods. There were a few internal methods that I missed when setting up the download logic leading to issues where `.login()` was not called.

**Solution:**

1. Correct the missing propagation.
2. Updating the cache logic to cache in the class init rather than on login

**Testing:**

1. Tested locally to connect and query on an anonymously accessible table
2. Integration tests

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2,7 +2,6 @@
 The `Synapse` object encapsulates a connection to the Synapse service and is used for building projects, uploading and
 retrieving data, and recording provenance of data analysis.
 """
-
 import asyncio
 import collections
 import collections.abc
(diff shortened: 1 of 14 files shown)


BWMac (approved):
LGTM!

thomasyu888 (changes requested):
Thanks for the quick work, can we add an integration test to download a publicly accessible table?

BryanFauble:
> Thanks for the quick work, can we add an integration test to download a publicly accessible table?

Thanks @thomasyu888 

I tried to make an integration test around this, however, I cannot independently make a table publicly downloadable (ie: Without a login) based off this documentation: https://sagebionetworks.jira.com/wiki/spaces/GOV/pages/1108672767/How+to+Enable+Anonymous+Download+of+Synapse+Data

I did however manually test this with the Synapse homepage metric tables which are publicly downloadable to verify the functionality.

danlu1 (approved):
I confirmed that I can query anonymous viewable table from synapse now.

thomasyu888 (approved):
🔥 LGTM!

PR #1126, opened 2024-08-29
[SYNPY-1514] Handle Expired Pre-Signed URLs

BWMac:
**Problem:**

Due to the unordered nature of AsyncIO executing tasks async, when there is a large queue of files to download, the file may not be downloaded for a while after a pre-signed URL is retrieved. This shows up as a `403 Forbidden` error which needs to be handled for these cases to prevent downloads from failing.

**Solution:**

Update `download_from_url` to handle these `403` responses and try again by checking if the URL is expired when we receive a 403 response code, retrieving a new URL and retrying the download.

**Testing:**

- This was tested by:
  1. Reproducing the error faced by a [user](https://sagebionetworks.jira.com/browse/SYNSD-1188) by using the `random` package to have this failure occur.
  2. Hard-coding an expired URL into the code and running a `synapse get` CLI command
- new unit and integration tests were added

### DIFF
--- a/synapseclient/core/download/download_functions.py
+++ b/synapseclient/core/download/download_functions.py
@@ -774,6 +774,8 @@ def download_from_url(
                             verbose=client.debug,
                             **STANDARD_RETRY_PARAMS,
                         )
+                    else:
+                        raise
                 elif err.response.status_code == 404:
                     raise SynapseError(f"Could not download the file at {url}") from err
                 elif (
--- a/synapseclient/core/download/download_functions.py
+++ b/synapseclient/core/download/download_functions.py
@@ -404,13 +404,13 @@ async def download_by_file_handle(
 
     while retries > 0:
         try:
-            file_handle_result: Dict[str, str] = (
-                await get_file_handle_for_download_async(
-                    file_handle_id=file_handle_id,
-                    synapse_id=synapse_id,
-                    entity_type=entity_type,
-                    synapse_client=syn,
-                )
+            file_handle_result: Dict[
+                str, str
+            ] = await get_file_handle_for_download_async(
+                file_handle_id=file_handle_id,
+                synapse_id=synapse_id,
+                entity_type=entity_type,
+                synapse_client=syn,
             )
             file_handle = file_handle_result["fileHandle"]
             concrete_type = file_handle["concreteType"]
@@ -734,7 +734,7 @@ def download_from_url(
                 if is_synapse_uri(uri=url, synapse_client=client)
                 else None
             )
-            
+
             try:
                 response = with_retry(
                     lambda url=url, range_header=range_header, auth=auth: client._requests_session.get(
--- a/synapseclient/core/download/download_functions.py
+++ b/synapseclient/core/download/download_functions.py
@@ -749,21 +749,14 @@ def download_from_url(
                     **STANDARD_RETRY_PARAMS,
                 )
                 exceptions._raise_for_status(response, verbose=client.debug)
-            except (SynapseHTTPError, SSLZeroReturnError) as err:
-                # is this also casued by an expired URL?
-                override_url = False
-                if isinstance(err, SSLZeroReturnError):
-                    will_retry = True
-                    override_url = True
-                else:
-                    will_retry = err.response.status_code == 403
-                if will_retry:
+            except SynapseHTTPError as err:
+                if err.response.status_code == 403:
                     url_is_expired = datetime.datetime.now(
                         tz=datetime.timezone.utc
                     ) + PresignedUrlProvider._TIME_BUFFER >= _pre_signed_url_expiration_time(
                         url
                     )
-                    if url_is_expired or override_url:
+                    if url_is_expired:
                         response = get_file_handle_for_download(
                             file_handle_id=file_handle_id,
                             synapse_id=object_id,
--- a/synapseclient/core/download/download_functions.py
+++ b/synapseclient/core/download/download_functions.py
@@ -9,7 +9,6 @@ import shutil
 import sys
 import urllib.parse as urllib_urlparse
 import urllib.request as urllib_request
-from ssl import SSLZeroReturnError
 from typing import TYPE_CHECKING, Dict, Optional, Union
 
 from tqdm import tqdm
--- a/synapseclient/core/download/download_functions.py
+++ b/synapseclient/core/download/download_functions.py
@@ -404,13 +404,13 @@ async def download_by_file_handle(
 
     while retries > 0:
         try:
-            file_handle_result: Dict[
-                str, str
-            ] = await get_file_handle_for_download_async(
-                file_handle_id=file_handle_id,
-                synapse_id=synapse_id,
-                entity_type=entity_type,
-                synapse_client=syn,
+            file_handle_result: Dict[str, str] = (
+                await get_file_handle_for_download_async(
+                    file_handle_id=file_handle_id,
+                    synapse_id=synapse_id,
+                    entity_type=entity_type,
+                    synapse_client=syn,
+                )
             )
             file_handle = file_handle_result["fileHandle"]
             concrete_type = file_handle["concreteType"]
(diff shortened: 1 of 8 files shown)


synapseclient/core/download/download_functions.py
```diff
@@ -719,21 +735,46 @@ def _ftp_report_hook(
...
                 exceptions._raise_for_status(response, verbose=client.debug)
             except SynapseHTTPError as err:
-                if err.response.status_code == 404:
+                if err.response.status_code == 403:
+                    url_is_expired = datetime.datetime.now(
+                        tz=datetime.timezone.utc
+                    ) + PresignedUrlProvider._TIME_BUFFER >= _pre_signed_url_expiration_time(
+                        url
+                    )
+                    if url_is_expired:
```
thomasyu888, line 758: I know it's still a WIP, but just checking what happens if the url isn't expired

BWMac: Ah yeah, I was missing a raise. Added in latest commit https://github.com/Sage-Bionetworks/synapsePythonClient/blob/300b96cf781fc216bd3a98f690abbb53a3da00c9/synapseclient/core/download/download_functions.py#L777

BryanFauble:
There was also an SSL exception that was in the logs the user had provided. Can you also see if we can add that to the exceptions that are retried?

BWMac:
@BryanFauble We definitely could, but as written my solution to the 403 response codes is to check if the URL is expired and then if it is, retry. Will the URL also be expired in the case where `SSLZeroReturnError` is raised or should this case bypass the URL check/refresh?

BryanFauble:
> @BryanFauble We definitely could, but as written my solution to the 403 response codes is to check if the URL is expired and then if it is, retry. Will the URL also be expired in the case where `SSLZeroReturnError` is raised or should this case bypass the URL check/refresh?

I don't think a 403 is going to be raised. I think the solution is to add it to the standard retry params:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/2dda95416608604ec928905596c4680c6a5b06be/synapseclient/client.py#L190

In the retry function you can also pass the actual exception class too. I think we will want to add it as one of the things to retry, just like the connection exceptions that are raised. That way the ssl errors are treated with the same 5 minute retry logic.

synapseclient/core/download/download_functions.py
```diff
@@ -610,6 +619,8 @@ async def download_from_url_multi_threaded(
 def download_from_url(
     url: str,
     destination: str,
+    object_id: Optional[str],
```
BryanFauble, line 622: Could you update this to `entity_id` and `entity_type` respectively.

BryanFauble: Or since `object_type` isn't always and entity, maybe `file_handle_associate_type` to match: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/file/FileHandleAssociateType.html

synapseclient/core/download/download_functions.py
```diff
@@ -719,21 +735,48 @@ def _ftp_report_hook(
...
                 exceptions._raise_for_status(response, verbose=client.debug)
             except SynapseHTTPError as err:
-                if err.response.status_code == 404:
+                if err.response.status_code == 403:
+                    url_is_expired = datetime.datetime.now(
+                        tz=datetime.timezone.utc
+                    ) + PresignedUrlProvider._TIME_BUFFER >= _pre_signed_url_expiration_time(
+                        url
+                    )
+                    if url_is_expired:
```
BryanFauble, line 758: Should this check also be put before:

```
                response = with_retry(
                    lambda url=url, range_header=range_header, auth=auth: client._requests_session.get(
                        url=url,
                        headers=client._generate_headers(range_header),
                        stream=True,
                        allow_redirects=False,
                        auth=auth,
                    ),
                    verbose=client.debug,
                    **STANDARD_RETRY_PARAMS,
                )
```

I am thinking of the scenario where we know before we make the call if the URL happens to be expired we can just go ahead and get a new presigned URL at that point in time.

BryanFauble (approved):
Great changes! Just a few small areas that changes could be made, but shouldn't affect overall intention of the code.

synapseclient/core/download/download_functions.py
BWMac: In one of the [tests](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f368f1fbb08b33c87443adaa37941eee219cc39a/tests/integration/synapseclient/core/test_download.py#L253-L254), `download_from_url` is called with a URL that does not have an expiration time and so this errors out if we add this check before the initial attempt. Is that a real scenario that we need to handle?

BryanFauble: It is a real scenario because this is what the user had reported. The way you wrote the logic is that even if we know the URL is expired we will make the call -> Fail -> Get a new URL -> make the call again.

The logic you have works, however, we could trim out the Fail and get a new token.


As for integration tests failing - You could update the logic so that if the URL does not have an expiration date/time we won't get a new token because we won't know if the failure was because it was expired, or actually denied with an HTTP 403 for some other reason.

BryanFauble: > URL that does not have an expiration time and so this errors out if we add this check before the initial attempt. Is that a real scenario that we need to handle?

I get what you mean about this. Let's handle this situation like this:

1. If a URL does not have an expiration time we should not error
2. Make the initial call to get the data and even if a 403 error shows up let the transfer fail and do not try again

BWMac: Implemented checking for if a URL has an expiration [here](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/dc259ec854737ee7179c469b6fce6c8b221c3758/synapseclient/core/download/download_functions.py#L739-L747).

synapseclient/core/download/download_functions.py
```diff
@@ -719,21 +735,64 @@ def _ftp_report_hook(
...
+                        auth=auth,
+                    ),
+                    verbose=client.debug,
+                    **STANDARD_RETRY_PARAMS,
+                )
                 exceptions._raise_for_status(response, verbose=client.debug)
             except SynapseHTTPError as err:
-                if err.response.status_code == 404:
+                if err.response.status_code == 403:
+                    url_is_expired = datetime.datetime.now(
```
BryanFauble, line 769: Does the additional check you added:

```
                url_has_expiration = "Expires" in urllib_urlparse.urlparse(url).query
                url_is_expired = False
                if url_has_expiration:
```

Need to also be added here?

BWMac: Is it possible for the URL to expire between the first and second download attempts?

BryanFauble: > Is it possible for the URL to expire between the first and second download attempts?

Likely no. However from what you had mentioned:

> URL that does not have an expiration time and so this errors out if we add this check before the initial attempt.

If we have a URL that does not have an expiration time AND it got an HTTP 403 error back. Wouldn't that raise an exception here since the code is expecting that the URL has a expiration time on it?

BWMac: OK I added this and a test for the case when a URL does not have an expiration datetime.

BryanFauble (approved):
LGTM!

synapseclient/core/download/download_functions.py
```diff
@@ -719,21 +735,69 @@ def _ftp_report_hook(
...
-                    headers=client._generate_headers(range_header),
-                    stream=True,
-                    allow_redirects=False,
-                    auth=auth,
-                ),
-                verbose=client.debug,
-                **STANDARD_RETRY_PARAMS,
-            )
             try:
+                url_has_expiration = "Expires" in urllib_urlparse.urlparse(url).query
```
thomasyu888, line 739: Just a check here : is it worthwhile to put this in a helper function or is that just unnecessary overhead? (Since it's only used twice)

thomasyu888 (approved):
🔥 LGTM! This is neat, although the exception handling definitely looks more complex now.

synapseclient/core/download/download_functions.py
BWMac: I had the same thought but figured it wasn't worth it because of the changes to tests and new tests that would be needed.

thomasyu888: The one thought I had is whether this would be easier to test as a unit test.  That said - i think the test cases you added does cover for these scenarios - it may become an extra step to change code in both areas if things need to be updated.

PR #1127, opened 2024-09-03
[SYNPY-1509] Updates `NA` handling for `CSVFileTable.asDataFrame`

BWMac:
**Problem:**

We do not currently support passing `keep_default_na` and `na_values` to `CSVFileTable.asDataFrame` so we are only offering the default functionality of `pandas.read_csv` to users. For example, if someone wanted to have the string `"None"` from a Synapse table remain a string and not become `NA` when exported to a CSV file, they would be unable to.

**Solution:**

Add `**kwargs` as arguments to `CSVFileTable.asDataFrame` and pass them down to `pandas.read_csv` through `_csv_to_pandas_df`.

**Testing:**

I added unit tests for this new functionality and tested my changes with the following script:

```
import synapseclient
from synapseclient.table import CsvFileTable

syn = synapseclient.Synapse()
syn.login()

import pandas as pd

data = {
    'id': ['abc', '', 'None', None, 'NA'],
    'test_num': ['1', '2', '3', '4', '5'],
}

test_df = pd.DataFrame(data)
schema = synapseclient.table.Schema(
    name="test_table",
    parent="syn59478569",
    column_names=["id", "version"],
    column_types=["STRING", "STRING"],
)

test_table = CsvFileTable.from_data_frame(df=test_df, schema=schema)

# rows 2-5 should have blank Ids
test_table_df_before = test_table.asDataFrame()
test_table_df_before.to_csv("test_table_before.csv", index=False)

# rows 3 and 5 should have "None" and "NA" respectively in the Id column
test_table_df_no_default_na = test_table.asDataFrame(keep_default_na=False)
test_table_df_no_default_na.to_csv("test_table_no_default_na.csv", index=False)

# Row 3 should have "None" in the Id column, row 5 should have a blank Id
test_table_df_custom_na_values = test_table.asDataFrame(
    na_values=["NA"], keep_default_na=False
)
test_table_df_custom_na_values.to_csv("test_table_custom_na_values.csv", index=False)
```

**Notes:**

- I ran into an issue with `numpy` > 2.0 so I added a constraint to its versioning.
- `pandas.parser.CParserError` was deprecated so I replaced it with the newer `pandas.errors.ParserError`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -99,6 +99,7 @@ tests =
 
 pandas =
     pandas>=1.5,<3.0
+    numpy<2.0.0
 
 pysftp =
     pysftp>=0.2.8,<0.3
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -2469,7 +2469,10 @@ class CsvFileTable(TableAbstractBaseClass):
         return self
 
     def asDataFrame(
-        self, rowIdAndVersionInIndex=True, convert_to_datetime=False, **kwargs
+        self,
+        rowIdAndVersionInIndex: bool = True,
+        convert_to_datetime: bool = False,
+        **kwargs,
     ):
         """Convert query result to a Pandas DataFrame.
 
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -2463,7 +2463,7 @@ class CsvFileTable(TableAbstractBaseClass):
                                     (and row_etag if it exists)
             convert_to_datetime:    If set to True, will convert all Synapse DATE columns from UNIX timestamp
                                     integers into UTC datetime objects
-            na_values:            A list of strings to recognize as NA/NaN. 
+            na_values:            A list of strings to recognize as NA/NaN.
             keep_default_na:      Whether to include the default NaN values.
                                  https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html
 
@@ -2507,7 +2507,7 @@ class CsvFileTable(TableAbstractBaseClass):
                 na_values=na_values,
                 keep_default_na=keep_default_na,
             )
-        
+
         except pd.errors.ParserError:
             return pd.DataFrame()
 
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -407,6 +407,8 @@ def _csv_to_pandas_df(
     list_columns=None,
     rowIdAndVersionInIndex=True,
     dtype=None,
+    na_values=None,
+    keep_default_na=True,
 ):
     test_import_pandas()
     import pandas as pd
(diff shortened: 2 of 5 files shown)


setup.cfg
```diff
@@ -99,6 +99,7 @@ tests =
 
 pandas =
     pandas>=1.5,<3.0
+    numpy<2.0.0
```
BryanFauble, line 102: My only comment before this is ready for review:

Is there any way we can expand this out? Possibly adding in some logic we can remove later on where it does X if numpy < 2.0.0, and Y if numpy >= 2.0.0?


My concern is that as we add more of these constraints we'll run into compatibility issues with others using this package.

synapseclient/table.py
```diff
@@ -2445,14 +2470,23 @@ def _update_self(self, syn):
             self.etag = upload_to_table_result["etag"]
         return self
 
-    def asDataFrame(self, rowIdAndVersionInIndex=True, convert_to_datetime=False):
+    def asDataFrame(
+        self,
+        rowIdAndVersionInIndex=True,
+        convert_to_datetime=False,
+        na_values=None,
+        keep_default_na=True,
```
thomasyu888, line 2478: Does it make sense to take a kwargs here to support all pd.Dataframe kwargs?

setup.cfg
BWMac: I had the same concern. I will look more into this and post updates here. For context, we are experiencing the same problem as this [issue](https://github.com/numpy/numpy/issues/26710). The dev who raised the issue said that they are doing [this](https://github.com/numpy/numpy/issues/26710#issuecomment-2171874281) as a solution.

BWMac: But it's more complicated because the newer Pandas versions aren't compatible with python 3.8, So we need the pandas version constraints to remain the same I think. I'm not sure there is anything better to do other than keeping this numpy constraint until we no longer support python 3.8

synapseclient/table.py
BWMac: I don't see why not. I'll add this

setup.cfg
thomasyu888: ^ Can we create a ticket for that, we also should add support for python 3.12, but i already know there are going to be issues with that since people have encountered issues when using python 3.12 with the python client.

tests/integration/synapseclient/core/test_external_storage.py
```diff
@@ -304,7 +304,7 @@ async def test_sts_external_storage_location(self) -> None:
                     Bucket=bucket_name,
                     Key=remote_key,
                 )
-            assert "Access Denied" in str(ex_cm.value)
+            assert "AccessDenied" in str(ex_cm.value)
```
thomasyu888, line 307: Interesting, what changed about this?

thomasyu888 (approved):
🔥 LGTM!  Just a few comments.

tests/integration/synapseclient/core/test_external_storage.py
BWMac: It just looks like AWS changed the error returned 🤷🏻

setup.cfg
BWMac: https://sagebionetworks.jira.com/browse/SYNPY-1517

PR #1128, opened 2024-09-06
[FDS-2373] Update retry delay, add some otel traces, add retry to url creation

BryanFauble:
**Changes:**

1. Reduce retry delay in-between table queries.
2. Update versioning for OTEL components to prevent issues with version ranges being used within schematic.
3. Wrap endpoint querying with retry logic.
4. Adding some trace collection around dozing/waiting for async results from synapse.

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -5077,7 +5077,7 @@ class Synapse(object):
                 sleep = min(
                     self.table_query_max_sleep, sleep * self.table_query_backoff
                 )
-                doze(sleep, trace_span_name="Synapse::_waitForAsync")
+                doze(sleep, trace_span_name="Synapse::doze_for_async")
             else:
                 break
         else:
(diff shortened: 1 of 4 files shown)


synapseclient/client.py
```diff
@@ -393,16 +397,12 @@ def log_response(response: httpx.Response) -> None:
...
-            raise ValueError("debug must be set to a bool (either True or False)")
-        self.debug = debug
-
         self.silent = silent
         self._init_logger()  # initializes self.logger
 
         self.skip_checks = skip_checks
 
-        self.table_query_sleep = 2
+        self.table_query_sleep = 0.2  # Sleep for 200ms between table query retries
```
danlu1, line 405: Just out of curiosity, how does this interval get decided?

danlu1 (approved):
LGTM

synapseclient/client.py
BryanFauble: I can't speak to how it was originally decided however for this value I based it off how long some queries actually needed to wait during schematic integration tests. I found that a delay of 0.2 seconds meant that only 1 wait cycle was needed before results were needed. Any shorter and several calls were wasted waiting for Synapse to finish providing the results, any longer and we were wasting time doing nothing.

PR #1129, opened 2024-09-13
[FDS-2386] Global client caching, Telemetry auto instruementation

BryanFauble:
**Problem:**

1. I found some more places where internally in the client we were not propagating along the `synapse_client` to further functions. When I started to disable client caching it was causing a few integration tests within schematic to fail (Note: The code that caused this was only present in an integration test, and not present in any code hit during a production request to schematic).
2. There was a lot of code added to handle otel context propagation to threads in the threading library.
3. When the client is used in the context of a multi-user environment it is very easy to not set `cache_client = False` when creating a new Synapse object. As a result a client can unintentionally be cached.

**Solution:**

1. Adding a global settings to disable client caching in the `Synapse` class
2. Adding a global setting to enable otel instrumentation in the client.
3. Using the otel auto instrumentation libraries to handle creating traces for several libraries.
4. I updated all of the integration tests to not use any cache Synapse class instance, and instead everything needs to pass it along - This will be used to verify the code is correctly propagating the instance along when it needs to.

**Testing:**

1. I am testing this actively with Schematic, along with integration tests that I verified ran locally/in GH.


synapseclient/models/file.py
```diff
@@ -1025,6 +1021,7 @@ async def get_async(
             if self.path and os.path.isfile(self.path)
             else self.path,
             md5=self.content_md5,
+            synapse_client=syn,
```
BryanFauble, line 1024: One of the spots that prompted me to make this change

synapseclient/models/folder.py
```diff
@@ -254,7 +251,11 @@ async def store_async(
                     entity=self, failure_strategy=None, synapse_client=synapse_client
                 )
             )
-            and (existing_folder := await Folder(id=existing_folder_id).get_async())
+            and (
+                existing_folder := await Folder(id=existing_folder_id).get_async(
+                    synapse_client=synapse_client
```
BryanFauble, line 256: One of the spots that prompted me to make this change

synapseclient/models/folder.py
```diff
@@ -338,6 +336,7 @@ async def get_async(
         await get_from_entity_factory(
             entity_to_update=self,
             synapse_id_or_path=entity_id,
+            synapse_client=synapse_client,
```
BryanFauble, line 339: One of the spots that prompted me to make this change

synapseclient/models/folder.py
```diff
@@ -427,34 +424,30 @@ async def copy_async(
...
+                setProvenance=file_copy_activity,
             ),
         )
 
         new_folder_id = source_and_destination.get(self.id, None)
         if not new_folder_id:
             raise SynapseError("Failed to copy folder.")
         folder_copy = await (
-            await Folder(id=new_folder_id).get_async()
+            await Folder(id=new_folder_id).get_async(synapse_client=syn)
```
BryanFauble, line 445: One of the spots that prompted me to make this change

synapseclient/models/project.py
```diff
@@ -475,35 +468,29 @@ async def copy_async(
...
+                setProvenance=file_copy_activity,
             ),
         )
 
         new_project_id = source_and_destination.get(self.id, None)
         if not new_project_id:
             raise SynapseError("Failed to copy project.")
         project_copy = await (
-            await Project(id=new_project_id).get_async()
+            await Project(id=new_project_id).get_async(synapse_client=syn)
```
BryanFauble, line 490: One of the spots that prompted me to make this change

synapseclient/client.py
```diff
@@ -596,17 +534,49 @@ def get_client(cls, synapse_client: typing.Union[None, "Synapse"]) -> "Synapse":
...
+                the `synapse_client` kwarg is available, and no instance has been cached
+                during the Synapse constructor.
         """
         if synapse_client:
             return synapse_client
 
         if not cls._synapse_client:
             raise SynapseError(
-                "No instance has been created - Please use login() first"
+                "No instance has been passed into a function where the `synapse_client` kwarg is available, and no instance has been cached during the Synapse constructor."
```
jaymedina, line 546: nit: The wording is a little confusing - I would re-write it as: 
> No Synapse client instance was provided, and no cached instance is available. Ensure that either an instance is passed into `synapse_client`, or a cached instance is available.

But this is subjective so I'll leave it up to you.

synapseclient/client.py
```diff
@@ -596,17 +534,49 @@ def get_client(cls, synapse_client: typing.Union[None, "Synapse"]) -> "Synapse":
...
+        """Determines whether OpenTelemetry is enabled for the Synapse client. This is
+        used to know whether or not this library will automatically kick off the
+        instruementation of several dependent libraries including:
+
+            - threading
+            - urllib
+            - requests
+            - httpx
+        """
+        if enable_open_telemetry and not Synapse._enable_open_telemetry:
```
jaymedina, line 574: I would add some comments here or in the docstring to clarify the logic a bit more:

```
# Turns OpenTelemetry on when it was previously off
...
# Turns OpenTelemetry off when it was previously on
...
```

BryanFauble: Added:

        When OpenTelemetry is enabled it will automatically start the instrumentation
        of these libraries. When it is disabled it will automatically stop the
        instrumentation of these libraries.

synapseclient/models/folder.py
```diff
@@ -427,34 +424,30 @@ async def copy_async(
...
         ).sync_from_synapse_async(
             download_file=False,
             synapse_client=synapse_client,
```
jaymedina, line 448: Shouldn't this be `synapse_client=syn` like above?

jaymedina (approved):
One last minor comment, apart from that everything looks good pending a passing build🔥

synapseclient/models/folder.py
BryanFauble: They end up working the exact same way because of how `syn = Synapse.get_client(synapse_client=synapse_client)` is used. However, I made a change here to make this consistent

thomasyu888 (approved):

PR #1130, opened 2024-09-17
[FDS-2386] Add owner id to the credentials to be used later on

BryanFauble:
**Problem:**

1. Within schematic there is a need to know the ID of the user that is logged in as it is passed to the Synapse servers to lookup data the user has access to. What needed to be done was another HTTP call to get the user profile and then extract the ownerId from that call.

**Solution:**

1. When logging into Synapse and getting the user profile stash the ownerId onto the credentials object so that schematic can use this and skip an HTTP call.

**Testing:**

1. Unit test
2. I have been using this code locally within schematic.

### DIFF
--- a/tests/unit/synapseclient/core/credentials/unit_test_cred_data.py
+++ b/tests/unit/synapseclient/core/credentials/unit_test_cred_data.py
@@ -13,6 +13,7 @@ class TestSynapseAuthTokenCredentials:
         self.username = "ahhhhhhhhhhhhhh"
         self.auth_token = "opensesame"
         self.displayname = "hhhhhaaaa"
+        self.owner_id = 123
         self.credentials = SynapseAuthTokenCredentials(
             self.auth_token, username=self.username, displayname=self.displayname
         )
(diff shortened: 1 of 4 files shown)


tests/unit/synapseclient/core/credentials/unit_test_cred_data.py
```diff
@@ -57,7 +58,8 @@ def test_repr(self):
             f"SynapseAuthTokenCredentials("
             f"username='{self.username}', "
             f"displayname='{self.displayname}', "
-            f"token='{self.auth_token}')" == repr(self.credentials)
+            f"token='{self.auth_token}')" == repr(self.credentials),
```
jaymedina, line 61: Update the syntax: 
`f"token='{self.auth_token}', "`

tests/unit/synapseclient/core/credentials/unit_test_cred_data.py
```diff
@@ -57,7 +58,8 @@ def test_repr(self):
...
+            f"owner_id={self.owner_id}",
```
jaymedina, line 62: and:
`f"owner_id={self.owner_id})",`

jaymedina (approved):

jaymedina:
Nice work! 🚀

PR #1131, opened 2024-09-20
[SYNSD-1233] Adding in a try/catch for HTTP 409 conflicts on entity creation

BryanFauble:
**Problem:**

1. If an HTTP Read error occurs when receiving back the response from creating a Synapse entity an HTTP 409 error will be thrown when the request is retried.

**Solution:**

1. Try out removing the `ReadError` and `ReadTimeout` as exceptions that are retried. Instead let the `SynapseHTTPError` be raised and wrap the POST call in a try/except that will get the entity from Synapse if it was a conflict, then compare the results to what the expected state was.

**Testing:**

1. I did very light manual testing for this by running the `syncToSynapse` command and I manually triggered the exception to be raised.

### DIFF
--- a/synapseclient/api/entity_services.py
+++ b/synapseclient/api/entity_services.py
@@ -9,6 +9,7 @@ from typing import TYPE_CHECKING, Any, Dict, List, Optional, Union
 from async_lru import alru_cache
 
 from synapseclient.core.exceptions import SynapseHTTPError
+from synapseclient.core.retry import RETRYABLE_CONNECTION_EXCEPTIONS_NO_READ_ISSUES
 
 if TYPE_CHECKING:
     from synapseclient import Synapse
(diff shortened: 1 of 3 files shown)


tests/integration/synapseclient/models/async/test_file_async.py
```diff
@@ -82,6 +85,82 @@ async def test_store_in_project(self, project_model: Project, file: File) -> Non
...
+        original_read = httpcore._backends.anyio.AnyIOStream.read
+        call_count = 0
+
+        async def new_read(*args, **kwargs) -> bytes:
+            nonlocal call_count
+            async_network_stream: httpcore.AsyncNetworkStream = (
+                spy_tls_stream.spy_return
+            )
+            call_count = call_count + 1
+            # This is very brittle, however, I do not see data to determine what the
```
thomasyu888, line 109: 👀 Does that mean if we add functions called in the async stream that we would have edit this call_count?

BryanFauble: Yeah, if more or less calls occur during the `store_async` function we would need to adjust this call count.

This specific way of replacing the read isn't working on GH actions. Though I did confirm the logic is working on my local machine. This needs to be fixed, but more digging into this is needed. I can get the hostname/IP addresses used, but being this "far" down in the libraries and working directly in httpcore, I lose a lot of context like request headers and the uri.

PR #1132, opened 2024-09-20
[SYNPY-1521] Fixes `asDataFrame` `kwarg` Collision Issue

BWMac:
**Problem:**

In #1127, I added the ability to pass `pandas.read_csv` keyword arguments to `CsvFileTable.asDataFrame` to give users the ability to have more control over the dataframe that was produced (such as determining which values should be treated as NA). An unintended side-effect of this was that users could unintentionally pass a keyword argument that is already set in the code that calls `read_csv` in `_csv_to_pandas_df`, causing an error like this one:
```
    446 # "ValueError: Only length-1 line terminators supported"
--> 447 df = pd.read_csv(
    448     filepath,
    449     dtype=dtype,
    450     sep=separator,
    451     lineterminator=line_terminator if len(line_terminator) == 1 else None,
    452     quotechar=quote_char,
    453     escapechar=escape_char,
    454     header=0 if contain_headers else None,
    455     skiprows=lines_to_skip,
    456     **kwargs,
    457 )
    458 # parse date columns if exists
    459 if date_columns:

TypeError: pandas.io.parsers.readers.read_csv() got multiple values for keyword argument 'sep'
```

**Solution:**

Add logic to `_csv_to_pandas_df` which creates a keyword arguments dictionary which is then updated to resolve any conflicts while giving precedence to the arguments passed intentionally by the user.

**Testing:**

- Tested that the change behaves as expected (works with duplicate keyword arguments) using this script:
```
import synapseclient
from synapseclient.table import CsvFileTable

syn = synapseclient.Synapse()
syn.login()

import pandas as pd

data = {
    'id': ['1', '2', '3', '4', '5'],
    'test_num': ['1', '2', '3', '4', '5'],
}

test_df = pd.DataFrame(data)
schema = synapseclient.table.Schema(
    name="test_table",
    parent="syn59478569",
    column_names=["id", "version"],
    column_types=["STRING", "STRING"],
)

test_table = CsvFileTable.from_data_frame(df=test_df, schema=schema)

# Data should be read in as expected without errors

test_table_df_sep = test_table.asDataFrame(sep=",")
test_table_df_sep.to_csv("test_table_with_sep.csv", index=False)

test_table_df_quote_char = test_table.asDataFrame(quotechar='"')
test_table_df_quote_char.to_csv("test_table_with_quote_char.csv", index=False)
```
- Added unit tests for `_csv_to_pandas_df`. Previously there were none.

### DIFF
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -418,7 +418,7 @@ def _csv_to_pandas_df(
                     Passed as `sep` to pandas. If `sep` is supplied as a `kwarg`
                     it will be used instead of this `separator` argument.
         quote_char: The quote character for the file,
-                    Defaults to `DEFAULT_QUOTE_CHARACTER`. 
+                    Defaults to `DEFAULT_QUOTE_CHARACTER`.
                     Passed as `quotechar` to pandas. If `quotechar` is supplied as a `kwarg`
                     it will be used instead of this `quote_char` argument.
         escape_char: The escape character for the file,
@@ -426,7 +426,7 @@ def _csv_to_pandas_df(
         contain_headers: Whether the file contains headers,
                     Defaults to `True`.
         lines_to_skip: The number of lines to skip at the beginning of the file,
-                        Defaults to `0`. Passed as `skiprows` to pandas. 
+                        Defaults to `0`. Passed as `skiprows` to pandas.
                         If `skiprows` is supplied as a `kwarg`
                         it will be used instead of this `lines_to_skip` argument.
         date_columns: The names of the date columns in the file
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -34,7 +34,7 @@ import os
 import re
 import tempfile
 from builtins import zip
-from typing import Dict, List, Tuple, TypeVar, Union, Optional, Any
+from typing import Any, Dict, List, Optional, Tuple, TypeVar, Union
 
 from synapseclient.core.constants import concrete_types
 from synapseclient.core.exceptions import SynapseError
@@ -408,7 +408,7 @@ def _csv_to_pandas_df(
     rowIdAndVersionInIndex: bool = True,
     dtype: Optional[Dict[str, Any]] = None,
     **kwargs,
-) -> "pd.DataFrame":
+):
     """
     Convert a csv file to a pandas dataframe
 
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -447,7 +447,6 @@ def _csv_to_pandas_df(
         "escapechar": escape_char,
         "header": 0 if contain_headers else None,
         "skiprows": lines_to_skip,
-        **kwargs,
     }
     pandas_args.update(kwargs)
 
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -450,7 +450,7 @@ def _csv_to_pandas_df(
         **kwargs,
     }
     pandas_args.update(kwargs)
-    
+
     # assign line terminator only if for single character
     # line terminators (e.g. not '\r\n') 'cause pandas doesn't
     # longer line terminators. See: <https://github.com/pydata/pandas/issues/3501>
@@ -458,9 +458,9 @@ def _csv_to_pandas_df(
     df = pd.read_csv(
         filepath,
         lineterminator=line_terminator if len(line_terminator) == 1 else None,
-        **pandas_args
-        )
-    
+        **pandas_args,
+    )
+
     # parse date columns if exists
     if date_columns:
         df = _convert_df_date_cols_to_datetime(df, date_columns)
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -440,21 +440,27 @@ def _csv_to_pandas_df(
 
     line_terminator = str(os.linesep)
 
+    pandas_args = {
+        "dtype": dtype,
+        "sep": separator,
+        "quotechar": quote_char,
+        "escapechar": escape_char,
+        "header": 0 if contain_headers else None,
+        "skiprows": lines_to_skip,
+        **kwargs,
+    }
+    pandas_args.update(kwargs)
+    
     # assign line terminator only if for single character
     # line terminators (e.g. not '\r\n') 'cause pandas doesn't
     # longer line terminators. See: <https://github.com/pydata/pandas/issues/3501>
     # "ValueError: Only length-1 line terminators supported"
     df = pd.read_csv(
         filepath,
-        dtype=dtype,
-        sep=separator,
         lineterminator=line_terminator if len(line_terminator) == 1 else None,
-        quotechar=quote_char,
-        escapechar=escape_char,
-        header=0 if contain_headers else None,
-        skiprows=lines_to_skip,
-        **kwargs,
-    )
+        **pandas_args
+        )
+    
     # parse date columns if exists
     if date_columns:
         df = _convert_df_date_cols_to_datetime(df, date_columns)
--- a/synapseclient/table.py
+++ b/synapseclient/table.py
@@ -415,14 +415,20 @@ def _csv_to_pandas_df(
     Arguments:
         filepath: The path to the file.
         separator: The separator for the file, Defaults to `DEFAULT_SEPARATOR`.
+                    Passed as `sep` to pandas. If `sep` is supplied as a `kwarg`
+                    it will be used instead of this `separator` argument.
         quote_char: The quote character for the file,
-          Defaults to `DEFAULT_QUOTE_CHARACTER`.
+                    Defaults to `DEFAULT_QUOTE_CHARACTER`. 
+                    Passed as `quotechar` to pandas. If `quotechar` is supplied as a `kwarg`
+                    it will be used instead of this `quote_char` argument.
         escape_char: The escape character for the file,
                     Defaults to `DEFAULT_ESCAPSE_CHAR`.
-                    contain_headers: Whether the file contains headers,
+        contain_headers: Whether the file contains headers,
                     Defaults to `True`.
         lines_to_skip: The number of lines to skip at the beginning of the file,
-                        Defaults to `0`.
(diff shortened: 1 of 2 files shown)


synapseclient/table.py
```diff
@@ -397,16 +397,16 @@ def _convert_df_date_cols_to_datetime(
...
+    quote_char: str = DEFAULT_QUOTE_CHARACTER,
+    escape_char: str = DEFAULT_ESCAPSE_CHAR,
+    contain_headers: bool = True,
+    lines_to_skip: int = 0,
+    date_columns: Optional[List[str]] = None,
+    list_columns: Optional[List[str]] = None,
+    rowIdAndVersionInIndex: bool = True,
+    dtype: Optional[Dict[str, Any]] = None,
     **kwargs,
 ):
```
jaymedina, line 411: nit: Add a pandas df return type here

BWMac: I can't add this unfortunately because pandas is imported within the function definition so doing that causes an error

tests/unit/synapseclient/unit_test_tables.py
```diff
@@ -251,6 +253,198 @@ def test_convert_df_date_cols_to_datetime() -> None:
...
+        )
+
+        # AND I expect the returned DataFrame to be
+        # the same as the original DataFrame (file)
+        pd.testing.assert_frame_equal(df, expected_df)
+
+
+def test_csv_to_pandas_df_with_kwargs() -> None:
+    # GIVEN a pandas DataFrame (CSV file stand-in)
+    expected_df = pd.DataFrame({"col1": [1, 2, 3], "col2": ["a", "b", "c"]})
```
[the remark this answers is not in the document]
BWMac: I could, but this test is intended to just make sure that the `kwargs` are passed appropriately. I chose one argument that we expect to overwrite the default set in our code (`escapechar`) and one where no default is set by us (`keep_default_na`) which ensures that the issue reported in this bug has been resolved. 

Plus, I am mocking `pandas.read_csv` so just ensuring that the argument is passed to it should be enough IMHO since we don't want to test its functionality in this test.

synapseclient/table.py
jaymedina: OK this isn't a huge deal so I'll resolve this. I'm wondering why we're importing `pandas` this way instead of at the top, and why there's this extra test for importing pandas, it seems kind of odd. Tagging @BryanFauble in here for when he gets back

thomasyu888: I can chime in here, it's because pandas historically is an optional dependency of the client.

jaymedina: Good to know!

synapseclient/table.py
```diff
@@ -440,21 +440,27 @@ def _csv_to_pandas_df(
...
     line_terminator = str(os.linesep)
 
+    pandas_args = {
+        "dtype": dtype,
+        "sep": separator,
+        "quotechar": quote_char,
+        "escapechar": escape_char,
+        "header": 0 if contain_headers else None,
+        "skiprows": lines_to_skip,
+        **kwargs,
```
thomasyu888, line 450: Is this line necessary with the next line?

jaymedina (approved):
LGTM! 🔥

synapseclient/table.py
```diff
@@ -440,21 +440,27 @@ def _csv_to_pandas_df(
...
+    }
+    pandas_args.update(kwargs)
```
thomasyu888, line 452: What happens when you add a kwargs pandas doesn't accept? Does to throw a pandas error?

Not saying we need to catch it, just curious

BWMac: Yeah it would throw a `pandas` error like

```
  File "/Users/bmacdonald/Desktop/test_duplicate_kwargs.py", line 28, in <module>
    test_table_df_sep = test_table.asDataFrame(test=",")
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bmacdonald/Development/repos/synapsePythonClient/synapseclient/table.py", line 2526, in asDataFrame
    return _csv_to_pandas_df(
           ^^^^^^^^^^^^^^^^^^
  File "/Users/bmacdonald/Development/repos/synapsePythonClient/synapseclient/table.py", line 458, in _csv_to_pandas_df
    df = pd.read_csv(
         ^^^^^^^^^^^^
TypeError: read_csv() got an unexpected keyword argument 'test'
```

which I chose not to handle because if the arguments `read_csv` change we would have to update how we handle it

BWMac: `separator` is a default `_csv_to_pandas_df` argument so if you pass that to `asDataFrame` it will error with 
```
  File "/Users/bmacdonald/Desktop/test_duplicate_kwargs.py", line 28, in <module>
    test_table_df_sep = test_table.asDataFrame(separator=",", sep=",")
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/bmacdonald/Development/repos/synapsePythonClient/synapseclient/table.py", line 2526, in asDataFrame
    return _csv_to_pandas_df(
           ^^^^^^^^^^^^^^^^^^
TypeError: synapseclient.table._csv_to_pandas_df() got multiple values for keyword argument 'separator'
```

Whereas `sep` would overwrite the default `separator` argument because it is a `pandas.read_csv` argument rather than a `_csv_to_pandas_df` arg.

thomasyu888: Right, I guess my question is more of a "philosophical" one. Do we accept the `sep="\t"` or do we stick with `separator=","` because that is the defined argument and not `kwargs`?

That said, when this is refactored, I envision there to not be a `seperator` parameter at all.

BWMac: I was approaching this with a bias towards intentionality. If a user chooses to set `sep="\t"`, that should override the default `separator` argument set in `_csv_to_pandas_df` because that is probably what a user would expect to happen. If we enforce the default behavior we would need to provide some message to say that this is happening or there would be a confusing silent failure happening from the user's perspective.

I also imagine that once this is refactored this decision will be abstracted away, anyway.

synapseclient/table.py
BWMac: Nope, removed it in the latest commit.

synapseclient/table.py
thomasyu888: Right, what are the changes a user would set

`separator` to be something different? Should we add a blurb saying that the kwargs take precedence?

BryanFauble: > Right, what are the chances a user would set `separator` to be something different? Should we add a blurb saying that the kwargs take precedence? Like
> 
> Separator='L' Sep='M'
> 
> It would take sep

+1, perhaps we add to each function argument what `kwargs` would replace it. See my example for `separator` for my initial thought:

```
    """
    Convert a csv file to a pandas dataframe
    Arguments:
        filepath: The path to the file.
        separator: The separator for the file, Defaults to `DEFAULT_SEPARATOR`. Passed as `sep` to pandas. If `sep` is supplied as a `kwarg` it will be used instead of this `separator` argument.
        quote_char: The quote character for the file,
          Defaults to `DEFAULT_QUOTE_CHARACTER`.
        escape_char: The escape character for the file,
                    Defaults to `DEFAULT_ESCAPSE_CHAR`.
                    contain_headers: Whether the file contains headers,
                    Defaults to `True`.
        lines_to_skip: The number of lines to skip at the beginning of the file,
                        Defaults to `0`.
        date_columns: The names of the date columns in the file
        list_columns: The names of the list columns in the file
        rowIdAndVersionInIndex: Whether the file contains rowId and
                                version in the index, Defaults to `True`.
        dtype: The data type for the file, Defaults to `None`.
        **kwargs: Additional keyword arguments to pass to pandas.read_csv. See
                    https://pandas.pydata.org/docs/reference/api/pandas.read_csv.html
                    for complete list of supported arguments.
    Returns:
        A pandas dataframe
    """
```

BWMac: Well users shouldn't be calling `_csv_to_pandas_df` directly, it's called by `asDataFrame` so if they set `separator` and `sep` as `kwargs` in `asDataFrame` you would get this error [above](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1132#discussion_r1771956172). You shouldn't really encounter a situation where someone is doing that though because that would be a pretty weird thing to do. We can add something saying that the `kwargs` take precendence but it would be kind of confusing potentially because `_csv_to_pandas_df` and its default arguments are somewhat hidden.

BryanFauble (approved):
Logic looks good to me! Thanks, just 1 comment on how we could improve the docstring.

synapseclient/table.py
BWMac: We could do what @BryanFauble suggests, but then we are maintaining documentation on `pandas` arguments which could potentially change in a future version which I don't love. But happy to do it if we think it's necessary.

BryanFauble: @BWMac Should we add these types of comments to the top level where a user enters into this logic. That way everything is explicitly stated for how the function arguments are going to be evaluated.

Some areas at a glance:

1. https://github.com/Sage-Bionetworks/synapsePythonClient/blob/857154e45ddd9f85425b7417d349786da633e8f2/synapseclient/table.py#L2414-L2427
2. https://github.com/Sage-Bionetworks/synapsePythonClient/blob/857154e45ddd9f85425b7417d349786da633e8f2/synapseclient/table.py#L2476-L2496

BryanFauble: > so if they set separator and sep as kwargs in asDataFrame

This might be possible because one is set in the class instance constructor, and the other is passed in via the function argument.

BWMac: Ah right that's a really good point. I like the idea of putting this information closer to the top level so users will see it. I will take a look later and make some updates to documentation to help users navigate this. Thanks for the input @BryanFauble and @thomasyu888 !

BWMac: Updated docstrings in latest commit

PR #1134, opened 2024-09-25
[SYNPY-1517] Deprecate Python 3.8 Support

BWMac:
**Description:**
This PR deprecates support for Python 3.8 in the Python Client. It also adjusts the `pandas` versions required in the project and removes the `numpy` version specification because we no longer have to allow for Python 3.8. 

**Documentation:**
I have also updated the relevant documentation to no longer indicate support for Python 3.8.

I have done my best to find all of the places where documentation needs to be adjusted, but please let me know if I have missed anything.


synapseutils/sync.py
```diff
@@ -584,7 +584,7 @@ async def _upload_item_async(
 
 @deprecated(
     version="4.4.0",
-    reason="To be removed in 5.0.0. This is being replaced by `generate_manifest`.",
```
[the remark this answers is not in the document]
BWMac: I guess my perspective is if the goal of including these decorators is to get users to start using the new implementations and we keep them in it gives people more time to make that change. And in the meantime we can look into the potential SYNR impacts before 6.0 is released. I don't feel too strongly about it though.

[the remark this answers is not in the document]
thomasyu888: Makes sense, couple of things to look into

1. Could we look into if other open source packages do a major release when dropping support for older python versions?
2. If true for 1, then the October of each year would be a major release.

README.md
```diff
@@ -36,9 +36,9 @@ or by sending an email to [python-announce+subscribe@sagebase.org](mailto:python
 Installation
 ------------
 
-The Python Synapse client has been tested on 3.8, 3.9, 3.10 and 3.11 on Mac OS X, Ubuntu Linux and Windows.
+The Python Synapse client has been tested on versions 3.9, 3.10, 3.11 and 3.12 on Mac OS X, Ubuntu Linux and Windows.
 
-**Starting from Synapse Python client version 3.0, Synapse Python client requires Python >= 3.8**
+**Starting from Synapse Python client version 3.0, Synapse Python client requires Python >= 3.9**
```
jaymedina, line 41: The `Python >= 3.9` requirement will begin at Synapse Python client version 5.0, right?

BWMac: This was going to be released with 5.0, but now it won't so I will revert

setup.cfg
```diff
@@ -63,6 +62,7 @@ install_requires =
     tqdm>=4.66.2,<5.0
     async-lru~=2.0.4
     psutil~=5.9.8
+    importlib_resources~=6.0.0
```
BryanFauble, line 65: Why is this needed over `import importlib.resources`?

setup.cfg
```diff
@@ -100,11 +100,10 @@ tests =
     pytest-rerunfailures~=12.0
     func-timeout~=4.3
     pytest-cov~=4.1.0
-    pandas>=1.5,<3.0
+    pandas>=2.1.2,<3.0
 
 pandas =
-    pandas>=1.5,<3.0
-    numpy<2.0.0
+    pandas>=2.1.2,<3.0
```
BryanFauble, line 106: Does pandas 1.x no longer work?

setup.cfg
```diff
@@ -43,7 +42,7 @@ classifiers =
 zip_safe = False
 include_package_data = True
 packages = find:
-python_requires = >=3.8
+python_requires = >=3.9
```
thomasyu888, line 45: Python 3.13 drops soon, we may want to restrict that

setup.cfg
BWMac: I did that because of this [comment](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/97085dea6ae23478e1021c27938992cf439a62f4/synapseclient/core/version_check.py#L168), but maybe that was a typo and it was meant to still be importlib.resources?

BryanFauble: Let's see if we can continue to use the existing library without adding extra dependencies. I was looking at the pypi page it was mentioned it was for backporting functionality for old python versions, which isn't the situation we are in.

setup.cfg
BryanFauble: If newer versions don't support python 3.8, that's fine since we are removing it.

Does 1.5.x support python 3.9 and above?

I would like to keep the ranges as big as possible.

BWMac: 1.5.x supports 3.8-3.11

BryanFauble: With that being the case, should we revert this change then, and continue to support `pandas>=1.5,<3.0`?

pip should handle solving for the available dependencies in this case; we are not excluding anyone who still wants/needs to use v1 of pandas.

synapseclient/__init__.py
```diff
@@ -1,6 +1,6 @@
-import importlib.resources
 import json
 
+import importlib_resources
```
BryanFauble, line 3: This was the library we didn't want to use right?

.github/workflows/build.yml
```diff
@@ -78,7 +78,7 @@ jobs:
           path: |
             ${{ steps.get-dependencies.outputs.site_packages_loc }}
             ${{ steps.get-dependencies.outputs.site_bin_dir }}
-          key: ${{ runner.os }}-${{ matrix.python }}-build-${{ env.cache-name }}-${{ hashFiles('setup.py') }}-v18
+          key: ${{ runner.os }}-${{ matrix.python }}-build-${{ env.cache-name }}-${{ hashFiles('setup.py') }}-v19
```
BryanFauble, line 81: Bump once again to `v20` since the cache is currently using the version that had the library we removed

synapseclient/__init__.py
BWMac: 😅

BWMac: fixed

BryanFauble (approved):
Thanks for updating some of the docs I should have tackled when adding 3.12 :D

PR #1135, opened 2024-09-25
[SYNPY-1513] Validate input submission ID in `getSubmission(...)`

jaymedina:
## problem

The endpoint `evaluation/submission/{id}` was receiving communications with submission IDs that had decimal points in them (e.g. `123.0`). This was obstructing an ingestion process in Snowflake, and should not exist in the first place. Submission IDs are not ID types that support decimal points.

## solution

In `getSubmission(...)` and other related methods in `client.py`, the `id_of(...)` method from `utils.py` is utilized to retrieve a valid entity ID. This method was being used for submission IDs, but was effectively useless, as it would just return the same string object or numbers object (as a string) without verifying its format.

The solution involves creating a separate method called `validate_submission_id` and replacing `id_of` with it, to enforce the no-decimal rule in the input.

- [x] Create `utils.validate_submission_id` to handle decimals in input IDs
- [x] Add/update relevant unit and integration tests

### Test `getSubmission`
```
In [1]: import synapseclient

In [2]: syn = synapseclient.login()
Welcome, Jenny Medina!

# Testing an expected input in `getSubmission`
In [3]: syn.getSubmission("9745366")
Out[3]: 
{'id': '9745366',
 [ ... ]
 'filePath': None}

# Testing an input with a decimal in `getSubmission`
In [4]: syn.getSubmission("9745366.0")
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 syn.getSubmission("9745366.0")

[ ... ]

ValueError: Invalid submission ID: 9745366.0. ID can either be an integer or a string with no decimals.

In [5]: 
```

### Test `getSubmissionStatus`
```
In [5]: syn.getSubmissionStatus("9745366")
Out[5]: 
{'id': '9745366',
[ ... ]
 'statusVersion': 7}

In [6]: syn.getSubmissionStatus("9745366.0")
ValueError                                Traceback (most recent call last)
Cell In[6], line 1
----> 1 syn.getSubmissionStatus("9745366.0")

[ ... ]

ValueError: Invalid submission ID: 9745366.0. ID can either be an integer or a string with no decimals.

In [7]: 
```

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -4812,10 +4812,11 @@ class Synapse(object):
         self, id: typing.Union[str, int, collections.abc.Mapping], **kwargs
     ) -> Submission:
         """
-        Gets a [synapseclient.evaluation.Submission][] object by its id.
+        Gets a [synapseclient.evaluation.Submission][] object based on a given ID
+        or previous [synapseclient.evaluation.Submission][] object.
 
         Arguments:
-            id: The id of the submission to retrieve
+            id: The ID of the submission to retrieve or a [synapseclient.evaluation.Submission][] object
 
         Returns:
             A [synapseclient.evaluation.Submission][] object
@@ -4856,13 +4857,13 @@ class Synapse(object):
         return submission
 
     def getSubmissionStatus(
-        self, submission: typing.Union[str, int]
+        self, submission: typing.Union[str, int, collections.abc.Mapping]
     ) -> SubmissionStatus:
         """
-        Downloads the status of a Submission.
+        Downloads the status of a Submission given its ID or previous [synapseclient.evaluation.Submission][] object.
 
         Arguments:
-            submission: The submission to lookup
+            submission: The submission to lookup (ID or [synapseclient.evaluation.Submission][] object)
 
         Returns:
             A [synapseclient.evaluation.SubmissionStatus][] object
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -4808,7 +4808,9 @@ class Synapse(object):
             if next_page_token is None:
                 break
 
-    def getSubmission(self, id: typing.Union[str, int], **kwargs) -> Submission:
+    def getSubmission(
+        self, id: typing.Union[str, int, collections.abc.Mapping], **kwargs
+    ) -> Submission:
         """
         Gets a [synapseclient.evaluation.Submission][] object by its id.
 
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -114,6 +114,7 @@ from synapseclient.core.utils import (
     is_integer,
     is_json,
     require_param,
+    validate_submission_id,
 )
 from synapseclient.core.version_check import version_check
 
@@ -4807,7 +4808,7 @@ class Synapse(object):
             if next_page_token is None:
                 break
 
-    def getSubmission(self, id, **kwargs):
+    def getSubmission(self, id: typing.Union[str, int], **kwargs) -> Submission:
         """
         Gets a [synapseclient.evaluation.Submission][] object by its id.
 
@@ -4823,7 +4824,7 @@ class Synapse(object):
              on the *downloadFile*, *downloadLocation*, and *ifcollision* parameters
         """
 
-        submission_id = id_of(id)
+        submission_id = validate_submission_id(id)
         uri = Submission.getURI(submission_id)
         submission = Submission(**self.restGET(uri))
 
@@ -4852,7 +4853,7 @@ class Synapse(object):
 
         return submission
 
-    def getSubmissionStatus(self, submission):
+    def getSubmissionStatus(self, submission: typing.Union[str, int]) -> SubmissionStatus:
         """
         Downloads the status of a Submission.
 
@@ -4863,7 +4864,7 @@ class Synapse(object):
             A [synapseclient.evaluation.SubmissionStatus][] object
         """
 
-        submission_id = id_of(submission)
+        submission_id = validate_submission_id(submission)
         uri = SubmissionStatus.getURI(submission_id)
         val = self.restGET(uri)
         return SubmissionStatus(**val)
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -4853,7 +4853,9 @@ class Synapse(object):
 
         return submission
 
-    def getSubmissionStatus(self, submission: typing.Union[str, int]) -> SubmissionStatus:
+    def getSubmissionStatus(
+        self, submission: typing.Union[str, int]
+    ) -> SubmissionStatus:
         """
         Downloads the status of a Submission.
 
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -242,6 +242,32 @@ def id_of(obj: typing.Union[str, collections.abc.Mapping, numbers.Number]) -> st
     raise ValueError("Invalid parameters: couldn't find id of " + str(obj))
 
 
+def validate_submission_id(submission_id: typing.Union[str, int]) -> str:
+    """
+    Ensures that a given submission ID is either an integer or a string that
+    can be converted to an integer. Version notation is not supported for submission
+    IDs, therefore decimals are not allowed.
+
+    Arguments:
+        submission_id: The submission ID to validate
+
+    Returns:
+        The submission ID as a string
+
+    Raises:
+        ValueError: if the submission ID is invalid
+
+    """
+    if isinstance(submission_id, int):
+        return str(submission_id)
+    elif isinstance(submission_id, str) and submission_id.isdigit():
+        return submission_id
+    else:
+        raise ValueError(
+            f"Invalid submission ID: {submission_id}. ID can either be an integer or a string with no decimals."
+        )
+
+
 def concrete_type_of(obj: collections.abc.Mapping):
     """
     Return the concrete type of an object representing a Synapse entity.
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -53,7 +53,6 @@ SLASH_PREFIX_REGEX = re.compile(r"\/[A-Za-z]:")
 # Set up logging
 LOGGER_NAME = DEFAULT_LOGGER_NAME
 LOGGER = logging.getLogger(LOGGER_NAME)
-logging.getLogger("py.warnings").handlers = LOGGER.handlers
 
 
 def md5_for_file(
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -272,7 +272,7 @@ def validate_submission_id(
     elif isinstance(submission_id, collections.abc.Mapping):
         syn_id = _get_from_members_items_or_properties(submission_id, "id")
         if syn_id is not None:
-            return str((float(syn_id)))
+            return validate_submission_id(syn_id)
     else:
         int_submission_id = int(float(submission_id))
         LOGGER.warning(
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -27,10 +27,11 @@ import warnings
 import zipfile
 from dataclasses import asdict, is_dataclass
 from typing import TYPE_CHECKING, TypeVar
-from synapseclient.core.logging_setup import DEFAULT_LOGGER_NAME
 
 import requests
 from opentelemetry import trace
+
+from synapseclient.core.logging_setup import DEFAULT_LOGGER_NAME
 
 if TYPE_CHECKING:
     from synapseclient.models import File, Folder, Project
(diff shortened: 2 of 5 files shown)


jaymedina:
Hey @BryanFauble @thomasyu888 , before I move forward with this solution I'd like to run it by you both since it is a breaking change (see latest tests) and involves a little bit of re-structuring everywhere where `id_of` is called. When you have some time, see my comment in "Design considerations" for my reasoning, and let me know what you think.

Thanks!

BryanFauble:
@jaymedina 

> it made more sense to refactor this ID-handling method to correct for decimals in all these other entities, since the only entity type that can support decimals is File entities.

The issue with this is that is we might not know ahead of time if an ID is for an entity is a File, or some other entity in Synapse. Take the example where `syn.get()` is a factory that returns back several types of entities/resources, and we don't know what the entity type is until after we call Synapse, after we have called `id_of`.

I tested out the dot notation with a project, and it seems like on the backend these entities do have a version applied to them, but it can only ever be version 1, "syn54022402.2" fails:

```


syn = Synapse()
syn.login()

project = syn.get("syn54022402")
print(project)
project = syn.get("syn54022402.1")
print(project)
```

I think at the moment it just makes sense to worry about what's in front of us and only update `syn.getSubmission()` with the new logic.

Should Synapse itself should be performing this data check?

jaymedina:
> Should Synapse itself should be performing this data check?

That's a really good point @BryanFauble and I guess it depends on the scope of responsibility within synpy. Is the current expectation when we develop on this client that synpy validates inputs before communicating with the API or do we leave that responsibility with the API to handle these cases (in this context, stripping the decimal from submission IDs)? Validating on the synpy side feels more flexible and would prevent unnecessary communications with Synapse, but also adds complexity to the codebase like what's going to be in this PR.

Right now I lean towards keeping small validation logic like this in synpy (under `utils.py`) because of the flexibility of being able to write our own warning/error messages, and the low overhead of our code review process.

If we go with handling the inputs on the client side, I would go back to my previous solution of handling the submission IDs with a new function in `utils.py`, since `id_of` is for synapse IDs and not submission IDs.

BryanFauble:
> Is the current expectation when we develop on this client that synpy validates inputs before communicating with the API or do we leave that responsibility with the API to handle these cases (in this context, a submission ID having a decimal)?

It's important for both. Think about the client/server relationship as it pertains to a Web UI and a backend server. The UI can have input validation, but the server needs to have that final strict enforcement of the rules.


> Right now I lean towards keeping small validation logic like this in synpy (under utils.py) because of the flexibility of being able to write our own warning/error messages, and the low overhead of our code review process.

I agree that we can handle this client side. There are a bunch of these similar checks in the code base already - For example making sure that an entity ID starts as `syn#####`.

> If we go with handling the inputs on the client side, I would go back to my previous solution of handling the submission IDs with a new function in utils.py, since id_of is for synapse IDs and not submission IDs.

I agree that this should be a new function, since the rules for a submission ID are different from those of an entity/synapse ID.

tests/integration/synapseclient/test_evaluations.py
```diff
@@ -110,7 +110,7 @@ async def test_evaluations(syn: Synapse, project: Project):
         submissions = syn.getSubmissions(ev, limit=num_of_submissions - 1)
         for submission in submissions:
             assert re.match("Submission \\d+", submission["name"])
-            status = syn.getSubmissionStatus(submission)
+            status = syn.getSubmissionStatus(submission.get("id"))
```
jaymedina, line 113: This used to work because `id_of` has some functionality to handle mapping objects like dictionaries. Because we are now validating for non-decimal submission IDs, we're no longer using `id_of`, meaning `getSubmissionStatus` and `getSubmission` no longer support mapping.

Supporting mapping was not part of their intended functionality based on the docstring for `getSubmission`:

```
"""
Gets a [synapseclient.evaluation.Submission][] object by its id.

Arguments:
    id: The id of the submission to retrieve
```

so this is not a feature change.

synapseclient/client.py
```diff
@@ -4823,7 +4824,7 @@ def getSubmission(self, id, **kwargs):
              on the *downloadFile*, *downloadLocation*, and *ifcollision* parameters
         """
 
-        submission_id = id_of(id)
+        submission_id = validate_submission_id(id)
```
thomasyu888, line 4827: Just making note that this should still work:

```
import synapseclient
syn = synapseclient.login()
syn.getSubmission(<id here>)
sub_obj = syn.getSubmission(<id here>)
syn.getSubmission(sub_obj)
```

thomasyu888: This is a perfect example of why type hints would've been great when this function was originally implemented 😅 
Furthermore, there _should be_ an integration test for this scenario.

jaymedina: Hmm, why would a person get a submission object with the submission object itself, if the submission object is what they're trying to retrieve? Hope that makes sense

synapseclient/core/utils.py
```diff
@@ -242,6 +242,32 @@ def id_of(obj: typing.Union[str, collections.abc.Mapping, numbers.Number]) -> st
...
+    Raises:
+        ValueError: if the submission ID is invalid
+
+    """
+    if isinstance(submission_id, int):
+        return str(submission_id)
+    elif isinstance(submission_id, str) and submission_id.isdigit():
+        return submission_id
+    else:
+        raise ValueError(
```
thomasyu888, line 266: Ah interesting, sorry if there was any miss-communication, what are your thoughts on doing something like...

```
str(int(submission_id))
```

tests/integration/synapseclient/test_evaluations.py
thomasyu888: This is a feature change, because people are no longer able to pass in the `Submission` object

thomasyu888: So I realize there is some confusion, but the docstring was basically wrong, it didn't just take id.  Even if that may have been the intended case for the function, the reality is we did support these the `Submission` object.

synapseclient/core/utils.py
jaymedina: I see, thanks for clarifying! I don't like the idea of silently transforming user inputs, so this wasn't my initial thought, but I can definitely patch it this way and throw a warning message to the user saying "Your ID has been changed", just in case a user intentionally fed a float submission ID thinking submission IDs also support dot notation.

I'll make this update in the morning, thanks!

thomasyu888: Agreed, there are vulnerabilities when doing `str(int(234234))`.  There are a couple of places in the code base that do something like `str(int(...))`.  

The problem with using `int` is that a user can technically accidentally pass in: `str(int(234234.3435346))`, which returns '234234'.  At this moment, I think that is _ok_.  I like your idea of pushing a `logger.warning`.

synapseclient/client.py
jaymedina: > My recommendation is to not add complexity to the validate_submission_id function to check for the Submission object,

I can see this pov and was thinking on those same lines, but I started fiddling with the code (see changes [here](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1135/commits/c33f42448868ac9f88439f2745e72ca9d8be8ff0)) and I don't think it adds much more complexity:

This added an extra conditional and introduced the same dictionary-handling function that's in `id_of`, which makes `validate_submission_id` have the same level of abstraction as `id_of`. From my pov, it makes `id_of` and `validate_submission_id` look more like analog functions, which is what we're going for here because that's essentially what they are.

```
if isinstance(submission_id, int):
    return str(submission_id)
elif isinstance(submission_id, str) and submission_id.isdigit():
    return submission_id
elif isinstance(submission_id, collections.abc.Mapping):
    syn_id = _get_from_members_items_or_properties(submission_id, 'id')
    if syn_id is not None:
        return str((float(syn_id)))
```

With these changes, we now have support for input Submission objects in `getSubmission` and `getSubmissionStatus`:

```
In [18]: sub = syn.getSubmissionStatus("9745366.0")
[WARNING] Submission ID '9745366.0' contains decimals which are not supported. Submission ID will be converted to '9745366'.

In [19]: sub2 = syn.getSubmissionStatus(sub)

In [20]: sub3 = syn.getSubmission(sub2)

In [21]: sub3
Out[21]: 
{'id': '9745366',
...
```

I haven't added tests for this because we're still deciding. but if we are keeping support for `getSubmissionBundles` and Submission objects as inputs for `getSubmission` and `getSubmissionStatus`, then I'm voting for keeping `validate_submission_id` in the state that it's in in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1135/commits/c33f42448868ac9f88439f2745e72ca9d8be8ff0

synapseclient/core/utils.py
jaymedina: Changes made https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1135/commits/aa632fbc3ffd8c7c9d42e64785f1134034effcd8

synapseclient/client.py
BryanFauble: > With these changes, we now have support for input Submission objects in getSubmission and getSubmissionStatus:

This is great, and what I would expect as of right now in the client.

Some thoughts on future state:

I would like for this to eventually evolve to work similarly to the function signature I've written for the `entity_factory.py` defined here: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/api/entity_factory.py#L26-L38 . The idea is that any `client/Synapse` class methods are only used as a factory for these classes, but then any operations take place of those objects themselves. For example you'd call `.get_submission()` or `.get_submissions()` and it's return a `Submission` object which would have a `.status` attribute on the class instance.

jaymedina: Thanks for the feedback all - I moved forward with this solution and am working on one last unit test that ensures the inputs of `getSubmissionStatus` and `getSubmission` work together, and then this will be good to go for a final review.

Also noticed we don't have much code coverage for `getSubmissionStatus` itself, so I made a [tech debt ticket](https://sagebionetworks.jira.com/browse/IBCDPE-1093) to cover this.

synapseclient/core/utils.py
```diff
@@ -242,6 +250,38 @@ def id_of(obj: typing.Union[str, collections.abc.Mapping, numbers.Number]) -> st
...
+    if isinstance(submission_id, int):
+        return str(submission_id)
+    elif isinstance(submission_id, str) and submission_id.isdigit():
+        return submission_id
+    elif isinstance(submission_id, collections.abc.Mapping):
+        syn_id = _get_from_members_items_or_properties(submission_id, "id")
+        if syn_id is not None:
+            return validate_submission_id(syn_id)
+    else:
+        int_submission_id = int(float(submission_id))
```
thomasyu888, line 277: What is the error if we pass in a random string?  Currently it's.
```
SynapseHTTPError: 404 Client Error: 
Submission 'fasdf' does not exist
```

I think the error will become:
```
----> 1 float("fasdf")

ValueError: could not convert string to float: 'fasdf'
```

which _could_ be fine

jaymedina: Good point, I've implemented a catch for non-digit values in a string input:

```
In [55]: sub = syn.getSubmission("test")
[...]
ValueError: could not convert string to float: 'test'

During handling of the above exception, another exception occurred:

[...]

ValueError: Submission ID 'test' is not a valid submission ID. Please use digits only.
```

synapseclient/core/utils.py
```diff
@@ -47,6 +50,11 @@
 
 SLASH_PREFIX_REGEX = re.compile(r"\/[A-Za-z]:")
 
+# Set up logging
+LOGGER_NAME = DEFAULT_LOGGER_NAME
```
thomasyu888, line 54: @BryanFauble does this have any impact on the logging for the client?

thomasyu888 (approved):
🔥 LGTM! I'll leave it to @BryanFauble for final review.

synapseclient/core/utils.py
BryanFauble: @jaymedina Any changes to the logging should be done in `synapseclient/core/logging_setup.py`

Can you give a small recap of what this change would be accomplishing?

jaymedina: My intent here was implementing a logger instance, similar to the way I was it being done in [retry.py](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/1918ec2a667570637fa9c16a459bb55879c1e1d5/synapseclient/core/retry.py#L117). I chose `DEFAULT_LOGGER_NAME` since we are just logging a `Warning` in this case.

BryanFauble: For this line:
`logging.getLogger("py.warnings").handlers = LOGGER.handlers`

This is changing some of the handlers on the logger instance you are grabbing, could this logic live in `logging_setup.py` instead?

jaymedina: Hi @BryanFauble - I did a bit more research on the logging setup and handlers, and realized this line is redundant because I am not using a native python warning, so there's no need to update the handlers with our custom config. I've removed this line and ran another test, and the warning is the same.

```
In [4]: sub = syn.getSubmission("9745366.0")
[WARNING] Submission ID '9745366.0' contains decimals which are not supported. Submission ID will be converted to '9745366'.
```

BryanFauble (approved):
This is great! Thanks for the thorough tests!

PR #1143, opened 2024-10-10
[FDS-2458] Attach the user.id of the logged in user to the trace span

BryanFauble:
**Problem:**

1. When the user ID is retrieved the data is not added to any trace, by providing this information we can use telemetry data to track a user action in the data.

**Solution:**

1. When the credentials are retrieved attach the user.id if the span is being recorded.

**Testing:**

1. I verified that the data is being passed along as expected:

image


thomasyu888 (approved):
🔥 LGTM!

PR #1145, opened 2024-11-01
[SYNPY-1423] update mkdocstrings dependency for 1.8.0

danlu1:
Problem:

1. When using mkdocstrings < 1.8.0, custom CSS is used to define symbol type before heading and in toc. 
2. We want to render cross-references for type annotations in signatures for modules from the standard library.
3. Docstrings in json_schema.py are not following Google Style.

Solution:

1. Remove the customization and implemented standard solution of showing symbol_type in headings and toc.
2. Implemented singnature cross-references
3. Update docstrings in json_schema.py


synapseclient/services/json_schema.py
```diff
@@ -144,9 +140,10 @@ def expand(self):
     def bind_to_object(self, synapse_id: str):
         """Bind schema to an entity
 
-        :param synapse_id: Synapse Id to bind json schema to.
-        :type synapse_id:  str
+        Args:
```
BryanFauble, line 143: I have usually preferred to use `Arguments:` over `Args:`, but they both work as far as mkdocs is concerned.

BryanFauble (approved):
I started up the mkdocs site locally and everything is looking good! Thanks for making the changes here.

I really like how `signature_crossrefs` is working and being able to click the signatures and be taken to the referenced types.

synapseclient/services/json_schema.py
danlu1: Yea, they both work but I will convert to use `Arguments`. `Args` is automatically generated by autoDocstring.

PR #1146, opened 2024-11-06
Add procps to Dockerfile to enable use of container in Nextflow

adamjtaylor:
This pull request includes a small change to the `Dockerfile`. The change adds the `procps` package to the list of installed packages. `ps` from `procps` is required by Nextflow. There are many use cases where having auto-built, latest version of synapsePythonClient in a Docker container will be valuable for users.

* [`Dockerfile`](diffhunk://#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R7): Added `procps` to the list of installed packages.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,7 @@ RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selectio
 RUN apt-get update \
     && apt-get install --no-install-recommends -y \
     gcc \
+    procps \
     python3-dev \
     python3-setuptools \
     python3-pip \


thomasyu888 (commented):
I am ok with this, I'll defer to the team if they had any other ideas for adding this dependency

BryanFauble (approved):
Change look good to me. I verified I could still build the docker container and the size of the container goes from 555.78MB -> 565.91MB (Marginal increase)

thomasyu888:
> Change look good to me. I verified I could still build the docker container and the size of the container goes from 555.78MB -> 565.91MB (Marginal increase)

Ah, let's merge then if marginal increase. @BryanFauble , thanks!

PR #1147, opened 2024-12-06
[SYNPY-1548] Swap to a FIFO queue

BryanFauble:
**Problem:**

1. AsyncIO has no order guarantee for the AsyncIO tasks that are created, it can lead to us requesting a auth token for a specific file and not using it until much later, leading it to be expired.
2. Logging and messaging around the download process was a bit all over the place without clearly indicating what message was for what Synapse ID.
3. The TQDM download bar (See: https://sagebionetworks.jira.com/browse/SYNPY-1507) was broken during concurrent file downloading and the sync from synapse logic didn't use a context managed download bar.

**Solution:**

1. Move to a FIFO queue for file downloads that works by indexing the container to be downloaded, and then starting a number of workers that will concurrently download each file. This makes it such that only a smaller number of files can enter the download process rather than everything.
2. Adding in additional retry logic around getting the file size, along with moving the logic to get the authenticated URL to the same thread where the file size is retrieved.
3. Update the messaging for file downloads to include a prefix of [ID:Name]:

image

4. Update the TQDM bar to track concurrent file downloads and only close the transfer bar when there are no active transfers. Also swapping the sync from synapse container logic to use a context managed transfer bar

**Testing:**

1. I have tested/benchmarked (Without significant changes) 1000 files 1 MiB each, and 100 files, 1000 MiB each
2. I re-did benchmarking for 100GB file downloads and verified no changes to benchmarking results
3. @thomasyu888 Also verified around 800GB of file downloads without error with a synapse project


synapseclient/core/transfer_bar.py
```diff
@@ -86,22 +90,28 @@ def shared_download_progress_bar(
...
+    if force_close or not _is_context_managed_download_bar():
         progress_bar: tqdm = getattr(_thread_local, "progress_bar_download", None)
-        if progress_bar is not None:
+        transfer_count: int = getattr(_thread_local, "transfer_count", 0)
+        transfer_count -= 1
+        if transfer_count < 0:
+            transfer_count = 0
+
+        _thread_local.transfer_count = transfer_count
+        if progress_bar is not None and not transfer_count:
```
BryanFauble, line 114: See https://sagebionetworks.jira.com/browse/SYNPY-1507 - There was an issue where using `asyncio.gather` to execute a bunch of download tasks in parallel would cause the progress bar to get closed, and no longer show. By tracking the files being transferred and increment when we open a new bar/decrement when closing a bar we can close the bar when the last transfer occurs. It allows the bar to remain open for the duration of the transfer and maintain it's progress/context throughout.

synapseclient/models/mixins/storable_container.py
```diff
@@ -54,6 +55,42 @@ class StorableContainer(StorableContainerSynchronousProtocol):
...
+            except Exception as ex:
+                result = ex
+
+            self._resolve_sync_from_synapse_result(
+                result=result,
+                failure_strategy=failure_strategy,
+                synapse_client=synapse_client,
+            )
+
+            queue.task_done()
```
BryanFauble, line 92: By taking this worker approach we are limiting the number of concurrent file transfers. Asyncio also provides a semaphore (https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore) - But, this First in, First out queue seemed like the more intuitive approach to solving this problem.

tests/unit/synapseutils/unit_test_synapseutils_sync.py
```diff
@@ -408,8 +408,8 @@ def test_sync_from_synapse_download_file_is_false(syn: Synapse) -> None:
         new_callable=AsyncMock,
         side_effect=[
             mocked_project_rest_api_dict(),
-            mocked_folder_rest_api_dict(),
             mocked_file_rest_api_dict(),
+            mocked_folder_rest_api_dict(),
         ],
```
BryanFauble, line 413: Due to swapping over to using an AsyncIO queue the files are instantly downloaded as they're added to the queue, instead of waiting for the code to execute all of the tasks as Folder tasks are added to the list. It changes the order of operations slightly, and as a result - Broke these unit tests.

synapseclient/core/download/download_async.py
```diff
@@ -232,40 +191,47 @@ def _pre_signed_url_expiration_time(url: str) -> datetime:
...
+        url_provider: A URL provider for the presigned urls
         debug: A boolean to specify if debug mode is on
 
     Returns:
         The size of the file in bytes
     """
-    with syn._requests_session_storage.stream("GET", url) as response:
+    with syn._requests_session_storage.stream(
+        method="GET", url=presigned_url_provider.get_info().url
+    ) as response:
```
BryanFauble, line 234: Moving the logic to here aligns with how the logic works when streaming each individual part of the files. This is executing in a different thread so the presigned URL should be instantly used.

I also wrapped this in a retry block as well to ensure it'll remain functional.

synapseclient/core/download/download_functions.py
```diff
@@ -268,9 +276,13 @@ async def download_file_entity_model(
...
                 os.makedirs(download_location)
             client.logger.info(
-                f"Copying existing file from {cached_file_path} to {download_path}"
+                f"[{file.id}:{file_name}]: Copying existing file from {cached_file_path} to {download_path}"
             )
             shutil.copy(cached_file_path, download_path)
+        else:
+            client.logger.info(
+                f"[{file.id}:{file_name}]: Found existing file at {download_path}, skipping download."
+            )
```
BryanFauble, line 285: In lieu of structured logging setup this is at least a start to get to the point where logging is clearer within the client. I updated a bunch of the messages around the download process so that it's clear what Synapse ID/file/entity produced the message.

BWMac (approved):
LGTM! Awesome update

thomasyu888 (approved):
🔥 Nice use of FIFO queue so that it's not just a completely random ordering of tasks being executed.  Does this slow down download speeds a bit?

PR #1151, opened 2025-01-07
[SYNPY-1551] Tables refactor

BryanFauble:
**NOTE:**

1. A number of changes/fixes are in a follow-up pull request [available here](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1167) <---- **Should probably be reviewed before this pull request**.
2. Until ^ is merged the tests in this PR will fail (Since the fixes are introduced in ^)

**Problem:**

1. Working with tables in the current implementation is a difficult task. Many different functions must be strung together to accomplish some tasks, features like renaming columns is missing.

**Solution:**

Documentation for the Tables API can be found on read the docs [available here](https://synapsepythonclient--1167.org.readthedocs.build/en/1167/reference/experimental/sync/table/), and for the async [interface here](https://synapsepythonclient--1167.org.readthedocs.build/en/1167/reference/experimental/async/table/)

1. Updating the initial Tables OOP model to bring forward a number of changes:

The main functions that exist on the Table to interact with required functionality include:

- store
- store_rows
- get
- delete
- delete_rows
- query
- query with a part mask

Convenience functions (Functions that smooth out setting attributes, but require a call to `.store()` to persist to Synapse)

- delete_column
- add_column
- reorder_column

1. Once the logic for these functions is nailed down the same ideas may be applied to both `EntityView` and `Dataset` concepts in Synapse. They're similar in concept to a Table, but serve different purposes for users. When we start to implement those 2 models we will be able to extract out common logic and re-use it across all 3 models.

**Testing:**

1. Integration tests/unit tests (Some available in [this PR](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1167))
3. The DPE team and I spent time to test out and break these changes across a variety of different testing noted in [this Jira epic](https://sagebionetworks.jira.com/browse/SYNPY-1559). 

TODO:

- [x] Get feedback on proposed changes and smooth out rough edges
- [x] Add additional examples and doc strings to all functions
- [x] Implement logic to support loading results from Synapse directly into a pandas dataframe (Without requiring the usage of an intermediate CSV file written to disk): FUTURE TICKET
- [x] Batch up inserting/updating rows for any number of requested changes
- [x] Implement querying and returning results via a DataFrame
- [x] Add verbose integration tests for all logic
- [x] Add verbose unit tests for all logic
- [x] Swap the submission of data to Synapse to use a single transaction via this api: https://rest-docs.synapse.org/rest/POST/entity/id/table/transaction/async/start.html
- [x] Implement a dry-run feature to both `store` and `store_rows` that prints out the changes to the schema that will be applied (Determine if we could do a diff of the data, if it's feasible to implement - If not, do not implement)
- [x] Factor the `delete_rows` logic, or document the process well. Current behavior requires both a row_id and version_number to delete sourced from query results. As querying for data changed, so will the delete logic.
- [x] Implement storing of rows for each of the formats we are going to let users submit data: Csv, list of lists, dictionary, pandas dataframe
- [x] Implement a `_columns_to_delete` concept within the models to prevent unintentional deletion of columns and require that deleting a column go through the `delete_column` function.


thomasyu888:
My only comment for now: should we have the concept of "snapshot" which just mints a version.

BryanFauble:
> My only comment for now: should we have the concept of "snapshot" which just mints a version.

Good call! I implemented that logic in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1151/commits/cf2c69a6c490f45d565875f67ae5130d3cbf2995

Changes are shown on the doc site here: https://synapsepythonclient--1151.org.readthedocs.build/en/1151/reference/oop/table_refactor/#synapseclient.models.Table.snapshot

I also added some logic to clean the bearer token out of debug log messages (I saw this while looking into this logic).

synapseclient/models/table.py
```diff
@@ -364,10 +408,14 @@ class Table(TableSynchronousProtocol, AccessControllable):
         modified_on: The date this table was last modified.
             In YYYY-MM-DD-Thh:mm:ss.sssZ format
         modified_by: The ID of the user that last modified this table.
-        version_number: The version number issued to this version on the object.
-        version_label: The version label for this table
-        version_comment: The version comment for this table
-        is_latest_version: If this is the latest version of the object.
+        version_number: (Read Only) The version number issued to this version on the
+            object. Use this the `.snapshot()` method to create a new version of the
```
BWMac, line 412:
```suggestion
            object. Use the `.snapshot()` method to create a new version of the
```

synapseclient/models/table.py
```diff
@@ -353,8 +358,47 @@ class Table(TableSynchronousProtocol, AccessControllable):
...
+            and `.delete_column()` methods on your table class instance.
+
+
+            Note that the keys in this dictionary should match the column names as they are in
+            Synapse. However, know that the name attribute of the Column object is used for
+            all interactions with the Synapse API. The OrderedDict key is purely for the usage
+            of this interface. For example, if you wish to rename a column you may do so by
+            changing the name attribute of the Column object. The key in the OrderedDict does
+            not need to be changed. The next time you store the table the column will be updated
+            in Synapse with the new name and the key in the OrderedDict will be updated.
```
BWMac, line 401: Is there a reason we can't update the key when the name of the column is changed in the Column object it is mapped to? Or do you see it as advantageous to still be using the old key name?

It could be a little awkward if a user wants to perform multiple operations on the Table/Column before `store`-ing it again

synapseclient/models/table.py
```diff
@@ -383,6 +431,186 @@ class Table(TableSynchronousProtocol, AccessControllable):
...
+
+            table = Table(
+                name="my_table",
+                parent_id="syn1234",
+                columns=columns
+            )
+
+            table.store()
+
+    Example: Creating a table with an OrderedDict of columns
```
BWMac, line 589: Is there a tangible difference between creation with a `Dict` of columns vs and `OrderedDict`? Will the order of a regular `Dict` also be respected in terms of the order that columns are displayed?

synapseclient/models/table.py
```diff
@@ -397,49 +625,96 @@ class Table(TableSynchronousProtocol, AccessControllable):
...
 
     version_label: Optional[str] = None
-    """The version label for this table"""
+    """(Read Only) The version label for this table. Use this the `.snapshot()` method
+    to create a new version of the table."""
 
     version_comment: Optional[str] = None
-    """The version comment for this table"""
+    """(Read Only) The version comment for this table. Use this the `.snapshot()` method
+    to create a new version of the table."""
```
BWMac, line 707:
```suggestion
    version_number: Optional[int] = field(default=None, compare=False)
    """(Read Only) The version number issued to this version on the object. Use the
    `.snapshot()` method to create a new version of the table."""

    version_label: Optional[str] = None
    """(Read Only) The version label for this table. Use the `.snapshot()` method
    to create a new version of the table."""

    version_comment: Optional[str] = None
    """(Read Only) The version comment for this table. Use the `.snapshot()` method
    to create a new version of the table."""
```

BWMac (commented):
Left some quesions/discussion points. Overall I like the feel of the new interface and I think it's much more straightforward and user-friendly. 

I'm curious how you are thinking about the interface for storing/deleting/alterring rows in this new framework. Will there be an in-memory representation of the contents of the table stored on the `Table` object? What would an operation look like if a user wanted to change the name of a column and alter row content in between `store` calls?

synapseclient/models/table.py
```diff
@@ -321,8 +283,8 @@ async def store_async(
         Returns:
             The Column instance stored in synapse.
         """
-        # TODO - We need to add in some validation before the store to verify we have enough
-        # information to store the data
+        # TODO - Update the storage of columns to go through this API: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/TableSchemaChangeRequest.html
+        # TODO: As of right now since we are not using this API the columns do not maintain the data that is stored in them. This is a limitation of the current implementation.
```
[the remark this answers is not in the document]
BryanFauble: I am planning to finish out any additional implementation as a separate task (This PR will not be merged until everything is ready).

The idea with this PR is that it's proposing many of the interfaces that we will be fully implementing later on, allowing for a bunch of changes to take place based on feedback.

synapseclient/models/table.py
```diff
@@ -341,6 +303,63 @@ async def store_async(
         return self
 
 
+class SchemaStorageStrategy(str, Enum):
```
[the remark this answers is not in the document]
BryanFauble: I wanted to keep the option open in the future to add additional options. If I were to use a boolean flag on the method then every change would be an update to the interface (ie: A new boolean). Using an enum flag like this let's us add options without updating the parameters the methods accept.

synapseclient/models/table.py
BryanFauble: `Dict` should retain order since python 3.7+, but since it didn't before I wouldn't assume that the behavior will remain consistent forever.

OrderedDict is the guaranteed way to provide the order, `Dict` seems like it's a "It will probably respect order" situation

BryanFauble:
@BWMac 

> I'm curious how you are thinking about the interface for storing/deleting/alterring rows in this new framework.

- store_rows (Also used for update): Pass the values to store and insert them into Synapse. In the case of altering rows we would query for the rows to update first, and then add the row_id and row_version columns to the data. We would be doing the same kind of logic as this: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/5a377605c3afcc0143162a93733eade110741a40/synapseclient/table.py#L2219-L2332
- delete_rows: Let's you run a query and delete all matching rows

> Will there be an in-memory representation of the contents of the table stored on the `Table` object? 

I was not planning to store a reference to the content on the `Table` object. The content of the table depends on what you query for. My intention is that any time you want to get data you'd run a fresh query and pass the results around. My concern around storing this on the `Table` object is that if I keep a reference to the result the garbage collector wouldn't clean it up until the reference to the `Table` was removed. I can see cases where you might want to query a table multiple times for different things - get results and act on those results, query again and act on those results.... I'd rather that the data is intentionally passed to what needs it.


I had set `query` to be a staticmethod on the `Table` class. The thing that would prevent me from want to make it a non-staticmethod is that you would still need to provide the table ID in the `query` you execute (ie: `select * from syn123` [examples](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/web/controller/TableExamples.html)). Meaning it'd look like this in code if that were the case:

```
my_table = Table(id="syn1234")
query_results = my_table.query(query="select * from syn1234")
```

We need to duplicate the `syn1234` twice in this case. This is an issue present in my proposed `delete_rows` method too:

```
from synapseclient.models import Table

syn = Synapse()
syn.login()

Table(id="syn1234").delete_rows(query="SELECT ROW_ID, ROW_VERSION FROM syn1234 WHERE foo = 'asdf'")
```

I had considered something like being able to use a pre-defined string within the query that gets replaced with the ID of the table, ie:

```
from synapseclient.models import Table

syn = Synapse()
syn.login()

# Current expectation
query_results = Table.query(query="SELECT ROW_ID, ROW_VERSION FROM syn1234 WHERE foo = 'asdf'")

# 'pre-defined' string idea
query_results = Table(id="syn1234").query(query="SELECT ROW_ID, ROW_VERSION FROM :id WHERE foo = 'asdf'")
```

The problem with this thought is that you couldn't ever take this query and execute it in the Synapse UI, or other client that doesn't contain this exact behavior.


> What would an operation look like if a user wanted to change the name of a column and alter row content in between `store` calls?

The syntax should look something like this:

```
# Query for the rows to update
content_to_update = Table.query(query="select * from syn1234 where foo = 'bar'")

# Update those rows
content_to_update["some_content"] = ["some-new-values"]

# include_columns to rename an existing column
my_table = Table(id="syn1234").get(include_columns=True)
my_table.columns["some_content"].name = "some_new_column_name"

# The `store_rows` function could look for the new column in the pandas dataframe; If present - Use the data from that column, if not present - Look for the previous column name
my_table.store_rows(values=content_to_update)
```

**Question**: Do you think we should expect that the user is updating the name of the column in the pandas dataframe? Should we provide a `rename_column` function that takes in: `old_column_name: str, new_column_name: str, rows: pd.Dataframe = None` that would 1) Update the `.columns` attribute dictionary key, 2) Update the `Column` class instance `.name` attribute, 3) Update the name of the column in the passed in Dataframe?

synapseclient/models/table.py
```diff
@@ -489,64 +837,152 @@ def fill_from_dict(
...
 
-    @otel_trace_method(
-        method_to_trace_name=lambda _, **kwargs: f"Store_rows_by_csv: {kwargs.get('csv_path', None)}"
-    )
-    async def store_rows_from_csv_async(
-        self, csv_path: str, *, synapse_client: Optional[Synapse] = None
-    ) -> str:
-        """Takes in a path to a CSV and stores the rows to Synapse.
+    # TODO: Finish implementation
+    async def store_rows_async(
```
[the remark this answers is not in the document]
BryanFauble: Great question. I ended up putting this into a new method for that specific logic since partial updates are quite a bit different for how it needs to be implemented. It's in a `upsert_rows` method detailed here:

https://synapsepythonclient--1151.org.readthedocs.build/en/1151/reference/oop/table_refactor/#synapseclient.models.Table.upsert_rows

I also got an implementation working (At least the partial update portion) to test out how Synapse does it.

synapseclient/models/__init__.py
```diff
@@ -13,11 +13,11 @@
 from synapseclient.models.services import FailureStrategy
 from synapseclient.models.table import (
     Column,
+    ColumnExpansionStrategy,
     ColumnType,
-    CsvResultFormat,
     FacetType,
-    Row,
     RowsetResultFormat,
```
thomasyu888, line 19: ```ImportError: cannot import name 'RowsetResultFormat' from 'synapseclient.models.table' (/Users/tyu/sage/synapsePythonClient/synapseclient/models/table.py)```

synapseclient/models/table.py
```diff
@@ -383,6 +487,186 @@ class Table(TableSynchronousProtocol, AccessControllable):
...
+            )
+
+            table = Table(
+                name="my_table",
+                parent_id="syn1234",
+            )
+
+            # The call to `store_rows` will also call `.store()` on the table too,
+            # meaning if the table does not exist it will be created.
+            table.store_rows(values=my_data, schema_storage_strategy=SchemaStorageStrategy.INFER_FROM_DATA)
```
thomasyu888, line 532: There is an issue with this function.

```
SynapseHTTPError                          Traceback (most recent call last)
Cell In[2], line 1
----> 1 table.store_rows(values=my_data, schema_storage_strategy=SchemaStorageStrategy.INFER_FROM_DATA)

File ~/sage/synapsePythonClient/synapseclient/core/async_utils.py:72, in ClassOrInstance.__get__.<locals>.f(*args, **kwds)
     70 def f(*args, **kwds):
     71     if obj is not None:
---> 72         return self.fn(obj, *args, **kwds)
     73     else:
     74         return self.fn(cls, *args, **kwds)

File ~/sage/synapsePythonClient/synapseclient/core/async_utils.py:146, in async_to_sync.<locals>.create_method.<locals>.newmethod(self, *args, **kwargs)
    140 synapse_client = Synapse.get_client(
    141     getattr(kwargs, "synapse_client", None)
    142 )
    143 synapse_client.logger.exception(
    144     f"Error occurred while running {async_method_name} on {self.__class__}."
    145 )
--> 146 raise ex

File ~/sage/synapsePythonClient/synapseclient/core/async_utils.py:135, in async_to_sync.<locals>.create_method.<locals>.newmethod(self, *args, **kwargs)
    133         return loop.run_until_complete(wrapper(*args, **kwargs))
    134     else:
--> 135         return asyncio.run(wrapper(*args, **kwargs))
    137 except Exception as ex:
    138     from synapseclient import Synapse

File ~/anaconda3/envs/synapse/lib/python3.10/asyncio/runners.py:44, in run(main, debug)
     42     if debug is not None:
     43         loop.set_debug(debug)
---> 44     return loop.run_until_complete(main)
     45 finally:
     46     try:

File ~/anaconda3/envs/synapse/lib/python3.10/asyncio/base_events.py:649, in BaseEventLoop.run_until_complete(self, future)
    646 if not future.done():
    647     raise RuntimeError('Event loop stopped before Future completed.')
--> 649 return future.result()

File ~/sage/synapsePythonClient/synapseclient/core/async_utils.py:122, in async_to_sync.<locals>.create_method.<locals>.newmethod.<locals>.wrapper(*args, **kwargs)
    120 async def wrapper(*args, **kwargs):
    121     """Wrapper for the function to be called in an async context."""
--> 122     return await getattr(self, async_method_name)(*args, **kwargs)

File ~/sage/synapsePythonClient/synapseclient/models/table.py:1625, in Table.store_rows_async(self, values, schema_storage_strategy, column_expansion_strategy, dry_run, synapse_client, **kwargs)
   1621     changes.append(upload_request)
   1622     transaction_request = TableUpdateTransactionRequest(
   1623         entity_id=self.id, changes=changes
   1624     )
-> 1625     client._waitForAsync(
   1626         uri=uri, request=transaction_request.to_synapse_request()
   1627     )
   1628 else:
   1629     raise ValueError(
   1630         "Don't know how to make tables from values of type %s." % type(values)
   1631     )

File ~/anaconda3/envs/synapse/lib/python3.10/contextlib.py:79, in ContextDecorator.__call__.<locals>.inner(*args, **kwds)
     76 @wraps(func)
     77 def inner(*args, **kwds):
     78     with self._recreate_cm():
---> 79         return func(*args, **kwds)

File ~/sage/synapsePythonClient/synapseclient/client.py:5041, in Synapse._waitForAsync(self, uri, request, endpoint)
   5039 lastMessage, lastProgress, lastTotal, progressed = "", 0, 1, False
   5040 while time.time() - start_time < self.table_query_timeout:
-> 5041     result = self.restGET(
   5042         uri + "/get/%s" % async_job_id["token"], endpoint=endpoint
   5043     )
   5044     if result.get("jobState", None) == "PROCESSING":
   5045         progressed = True

File ~/sage/synapsePythonClient/synapseclient/client.py:6127, in Synapse.restGET(self, uri, endpoint, headers, retryPolicy, requests_session, **kwargs)
   6105 def restGET(
   6106     self,
   6107     uri,
   (...)
   6112     **kwargs,
   6113 ):
   6114     """
   6115     Sends an HTTP GET request to the Synapse server.
   6116 
   (...)
   6125         JSON encoding of response
   6126     """
-> 6127     response = self._rest_call(
   6128         "get", uri, None, endpoint, headers, retryPolicy, requests_session, **kwargs
   6129     )
   6130     return self._return_rest_body(response)

File ~/sage/synapsePythonClient/synapseclient/client.py:6102, in Synapse._rest_call(self, method, uri, data, endpoint, headers, retryPolicy, requests_session, **kwargs)
   6090 requests_method_fn = getattr(requests_session, method)
   6091 response = with_retry(
   6092     lambda: requests_method_fn(
   6093         uri,
   (...)
   6100     **retryPolicy,
   6101 )
-> 6102 self._handle_synapse_http_error(response)
   6103 return response

File ~/sage/synapsePythonClient/synapseclient/client.py:6016, in Synapse._handle_synapse_http_error(self, response)
   6013 """Raise errors as appropriate for returned Synapse http status codes"""
   6015 try:
-> 6016     exceptions._raise_for_status(response, verbose=self.debug)
   6017 except exceptions.SynapseHTTPError as ex:
   6018     # if we get a unauthenticated or forbidden error and the user is not logged in
   6019     # then we raise it as an authentication error.
   6020     # we can't know for certain that logging in to their particular account will grant them
   6021     # access to this resource but more than likely it's the cause of this error.
   6022     if response.status_code in (401, 403) and not self.credentials:

File ~/sage/synapsePythonClient/synapseclient/core/exceptions.py:194, in _raise_for_status(response, verbose)
    191     except:  # noqa
    192         message += f"\n{UNABLE_TO_APPEND_RESPONSE}"
--> 194 raise SynapseHTTPError(message, response=response)

SynapseHTTPError: 400 Client Error: 
Invalid Entity ID: None

```

synapseclient/models/table.py
```diff
@@ -489,226 +909,2209 @@ def fill_from_dict(
...
+            changes = []
+            if schema_change_request:
+                changes.append(schema_change_request)
+            changes.append(upload_request)
+            transaction_request = TableUpdateTransactionRequest(
+                entity_id=self.id, changes=changes
+            )
+            client._waitForAsync(
+                uri=uri, request=transaction_request.to_synapse_request()
+            )
```
thomasyu888, line 1627: Regarding the 

```
SynapseHTTPError: 400 Client Error: 
Invalid Entity ID: None
```
bug, I'm guessing that it's something about `self.id not being totally set.

synapseclient/models/table.py
```diff
@@ -489,226 +909,2209 @@ def fill_from_dict(
...
+
+                from synapseclient import Synapse
+                from synapseclient.models import Table
+
+                syn = Synapse()
+                syn.login()
+
+                table = Table(
+                    id="syn1234"
+                ).get(include_columns=True)
```
thomasyu888, line 2110: I wonder if this would be confusing for our users for `include_columns` default behavior to be false.

synapseclient/models/table.py
```diff
@@ -489,226 +909,2209 @@ def fill_from_dict(
...
+        result = {
+            "entity": entity,
+        }
+        delete_none_keys(result)
+        return result
+
+    async def upsert_rows_async(
+        self,
+        values: pd.DataFrame,
+        upsert_columns: List[str],
```
thomasyu888, line 951: discussion: I wonder if renaming this to "primary_keys" would be more synonymous with database language instead of introducing our own language.

synapseclient/models/__init__.py
BryanFauble: Try this again. I missed committing this change from my local

synapseclient/models/table.py
BryanFauble: Can you share the script that you got this error with? I had intended that the Table would be created prior to calling `store_rows_async`, so I am wondering if you called it on a table that hadn't been stored to Synapse yet. Otherwise, this portion should have found the ID:

```
        if (
            (not self._last_persistent_instance)
            and (
                existing_id := await get_id(
                    entity=self, synapse_client=synapse_client, failure_strategy=None
                )
            )
            and (
                existing_entity := await Table(id=existing_id).get_async(
                    include_columns=True, synapse_client=synapse_client
                )
            )
        ):
            merge_dataclass_entities(
                source=existing_entity,
                destination=self,
            )
```

I can look to raise an exception if you're trying to call the function without a table that's already stored.

synapseclient/models/table.py
BryanFauble: I agree that it'll make sense to leave this as `True` by default

synapseclient/models/table.py
BryanFauble: I agree with this renaming

synapseclient/models/table.py
thomasyu888: It was this example:

```
import pandas as pd

from synapseclient.models import Table, SchemaStorageStrategy

syn = Synapse()
syn.login()

my_data = pd.DataFrame(
    {
        "my_string_column": ["a", "b", "c", "d"],
        "my_integer_column": [1, 2, 3, 4],
        "my_double_column": [1.0, 2.0, 3.0, 4.0],
        "my_boolean_column": [True, False, True, False],
    }
)

table = Table(
    name="my_table",
    parent_id="syn1234",
).store()

table.store_rows(values=my_data, schema_storage_strategy=SchemaStorageStrategy.INFER_FROM_DATA)

# Prints out the stored data about this specific column
print(table.columns["my_string_column"])
```

thomasyu888: oh wait.. I might have not called the `.store()` function on the table...

thomasyu888: Ah yes, having `.store()` first on the table object worked

BryanFauble: Sounds good. I updated the example code to include that:
```
            import pandas as pd

            from synapseclient.models import Table, SchemaStorageStrategy

            syn = Synapse()
            syn.login()

            my_data = pd.DataFrame(
                {
                    "my_string_column": ["a", "b", "c", "d"],
                    "my_integer_column": [1, 2, 3, 4],
                    "my_double_column": [1.0, 2.0, 3.0, 4.0],
                    "my_boolean_column": [True, False, True, False],
                }
            )

            table = Table(
                name="my_table",
                parent_id="syn1234",
            ).store() # <---- This was updated

            table.store_rows(values=my_data, schema_storage_strategy=SchemaStorageStrategy.INFER_FROM_DATA)

            # Prints out the stored data about this specific column
            print(table.columns["my_string_column"])
```

thomasyu888 (commented):
A high level note before I forget:

What happens currently when you store a string that is greater than the string length set on the Table schema?

BryanFauble:
> A high level note before I forget:
> 
> What happens currently when you store a string that is greater than the string length set on the Table schema?

@thomasyu888 


# For Tables:

An exception is raised from the Synapse API:

```
synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error: Value at [0,3] was not a valid STRING. String 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' exceeds the maximum length of 50 characters.
```

When this occurs the transaction will be rolled-back. This includes any schema changes, other updates, and other inserts you might have been trying to accomplish in the same call. Example code:

```
my_table = Table(id="syn53064523").get()

query_results = my_table.query(query="SELECT * FROM syn53064523 WHERE my_string_column = 'REBHB'")

query_results.loc[query_results['my_string_column'] == 'REBHB',
                  'another_string_column'] = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"

query_results['new_column'] = ["new_value"]

my_table.store_rows(values=query_results, schema_storage_strategy=SchemaStorageStrategy.INFER_FROM_DATA)
```

# For Datasets:

Datasets work a little differently when you define columns for them. The columns that are present are Annotations that will get placed onto the File itself. Because of this, setting the column to a value that is not valid for the Schema does not error out. Instead, it will accept the transaction, but fail to load the table when you access it in the UI:

image

synapseclient/api/entity_services.py
```diff
@@ -215,6 +217,80 @@ async def delete_entity_generated_by(
     )
 
 
+async def delete_entity(
+    entity_id: str,
+    version_number: int = None,
```
andrewelamb, line 222: When testing this if you attempt to delete a version of an entity that doesn't exist, nothing happens, but there isn't a warning or error. Is that expected?

synapseclient/api/table_services.py
```diff
@@ -0,0 +1,71 @@
...
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+    from synapseclient.models import Column
+
+
+async def get_columns(
+    table_id: str,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> List["Column"]:
```
andrewelamb, line 18: Why is Column in quotes here?

synapseclient/api/table_services.py
```diff
@@ -0,0 +1,71 @@
...
+
+import json
+from typing import TYPE_CHECKING, List, Optional
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+    from synapseclient.models import Column
+
+
+async def get_columns(
```
andrewelamb, line 14: I tried giving this function the synapse ID for a project it returned an empty list. Perhaps if the entity is not a table this should be an exception?

synapseclient/api/table_services.py
BryanFauble: When the typing is in quotes it means that it was imported behind this check:

```

if TYPE_CHECKING:
    from synapseclient.models import Column
```

You have to use quotes because it's only for static type checking. It's Pythons mechanism to allow for typing to work while preventing circular imports from occurring.

synapseclient/api/table_services.py
BryanFauble: I don't think that the check should exist in this code. If anything it should be a check on the Synapse Rest API.

1) The issue with adding the check in here is that we would need to perform a separate HTTP request in order to verify what the Synapse ID is that you passed along. This adds extra overheard every time someone would like to call this specific function, since it lacks any other context except for what is passed into it.
2) These functions in the `/api/*` are meant for power users wanting to use the API's directly, and us to be able to build the client functionality. We can think of this as a hierarchy like:

image


Thoughts?

synapseclient/api/entity_services.py
BryanFauble: This is consistent behavior with the current functionality:
```
import asyncio
from synapseclient.api import delete_entity

syn = Synapse()
syn.login()


async def main():
    res = await delete_entity(entity_id="syn53604544", version_number=99)
    print(res)

    res2 = syn.delete(obj="syn53604544", version=99)
    print(res2)

asyncio.run(main())
```

synapseclient/models/mixins/table_operator.py
```diff
@@ -0,0 +1,2836 @@
...
+
+            syn = Synapse()
+            syn.login()
+
+            table = Table(
+                id="syn1234"
+            ).get(include_columns=True)
+
+            table.add_column(
+                Column(name="my_column", column_type=ColumnType.STRING)
```
andrewelamb, line 680: It looks like the `name` parameter might have gotten changed to `column` at some point?

synapseclient/models/dataset.py
```diff
@@ -0,0 +1,417 @@
+import dataclasses
+from collections import OrderedDict
+from copy import deepcopy
+from dataclasses import dataclass, field
+from datetime import date, datetime
+from typing import Any, Dict, List, Optional, Union
+
```
BryanFauble, line 7: Brad will be finishing out the Dataset in a follow-up pull request. It's in an incomplete state at the moment and is only partially functional.

synapseclient/models/mixins/table_operator.py
```diff
@@ -0,0 +1,2836 @@
...
+        elif isinstance(values, DataFrame) or isinstance(values, str):
+            values = values.copy()
+        else:
+            raise ValueError(
+                "Don't know how to make tables from values of type %s." % type(values)
+            )
+
+        all_columns_from_df = [f'"{column}"' for column in values.columns]
+
+        # TODO: Chunk this up into smaller queries. Running this when upserting 1 million rows causes an HTTP 403 to be raised from Synapse
```
BryanFauble, line 1903: Resolved in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1167

synapseclient/api/table_services.py
andrewelamb: That's an excellent point. Perhaps we should ask the platform team for this check? In Schematic several endpoints ask for two synapse entities, usually a fileview and something else. A common user error is two switch those, and the resulting error message has not been useful.

BryanFauble: A better check may be to enforce in this code:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/93c450bbc1d228b978b9bfa701ab64983555d798/synapseclient/api/entity_factory.py#L352-L365

This is a "generic" factory function that gets an entity from Synapse and pushes it into the relevant dataclass in Python (Project, Table, File, ect...). If the `.get()` type of object that you're calling from doesn't match, we could raise an exception here.

ie:

```
my_table = Table(id="ID of a project").get() # Raise an exception
my_table = Table(id="ID of a table").get() # Proceed as expected
```

BryanFauble: I created https://sagebionetworks.jira.com/browse/SYNPY-1572 for this

thomasyu888 (approved):
🔥 LGTM!  Awesome work here

synapseclient/models/mixins/table_operator.py
```diff
@@ -0,0 +1,2836 @@
...
+                if column.name:
+                    if column.name in results:
+                        raise ValueError(f"Duplicate column name: {column.name}")
+                    results[column.name] = column
+                else:
+                    column.name = key
+                    if key in results:
+                        raise ValueError(f"Duplicate column name: {key}")
+                    results[key] = column
+            return results
```
BWMac, line 970: Could these be combined into one block that checks for either type of dictionary?

BWMac (commented):
Looks good. A few minor comments/questions. Just waiting on unit tests for sync Table functions.

BWMac (approved):
Pre-approving. Awesome work!

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -5,12 +5,18 @@
...
-from synapseclient.core.constants.concrete_types import AGENT_CHAT_REQUEST
+from synapseclient.core.constants.concrete_types import (
+    AGENT_CHAT_REQUEST,
+    TABLE_UPDATE_TRANSACTION_REQUEST,
+)
 from synapseclient.core.exceptions import SynapseError, SynapseTimeoutError
 
 ASYNC_JOB_URIS = {
     AGENT_CHAT_REQUEST: "/agent/chat/async",
+    TABLE_UPDATE_TRANSACTION_REQUEST: "/entity/{entityId}/table/transaction/async",
```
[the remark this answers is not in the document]
BryanFauble: I wasn't planning to accomplish this via an f-string, this is actually using the [string.format](https://www.w3schools.com/python/ref_string_format.asp) functionality where:

> The placeholder is defined using curly brackets: {}.


The reason for this is because of how it is being used. In the `send_job_async` we look up the `ASYNC_JOB_URIS` based off the request type:

```
request_type = request.get("concreteType")

uri = ASYNC_JOB_URIS[request_type]
```

I then go on to do this check, and the `uri.format` call:

```
    if "{entityId}" in uri:
        if "entityId" not in request:
            raise ValueError(f"Attempting to send job with missing id in uri: {uri}")
        uri = uri.format(entityId=request["entityId"])
```

I'm not sure there is an easy way to accomplish this with an f string because of the dictionary lookup we perform based on `request_type`

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -320,8 +326,14 @@ async def send_job_async(
         raise ValueError(f"Unsupported request type: {request_type}")
 
     client = Synapse.get_client(synapse_client=synapse_client)
+    uri = ASYNC_JOB_URIS[request_type]
+    if "{entityId}" in uri:
```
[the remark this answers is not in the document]
BryanFauble: See https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1151#discussion_r1974360881

This is where we can add in additional checks for URLs where we need to replace content in them. It's also checking for the literal string `{entityId}` including the brackets (The format to use the `.format()` function on strings)

synapseclient/models/mixins/asynchronous_job.py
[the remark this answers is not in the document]
BWMac: OK sounds good to me!

synapseclient/models/mixins/table_operator.py
```diff
@@ -0,0 +1,2836 @@
...
+    to perform a number of operations on the entity such as getting, deleting, or
+    updating columns, querying for data, and more."""
+
+    id: None = None
+    name: None = None
+    parent_id: None = None
+    activity: None = None
+    version_number: None = None
+    _last_persistent_instance: None = None
+    _columns_to_delete: Dict = None
```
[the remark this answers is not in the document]
BryanFauble: Changes made in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1168

synapseclient/models/mixins/table_operator.py
BryanFauble: Changes made in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1168

PR #1152, opened 2025-01-08
[SYNPY-1544] Synapse Agent OOP Model

BWMac:
**Description:**

This PR is the implementation of an OOP model for interacting with the new Synapse Agent API [endpoints](https://rest-docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.AgentController). The design for the model implementation can be found in [Confluence](https://sagebionetworks.jira.com/wiki/spaces/DPE/pages/3833954306/SYNPY-1544+Synapse+Agent+OOP+Model+Mini-Design).

The core implementation includes (from "bottom to top"):
- `synapseclient.api.agent_services`:
     - Low-level functions for interacting with the API endpoints that are specific to using Synapse Agents.
-  `synapseclient.models.AgentPrompt`:
     - A dataclass representing a prompt sent to a Synapse Agent within a chat session.
     - Holds prompt, response and all necessary metadata for tracking an individual prompt.
     - Only has an async interface for its primary business logic - this class is used by the higher-level `AgentSession` class and is not intended to be used directly.
     - The concept of a "prompt" in the Synapse API is built on the more general concept of an "asynchronous job" which is also used for operations such as large data transfers and the minting of DOIs.
          - https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/asynch/AsynchronousResponseBody.html
          - https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/asynch/AsynchronousRequestBody.html
          - https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/asynch/AsynchronousJobStatus.html
          - Previously, we did not have a generalized solution for creating OOP models on top of the asynchronous job functionality, and so we (@BryanFauble helped out quite a bit here) implemented the `AsynchronousCommunicator` Mixin which includes reusable logic for starting an async job, waiting for it to complete, and returning the response.
- `synapseclient.models.AgentSession`:
     - a dataclass representing a chat session with a Synapse Agent.
     - Includes functionality for directly interacting with chat sessions with a Synapse Agent.
     - The `AgentSession` logic is used in the `Agent` class, and we will generally recommend that users utilize the `Agent` class as the primary interface.
     - Asynchronous and Synchronous interfaces are available.
- `synapseclient.models.Agent`:
     - a dataclass representing a Synapse Agent you can interact with. 
     - Includes functionality for registering a custom agent, starting sessions with agents, sending prompts and also managing more than one session at a time.
     - The basic use case includes one session at a time where a session started by a user (either explicitly with `start_session` or implicitly by just calling `prompt`) becomes the default for prompting, but examples for conducting more than one session at a time can be found in the POC script and class docstring.

**Notes:**
- All needed unit and integration tests for synchronous and asynchronous implementations  are included.
- Example usage of the models can be found in the class docstrings and in the POC script included in the PR.
- I took the opportunity to refactor the `Experimental` section of the documentation, breaking up each model into it's own markdown script which makes maintaining the docs for OOP models a little easier.


synapseclient/api/agent_services.py
```diff
@@ -0,0 +1,259 @@
...
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/AgentRegistration.html>
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Request matching <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/AgentRegistrationRequest.html>
+    request = {
+        "awsAgentId": cloud_agent_id,
+        "awsAliasId": cloud_alias_id if cloud_alias_id else None,
```
BryanFauble, line 42: Will this request work if you pass in None? My assumption is that the key in the http body wouldn't be sent if there is no value

synapseclient/api/agent_services.py
```diff
@@ -0,0 +1,259 @@
...
+        agent_registration_id: The ID of the agent registration to start the session for.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Request matching <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/CreateAgentSessionRequest.html>
```
BryanFauble, line 90: Might make sense to put this in the docstring so it can be referenced  to understand what access_level is supposed to be

synapseclient/api/agent_services.py
```diff
@@ -0,0 +1,259 @@
...
+        The response matching
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/asynch/AsyncJobId.html>
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Request matching <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/AgentChatRequest.html>
+    request = {
+        "concreteType": "org.sagebionetworks.repo.model.agent.AgentChatRequest",
```
BryanFauble, line 177: There is a constant file to throw this in to:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/constants/concrete_types.py

synapseclient/api/agent_services.py
```diff
@@ -0,0 +1,259 @@
...
+    Raises:
+        TimeoutError: If the response is not ready after 1 minute
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    start_time = asyncio.get_event_loop().time()
+    TIMEOUT = 60
+
+    while True:
```
BryanFauble, line 215: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8bd5e195c389e3a83bd1308be3c6a42f25d87b8f/synapseclient/client.py#L5029

This function exists on the client, but it not built with asyncio in mind. There is some additional error handling it does for the failed state. I could see an asyncio compliant version being written and used here.

Maybe it could go in this file? https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8bd5e195c389e3a83bd1308be3c6a42f25d87b8f/synapseclient/core/async_utils.py

synapseclient/api/agent_services.py
```diff
@@ -0,0 +1,259 @@
...
+    prompt_id: str,
+    newer_than: Optional[int] = None,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """
+    Gets the trace of a prompt.
+
+    Arguments:
+        prompt_id: The token of the prompt to get the trace for.
+        newer_than: The timestamp to get trace results newer than. Defaults to None (all results).
```
BryanFauble, line 239: Link to the rest docs that define what format the timestamp should be in

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+    get_trace,
+    register_agent,
+    send_prompt,
+    start_session,
+    update_session,
+)
+from synapseclient.core.async_utils import otel_trace_method
+
+
+class AgentType(Enum):
```
BryanFauble, line 20: Usually I've have been using (str, Enum) types for these. Mostly so they emulate string like behavior and print to console nicely. In python 3.11 they introduced a StrEnum for this. Small ai code difference between the 2:

```python
from enum import Enum

class ColumnType(str, Enum):
    INTEGER = "integer"
    FLOAT = "float"

# Behaves like a string
print(ColumnType.INTEGER == "integer")  # True
print(len(ColumnType.INTEGER))         # 7
print(ColumnType.INTEGER.upper())      # INTEGER

# JSON serialization
import json
print(json.dumps({"type": ColumnType.INTEGER}))  # {"type": "integer"}
```

Vs

```python
from enum import Enum

class ColumnType(Enum):
    INTEGER = "integer"
    FLOAT = "float"

# Comparison needs `.value`
print(ColumnType.INTEGER == "integer")  # False
print(ColumnType.INTEGER.value == "integer")  # True

# JSON serialization needs customization
import json
print(json.dumps({"type": ColumnType.INTEGER.value}))  # {"type": "integer"}
```

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+    """
+    Enum representing the type of agent as defined in
+        <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/AgentType.html>
+    'BASELINE' is a default agent provided by Synapse.
```
BryanFauble, line 24: Nit. Can be a list. Syntax is a empty newline before and each entry starts with `-`.

Might need to make sure it renders properly on my docs.

Related to ^
These will be "Less" experimental APIs, so we can add a new non-experimental section to https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8bd5e195c389e3a83bd1308be3c6a42f25d87b8f/mkdocs.yml#L55 for this. Not sure on the title, maybe "AI Agent" or something?

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+        id: The unique ID of the agent session. Can only be used by the user that created it.
+        access_level: The access level of the agent session.
+            One of PUBLICLY_ACCESSIBLE, READ_YOUR_PRIVATE_DATA, or WRITE_YOUR_PRIVATE_DATA.
+        started_on: The date the agent session was started.
+        started_by: The ID of the user who started the agent session.
+        modified_on: The date the agent session was last modified.
+        agent_registration_id: The registration ID of the agent that will be used for this session.
+        etag: The etag of the agent session.
+    """
+
```
BryanFauble, line 81: Check out the tables refactor PR for examples of the syntax for an Examples section. I'd like for the docstring to have scripts (Copy/pastable) that folks could run and get started.

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+        self.started_by = synapse_agent_session.get("startedBy", None)
+        self.modified_on = synapse_agent_session.get("modifiedOn", None)
+        self.agent_registration_id = synapse_agent_session.get(
+            "agentRegistrationId", None
+        )
+        self.etag = synapse_agent_session.get("etag", None)
+        return self
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Start_Session: {self.id}"
```
BryanFauble, line 133: Will id be filled in at this point in time before a session has started?

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+    ) -> "AgentSession":
+        """Gets an agent session.
+
+        Arguments:
+            synapse_client: The Synapse client to use for the request. If None, the default client will be used.
+
+        Returns:
+            The retrieved AgentSession object.
+        """
+        syn = Synapse.get_client(synapse_client=synapse_client)
```
BryanFauble, line 168: If you don't need to use the instance of the client you may pass the `synapse_client` along to `get_session`

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+
+        Arguments:
+            synapse_client: The Synapse client to use for the request. If None, the default client will be used.
+
+        Returns:
+            The updated AgentSession object.
+        """
+        syn = Synapse.get_client(synapse_client=synapse_client)
+
+        self.access_level = access_level
```
BryanFauble, line 194: My expectation was that folks would be updating attributes on the instance and then calling the store/update functions, rather than passing the updated values into those functions.

My idea is that I wanted any arguments you pass into the function to determine how it would execute, rather than the data it would be acting on.

In short the idea being:
```
agent_session.access_level = new_value
agent_session.update()
```

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+            access_level=self.access_level.value,
+            synapse_client=syn,
+        )
+        return self.fill_from_dict(session_response)
+
+    @otel_trace_method(method_to_trace_name=lambda self, **kwargs: f"Prompt: {self.id}")
+    async def prompt_async(
+        self,
+        *,
+        prompt: str,
```
BryanFauble, line 206: `*` forces things to be kwargs. It is acceptable that you may want that, but i wanted to call it out if that was not the intention.

I always want synapse_client to be a kwarg so it will never end up after a positional argument

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+        method_to_trace_name=lambda self, **kwargs: f"Get_Agent_Session: {self.registration_id}"
+    )
+    async def get_session_async(
+        self, *, session_id: str, synapse_client: Optional[Synapse] = None
+    ) -> "AgentSession":
+        syn = Synapse.get_client(synapse_client=synapse_client)
+        session = await AgentSession(id=session_id).get_async(synapse_client=syn)
+        if session.id not in self.sessions:
+            self.sessions[session.id] = session
+        self.current_session = session.id
```
BryanFauble, line 394: Rather than storing a string of the id you could just store the session object itself. Both places (list of sessions) and current_session should just have pointers to the same object instance since they are not deep copies of one another.

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+        return session
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Prompt_Agent_Session: {self.registration_id}"
+    )
+    async def prompt(
+        self,
+        *,
+        session_id: Optional[str] = None,
```
BryanFauble, line 403: Id think you would rather pass in a AgentSession object here

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+        prompt: str,
```
BryanFauble, line 404: Required args cannot show up after optional args. It should be fine to put this as the first argument

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+            synapse_client=syn,
+        )
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Get_Agent_Session_Chat_History: {self.registration_id}"
+    )
+    def get_chat_history(self) -> List[AgentPrompt]:
+        """Gets the chat history for the current session."""
+        # TODO: Is this the best way to do this?
+        return self.sessions[self.current_session].chat_history
```
BryanFauble, line 447: With the suggestion above this could swap to

`self.current_session.chat_history if self.current_session else None`

BryanFauble: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1152/files#r1907958583

synapseclient/api/agent_services.py
BWMac: It will work because the `awsAliasId` is optional in the [API endpoint](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/AgentRegistrationRequest.html) itself, with a set default value. I guess there isn't really a reason to not just pass that default value here though.

synapseclient/models/agent.py
BWMac: Are you sure this should not be considered experimental? The API endpoints are brand new and could potentially change/be added to which may lead to updates in the OOP interface

synapseclient/models/agent.py
BWMac: Good catch, so then should there just not be an identifier in that f-string?

synapseclient/models/agent.py
BWMac: Ah right, I'm doubling up on those `get_client` calls in the lower-level functions too

synapseclient/models/agent.py
```diff
@@ -0,0 +1,447 @@
...
+        """
+        syn = Synapse.get_client(synapse_client=synapse_client)
+        session_response = await start_session(
+            access_level=self.access_level.value,
+            agent_registration_id=self.agent_registration_id,
+            synapse_client=syn,
+        )
+        return self.fill_from_dict(session_response)
+
+    @otel_trace_method(
```
thomasyu888, line 154: A bit of a tangent, but could we leverage signoz cloud during python client development?

synapseclient/models/agent.py
BWMac: I think using `update` here is fine because it's the terminology used in the [API docs](https://rest-docs.synapse.org/rest/PUT/agent/session/sessionId.html). Plus an Agent session isn't a Synapse Entity so conceptually it's different to interact with.

synapseclient/models/agent.py
BryanFauble: Yes

synapseclient/models/agent.py
BryanFauble: You could not put an identifier in.

You could also add and attribute to the captured span if you wanted to be able to search for the information later on

synapseclient/models/agent.py
BryanFauble: Good point. Throw it under there as a new section

synapseclient/api/agent_services.py
BryanFauble: It's optional in the API, but i wasn't certain if it would send the literal string "None" here, or if it drops the field from the body, or it leaves it as blank when sending the request

BWMac: Oh I see. It leaves it blank as far as I can tell

BryanFauble: Double check the behavior here. Blank does have a different meaning than the absence of the key (null).

It might not affect this api but i know I other areas of the code none keys are getting popped out of the body before the request gets sent

BWMac: Would it make more sense to just pass the API default as the default here too and not have to worry about it?

BryanFauble: If we send the default in the python client, and the default changes server side - Anyone staying on that version of the client will continue to send in (old) defaults. I would rather we don't send the key at all and only send it if it's a non-None value.

That way if the default changes (server side) the client side code has no changes.

synapseclient/api/agent_services.py
BWMac: `_waitForAync` does both the POST call (starting the job) and the GET call (retrieving the result). Should the functionality I have written here work that way too? I was viewing it as 1 function per API call in agent_services originally.

synapseclient/models/agent.py
BWMac: Would it be its own section, or just fall under this [one](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8bd5e195c389e3a83bd1308be3c6a42f25d87b8f/mkdocs.yml#L78C11-L78C62)?

synapseclient/api/agent_services.py
BryanFauble: They can be 2 separate functions that handle each part. In the Agent script you'll chain them together

synapseclient/models/agent.py
BryanFauble: Create a new entry under `- Experimental:` for this

synapseclient/client.py
```diff
@@ -6373,20 +6373,17 @@ async def rest_get_async(
...
-                endpoint,
-                headers,
-                retry_policy,
-                requests_session_async_synapse,
-                **kwargs,
-            )
-            return self._return_rest_body(response)
-        except Exception:
-            self.logger.exception("Error in rest_get_async")
+        response = await self._rest_call_async(
```
BWMac, line 6376: This error handling was preventing Synapse API errors from being surfaced to client users. All tests passed after making this change.

synapseclient/models/services/storable_entity_components.py
```diff
@@ -4,10 +4,9 @@
 
 from synapseclient import Synapse
 from synapseclient.core.exceptions import SynapseError
-from synapseclient.models import Annotations
 
 if TYPE_CHECKING:
```
BWMac, line 8: This was causing a circular import error

mkdocs.yml
```diff
@@ -75,8 +75,29 @@ nav:
       - Core: reference/core.md
       - REST Apis: reference/rest_apis.md
       - Experimental:
-          - Object-Orientated Models: reference/oop/models.md
-          - Async Object-Orientated Models: reference/oop/models_async.md
+          - Mixins:
```
thomasyu888, line 78: This is really nice - thanks for this on the fly refactor

synapseclient/api/agent_services.py
```diff
@@ -0,0 +1,182 @@
...
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(uri=f"/agent/registration/{registration_id}")
+
+
+async def start_session(
+    access_level: str,
+    agent_registration_id: str,
+    synapse_client: Optional["Synapse"] = None,
```
BryanFauble, line 71: I had been forcing `synapse_client` to be passed in by a kwargument. It would allow us to add new arguments to functions and force folks to re-order arguments:

```
async def start_session(
    access_level: str,
    agent_registration_id: str,
    *,
    synapse_client: Optional["Synapse"] = None,
```

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+
+        from synapseclient import Synapse
+        from synapseclient.models.agent import AgentSession, AgentSessionAccessLevel
+
+        syn = Synapse()
+        syn.login()
+
+        my_session = AgentSession(agent_registration_id=my_agent_registration_id).start(
+            synapse_client=syn
+        )
```
BryanFauble, line 159: Since `synapse_client` is optional (And only required for advanced use cases where you might be using it in a service like Schematic), I had been leaving it out of examples for the most part.

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+
+    Example: Start a session and send a prompt.
+
+        from synapseclient import Synapse
+        from synapseclient.models.agent import AgentSession, AgentSessionAccessLevel
+
+        syn = Synapse()
+        syn.login()
+
+        my_session = AgentSession(agent_registration_id=my_agent_registration_id).start(
```
BryanFauble, line 157: `my_agent_registration_id` should be specified in this example

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+        modified_on: The date the agent session was last modified.
+        agent_registration_id: The registration ID of the agent that will
+            be used for this session.
+        etag: The etag of the agent session.
+
+    Note: It is recommended to use the `Agent` class to start and get sessions,
+    but you are free to use this class directly if you wish.
+
+    Example: Start a session and send a prompt.
+
```
BryanFauble, line 150: The formatting of these examples isn't exactly correct:
https://synapsepythonclient--1152.org.readthedocs.build/en/1152/reference/experimental/sync/agent/#synapseclient.models.Agent

image

It must be in this format (Copied from another spot in the code):

```
        Example: Using this function
            Download file into cache

                from synapseclient.api import get_from_entity_factory

                syn = Synapse()
                syn.login()

                entity = await get_from_entity_factory(synapse_id_or_path='syn1906479')
                print(entity.name)
                print(entity.path)
```

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+        Returns:
+            The registered or existing Agent object.
+        """
+        syn = Synapse.get_client(synapse_client=synapse_client)
+        agent_response = await register_agent(
+            cloud_agent_id=self.cloud_agent_id,
+            cloud_alias_id=self.cloud_alias_id,
+            synapse_client=syn,
+        )
```
BryanFauble, line 499: You can drop the extra call to `get_client` since you're just passing along whatever was given

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -0,0 +1,372 @@
...
+    async def send_job_and_wait_async(
+        self,
+        post_exchange_args: Optional[Dict[str, Any]] = None,
+        *,
+        synapse_client: Optional[Synapse] = None,
+    ) -> "AsynchronousCommunicator":
+        """Send the job to the Asynchronous Job service and wait for it to complete.
+
+        Arguments:
+            post_exchange_args: Additional arguments to pass to the request.
```
BryanFauble, line 58: Could you update what I wrote here to be a little more verbose on how this is intended to be used? Maybe an example too?

Could you also update the language for the `synapse_client` docstring to match the other areas of the code?

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -0,0 +1,372 @@
...
+    def fill_from_dict(self, async_job_status: dict) -> "AsynchronousJobStatus":
+        """Converts a response from the REST API into this dataclass.
+
+        Arguments:
+            async_job_status: The response from the REST API.
+
+        Returns:
+            A AsynchronousJobStatus object.
+        """
+        self.state = async_job_status.get("jobState", None)
```
BryanFauble, line 198: We'll need to instantiate a AsynchronousJobState object here

synapseclient/models/protocols/agent_protocol.py
```diff
@@ -0,0 +1,163 @@
...
+
+class AgentSessionSynchronousProtocol(Protocol):
+    """Protocol for the methods of the AgentSession class that have synchronous counterparts
+    generated at runtime."""
+
+    def start(self, *, synapse_client: Optional[Synapse] = None) -> "AgentSession":
+        """Starts an agent session.
+
+        Arguments:
+            synapse_client: The Synapse client to use for the request.
```
BryanFauble, line 20: To update for consistent verbiage:
```
            synapse_client: If not passed in and caching was not disabled by
                    `Synapse.allow_client_caching(False)` this will use the last created
                    instance from the Synapse class constructor.
```

synapseclient/models/services/storable_entity_components.py
BryanFauble: This is causing errors in the test run because the code in this py script is expecting to be able to use `Annotations` beyond it being a type check:
```
 name 'Annotations' is not defined
  File "/home/runner/work/synapsePythonClient/synapsePythonClient/synapseclient/models/services/storable_entity_components.py", line 38, in wrap_coroutine
    return await task
  File "/opt/hostedtoolcache/Python/3.9.21/x64/lib/python3.9/asyncio/futures.py", line 287, in __await__
    return self.result()  # May raise too.
  File "/opt/hostedtoolcache/Python/3.9.21/x64/lib/python3.9/asyncio/futures.py", line 201, in result
    raise self._exception
  File "/opt/hostedtoolcache/Python/3.9.21/x64/lib/python3.9/asyncio/tasks.py", line 256, in __step
    result = coro.send(None)
  File "/home/runner/work/synapsePythonClient/synapsePythonClient/synapseclient/models/services/storable_entity_components.py", line 245, in _store_activity_and_annotations
    result = await Annotations(
NameError: name 'Annotations' is not defined
name 'Annotations' is not defined
NoneType: None
```


In the spots it's being used we'll need to use the lazy import approach

docs/reference/experimental/sync/agent.md
```diff
@@ -0,0 +1,34 @@
...
+## API Reference
+
+::: synapseclient.models.Agent
+    options:
+        members:
+        - register
+        - get
+        - start_session
+        - get_session
+        - prompt
```
BryanFauble, line 22: These functions are not showing on the doc site:
https://synapsepythonclient--1152.org.readthedocs.build/en/1152/reference/experimental/sync/agent/#synapseclient.models.Agent-functions

I think because it's missing `inherited_members: true`

tests/integration/synapseclient/models/async/test_agent_async.py
```diff
@@ -0,0 +1,211 @@
...
+from synapseclient.core.constants.concrete_types import AGENT_CHAT_REQUEST
+from synapseclient.models.agent import (
+    Agent,
+    AgentPrompt,
+    AgentSession,
+    AgentSessionAccessLevel,
+)
+
+AGENT_AWS_ID = "QOTV3KQM1X"
+AGENT_REGISTRATION_ID = "29"
```
BryanFauble, line 15: Add a comment what these map to

tests/integration/synapseclient/models/async/test_agent_async.py
```diff
@@ -0,0 +1,211 @@
...
+        )
+        # WHEN I start a session
+        await agent_session.start_async(synapse_client=self.syn)
+        # AND I update the access level of the session
+        agent_session.access_level = AgentSessionAccessLevel.READ_YOUR_PRIVATE_DATA
+        await agent_session.update_async(synapse_client=self.syn)
+        # THEN I expect the access level to be updated
+        assert (
+            agent_session.access_level == AgentSessionAccessLevel.READ_YOUR_PRIVATE_DATA
+        )
```
BryanFauble, line 100: A better verification of this is to make sure that the data was stored in Synapse. Consider if this silently failed and did not persist the data to Synapse. Your instance would have the updated attribute value, but the next time the session is retrieved it could be using the previous access level.

Perform a new read with a new instance after calling `update_async` and verify on that newly read instance that it contains the right access_level

tests/integration/synapseclient/models/async/test_agent_async.py
```diff
@@ -0,0 +1,211 @@
...
+        assert agent_session.chat_history[0].response is not None
+        assert agent_session.chat_history[0].trace is not None
+
+
+class TestAgent:
+    """Integration tests for the synchronous methods of the Agent class."""
+
+    def get_test_agent(self) -> Agent:
+        return Agent(
+            cloud_agent_id="QOTV3KQM1X",
```
BryanFauble, line 124: You have the `AGENT_AWS_ID` constant available

BryanFauble (changes requested):
Great work! See my comments and re-request a review when you're ready

synapseclient/api/agent_services.py
BWMac: Updated in all agent_services functions

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -0,0 +1,372 @@
...
+                f"{job_status.error_message}\n{job_status.error_details}",
+            )
+        else:
+            break
+    else:
+        raise SynapseTimeoutError(
+            f"Timeout waiting for query results: {time.time() - start_time} seconds"
+        )
+
+    return result
```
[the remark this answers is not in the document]
BWMac: We could, but I think it may make more sense as is since we are adding the key/value pairs of the result here to another dictionary where it is called so it would just have to get converted right back to a dictionary anyway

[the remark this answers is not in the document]
BryanFauble: At the moment that is true that is how it is working, but should `job_id` be added as an attribute to `AsynchronousJobStatus`, and `AsynchronousJobStatus` be returned from `send_job_and_wait_async` and `get_job_async`?

Inside `get_job_async` construct the `AsynchronousJobStatus` and attach the Job ID

docs/reference/experimental/sync/agent.md
BWMac: I got the functions to show up for `Agent`, but they aren't showing up for `Activitiy`, `Team`, or `UserProfile` in the synchronous section and adding `inherited_members: true` didn't work 🤔

BWMac: @BryanFauble It looks like those aren't showing up in the current version of the [docs](https://python-docs.synapse.org/reference/oop/models/) too. Was there a reason why that wasn't happening before that I'm not aware of?

synapseclient/models/mixins/asynchronous_job.py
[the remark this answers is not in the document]
BWMac: `job_id` already is on `AsynchronousJobStatus` as `id`. I think this makes some sense to do, but then does `AgentPrompt` (and future `AsynchronousCommunicator`-inheriting classes) then have a sort of `fill_from_async_job_status` function instead of `fill_from_dict`?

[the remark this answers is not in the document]
BWMac: After looking into this a bit more, this won't work as suggested. The reason is that the response bodies are different while the job is being worked on by the server ([status model](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/asynch/AsynchronousJobStatus.html) versus when the job has been completed ([chat response model](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/AgentChatResponse.html)) That job status API response will always be the same across asynchronous job types, allowing for `AsynchronousJobStatus` to be a universal "internal handler" in this function for determining whether a job is complete or not. But the response will differ depending on the job type.

BryanFauble:
@BWMac A bunch of tests seem to be failing with:
```
FAILED tests/integration/synapseclient/models/async/test_agent_async.py::TestAgentPrompt::test_send_job_and_wait_async_with_post_exchange_args - synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error: AgentRegistrationId='29' does not exist
```

Since the integration tests run against the DEV Instance of Synapse, does this agent need to be created there, and tests updated with this config?

docs/reference/experimental/sync/agent.md
BryanFauble: > Was there a reason why that wasn't happening before that I'm not aware of?

Nope. Just that I missed it. What you have currently looks great!

docs/reference/experimental/async/activity.md
```diff
@@ -0,0 +1,20 @@
+# Activity
```
BryanFauble, line 1: This is the warning message I originally had when going to this page (https://python-docs.synapse.org/reference/oop/models/), should we add this back-in somewhere?

> Contained within this file are experimental interfaces for working with the Synapse Python
Client. Unless otherwise noted these interfaces are subject to change at any time. Use
at your own risk.

mkdocs.yml
```diff
@@ -75,8 +75,29 @@ nav:
...
+              - UserProfile: reference/experimental/sync/user_profile.md
+          - Asynchronous:
+              - Agent: reference/experimental/async/agent.md
+              - Project: reference/experimental/async/project.md
+              - Folder: reference/experimental/async/folder.md
+              - File: reference/experimental/async/file.md
+              - Table: reference/experimental/async/table.md
+              - Activity: reference/experimental/async/activity.md
+              - Team: reference/experimental/async/team.md
+              - UserProfile: reference/experimental/async/user_profile.md
```
BryanFauble, line 100: What do think about putting `Mixins` as the last in the list?

I was also thinking about 1 additional change to drop the additional navigation into the additional `Synchronous` directory. The purpose would be so that all the object models aren't hidden when scrolling down the page.

```
          - Agent: reference/experimental/sync/agent.md
          - ...
          - ...
          - Asynchronous:
              - Agent: reference/experimental/async/agent.md
              - ...
```
Current page:
image

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+            synapse_client=synapse_client, post_exchange_args={"newer_than": newer_than}
+        )
+
+        self.chat_history.append(agent_prompt)
+
+        if print_response:
+            print(f"PROMPT:\n{prompt}\n")
+            print(f"RESPONSE:\n{agent_prompt.response}\n")
+            if enable_trace:
+                print(f"TRACE:\n{agent_prompt.trace}")
```
BryanFauble, line 359: Should these be raw `print` statements, instead of a `syn.logger.info()` message?

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+
+            syn = Synapse()
+            syn.login()
+
+            my_agent = Agent()
+            my_agent.prompt(
+                prompt="Can you tell me about the AD Knowledge Portal dataset?",
+                enable_trace=True,
+                print_response=True,
+            )
```
BryanFauble, line 390: Include imports in the example

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+    Example: Register and chat with a custom agent
+        **Only available for internal users (Sage Bionetworks employees)**
+
+        Alternatively, you can register a custom agent and chat with it provided
+        you have already created it.
+
+            syn = Synapse()
+            syn.login(silent=True)
+
+            my_agent = Agent(cloud_agent_id=AWS_AGENT_ID)
```
BryanFauble, line 401: `AWS_AGENT_ID` isn't defined in this example

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+            my_agent.register()
+
+            my_agent.prompt(
+                prompt="Hello",
+            enable_trace=True,
+            print_response=True,
+        )
```
BryanFauble, line 408: Tab in for formatting

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+@dataclass
+@async_to_sync
+class Agent(AgentSynchronousProtocol):
+    """Represents a [Synapse Agent Registration](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/agent/AgentRegistration.html)
+
+    Attributes:
+        cloud_agent_id: The unique ID of the agent in the cloud provider.
+        cloud_alias_id: The alias ID of the agent in the cloud provider.
+                    Defaults to 'TSTALIASID' in the Synapse API.
+        synapse_registration_id: The ID number of the agent assigned by Synapse.
```
BryanFauble, line 371: Should be `registration_id` instead of `synapse_registration_id`

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+    """The date the agent was registered."""
+
+    type: Optional[AgentType] = None
+    """The type of agent. One of either BASELINE or CUSTOM."""
+
+    sessions: Dict[str, AgentSession] = field(default_factory=dict)
+    """A dictionary of AgentSession objects, keyed by session ID."""
+
+    current_session: Optional[AgentSession] = None
+    """The current session. Prompts will be sent to this session by default."""
```
BryanFauble, line 460: These fields are not in the class docstring so they're missing from this table: https://synapsepythonclient--1152.org.readthedocs.build/en/1152/reference/experimental/sync/agent/#synapseclient.models.Agent

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+            agent_registration: The response from the REST API.
+
+        Returns:
+            The Agent object.
+        """
+        self.cloud_agent_id = agent_registration.get("awsAgentId", None)
+        self.cloud_alias_id = agent_registration.get("awsAliasId", None)
+        self.registration_id = agent_registration.get("agentRegistrationId", None)
+        self.registered_on = agent_registration.get("registeredOn", None)
+        self.type = agent_registration.get("type", None)
```
BryanFauble, line 476: This should be updated to create an Enum of `AgentSession`. Right now this would set `self.type` to a type `str`

synapseclient/models/agent.py
```diff
@@ -0,0 +1,629 @@
...
+        """Starts an agent session.
+
+        Arguments:
+            synapse_client: If not passed in and caching was not disabled by
+                    `Synapse.allow_client_caching(False)` this will use the last created
+                    instance from the Synapse class constructor.
+
+        Returns:
+            The new AgentSession object.
+        """
```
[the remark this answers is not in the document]
BWMac: Should there be duplicate examples in both the class and function docstrings? I think for the most part I put examples of all available functionality in the class ones.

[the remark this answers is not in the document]
BryanFauble: I had been planning to use the class as the more simple examples, and the function docstring to show more advanced use cases or all options.

If there isn't a more advanced use case, it'd probably be a copy

synapseclient/models/agent.py
```diff
@@ -0,0 +1,651 @@
...
+            synapse_client=synapse_client, post_exchange_args={"newer_than": newer_than}
+        )
+
+        self.chat_history.append(agent_prompt)
+
+        if print_response:
+            synapse_client.logger.info(f"PROMPT:\n{prompt}\n")
+            synapse_client.logger.info(f"RESPONSE:\n{agent_prompt.response}\n")
+            if enable_trace:
+                synapse_client.logger.info(f"TRACE:\n{agent_prompt.trace}")
```
BryanFauble, line 354: This will break if `synapse_client` is none. This needs:
`client = Synapse.get_client(synapse_client=synapse_client)`

and use that instance

synapseclient/models/agent.py
```diff
@@ -0,0 +1,882 @@
...
+
+                syn = Synapse()
+                syn.login()
+
+                my_session = await AgentSession(agent_registration_id="foo").start_async()
+                await my_session.prompt_async(
+                    prompt="Hello",
+                    enable_trace=True,
+                    print_response=True,
+                )
```
BryanFauble, line 278: At the moment these examples wouldn't run (Ignoring the valid `agent_registration_id` since `await` won't be valid here without being inside the context of an asyncio.run command:

```
                import asyncio
                from synapseclient.models.agent import AgentSession

                syn = Synapse()
                syn.login()

                async def main():
                  my_session = await AgentSession(agent_registration_id="foo").start_async()
                  await my_session.prompt_async(
                      prompt="Hello",
                      enable_trace=True,
                      print_response=True,
                  )
                asyncio.run(main())
```

Is the format I had been following

BryanFauble (approved):
Awesome changes! This all looks good to me.

When all the tests are passing we can move forward on merging the changes.

tests/integration/synapseclient/models/synchronous/test_agent.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+class TestAgentSession:
+    """Integration tests for the synchronous methods of the AgentSession class."""
+
+    @pytest.fixture(autouse=True, scope="function")
+    def init(self, syn: Synapse) -> None:
+        self.syn = syn
+
+    def test_start(self) -> None:
```
BryanFauble, line 23: Can you make all of the integration tests `async def ...`

The reason for this is described here: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/CONTRIBUTING.md#lessons-learned-for-integration-testing

PR #1153, opened 2025-01-14
[SYNPY-1544] potential changes to mixin

BryanFauble:
@BWMac and I are working together to discuss ways in which we will implement this feature

### DIFF
--- a/synapseclient/models/mixins/asynchronous_job.py
+++ b/synapseclient/models/mixins/asynchronous_job.py
@@ -221,12 +221,15 @@ async def send_job_and_wait_async(
         SynapseTimeoutError: If the job does not complete within the timeout.
     """
     job_id = await send_job_async(request=request, synapse_client=synapse_client)
-    return await get_job_async(
-        job_id=job_id,
-        request_type=AGENT_CHAT_REQUEST,
-        synapse_client=synapse_client,
-        endpoint=endpoint,
-    )
+    return {
+        "jobId": job_id,
+        **await get_job_async(
+            job_id=job_id,
+            request_type=AGENT_CHAT_REQUEST,
+            synapse_client=synapse_client,
+            endpoint=endpoint,
+        ),
+    }
 
 
 async def send_job_async(
--- a/synapseclient/models/mixins/asynchronous_job.py
+++ b/synapseclient/models/mixins/asynchronous_job.py
@@ -3,13 +3,59 @@ import json
 import time
 from dataclasses import dataclass
 from enum import Enum
-from typing import TYPE_CHECKING, Any, Dict, Optional
+from typing import Any, Dict, Optional, Self
 
+from synapseclient import Synapse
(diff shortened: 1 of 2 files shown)


synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -3,13 +3,57 @@
...
+        """
+        Converts a response from the REST API into this dataclass.
+
+        Arguments:
+            synapse_response: The response from the REST API.
+
+        Returns:
+            An instance of this class.
+        """
+        raise NotImplementedError("fill_from_dict must be implemented.")
```
BWMac, line 34: These would just be blank right? [Example](https://github.com/Sage-Bionetworks-Workflows/py-orca/blob/09bf1109b65c22960c1582ba96756796aaaa0d71/src/orca/services/base/client_factory.py#L40) from py-orca
```suggestion
```

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -151,129 +195,156 @@ def fill_from_dict(self, async_job_status: dict) -> "AsynchronousJobStatus":
         return self
 
 
-class AsynchronousJob:
+async def send_job_and_wait_async(
```
BWMac, line 198: should these be in a separate script?

BWMac (approved):
Thank you, this is much better than what I had previously. I like using AgentPrompt as the async communicator, makes much more sense.

BWMac (approved):
Thank you, this is much better than what I had previously. I like using AgentPrompt as the async communicator, makes much more sense.

synapseclient/models/mixins/asynchronous_job.py
BryanFauble: I think it may be either filled in or blank. If it has this `NotImplementedError` I believe if you used the mixin and didn't provide an implementation it'll raise the exception.

In the other case where you leave it blank I assume that the function would execute and return None without an error.

synapseclient/models/mixins/asynchronous_job.py
BryanFauble: Good point. This would probably make sense to live within a py script at `synapseclient.api`

synapseclient/models/mixins/asynchronous_job.py
BWMac: Ah I didn't realize that. In that case having the error is better for sure.

synapseclient/models/mixins/asynchronous_job.py
BWMac: I can do this in the main feature branch

BWMac: @BryanFauble Actually, maybe that isn't possible. If we move `get_job_async` we need to move the AsynchronousJobStatus and AsynchronousJobState classes to avoid a circular import issue

PR #1155, opened 2025-01-28
[SYNPY-1544] Fixes docstring

BWMac:
This small PR updates an example docstring for the required access level to add annotations to a Synapse entity with a Synapse Agent prompt.

### DIFF
--- a/synapseclient/models/agent.py
+++ b/synapseclient/models/agent.py
@@ -838,6 +838,9 @@ class Agent(AgentSynchronousProtocol):
 
                 async def main():
                     my_agent = Agent()
+                    await my_agent.start_session_async(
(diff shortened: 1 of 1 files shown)


synapseclient/models/protocols/agent_protocol.py
```diff
@@ -337,6 +337,9 @@ def prompt(
             syn.login()
 
             my_agent = Agent()
+            my_agent.start_session(
+                access_level=AgentSessionAccessLevel.WRITE_YOUR_PRIVATE_DATA
```
BryanFauble, line 341: Can you add the import for this enum?

BryanFauble: And the import for Agent in this example

BryanFauble (approved):
LGTM! We don't need to wait for the build on this one

PR #1157, opened 2025-01-30
[SYNPY-1557] Sync a Linked Folder Bug

BWMac:
**Description**

A user recently reported unexpected behavior when using the following command:
```
synapse get -r --followLink --downloadLocation my_folder syn123
```
The folder being recursively downloaded (synced) contained a link to another Synapse folder, and an error was being thrown:
```
FileNotFoundError: [Errno 2] No such file or directory: 'my_folder/my_linked_subfolder/SYNAPSE_METADATA_MANIFEST.tsv'
```
In short, the Synapse client was unable to properly mirror the folder structure with a linked folder in it. We would expect that linked folder to be created as a subfolder containing the files that it does on Synapse. 

In order to enable this behavior, I first wrote an [integration test](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/604bd21c10348e7bef9d0acdf5591cd95a9cb44c/tests/integration/synapseutils/test_synapseutils_sync.py#L2085) to check that the client was giving the desired behavior in this situation with the expectation that the test would throw the same error until a fix was implemented. Then, I dug into the code and discovered that [in](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/604bd21c10348e7bef9d0acdf5591cd95a9cb44c/synapseclient/models/mixins/storable_container.py#L650) `_follow_link`, we were previously not tracking the target name for the linked entity, which was causing a locally created folder mirroring the linked Synapse folder not to be created by the child task, resulting in the error. After adding that tracking ([passing](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/604bd21c10348e7bef9d0acdf5591cd95a9cb44c/synapseclient/models/mixins/storable_container.py#L697) the `name` on to `_create_task_for_child`) within `_follow_link`, the test passed and I was able to successfully sync the folder from the original user report.

Previous behavior for syncing "normal" child folders and files was not adversely effected by the change. I tested this manually on my own test project, on the location reported by the user, and through ensuring that the existing integration tests still pass.

### DIFF
--- a/synapseclient/models/agent.py
+++ b/synapseclient/models/agent.py
@@ -838,6 +838,9 @@ class Agent(AgentSynchronousProtocol):
 
                 async def main():
                     my_agent = Agent()
+                    await my_agent.start_session_async(
+                        access_level=AgentSessionAccessLevel.WRITE_YOUR_PRIVATE_DATA
+                    )
                     await my_agent.prompt_async(
                         prompt="Add the annotation 'test' to the file 'syn123456789'",
                         enable_trace=True,
--- a/synapseclient/models/agent.py
+++ b/synapseclient/models/agent.py
@@ -647,7 +647,7 @@ class Agent(AgentSynchronousProtocol):
 
                 import asyncio
                 from synapseclient import Synapse
-                from synapseclient.models.agent import Agent
+                from synapseclient.models import Agent, AgentSessionAccessLevel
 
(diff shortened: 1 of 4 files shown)


BryanFauble (approved):
Thanks for the patch!

thomasyu888 (approved):
🔥 Great find.

PR #1158, opened 2025-01-30
[SYNPY-1544] Return the AgentPrompt when calling the prompt function

BryanFauble:
**Problem:**

1. When calling the `prompt` function it was a bit awkward to get the response out of the chat history in order to programmatically use the result.

**Solution:**

1. Return back the AgentPrompt when calling the `prompt` function

**Testing:**

1. I manually verified that I was getting back the expected object from the prompt call


BWMac (approved):
Great change, LGTM!

PR #1165, opened 2025-02-12
[SYNPY-1547] `parentWikiId=""` Bug

BWMac:
**Description:**

A user reported an unhandled error when trying to store a Wiki object in Synapse like so (where syn123 is instead a valid Synapse project ID that the user has access to):
```
from synapseclient import Wiki
import synapseclient

foo = Wiki(owner="syn123", title="test", parentWikiId="")
syn.store(foo)
```
This happened because the API does not treat that value of `parentWikiId` like a `None` value and instead throws an error. To prevent that error from happening, I added logic in `Wiki.__init__` that converts any falsey value that is provided for `parentWikiId` to `None`. 

Additionally, the user was receiving this error message:
```
AttributeError: 'SynapseHTTPError' object has no attribute 'message'
```
rather than a message passed on from the API which I addressed by updating logic [here](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/20be364ab981e3c577d337beff9ffc45aa0d5fe8/synapseclient/client.py#L4999) to check the response text.

**Testing:**
After these changes were made, the script above was able to be run successfully. I also added a unit test to ensure that `parentWikiId=""` is treated like `parentWikiId=None` and an integration test to ensure that Wiki objects created with `parentWikiId=""` are stored successfully in Synapse.

### DIFF
--- a/synapseclient/wiki.py
+++ b/synapseclient/wiki.py
@@ -122,6 +122,12 @@ class Wiki(DictObject):
                 kwargs["attachmentFileHandleIds"].append(handle)
             del kwargs["fileHandles"]
 
+        # if parentWikiId is a falsey value (empty string, None, etc),
+        # standardize it to None
+        if "parentWikiId" in kwargs:
+            if not kwargs["parentWikiId"]:
+                kwargs["parentWikiId"] = None
+
         super(Wiki, self).__init__(kwargs)
         self.ownerId = id_of(self.owner)
         del self["owner"]
--- a/synapseclient/wiki.py
+++ b/synapseclient/wiki.py
@@ -124,9 +124,8 @@ class Wiki(DictObject):
 
         # if parentWikiId is a falsey value (empty string, None, etc),
         # standardize it to None
-        if "parentWikiId" in kwargs:
-            if not kwargs["parentWikiId"]:
-                kwargs["parentWikiId"] = None
+        if "parentWikiId" in kwargs and not kwargs["parentWikiId"]:
+            kwargs["parentWikiId"] = None
 
         super(Wiki, self).__init__(kwargs)
         self.ownerId = id_of(self.owner)
--- a/tests/integration/synapseclient/test_wikis.py
+++ b/tests/integration/synapseclient/test_wikis.py
@@ -126,3 +126,18 @@ async def test_wiki_version(syn, project):
     w2 = syn.getWiki(owner=wiki.ownerId, subpageId=wiki.id, version=1)
     assert "version 2" in w2.title
     assert "version 2" in w2.markdown
+
+
+async def test_wiki_with_empty_string_parent_wiki_id(syn, project):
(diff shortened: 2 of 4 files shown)


synapseclient/wiki.py
```diff
@@ -122,6 +122,12 @@ def __init__(self, **kwargs):
                 kwargs["attachmentFileHandleIds"].append(handle)
             del kwargs["fileHandles"]
 
+        # if parentWikiId is a falsey value (empty string, None, etc),
+        # standardize it to None
+        if "parentWikiId" in kwargs:
+            if not kwargs["parentWikiId"]:
```
BryanFauble, line 128: Nit: Could combine with the line above using an and statement

tests/integration/synapseclient/test_wikis.py
```diff
@@ -126,3 +126,18 @@ async def test_wiki_version(syn, project):
     w2 = syn.getWiki(owner=wiki.ownerId, subpageId=wiki.id, version=1)
     assert "version 2" in w2.title
     assert "version 2" in w2.markdown
+
+
+async def test_wiki_with_empty_string_parent_wiki_id(syn, project):
+    # WHEN a wiki is created with an empty string parentWikiId
```
BryanFauble, line 132: A cleaner test of this would be:

GIVEN a wiki with an empty string for ParentWikiId
WHEN I store the Wiki
THEN It's stored in Synapse <-- This check should be added by performing a read to verify Synapse has the data we expect it to have
AND ...

synapseclient/wiki.py
BWMac: Would that cause a `KeyError` if "parentWikiId" was not in `kwargs`?

tests/unit/synapseclient/unit_test_Wiki.py
```diff
@@ -76,3 +76,12 @@ def test_wiki_with_none_attachments(syn: Synapse):
...
         w = Wiki(owner="syn1", markdown="markdown", attachments=None)
         syn.store(w)
+
+
+def test_wiki_with_empty_string_parent_wiki_id(syn: Synapse):
+    # WHEN a wiki is created with an empty string parentWikiId
+    with patch.object(syn, "restPOST"):
+        w = Wiki(owner="syn1", markdown="markdown", parentWikiId="")
+        # THEN the parentWikiId is set to None
+        assert w.parentWikiId is None
```
BryanFauble, line 86: Should there be other assertions being made below since syn.store is being called after this? Such as verifying the restPOST function is called with the appropriate data it's supposed to Post to Synapse

synapseclient/wiki.py
BryanFauble: No because the second condition in an and statement is never evaluated if the first isn't True. Since the "in" condition would be False it wouldn't run into a KeyError.

BryanFauble (approved):
Changes look good logically, just a few minor points

PR #1167, opened 2025-02-18
[SYNPY-1551] Batch upsert and inserts

BryanFauble:
**Problem:**

1. When inserting, or updating a large number of rows we are hitting the limits imposed by the Synapse Rest API as defined in [this document](https://rest-docs.synapse.org/rest/POST/entity/id/table/transaction/async/start.html).
2. The error the current Table API runs into is `The provided CSV file exceeds the maximum size of 1 GB.` when inserting a large number of rows. It forces the users to batch up the CSV themselves.

**Solution:**

1. Batch up the request when querying for data during the table upsert process
3. Batch up the requests to modify rows or insert new rows via CSV to Synapse during the `store_rows` function

**Testing:**

- [x] Verifying that upserting a large number of rows works (Including a large number of updates, and inserts in the same transaction)
- [x] Verify that updating a large number of rows works after querying for a large number of rows works
- [x] Verify that inserting a large number of rows (> 1GB CSV file) works

Showing what the console looks like during the upload of a large CSV that is split into multiple chunks:
image


synapseclient/models/mixins/table_operator.py
```diff
@@ -1898,174 +1918,196 @@ async def main():
...
+            chunk_size_limit = 1.5 * 1024 * 1024  # 1.5MB
+            current_chunk_size = 0
+            chunk = []
+            for row in rows_to_update:
+                row_size = sys.getsizeof(row.to_synapse_request())
+                if current_chunk_size + row_size > chunk_size_limit:
+                    chunked_rows_to_update.append(chunk)
+                    chunk = []
+                chunk.append(row)
+                current_chunk_size += row_size
```
BryanFauble, line 2075: This is some of the largest changes here. This interface that we use to submit a `PartialRow` change is what limits requests to 2MB or less:

https://rest-docs.synapse.org/rest/POST/entity/id/table/transaction/async/start.html

In this case I am using 1.5MB to allow for some overhead room.

synapseclient/models/mixins/table_operator.py
```diff
@@ -1898,174 +1923,220 @@ async def main():
...
+                row_size = sys.getsizeof(row.to_synapse_request())
+                if current_chunk_size + row_size > chunk_size_limit:
+                    chunked_rows_to_update.append(chunk)
+                    chunk = []
+                    current_chunk_size = 0
+                chunk.append(row)
+                current_chunk_size += row_size
+
+            if chunk:
+                chunked_rows_to_update.append(chunk)
```
BryanFauble, line 2106: This is some of the largest changes here. This interface that we use to submit a PartialRow change is what limits requests to 2MB or less:

https://rest-docs.synapse.org/rest/POST/entity/id/table/transaction/async/start.html

In this case I am using 1.5MB to allow for some overhead room.

synapseclient/core/utils.py
```diff
@@ -1015,8 +1016,12 @@ def printTransferProgress(
...
         status,
     )
-    sys.stdout.write(text)
-    sys.stdout.flush()
+    # Used for backwards compatability if anyone happens to be using this function
+    if logger:
+        logger.info(text)
+    else:
+        sys.stdout.write(text)
+        sys.stdout.flush()
```
BryanFauble, line 1024: This is to make sure that the tqdm bar is playing nicely with these messages that get printed out for table operations. Internal to the "new" interface, we are still using some of the query functionality from the "old" client. Example of the console:
image

(It was much worse before :D )

synapseclient/models/protocols/table_operator_protocol.py
```diff
@@ -290,12 +290,15 @@ def upsert_rows(
...
 
-
         Limitations:
 
         - The number of rows that may be upserted in a single call should be
-            limited. Additional work is planned to support batching
-            the calls automatically for you.
+            kept to a minimum. There is significant overhead in the request to
+            Synapse for each row that is upserted. If you are upserting a large
+            number of rows a better approach may be to query for the data you want
```
BryanFauble, line 298: to update (Note for self)

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -370,14 +378,14 @@ async def get_job_async(
         SynapseTimeoutError: If the job does not complete or progress within the timeout interval.
     """
     client = Synapse.get_client(synapse_client=synapse_client)
-    start_time = asyncio.get_event_loop().time()
+    start_time = time.time()
```
BryanFauble, line 381: This kind of time was causing the message to appear oddly. Moving to `time.time`, seems to have resolved it, as it seems to wait the expected time.

image

synapseclient/core/upload/multipart_upload_async.py
```diff
@@ -684,15 +685,16 @@ def part_fn(part_number: int) -> bytes:
...
+    with logging_redirect_tqdm(loggers=[syn.logger]):
+        return await _multipart_upload_async(
+            syn,
+            dest_file_name,
+            upload_request,
+            part_fn,
+            md5_fn_util,
+            force_restart=force_restart,
+            storage_str=storage_str,
+        )
```
BryanFauble, line 697: I was noticing some weirdness when I had multiple progress bar running (One for the chunked upload of row data into a Table, and one for the actual upload of the file itself).

This doesn't totally fix it, but the behavior seems better with it like this.

synapseclient/models/mixins/table_operator.py
```diff
@@ -2428,63 +2699,250 @@ async def main():
...
+        client: Synapse,
+        job_timeout: int,
+        additional_changes: List[
+            Union[
+                "TableSchemaChangeRequest",
+                "UploadToTableRequest",
+                "AppendableRowSetRequest",
+            ]
+        ] = None,
+    ) -> None:
```
BryanFauble, line 2779: I'm not a super big fan of the code I wrote for this. I had looked at some alternative libraries but was hesitant to add another dependency here, especially when thinking about performance concerns that are outlined in [this comparison for splitting CSV files](https://www.mungingdata.com/python/split-csv-write-chunk-pandas/#dask).

From their investigation, the Python filesystem API was much quicker (Likely since under the hood it uses C). I didn't do a new analysis based on the code I wrote here as I've chalked it up in a "Good enough" category.

thomasyu888 (commented):
Awesome to see this - I did a partial review, but hoping the team can review the rest.

synapseclient/models/mixins/table_operator.py
```diff
@@ -2428,63 +2699,250 @@ async def main():
...
+                    escapechar=csv_table_descriptor.escape_character,
+                    lineterminator=csv_table_descriptor.line_end,
+                    quotechar=csv_table_descriptor.quote_character,
                 )
-            finally:
-                os.remove(filepath)
-            upload_request = UploadToTableRequest(
-                table_id=self.id, upload_file_handle_id=file_handle_id, update_etag=None
+                header = next(reader)
+                temp_dir = client.cache.get_cache_dir(file_handle_id=111111111)
```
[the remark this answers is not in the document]
thomasyu888: 1. I think the solution you outlined about using a `bytesio` object makes the most sense, I guess streaming the csv could be interesting, but that would require more thought.
2. I agree with writing these CSV to disk and leveraging the synapse cache is great.  I guess an alternative here is to use `NamedTemporaryFile`.

synapseclient/models/mixins/asynchronous_job.py
BryanFauble: The default timeout was 60. The configuration to change the default wasn't exposed up through the functions that called it.

synapseclient/models/mixins/table_operator.py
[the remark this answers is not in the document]
BryanFauble: > I think the solution you outlined about using a bytesio object makes the most sense, I guess streaming the csv could be interesting, but that would require more thought.


We need to move forward on implementing this. It looks like windows (At least in the github integration tests) aren't working for this solution:

> PermissionError: [Errno 13] Permission denied: 'C:\\Users\\runneradmin\\.synapseCache\\111\\111111111\\chunked_csv_for_synapse_store_rowsytdo3xfn.csv'

I will explore what the solution for this is going to look like.

synapseclient/models/mixins/table_operator.py
```diff
@@ -1197,6 +1223,21 @@ def to_synapse_request(self):
         delete_none_keys(result)
         return result
 
+    def size(self) -> int:
```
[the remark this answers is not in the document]
BryanFauble: There is a difference. `sys.getsizeof()` under-reports the size of the data compared to the number that Synapse calculates and uses. It ends up causing issues with the sizes that we send to Synapse since it ends up passing along too much data.

synapseclient/models/mixins/table_operator.py
```diff
@@ -1730,12 +1773,249 @@ def has_changed(self) -> bool:
...
+            The select statement that can be used to query Synapse to determine if a row
+            already exists in the
+        """
+
+        select_statement = "SELECT ROW_ID, "
+
+        if self.__class__.__name__ in CLASSES_THAT_CONTAIN_ROW_ETAG:
+            select_statement += "ROW_ETAG, "
+
+        select_statement += f"{', '.join(all_columns_from_df)} FROM {self.id} WHERE "
```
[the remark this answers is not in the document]
BryanFauble: This List is created like:
`all_columns_from_df = [f'"{column}"' for column in values.columns]`

So they will be all strings.

> is there the likelihood that the data type being passed in is a mixture of types?

Not for column names, but the values of the cells may be of different data types. In some cases (Like strings) we need to single quote them for the SQL select, otherwise like ints do not need it. That is what the lines just below this that are checking `ColumnType` is doing.

synapseclient/models/mixins/table_operator.py
```diff
@@ -1730,12 +1773,249 @@ def has_changed(self) -> bool:
...
+                    f"Column type {column_model.column_type} is not supported for primary_keys"
+                )
+            elif column_model.column_type in (
+                ColumnType.STRING,
+                ColumnType.MEDIUMTEXT,
+                ColumnType.LARGETEXT,
+                ColumnType.LINK,
+                ColumnType.ENTITYID,
+            ):
+                values_for_where_statement = [
```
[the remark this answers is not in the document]
BryanFauble: > Should this be a set to remove duplicates?

Good point.

> I have run into an issue of only being able to do a certain length of select .. from . .. where column in ('long, 'list'), but I forget the limit, i dont think we have to worry about it right this moment, but it is a concern

You're right. There is a limit, but I hadn't hit it my limited testing. The `rows_per_query` parameter in the `upsert_rows_async` method is what we/users will be able to lower when this becomes a concern.

synapseclient/core/upload/multipart_upload_async.py
BryanFauble: Great find. This is likely the issue we're encountering here. I will create a tech debt ticket to go back and look at these.

synapseclient/models/mixins/table_operator.py
```diff
@@ -1891,181 +2240,140 @@ async def main():
             values = DataFrame(values)
         elif isinstance(values, str):
             values = csv_to_pandas_df(filepath=values, **kwargs)
-        elif isinstance(values, DataFrame) or isinstance(values, str):
-            values = values.copy()
+        elif isinstance(values, DataFrame):
+            pass
```
BWMac, line 2244: is this missing some functionality?

synapseclient/models/mixins/table_operator.py
```diff
@@ -2428,63 +2799,249 @@ async def main():
...
+                        )
+                finally:
+                    temp_file.close()
+
+        else:
+            raise ValueError(
+                "Don't know how to make tables from values of type %s." % type(values)
             )
+
+    async def _chunk_and_upload_csv(
```
BWMac, line 2864: Should this function have its own docstring (separate from `_send_update`)?

BWMac (commented):
I was able to confirm that the batch uploading works by cranking the number of rows in the CSV file generated in the POC script way up. A few initial comments here and I'll circle back once the updates to the CSV chunking/streaming code are made

synapseclient/models/mixins/table_operator.py
BryanFauble: no. This is just to make sure it doesn't fall into the

```
        else:
            raise ValueError(
                "Don't know how to make tables from values of type %s." % type(values)
            )
```

block. If they're passing in a `DataFrame` we don't need to do anything to prep the `values` argument for usage in the method.

docs/scripts/object_orientated_programming_poc/oop_poc_table.py
```diff
@@ -16,7 +16,7 @@
 from datetime import date, datetime, timedelta, timezone
 
 import synapseclient
-from synapseclient.models import Column, ColumnType, CsvResultFormat, Row, Table
+from synapseclient.models import Column, ColumnType, Table
```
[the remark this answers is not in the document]
BryanFauble: I pulled this out to a constant at the top of the file

synapseclient/models/mixins/table_operator.py
BryanFauble: I completed this in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1168

synapseclient/models/mixins/table_operator.py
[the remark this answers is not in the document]
BryanFauble: It's not possible to use the function to delete rows. This is because to use the function, you need to supply one or more non-system-defined columns that you are selecting the data on. When you send an update request to [this API](https://rest-docs.synapse.org/rest/POST/entity/id/table/transaction/async/start.html) and include more than the ROW_ID and ROW_VERSION columns, it is always treated as an update to that row (And null out the data you do not supply). In order to delete a row, you'd need to send a request to that API, but only include ROW_ID and ROW_VERSION. That is how the `delete_rows` functionality is working.

synapseclient/core/upload/multipart_upload_async.py
BryanFauble: https://sagebionetworks.jira.com/browse/SYNPY-1488

synapseclient/models/mixins/table_operator.py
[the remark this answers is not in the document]
BryanFauble: This was done in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1168

BWMac (approved):

PR #1168, opened 2025-02-24
[SYNPY-1551] Upload CSV/DataFrame without loading entire 1Gb chunks into memory

BryanFauble:
**Problem:**

1. On order to append/update a large number of rows in a Synapse Table we must go through this [Async table update API](https://rest-docs.synapse.org/rest/POST/entity/id/table/transaction/async/start.html). This API has a limit of 1GB files that may be appended to a Synapse table, which means if a user would like to append more data it needs to be batched up across multiple requests.
2. In running Github tests we saw that Windows was failing to write the files as CSVs to the Synapse cache directory as the temporary staging for the upload process:
> PermissionError: [Errno 13] Permission denied: 'C:\Users\runneradmin\.synapseCache\111\111111111\chunked_csv_for_synapse_store_rowsytdo3xfn.csv'

**Solution:**

1. Checkout the Mermaid Sequence Diagrams [available on the doc site!](https://synapsepythonclient--1168.org.readthedocs.build/en/1168/reference/experimental/sync/table/#synapseclient.models.Table.store_rows)
2. Provide the ability for an (In-memory buffer) as the source for an upload to Synapse.
3. Modify the CSV Read/Write process for splitting the large CSV to determine chunk boundaries instead and read those boundaries during the file upload process (Changes in `upload_utils.py` and `multipart_upload_async.py`)
4. Modfiy the DataFrame Read/Write process to remove the need to write the DataFrame to a temporary CSV file. Instead, the DataFrame rows are read in 100 row increments to determine the row/byte boundary of each chunk being uploaded to Synapse. The upload process will then take into account the byte boundary when reading in the bytes from the DF.

**Testing:**

- [x] Integration test verifies that both a CSV and DataFrame can be uploaded in multiple chunks.
- [x] I verified that I could upload a ~7GB CSV file from a `t3a.micro` EC2 instance, and a >1GB DataFrame. I was doing this though different settings in the following python script:

```python
import asyncio
import os
import pandas as pd
from synapseclient.models import Table, Column, ColumnType
import random


syn = Synapse(debug=False)
syn.login()
AMOUNT_IN_ONE_GO = 100 

STORE_FROM_CSV = False


async def insert_rows():
    columns = [
        Column(name="column_index", column_type=ColumnType.INTEGER),
        Column(name="random_number", column_type=ColumnType.INTEGER),
        Column(name="random_string", column_type=ColumnType.STRING, maximum_size=1000),
        Column(name="random_string2", column_type=ColumnType.STRING, maximum_size=1000),
        Column(name="random_string3", column_type=ColumnType.STRING, maximum_size=1000),
        Column(name="random_string4", column_type=ColumnType.STRING, maximum_size=1000),
        Column(name="random_string5", column_type=ColumnType.STRING, maximum_size=1000),
        Column(name="random_string6", column_type=ColumnType.STRING, maximum_size=1000),
        Column(name="random_string7", column_type=ColumnType.STRING, maximum_size=1000),
    ]
    table = await Table(name="large-data", parent_id="syn52948289", columns=columns).store_async()

    # Random A-Z:
    random_strings = []
    random_characters = [chr(random.randint(65, 90)) for _ in range(1000)]
    random_string = "".join(random_characters)
    for _ in range(AMOUNT_IN_ONE_GO):
        random_strings.append(random_string)

    data = pd.DataFrame({
        "column_index": [i for i in range(AMOUNT_IN_ONE_GO)],
        "random_number": [random.randint(0, 100) for _ in range(AMOUNT_IN_ONE_GO)],
        "random_string": random_strings,
        "random_string2": random_strings,
        "random_string3": random_strings,
        "random_string4": random_strings,
        "random_string5": random_strings,
        "random_string6": random_strings,
        "random_string7": random_strings
    })
    if STORE_FROM_CSV:
        dir = os.path.expanduser("~/rows.csv")
        data.to_csv(dir, index=False, float_format="%.12g",)
        await table.store_rows_async(values=dir)
    else:
        await table.store_rows_async(values=data)


asyncio.run(insert_rows())
```


synapseclient/core/upload/upload_utils.py
```diff
@@ -2,9 +2,155 @@
...
+            the next chunk.
+        path_to_file_to_split: The path to the file that we are reading off disk and
+            uploading portions of to Synapse.
+        total_size_of_chunks_being_uploaded: The total size of the chunks that are being
+            uploaded. This is used to calculate the maximum number of bytes to read
+            from the file, accounting for the last chunk that may be smaller than the
+            chunk size.
+    """
+    header_bytes = None
+    if bytes_to_prepend and part_number == 1:
```
BWMac, line 41: This logical seems to imply that bytes_to_prepend can be `None`, but the type hints don't indicate it is optional. Does this check for an empty bytes object?

synapseclient/core/upload/upload_utils.py
```diff
@@ -2,9 +2,155 @@
...
+    bytes_to_prepend: bytes,
+    part_number: int,
+    part_size: int,
+    byte_offset: int,
+    total_size_of_chunks_being_uploaded: int,
+    line_start: int,
+    line_end: int,
+    to_csv_kwargs: Optional[Dict[str, Any]] = None,
+) -> bytes:
+    """Read the nth chunk from the file assuming that we are not going to be reading
```
BWMac, line 22: NIT: All of the docstrings in this script are missing `Returns` blocks

synapseclient/models/mixins/table_operator.py
```diff
@@ -2803,68 +2857,208 @@ async def main():
...
+        path_to_csv: str,
+        byte_chunk_offset: int,
+        md5: str,
+        csv_table_descriptor: CsvTableDescriptor,
+        job_timeout: int,
+        progress_bar: tqdm,
+        wait_for_update_semaphore: asyncio.Semaphore,
+        file_suffix: str,
+    ) -> None:
+        """
```
BWMac, line 2946: NIT: Incomplete docstring

BWMac (approved):
LGTM just a couple comments/questions

synapseclient/core/upload/upload_utils.py
BryanFauble: Moved this to be Optional

BWMac (approved):

PR #1171, opened 2025-02-27
[SYNPY-1553] Update to support python 3.13

BryanFauble:
**Problem:**

1. Python 3.13 is not supported

**Solution:**

1. Update the github pipeline to build 3.13 and run integration tests on that version.
2. Updating the docs and install files to reference 3.13 as a valid version.
3. Removing the usage of `cgi.parse_header` in favor of the `email.message` `Message` class as mentioned in (PEP 594) https://docs.python.org/3/whatsnew/3.13.html#removed-modules-and-apis

**Testing:**

- [x] Unit/Integration testing
- [x] Verified that before the changes I can run into the `cgi` error as reported by the original user, and after the change I was able to download a small (few byte) file as expected:

image

### DIFF
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -3,7 +3,6 @@ Utility functions useful in the implementation and testing of the Synapse client
 """
 
 import base64
-import cgi
 import collections.abc
 import datetime
 import errno
@@ -26,6 +25,7 @@ import uuid
 import warnings
 import zipfile
 from dataclasses import asdict, is_dataclass
+from email.message import Message
 from typing import TYPE_CHECKING, TypeVar
 
 import requests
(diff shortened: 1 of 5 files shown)


synapseclient/core/utils.py
```diff
@@ -173,8 +173,9 @@ def extract_filename(content_disposition_header, default_filename=None):
 
     if not content_disposition_header:
         return default_filename
-    value, params = cgi.parse_header(content_disposition_header)
-    return params.get("filename", default_filename)
+    message = Message()
+    message.add_header("content-disposition", content_disposition_header)
+    return message.get_filename(failobj=default_filename)
```
BryanFauble, line 178: I verified that there was unit test coverage that already covered this function.

BWMac (approved):

PR #1174, opened 2025-03-04
[SYNPY-1575] Support default columns

BryanFauble:
**Problem:**

1. Views among other types of tables in Synapse have a number of default columns that may be added to the Table.

**Solution:**

1. Query for the default columns of the view according to the https://rest-docs.synapse.org/rest/GET/column/tableview/defaults.html API
2. Allow the schema change logic to handle these default columns as an "add" to the table schema

**Testing:**

1. More testing via integration is needed, but this was verified with a FileView

### DIFF
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -47,7 +47,7 @@ from .file_services import (
     put_file_multipart_add,
     put_file_multipart_complete,
 )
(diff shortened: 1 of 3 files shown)


BryanFauble:
Will be incorporated with https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175

PR #1175, opened 2025-03-04
[SYNPY-1571] Finish `Dataset` Model

BWMac:
***This PR is a WIP***

### DIFF
--- a/synapseclient/api/table_services.py
+++ b/synapseclient/api/table_services.py
@@ -4,11 +4,25 @@ columns in the Synapse REST API.
 """
 
 import json
+from enum import Enum
 from typing import TYPE_CHECKING, List, Optional
 
 if TYPE_CHECKING:
     from synapseclient import Synapse
     from synapseclient.models import Column
+
+
+class ViewTypeMask(int, Enum):
+    FILE = 0x01
+    PROJECT = 0x02
+    TABLE = 0x04
+    FOLDER = 0x08
+    VIEW = 0x10
+    DOCKER = 0x20
+    SUBMISSION_VIEW = 0x40
+    DATASET = 0x80
+    DATASET_COLLECTION = 0x100
+    MATERIALIZED_VIEW = 0x200
 
 
 async def get_columns(
@@ -71,32 +85,40 @@ async def post_columns(
     return columns
 
 
-async def get_default_view_columns(
-    view_type: str,
+async def get_default_columns(
+    view_type_mask: Optional[ViewTypeMask] = None,
     *,
-    view_type_mask: str = None,
     synapse_client: Optional["Synapse"] = None,
 ) -> List["Column"]:
-    """Get the default view columns for a given view type.
+    """Get the default columns for a given view type. This will query the following API:
+    <https://rest-docs.synapse.org/rest/GET/column/tableview/defaults.html> in order to
+    retrieve this information.
 
     Arguments:
-        view_type: The type of view to get the default columns for.
-        view_type_mask: The mask of the view type to get the default columns for.
+        view_type_mask: The type of view to get the default columns for. Not required
+            in some cases like a submission view.
         synapse_client: If not passed in and caching was not disabled by
                 `Synapse.allow_client_caching(False)` this will use the last created
-                instance from the Synapse class constructor
+                instance from the Synapse class constructor.
+
+    Returns: The annotations set in Synapse.
     """
+
     from synapseclient import Synapse
     from synapseclient.models import Column
 
-    uri = f"/column/tableview/defaults?viewEntityType={view_type}"
     if view_type_mask:
-        uri += f"&viewTypeMask={view_type_mask}"
-    columns_response = await Synapse.get_client(
-        synapse_client=synapse_client
-    ).rest_get_async(uri)
+        result = await Synapse.get_client(synapse_client=synapse_client).rest_get_async(
+            f"/column/tableview/defaults?viewTypeMask={view_type_mask.value}",
+        )
+    else:
+        result = await Synapse.get_client(synapse_client=synapse_client).rest_get_async(
+            "/column/tableview/defaults",
+        )
 
-    return [
-        Column().fill_from_dict(synapse_column=column)
-        for column in columns_response["list"]
-    ]
+    columns = []
+
+    for column in result.get("list", []):
+        columns.append(Column().fill_from_dict(synapse_column=column))
+
+    return columns
--- a/synapseclient/api/table_services.py
+++ b/synapseclient/api/table_services.py
@@ -69,3 +69,34 @@ async def post_columns(
         column.fill_from_dict(result["list"][i])
 
     return columns
+
+
+async def get_default_view_columns(
+    view_type: str,
+    *,
+    view_type_mask: str = None,
+    synapse_client: Optional["Synapse"] = None,
+) -> List["Column"]:
+    """Get the default view columns for a given view type.
+
+    Arguments:
+        view_type: The type of view to get the default columns for.
+        view_type_mask: The mask of the view type to get the default columns for.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor
+    """
+    from synapseclient import Synapse
+    from synapseclient.models import Column
+
+    uri = f"/column/tableview/defaults?viewEntityType={view_type}"
+    if view_type_mask:
+        uri += f"&viewTypeMask={view_type_mask}"
+    columns_response = await Synapse.get_client(
+        synapse_client=synapse_client
+    ).rest_get_async(uri)
+
+    return [
+        Column().fill_from_dict(synapse_column=column)
+        for column in columns_response["list"]
+    ]
--- a/synapseclient/models/dataset.py
+++ b/synapseclient/models/dataset.py
@@ -387,15 +387,7 @@ class Dataset(AccessControllable, TableOperator, TableRowOperator):
             "versionLabel": self.version_label,
             "versionComment": self.version_comment,
             "isLatestVersion": self.is_latest_version,
-            "columnIds": (
-                [
-                    column.id
-                    for column in self._last_persistent_instance.columns.values()
-                ]
-                if self._last_persistent_instance
-                and self._last_persistent_instance.columns
-                else []
-            ),
+            "columnIds": [column.id for column in self.columns.values()],
             "isSearchEnabled": self.is_search_enabled,
             "items": (
                 [item.to_synapse_request() for item in self.items] if self.items else []
--- a/synapseclient/models/dataset.py
+++ b/synapseclient/models/dataset.py
@@ -3,11 +3,12 @@ from collections import OrderedDict
 from copy import deepcopy
 from dataclasses import dataclass, field
 from datetime import date, datetime
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Dict, List, Optional, Union, TYPE_CHECKING
 
 from typing_extensions import Self
 
 from synapseclient import Synapse
+from synapseclient.api import get_default_view_columns
 from synapseclient.core.async_utils import async_to_sync
 from synapseclient.core.constants import concrete_types
 from synapseclient.core.utils import delete_none_keys
@@ -18,6 +19,9 @@ from synapseclient.models.mixins.table_operator import (
     TableOperator,
     TableRowOperator,
 )
+
+if TYPE_CHECKING:
+    from synapseclient.models import File, Folder
 
 
 @dataclass()
@@ -383,15 +387,19 @@ class Dataset(AccessControllable, TableOperator, TableRowOperator):
             "versionLabel": self.version_label,
             "versionComment": self.version_comment,
             "isLatestVersion": self.is_latest_version,
-            "columnIds": [
-                column.id for column in self._last_persistent_instance.columns.values()
-            ]
-            if self._last_persistent_instance and self._last_persistent_instance.columns
-            else [],
+            "columnIds": (
+                [
+                    column.id
+                    for column in self._last_persistent_instance.columns.values()
+                ]
+                if self._last_persistent_instance
+                and self._last_persistent_instance.columns
+                else []
+            ),
             "isSearchEnabled": self.is_search_enabled,
-            "items": [item.to_synapse_request() for item in self.items]
-            if self.items
-            else [],
+            "items": (
+                [item.to_synapse_request() for item in self.items] if self.items else []
+            ),
             "size": self.size,
             "checksum": self.checksum,
             "count": self.count,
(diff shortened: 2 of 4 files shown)


synapseclient/api/table_services.py
```diff
@@ -69,3 +101,44 @@ async def post_columns(
         column.fill_from_dict(result["list"][i])
 
     return columns
+
+
+async def get_default_columns(
+    view_entity_type: Optional[ViewEntityType] = None,
+    view_type_mask: Optional[ViewTypeMask] = None,
```
BryanFauble, line 108: I think we'll need to make this a type `int`. Saying that the table is going to contain both `Files` and `Folders` is valid for the FileView. ie:

```
FILE = 0x01
FOLDER = 0x08

view_type_mask = FILE | FOLDER
```

synapseclient/models/dataset.py
```diff
@@ -403,15 +403,197 @@ def to_synapse_request(self):
...
+        client = Synapse.get_client(synapse_client=synapse_client)
+
+        if isinstance(item, EntityRef):
+            self._append_entity_ref(entity_ref=item)
+        elif isinstance(item, File):
+            file = File(
+                id=item.id, version_number=item.version_number, download_file=False
+            ).get()
+            self._append_entity_ref(
+                entity_ref=EntityRef(id=file.id, version=file.version_label)
```
BryanFauble, line 445: Should this be `version=file.version_number`?

synapseclient/models/dataset.py
```diff
@@ -403,15 +403,197 @@ def to_synapse_request(self):
...
+            for child in children:
+                if child["type"] == concrete_types.FILE_ENTITY:
+                    file = File(
+                        id=child["id"],
+                        version_number=child["versionNumber"],
+                        download_file=False,
+                    ).get()
+                    self._append_entity_ref(
+                        entity_ref=EntityRef(id=file.id, version=file.version_label)
+                    )
```
BryanFauble, line 458: I would think that you wouldn't need to get the `File` class instance in this case if you already have the `id` and `versionNumber`

synapseclient/api/table_services.py
BWMac: So then what we would pass to this function is the enum value or an expression like `ViewTypeMask.File.value | ViewTypeMask.Folder.value`?

BryanFauble: One option is to remove the Enum from these (Only accept an int). This is how https://synapsepythonclient.readthedocs.io/en/latest/reference/experimental/sync/table/#synapseclient.models.Table.query_part_mask was built


Or do it like your suggesting.

synapseclient/models/dataset.py
```diff
@@ -403,15 +412,138 @@ def to_synapse_request(self):
...
+
+        if isinstance(item, EntityRef):
+            self._append_entity_ref(entity_ref=item)
+        elif isinstance(item, File):
+            file = File(
+                id=item.id, version_number=item.version_number, download_file=False
+            ).get()
+            self._append_entity_ref(
+                entity_ref=EntityRef(id=file.id, version=file.version_number)
+            )
```
BryanFauble, line 455: There is one improvement here where if `item.version_number` is filled in we can use that value, but if `item.version_number` isn't filled in then we'll need to get the `File` in order to populate the `EntityRef`

synapseclient/models/dataset.py
```diff
@@ -403,15 +412,138 @@ def to_synapse_request(self):
...
+
+        client = Synapse.get_client(synapse_client=synapse_client)
+
+        if isinstance(item, EntityRef):
+            self._remove_entity_ref(item)
+        elif isinstance(item, File):
+            file = File(
+                id=item.id, version_number=item.version_number, download_file=False
+            ).get()
+            self._remove_entity_ref(EntityRef(id=file.id, version=file.version_label))
```
BryanFauble, line 516: `version_number` over `version_label`

synapseclient/models/dataset.py
```diff
@@ -403,15 +412,138 @@ def to_synapse_request(self):
...
+            for child in children:
+                if child["type"] == concrete_types.FILE_ENTITY:
+                    file = File(
+                        id=child["id"],
+                        version_number=child["versionNumber"],
+                        download_file=False,
+                    ).get()
+                    self._remove_entity_ref(
+                        EntityRef(id=file.id, version=file.version_label)
+                    )
```
BryanFauble, line 528: To remove needing to `.get()` the File

synapseclient/models/mixins/table_components.py
```diff
@@ -0,0 +1,1837 @@
...
+    @property
+    def has_changed(self) -> bool:
+        """Used to satisfy the usage in this mixin from the parent class."""
+
+    @property
+    def has_columns_changed(self) -> bool:
+        """Used to satisfy the usage in this mixin from the parent class."""
+
+
+class ViewAttributesMixin(TableAttributesMixin):
```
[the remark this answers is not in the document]
BWMac: Good point, I like `ViewBase`/`TableBase`

synapseclient/models/mixins/table_components.py
```diff
@@ -0,0 +1,1837 @@
...
+                    ) or not isna(cell_value):
+                        partial_change_values[column_id] = (
+                            _convert_pandas_row_to_python_types(
+                                cell=cell_value, column_type=column_type
+                            )
+                        )
+                    else:
+                        partial_change_values[column_id] = None
+            print(partial_change_values)
+            breakpoint()
```
BryanFauble: Debug logic to remove later

BWMac: removed in latest commit (my bad)

synapseclient/models/mixins/table_components.py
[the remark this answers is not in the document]
BWMac: Updated in latest [commit](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175/commits/7cac857c7e8a96ea310f1d7fa92f9d55782a8d06)

synapseclient/models/mixins/table_operator.py
```diff
@@ -3938,7 +3938,7 @@ async def store_async(
         if self.include_default_columns:
             default_columns = await get_default_columns(
                 view_entity_type=self.view_entity_type,
-                view_type_mask=self.view_type_mask,
+                view_type_mask=self.view_type_mask.value,
```
BryanFauble, line 3941: If `self.view_type_mask` is None this is going to raise an exception

synapseclient/models/mixins/table_components.py
```diff
@@ -0,0 +1,1831 @@
...
+    @property
+    def has_changed(self) -> bool:
+        """Used to satisfy the usage in this mixin from the parent class."""
+
+    @property
+    def has_columns_changed(self) -> bool:
+        """Used to satisfy the usage in this mixin from the parent class."""
+
+
+class ViewBase(TableBase):
```
BryanFauble, line 95: Since this is introducing dataclass fields (`from dataclasses import field`), does this need the `@dataclass` decorator?

synapseclient/models/protocols/table_operator_protocol.py
```diff
@@ -933,3 +933,58 @@ def delete_rows(
...
         return DataFrame()
+
+
+class ViewOperatorSynchronousProtocol(Protocol):
+    """A class that extends the TableOperator and TableRowOperator classes to add
+    appropriately handle View-like Synapse entities.
+
+    In the Synapse API, a View is a sub-category of the Table model which includes other Table-like
+    entities including: SubmissionView, EntityView, and Dataset.
+    """
```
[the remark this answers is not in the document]
BWMac: This was my plan and I do think it will work. I'll let you know if I encounter any issues with it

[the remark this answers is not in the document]
BWMac: @BryanFauble Thinking about this more, if we "reimplement" the mixin functions on each child class (Dataset, FileView, etc.) to get specific docstrings as we discussed in Slack, we would still need to have protocols for those child classes that encompass all of those methods, right?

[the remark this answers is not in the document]
BryanFauble: > we would still need to have protocols for those child classes that encompass all of those methods, right?

Yeah thats right. It's unfortunate that we have to do this extra leg work, but I think it will improve the user experience here.

> So then maybe we only have protocols for the entity classes, and not the mixins at all if we are going that route

I agree, that makes sense. That way we can do what you're suggesting in [this commit](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175/commits/5c6c0987344becfa5da1cc3e402269fa62d3bc04), for each of the Entity classes that have the `update_rows_async`.

One thing that we could do is instead of creating a separate `Protocol`, is just add the other non-async version of the method right below it. So it'd look like:

```
class Dataset:
    async def update_rows_async(
        self,
        values: DATA_FRAME_TYPE,
        primary_keys: List[str],
        dry_run: bool = False,
        *,
        rows_per_query: int = 50000,
        update_size_bytes: int = 1.9 * MB,
        insert_size_bytes: int = 900 * MB,
        job_timeout: int = 600,
        wait_for_eventually_consistent_view: bool = False,
        wait_for_eventually_consistent_view_timeout: int = 600,
        **kwargs,
    ) -> None:
        """Placeholder Docstring specific to the Dataset class' implementation
        of the update_rows_async method."""
        await super().update_rows_async(
            values=values,
            primary_keys=primary_keys,
            dry_run=dry_run,
            rows_per_query=rows_per_query,
            update_size_bytes=update_size_bytes,
            insert_size_bytes=insert_size_bytes,
            job_timeout=job_timeout,
            wait_for_eventually_consistent_view=wait_for_eventually_consistent_view,
            wait_for_eventually_consistent_view_timeout=wait_for_eventually_consistent_view_timeout,
            **kwargs,
        )

    def update_rows(
        self,
        values: DATA_FRAME_TYPE,
        primary_keys: List[str],
        dry_run: bool = False,
        *,
        rows_per_query: int = 50000,
        update_size_bytes: int = 1.9 * MB,
        insert_size_bytes: int = 900 * MB,
        job_timeout: int = 600,
        wait_for_eventually_consistent_view: bool = False,
        wait_for_eventually_consistent_view_timeout: int = 600,
        **kwargs,
    ) -> None:
        """Placeholder Docstring specific to the Dataset class' implementation
        of the update_rows_async method."""
        # THE FUNCTION IS REPLACED AT RUNTIME
```

My thought here is that we would be spreading the code out less by doing this.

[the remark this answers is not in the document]
BWMac: I don't think I love the idea of having the async and sync methods right next to each other in the same script for documentation purposes.  I like the way things are broken up now, where users just looking for the simplest way to interact with Synapse entities (sync interface) can look at one page of documentation specific to that, and advanced users can find the async support if their use case demands it.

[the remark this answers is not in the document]
BryanFauble: > I like the way things are broken up now, where users just looking for the simplest way to interact with Synapse entities (sync interface) can look at one page of documentation specific to that, and advanced users can find the async support if their use case demands it.

The way that users interact with the documentation pages would not change. The suggestion to place them next to eachother is purely for us. Recall that in the markdown pages we are explicitly telling mkdocs which functions to include on the specific doc page:



```
## API Reference


::: synapseclient.models.Table
    options:
        inherited_members: true
        members:
        - get
        - store
        - delete
        - query
        - query_part_mask
        - store_rows
        - upsert_rows
        - delete_rows
        - snapshot
        - delete_column
        - add_column
        - reorder_column
        - get_permissions
        - get_acl
        - set_permissions
```

vs

```
## API Reference

::: synapseclient.models.Table
    options:
        inherited_members: true
        members:
        - get_async
        - store_async
        - delete_async
        - query_async
        - query_part_mask_async
        - store_rows_async
        - upsert_rows_async
        - delete_rows_async
        - snapshot_async
        - delete_column
        - add_column
        - reorder_column
        - get_permissions
        - get_acl
        - set_permissions
```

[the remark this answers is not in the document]
BWMac: Ah right. OK let me try this out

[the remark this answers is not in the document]
BWMac: Unfortunately it seems like the async function logic replacing that of the sync function at runtime doesn't work if the sync function is defined in the same class definition rather than in a Protocol that the class inherits.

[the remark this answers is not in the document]
BWMac: We resolved this with an [update](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175/commits/2a96709fda682b6dd07e236a7af8df1cfea5ffae) to the `async_to_sync` decorator function.

synapseclient/models/mixins/table_components.py
```diff
@@ -2096,235 +1642,556 @@ async def upsert_rows_async(
...
+            job_timeout=job_timeout,
+            wait_for_eventually_consistent_view=wait_for_eventually_consistent_view,
+            wait_for_eventually_consistent_view_timeout=wait_for_eventually_consistent_view_timeout,
+            synapse_client=synapse_client,
+            **kwargs,
+        )
+
+
+@async_to_sync
+class ViewUpdateMixin(TableUpsertMixin):
```
BryanFauble, line 1796:
```suggestion
class ViewUpdateMixin:
```

synapseclient/models/mixins/table_components.py
```diff
@@ -2096,235 +1642,556 @@ async def upsert_rows_async(
...
+
+            synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor
+
+            **kwargs: Additional arguments that are passed to the `pd.DataFrame`
+                function when the `values` argument is a path to a csv file.
+        """
+        await super().upsert_rows_async(
+            values=values,
```
BryanFauble, line 1896:
```suggestion
        await _upsert_rows_async(
            entity=self,
            values=values,
```

BWMac:
Still working on getting the synchronous integration tests to pass ([something weird](https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/13901764090/job/38895109445) is happening on this branch), but figured I could open this up for a first round of review in the meantime.

docs/reference/experimental/async/dataset.md
```diff
@@ -0,0 +1,31 @@
+# Dataset
+
+Contained within this file are experimental interfaces for working with the Synapse Python
+Client. Unless otherwise noted these interfaces are subject to change at any time. Use
+at your own risk.
+
+## API reference
+
+::: synapseclient.models.EntityRef
```
BryanFauble, line 9: Should this be ordered after the Dataset? I imagine most folks will care more about Dataset first than the EntityRef

docs/scripts/object_orientated_programming_poc/oop_poc_dataset.py
```diff
@@ -0,0 +1,76 @@
+import pandas as pd
```
BryanFauble, line 1: For the FileView I had considered making this script but I felt that jumping right in and making the tutorial was more effective. If you have time it'd be great if a tutorial could be created, otherwise this is sufficient for this PR.

https://synapsepythonclient--1181.org.readthedocs.build/en/1181/tutorials/python/fileview/

synapseclient/api/table_services.py
```diff
@@ -69,3 +101,69 @@ async def post_columns(
...
+
+        To get the default columns for a File or a Folder, you can use:
+
+        ```python
+        view_type_mask = ViewTypeMask.FILE.value | ViewTypeMask.FOLDER.value
+        columns = await get_default_columns(
+            view_entity_type=ViewEntityType.DATASET,
+            view_type_mask=view_type_mask,
+        )
+        ```
```
BryanFauble, line 139: Could the `import` statements be added to the examples here.

synapseclient/models/dataset.py
```diff
@@ -3,63 +3,579 @@
...
+
+    @staticmethod
+    def query(
+        query: str,
+        include_row_id_and_row_version: bool = True,
+        convert_to_datetime: bool = False,
+        *,
+        synapse_client: Optional[Synapse] = None,
+        **kwargs,
+    ) -> DATA_FRAME_TYPE:
```
BryanFauble, line 380: I don't think that this is needed since I am going to be keeping `QueryMixinSynchronousProtocol` logic added here:

https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1181/files#r1991825702

synapseclient/models/dataset.py
```diff
@@ -403,15 +1069,644 @@ def to_synapse_request(self):
...
+        return await super().snapshot_async(
+            comment=comment,
+            label=label,
+            activity=activity,
+            synapse_client=synapse_client,
+        )
+
+    @staticmethod
+    async def query_async(
+        query: str,
```
BryanFauble, line 1652: Same comment https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175/files#r1999374719

synapseclient/models/table.py
```diff
@@ -1,30 +1,1088 @@
...
+            syn.login()
+
+            Table(id="syn4567").delete()
+            ```
+        """
+        return None
+
+    @staticmethod
+    def query(
+        query: str,
```
BryanFauble, line 173: We may remove this https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175/files#r1999374719

synapseclient/models/table.py
```diff
@@ -1,30 +1,1088 @@
...
+            ```
+        """
+        from pandas import DataFrame
+
+        return DataFrame()
+
+    @staticmethod
+    def query_part_mask(
+        query: str,
+        part_mask: int,
```
BryanFauble, line 261: We may remove this https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175/files#r1999374719

docs/scripts/object_orientated_programming_poc/oop_poc_dataset.py
BWMac: I'll go ahead and add this!

BWMac: [added](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175/commits/ff4d0c69199f06d257db95c73781f952520d6a40)

docs/reference/experimental/async/dataset.md
BWMac: agreed, fixed!

synapseclient/api/table_services.py
```diff
@@ -4,13 +4,45 @@
...
 
+class ViewEntityType(str, Enum):
+    """String enum representing the type of view. This is used to determine
+    the default columns that are added to the table.
+    As defined in the Synapse REST API:
+    <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/ViewEntityType.html>
+    """
+
+    ENTITY_VIEW = "entityview"
+    SUBMISSION_VIEW = "submissionview"
```
thomasyu888, line 23: I wonder if submissionviews will have a different set of default columns.

synapseclient/models/mixins/table_components.py
```diff
@@ -442,6 +452,147 @@ async def store_async(
...
+                    "Names may only contain: letters, numbers, spaces, underscores, "
+                    "hyphens, periods, plus signs, apostrophes, and parentheses."
+                )
+        return await super().store_async(
+            dry_run=dry_run, job_timeout=job_timeout, synapse_client=synapse_client
+        )
+
+
+@async_to_sync
+class DeleteMixin:
```
thomasyu888, line 544: I imagine these sort of mixins will be relevant across all entity types. Is there something special about Table deletions?

BWMac: I don't think there is anything too specific about Tables/Views in terms of how you delete them. In the future if we want this could be extended to other entity types.

PR #1176, opened 2025-03-05
[SYNPY-1575] Support waiting for evetually consistent view after changes are made

BryanFauble:
**Problem:**

1. When changes are made to a view in Synapse an eventually consistent process is kicked off in order to bring that view up to date. If you query the view directly after the changes are made you will not see your changes reflected in that view.

**Solution:**

1. During the upsert method call allow users to specify if they would like to wait for their changes to show up in the view before returning control.

**Testing:**

1. I wrote an integration test (For FileViews that are not yet committed) to verify the functionality before and after

### DIFF
--- a/synapseclient/models/mixins/table_operator.py
+++ b/synapseclient/models/mixins/table_operator.py
@@ -1900,7 +1900,7 @@ class TableRowOperator(TableRowOperatorSynchronousProtocol):
         Returns:
             A tuple containing a list of PartialRow objects that will be used to update
             rows in Synapse, a list of the indexs of the rows in the original
-            DataFrame that have changes, a list of the indexs of the rows in the
+            DataFrame that have changes, a list of the indexes of the rows in the
             original DataFrame that do not have changes, and a list of the etags for
             the rows that have changes.
         """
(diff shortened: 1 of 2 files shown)


synapseclient/models/mixins/table_operator.py
```diff
@@ -1898,15 +1900,17 @@ def _construct_partial_rows_for_upsert(
         Returns:
             A tuple containing a list of PartialRow objects that will be used to update
             rows in Synapse, a list of the indexs of the rows in the original
-            DataFrame that have changes, and a list of the indexs of the rows in the
-            original DataFrame that do not have changes.
+            DataFrame that have changes, a list of the indexs of the rows in the
+            original DataFrame that do not have changes, and a list of the etags for
+            the rows that have changes.
```
BWMac, line 1905:
```suggestion
            DataFrame that have changes, a list of the indexes of the rows in the
            original DataFrame that do not have changes, and a list of the etags for
            the rows that have changes.
```

synapseclient/models/mixins/table_operator.py
```diff
@@ -2152,6 +2161,14 @@ async def upsert_rows_async(
                 is reached a `SynapseTimeoutError` will be raised.
                 The default is 600 seconds
 
+            wait_for_eventually_consistent_view: Only used if the table is a view. If
+                set to True this will wait for the view to be reflect any changes that
+                you've made to the view. This is useful if you need to query the view
+                after making changes to the data.
```
BWMac, line 2167:
```suggestion
            wait_for_eventually_consistent_view: Only used if the table is a view. If
                set to True this will wait for the view to reflect any changes that
                you've made to the view. This is useful if you need to query the view
                after making changes to the data.
```

BWMac (approved):
LGTM, thanks

PR #1177, opened 2025-03-06
[SYNPY-1488] Patch nested tqdm progress bars and messages to logger

BryanFauble:
**Problem:**

1. Nested TQDM progress bars by default overwrite other progress bars. This leads to loss of information and poor looking logs.
2. There are a significant number of print, and sysout messages getting printed to the console. When messages are printed in this way then it will break the formatting of the tqdm progress bar.

**Solution:**

1. Updating the TQDM progress bars to include `leave=None`. This makes it so the behavior for the progress bars it to be removed if it isn't the highest level parent progress bar. It cleans up the bars significantly.
2. Updating many print and sysout messages to use the synapse class logger to print out exceptions, warnings, and info messages.

**Testing:**

1. The following are 2 short clips that show the console log during the creation, querying, and update of a FileView within Synapse during an integration test run:

Before Changes: https://www.loom.com/share/b4594dc4f83142c784f43aa7561d9652?sid=99de514e-9177-4453-b161-54c928af2cec
After Changes: https://www.loom.com/share/0dead7ad188a4bd191a8dfd2cf31ee0c?sid=365c3fd9-1a14-4c65-9499-7e120834d646

1. I also verified that the syncFromSynapse and syncToSynapse looks correct:

```python
from synapseutils import syncFromSynapse, syncToSynapse

syn = Synapse(debug=False)
syn.login()

syncFromSynapse(syn=syn, entity="syn60584018", path=".")
syncToSynapse(syn=syn, manifestFile="/home/ec2-user/SYNAPSE_METADATA_MANIFEST.tsv", dryRun=True)
```

image

image

1. Finally, I verified that exceptions and the messages related to those exceptions occur as a result of running CLI commands:
(Before, only the message without any stacktrace would print)
image

(After, it will also print the stack trace along with the given exception)
image

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -5171,8 +5171,8 @@ class Synapse(object):
                 + result.get("errorDetails", None),
                 asynchronousJobStatus=result,
             )
-
-        progress_bar.update(progress_bar.total - progress_bar.n)
+        if progress_bar.total and progress_bar.n:
+            progress_bar.update(progress_bar.total - progress_bar.n)
         progress_bar.refresh()
         progress_bar.close()
         return result
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -43,6 +43,7 @@ from opentelemetry.instrumentation.requests import RequestsInstrumentor
 from opentelemetry.instrumentation.threading import ThreadingInstrumentor
 from opentelemetry.instrumentation.urllib import URLLibInstrumentor
 from opentelemetry.trace import Span
+from tqdm import tqdm
 
 import synapseclient
 import synapseclient.core.multithread_download as multithread_download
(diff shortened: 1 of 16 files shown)


synapseclient/client.py
```diff
@@ -5129,22 +5130,27 @@ def _waitForAsync(self, uri, request, endpoint=None):
         # https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/asynch/AsynchronousJobStatus.html
         sleep = self.table_query_sleep
         start_time = time.time()
-        lastMessage, lastProgress, lastTotal, progressed = "", 0, 1, False
+        lastMessage, lastProgress, lastTotal = "", 0, 1
```
BWMac, line 5133: Is this an opportunity to switch some variables over to camel-case? or are you just leaving all of the older logic alone with respect to formatting/style?

BWMac (approved):
LGTM, progress bar looks good in the video and when I run this branch locally.

synapseclient/client.py
BryanFauble: For this PR, my focus was solely on ensuring that things use tqdm and the logger.

This code will eventually need to be patched, but a significant portion of it will also be deprecated or removed. I have a ticket to go through and deprecate everything that has a replacement, as well as to take stock of the items that don’t.

PR #1179, opened 2025-03-07
[SYNPY-1571] Migrate table to new mixin structure

BryanFauble:
**Problem:**

1. There is a different way that @BWMac and I wanted to go for defining new functionality for mixins that are re-used for the Tables/Views.

**Solution:**

1. Moving over the rest of the functionality from `table_operator.py` into the 2 separate `table_components.py` which contain 1) Mixins, and 2) The Data Models like `Column`


synapseclient/models/mixins/table_components.py
```diff
@@ -64,23 +122,6 @@ class TableBase:
...
-        """Used to satisfy the usage in this mixin from the parent class."""
-
-    @property
-    def has_changed(self) -> bool:
-        """Used to satisfy the usage in this mixin from the parent class."""
-
-    @property
-    def has_columns_changed(self) -> bool:
-        """Used to satisfy the usage in this mixin from the parent class."""
-
```
BryanFauble, line 83: I was having issues where keeping these in was causing `self` to evaluate to `None` at runtime, and in the debugger it was hitting these types of functions (In the few spots I deleted these).

We'll need to figure out if we need to keep these or not

synapseclient/models/table_components.py
```diff
@@ -0,0 +1,670 @@
+import os
+from dataclasses import dataclass, field, replace
+from enum import Enum
+from typing import Any, Dict, List, Optional, TypeVar, Union
+
+from typing_extensions import Self
+
+from synapseclient import Column as Synapse_Column
```
BryanFauble, line 8: This python module got all the `dataclass` classes that I originally had in `table_operator.py`

synapseclient/models/mixins/table_components.py
BWMac: I liked having these in as a reminder to implement them in a new model, but I suppose we don't technically _need_ them.

BryanFauble:
> Thanks for doing all of this! Looks like some of the paths the docs generator relies on are broken now. If you want to fix before merging that works or I can do it in the primary feature branch since there are other changes to be made there anyway.

@BWMac ill merge these changes in and we can make a new commit or pr with the fixes for those

PR #1180, opened 2025-03-11
[SYNPY-1575] Prevent upsert from showing

BryanFauble:
**Problem:**

1. The `upsert` functionality is not available within a `View` type FileView/Dataset, and it should be hidden. The idea is that it is replaced by an method called `update` so it does not give the impression that you may insert rows via the function call.
2. The current implementation was causing both `upsert` and `update` functions to be available:
image

**Solution:**

1. Swap the upsert logic to be a function that can be called. Calling that function from both the upsert/update mixins.
2. Remove the dependency of the update mixin that extends the upsert mixin.

**Testing:**

1. Will be reviewing test runs. I also verified that I was able to load the mkdoc pages

### DIFF
--- a/synapseclient/models/mixins/table_components.py
+++ b/synapseclient/models/mixins/table_components.py
@@ -495,13 +495,18 @@ class ViewStoreMixin(TableStoreMixin):
         client = Synapse.get_client(synapse_client=synapse_client)
 
         if self.include_default_columns:
+            view_type_mask = None
+            if self.view_type_mask:
+                if isinstance(self.view_type_mask, ViewTypeMask):
+                    view_type_mask = self.view_type_mask.value
+                else:
+                    view_type_mask = self.view_type_mask
+
             default_columns = await get_default_columns(
                 view_entity_type=(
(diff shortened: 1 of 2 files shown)


synapseclient/models/mixins/table_components.py
```diff
@@ -495,13 +495,18 @@ async def store_async(
         client = Synapse.get_client(synapse_client=synapse_client)
 
         if self.include_default_columns:
+            view_type_mask = None
+            if self.view_type_mask:
+                if isinstance(self.view_type_mask, ViewTypeMask):
+                    view_type_mask = self.view_type_mask.value
+                else:
+                    view_type_mask = self.view_type_mask
+
```
BryanFauble, line 504: These changes crept in, I didn't intend for them to show up here, but it's fine that they are.

In some cases users wanting to use a single value here can use the enum, but in the case of `FileView` a user may select 2 or more different masks to use. In that case it's an int instead of the Enum.

BWMac (approved):
LGTM, thanks!

PR #1181, opened 2025-03-11
[SYNPY-1575] Introduce EntityView model

BryanFauble:
**Problem:**

1. The `FileView` aka `EntityView` was not created in an OOP fashion.
2. There are a number of small bugs that I found while implementing this feature related to the work done in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1175 and more generally the Table refactor: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1151

**Solution:**

1. Creating a EntityView model based on this Synapse object: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/EntityView.html
2. Support the functionality for Views to be able to interact with the EntityViews
3. Creating a tutorial and doc pages for the EntityView
4. Patching an issue with re-ordering of Columns that I had found in testing. Essentially the problem is what when an existing View was stored the default columns were being ordered to the front of the table. To patch this issue I moved where default columns are being added to the `.columns` attribute to be after we get the existing object state from Synapse.
5. Moving the async job to use the TQDM progress bar to prevent it from being printed out the console.
6. Creating method wrappers in the EntityView class to allow for specific examples for that class type to be shown.

**Testing:**

1. I wrote a number of integration tests to verify the async and non-async functionality.
2. I wrote a tutorial that showcases how to use this EntityView model
3. I ran through a significant amount of testing during development to verify functionality to creating, updating, adding, removing, data and columns


synapseclient/models/fileview.py
```diff
@@ -0,0 +1,1104 @@
...
+    ViewSnapshotMixin,
+    ViewStoreMixin,
+    ViewUpdateMixin,
+)
+from synapseclient.models.table_components import Column
+
+DATA_FRAME_TYPE = TypeVar("pd.DataFrame")
+
+
+class FileViewSynchronousProtocol(Protocol):
```
BWMac, line 32: I wasn't planning on Dataset having a sync protocol as we discussed in Slack, should I switch to follow this pattern?

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -399,54 +401,63 @@ async def get_job_async(
...
-            )
+                progressed = (
+                    job_status.progress_message != last_message
+                    or last_progress != job_status.progress_current
+                )
+                if progress_tracking and progressed:
+                    progress_bar.update(job_status.progress_current - last_progress)
+                    last_message = job_status.progress_message
+                    last_progress = job_status.progress_current
+                    last_total = job_status.progress_total
```
BWMac, line 443: Is `last_total` still used anywhere?

docs/tutorials/python/file_view.md
```diff
@@ -1,4 +1,138 @@
-# File Views
-[See the current available tutorial](../python_client.md#file-views)
+# FileViews
```
BWMac, line 1: NIT: Should the files with information/scripts for FileViews be named like `fileview` or `file_view`? Should that be kept consistent?

synapseclient/models/mixins/table_components.py
```diff
@@ -1908,8 +1941,164 @@ async def update_rows_async(
         )
 
 
+class QueryMixinSynchronousProtocol(Protocol):
```
BWMac, line 1944: Why do we need sync protocols for query and snapshots?

BWMac (commented):
Do there need to be unit tests for FileViews, Datasets, etc. specifically?

synapseclient/models/mixins/table_components.py
BryanFauble: In this case `query` didn't have Class specific examples in it. By having a sync protocol for it that is used in the Mixin we don't need to do the work of creating a specific example for each class that is using it.


This is the example code:
```
            from synapseclient.models import query
            syn = Synapse()
            syn.login()
            results = query(query="SELECT * FROM syn1234")
            print(results)
```

The one thing this doesn't handle for, and isn't well advertised, is that the `query` function may also be used in the Context of your class instance like:


```
from synapseclient.models import Table # OR Dataset, FileView
syn = Synapse()
syn.login()

table = Table(id="syn1234").get()
results = table .query(query="SELECT * FROM syn1234")
print(results)
```

synapseclient/api/entity_factory.py
```diff
@@ -363,6 +366,14 @@ class type. This will also download the file if `download_file` is set to True.
         entity = entity_to_update.fill_from_dict(
             entity=entity_bundle["entity"], set_annotations=False
         )
+    elif entity["concreteType"] == concrete_types.ENTITY_VIEW:
+        if not entity_to_update:
+            from synapseclient.models.fileview import FileView
```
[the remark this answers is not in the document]
BryanFauble: Good point. In the currently client the class is [EntityViewSchema](https://python-docs.synapse.org/tutorials/python_client/?h=fileview#file-views),. 

I'll post a short poll in Slack and get a vote on this for what these new model should be called.

synapseclient/models/fileview.py
```diff
@@ -0,0 +1,1104 @@
...
+            else OrderedDict()
+        )
+        self._last_persistent_instance.annotations = (
+            deepcopy(self.annotations) if self.annotations else {}
+        )
+        self._last_persistent_instance.scope_ids = (
+            deepcopy(self.scope_ids) if self.scope_ids else set()
+        )
+
+    def fill_from_dict(self, entity: Dict, set_annotations: bool = True) -> "FileView":
```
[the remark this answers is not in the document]
BryanFauble: We can probably generalize a portion of this since there are some common fields on the Entity that get shared. Not everything can be generalized though. In tables there are `columnIds`, but not on file entities. On Views there is `scopeIds`, but not on datasets. The exceptions are what makes it annoyingly complex.

BryanFauble:
> Do there need to be unit tests for FileViews, Datasets, etc. specifically?

Good point. I'll add something similar to [this unit test functionality](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/synpy-1575-fileview-updates/tests/unit/synapseclient/models/unit_test_table.py) that I had written for the Table.

synapseclient/models/mixins/table_components.py
BryanFauble: In the case of `snapshot`, I need to add examples there in the docstring, and I should probably add this to this `FileViewSynchronousProtocol` instead of having a `ViewSnapshotMixinSynchronousProtocol` for this.

synapseclient/models/fileview.py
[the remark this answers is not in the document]
BWMac: We could probably have some re-useable logic that will convert the camel case keys in an API response to snake case for the class attributes. That way you can reuse that function across classes as long as the naming is consistent. One issue with that is that some of our OOP models are actually a combination of 2 or more API response models. For example, the asynchronous job functionality classes have request AND response attributes. So not all classes are filled in the same way at a level higher than just what keys we expect too.

docs/tutorials/python/file_view.md
```diff
@@ -1,4 +1,138 @@
...
+FileViews in Synapse allow you to create a virtual table that provides a unified view
```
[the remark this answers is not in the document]
BryanFauble: Updated in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1181/commits/6e593f81f591d54033118cc65dbb43679ed6994a

synapseclient/models/mixins/table_components.py
BryanFauble: Updated in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1181/commits/6e593f81f591d54033118cc65dbb43679ed6994a

synapseclient/models/mixins/asynchronous_job.py
BryanFauble: Updated in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1181/commits/6e593f81f591d54033118cc65dbb43679ed6994a

synapseclient/api/entity_factory.py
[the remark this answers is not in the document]
thomasyu888: The decision was to move towards EntityView. Thanks for posting that on slack!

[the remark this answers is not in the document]
BryanFauble: Updated in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1181/commits/d040c13389b6942d2b2ebf2aa096f9293ebaa422

BryanFauble:
@BWMac This could use a final review when you get a chance this week. Thanks in advance!

tests/integration/synapseclient/integration_test.py
```diff
@@ -86,7 +86,6 @@ async def testCustomConfigFile(schedule_for_cleanup):
         )
 
 
-@pytest.mark.flaky(reruns=3, only_rerun=["SynapseHTTPError"])
```
thomasyu888, line 89: Interesting, how have these been resolved?

tests/integration/synapseclient/models/async/test_entityview_async.py
```diff
@@ -0,0 +1,1202 @@
...
+        assert (
+            new_entityview_instance.columns["test_column2"].column_type
+            == ColumnType.INTEGER
+        )
+        assert len(new_entityview_instance.columns) == 2
+
+    async def test_create_entityview_with_invalid_column(
+        self, project_model: Project
+    ) -> None:
+        # GIVEN a entityview with an invalid column
```
thomasyu888, line 160: I may have missed this, but what happens when an entity view is created with the default columns specified?

thomasyu888 (approved):
🔥 LGTM! Awesome to see this work, I'll defer to brad for final review once he's back from vacay.

tests/integration/synapseclient/integration_test.py
BryanFauble: It has not been explicitly, but a global retry mechanism was put in place after this was written.

PR #1185, opened 2025-03-21
[SYNPY-1553] Removes Blank Auth Header

BWMac:
# **Problem:**

In the Synapse Python Client users should be able to start an Anonymous session like:
```
import synapseclient

syn = synapseclient.Synapse()
syn.login(authToken="")
```
However, `httpx` does not automatically handle cases where an empty authentication token is provided. So, currently users will be met with an error like:
```
httpx.LocalProtocolError: Illegal header value b'Bearer '
```
when they attempt to start an Anonymous session.

# **Solution:**

To fix this, need to not add the Authentication header to our request if the user has not provided a Synapse Authentication Token when using `syn.login()`. 

We also want to catch cases like these and inform users what they should do next after attempting to `get` an entity without having provided an authentication token.

# **Testing:**

I added a unit test to cover the new warning message in `_check_entity_restrictions` and made some small changes to other tests to accommodate this update.

**Note**: `warnings.warn` is deprecated in Python 3.13. I updated `_check_entity_restrictions` to use `logger.warning` instead of `.warn`


synapseclient/api/entity_factory.py
```diff
@@ -400,12 +400,19 @@ def _check_entity_restrictions(
     if restriction_information and restriction_information.get(
         "hasUnmetAccessRequirement", None
     ):
-        warning_message = (
-            "\nThis entity has access restrictions. Please visit the web page for this entity "
-            f'(syn.onweb("{synapse_id}")). Look for the "Access" label and the lock icon underneath '
-            'the file name. Click "Request Access", and then review and fulfill the file '
-            "download requirement(s).\n"
-        )
+        if not syn.credentials._token:
```
BryanFauble, line 403: Could we also check for `not syn.credentials`?

synapseclient/client.py
```diff
@@ -1445,17 +1445,26 @@ def _check_entity_restrictions(
...
-                "\nThis entity has access restrictions. Please visit the web page for this entity "
-                f'(syn.onweb("{id_of(entity)}")). Look for the "Access" label and the lock icon underneath '
-                'the file name. Click "Request Access", and then review and fulfill the file '
-                "download requirement(s).\n"
-            )
+        restriction_information = bundle.get("restrictionInformation", None)
+        if restriction_information and restriction_information.get(
+            "hasUnmetAccessRequirement", None
+        ):
+            if not self.credentials._token:
```
BryanFauble, line 1452: Check that self.credentials is not None

tests/unit/synapseclient/unit_test_client.py
```diff
@@ -1157,121 +1157,146 @@ def expected_request_json(token) -> str:
...
+                    "id": "syn123",
+                    "name": "anonymous",
+                    "concreteType": "org.sagebionetworks.repo.model.FileEntity",
+                    "parentId": "syn12345",
+                },
+                "entityType": "folder",
+                "restrictionInformation": {"hasUnmetAccessRequirement": True},
+            }
+            entity = "syn123"
+            self.syn.credentials = SynapseAuthTokenCredentials(token="")
```
BryanFauble, line 1243: Would be worthwhile to have a test where login wasn't called at all, or syn.credentials is None

BryanFauble (changes requested):

synapseclient/api/entity_factory.py
BWMac: no, `syn.credentials` is not None in these cases, but `token` (exposed as `_token`) is an empty string.

```
SynapseAuthTokenCredentials(username='anonymous', displayname='Anonymous', token='', owner_id='273950')
```

BWMac: Or did you mean do both checks?

BWMac: added that check and a test in my latest [commit](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1185/commits/a9d7bd47382d3c3753b8c5bca5514063df202f40)

BryanFauble (approved):
LGTM! Thanks

PR #1187, opened 2025-04-01
[SYNPY-1575] Correct regular expression for invalid column name

BryanFauble:
# **Problem:**

- The regular expression used to verify the names of the column will meet Synapse requirements didn't account for everything that was present in the exception message.

# **Solution:**

- Updating the regular expression and running a unit test with additional values

# **Testing:**

- Unit test to verify regular expression is picking up valid/invalid names


BWMac (approved):
LGTM!

PR #1188, opened 2025-04-02
Include a default timeout for HTTP requests library to 70 seconds

BryanFauble:
# **Problem:**

- Integration tests can hang for reasons that I do not understand, and they will only timeout after the 6 hour limit is finally reached.
- Based on [this documentation](https://requests.readthedocs.io/en/latest/user/advanced/#timeouts): 

> Most requests to external servers should have a timeout attached, in case the server is not responding in a timely manner. By default, requests do not time out unless a timeout value is set explicitly. Without a timeout, your code may hang for minutes or more.

Due to this what I suspect is occurring:

1. An HTTP request is sent to Synapse
2. A connection error occurs, but a message is not sent back to the SYNPY client code
3. The SYNPY client code hangs forever because it did not get a message back, AND, it never hits a timeout because there is no timeout set

# **Solution:**

- Set a default 70 second timeout for the code with an option for users to set a different value. This 70 second timeout is what I had already been using for the HTTPX library

# **Testing:**

- Will be reviewing Integration test runs in this PR

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -248,7 +248,6 @@ class Synapse(object):
         authEndpoint:          Location of authentication service
         fileHandleEndpoint:    Location of file service
         portalEndpoint:        Location of the website
-        serviceTimeoutSeconds: Wait time before timeout (currently unused)
         debug:                 Print debugging messages if True
         skip_checks:           Skip version and endpoint checks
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 LGTM!

PR #1191, opened 2025-04-14
[SYNPY-1497] refactored version check to use Pypi for version info

andrewelamb:
[SYNPY-1497]

# **Problem:**
When checking the Python version, `synapseclient.core.version_check.version_check` GitHub was being used to find the latest version. This caused a problem when a version had been pushed into the master branch, but hadn't been updated in Pypi yet.

In addition, while working on this, I found a bug. To compare package versions, the code converts the current and latest versions by spliting them into tuples where each version level is a string. For example, the current version of the client is "4.7.0". This gets converted into a tuple like: `("4", "7", "0')` . These tuples then get comapred: `("4", "6", "9') < ("4", "7", "0')` When tuple comparisons happen the first elements of each are compared, and if those are equal then second elements are compared and so on. 

The problem is that the comparisons are done as strings. This works fine if they are single digits: `"4" < "5"`. But this would be a problem if the version number was in the tens: `"10" < "5"`.

# **Solution:**
1. The `version_control` function now uses a Pypi link rather than Github to get version information.
2. When doing version comparisons the strings are converted to ints first.
3. There was some refactoring and improvement of docstrings in several functions.

# **Testing:**
New testing file: `tests/unit/synapseclient/core/unit_test_version_check.py`
All functions modified or added have tests here now, and some were moved here

New testing file: `tests/integration/synapseclient/core/test_version_check.py`
Tests for `version check` were moved here and simplified.

[SYNPY-1497]: https://sagebionetworks.jira.com/browse/SYNPY-1497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -821,7 +821,7 @@ class Synapse(object):
 
         # Check version before logging in
         if not self.skip_checks:
-            version_check()
+            version_check(self.logger)
 
         # Make sure to invalidate the existing session
         self.logout()
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -821,7 +821,7 @@ class Synapse(object):
 
         # Check version before logging in
         if not self.skip_checks:
-            version_check(self.logger)
+            version_check(logger=self.logger)
 
         # Make sure to invalidate the existing session
         self.logout()
--- a/synapseclient/core/version_check.py
+++ b/synapseclient/core/version_check.py
@@ -15,82 +15,172 @@ import importlib.resources
 import json
 import re
 import sys
+import urllib.request
+from typing import Optional
 
 import requests
 
 import synapseclient
 
 _VERSION_URL = "https://raw.githubusercontent.com/Sage-Bionetworks/synapsePythonClient/master/synapseclient/synapsePythonClient"  # noqa
+_PYPI_JSON_URL = "https://pypi.org/pypi/synapseclient/json"
+_RELEASE_NOTES_URL = "https://python-docs.synapse.org/news/"
 
 
 def version_check(
-    current_version=None, version_url=_VERSION_URL, check_for_point_releases=False
-):
+    current_version: Optional[str] = None,
+    check_for_point_releases: bool = False,
+    use_local_metadata: bool = False,
+) -> bool:
     """
     Gets the latest version information from version_url and check against the current version.
     Recommends upgrade, if a newer version exists.
 
-    Arguments:
-        current_version: The current version of the entity
-        version_url: The URL of the entity version
-        check_for_point_releases: Bool.
+    This wraps the _version_check function in a try except block.
+    The purpose of this is so that no exception caught running the version check stop the client from running.
+
+    Args:
+        current_version (Optional[str], optional): The current version of the package.
+          Defaults to None.
+          This argument is mainly used for testing.
+        check_for_point_releases (bool, optional):
+          Defaults to False.
+          If True, The whole package versions will be compared (ie. 1.0.0)
+          If False, only the major and minor package version will be compared (ie. 1.0)
+        use_local_metadata (bool, optional):
+          Defaults to False.
+          If True, importlib.resources will be used to get the latest version fo the package
+          If False, the latest version fo the package will be taken from Pypi
 
     Returns:
-        True if current version is the latest release (or higher) version, otherwise False.
+        bool: True if current version is the latest release (or higher) version, otherwise False.
     """
-
     try:
-        if not current_version:
-            current_version = synapseclient.__version__
-
-        version_info = _get_version_info(version_url)
-
-        current_base_version = _strip_dev_suffix(current_version)
-
-        # Check blacklist
-        if (
-            current_base_version in version_info["blacklist"]
-            or current_version in version_info["blacklist"]
+        if not _version_check(
+            current_version, check_for_point_releases, use_local_metadata
         ):
-            msg = (
-                "\nPLEASE UPGRADE YOUR CLIENT\n\nUpgrading your SynapseClient is"
-                " required. Please upgrade your client by typing:\n    pip install"
-                " --upgrade synapseclient\n\n"
-            )
-            raise SystemExit(msg)
-
-        if "message" in version_info:
-            sys.stderr.write(version_info["message"] + "\n")
-
-        levels = 3 if check_for_point_releases else 2
-
-        # Compare with latest version
-        if _version_tuple(current_version, levels=levels) < _version_tuple(
-            version_info["latestVersion"], levels=levels
-        ):
-            sys.stderr.write(
-                "\nUPGRADE AVAILABLE\n\nA more recent version of the Synapse Client"
-                " (%s) is available. Your version (%s) can be upgraded by typing:\n   "
-                " pip install --upgrade synapseclient\n\n"
-                % (
-                    version_info["latestVersion"],
-                    current_version,
-                )
-            )
-            if "releaseNotes" in version_info:
-                sys.stderr.write(
-                    "Python Synapse Client version %s release notes\n\n"
-                    % version_info["latestVersion"]
-                )
-                sys.stderr.write(version_info["releaseNotes"] + "\n\n")
             return False
 
     except Exception as e:
         # Don't prevent the client from running if something goes wrong
-        sys.stderr.write("Exception in version check: %s\n" % (str(e),))
+        sys.stderr.write(f"Exception in version check: {str(e)}\n")
         return False
 
     return True
+
+
+def _version_check(
+    current_version: Optional[str] = None,
+    check_for_point_releases: bool = False,
+    use_local_metadata: bool = False,
+) -> bool:
+    """
+    Gets the latest version information from version_url and check against the current version.
+    Recommends upgrade, if a newer version exists.
+
+    This has been split of from the version_check function to make testing easier.
+
+    Args:
+        current_version (Optional[str], optional): The current version of the package.
+          Defaults to None.
+          This argument is mainly used for testing.
+        check_for_point_releases (bool, optional):
+          Defaults to False.
+          If True, The whole package versions will be compared (ie. 1.0.0)
+          If False, only the major and minor package version will be compared (ie. 1.0)
+        use_local_metadata (bool, optional):
+          Defaults to False.
+          If True, importlib.resources will be used to get the latest version fo the package
+          If False, the latest version fo the package will be taken from Pypi
+
+    Returns:
+        bool: True if current version is the latest release (or higher) version, otherwise False.
+    """
+    if not current_version:
+        current_version = synapseclient.__version__
+    assert isinstance(current_version, str)
+
+    if use_local_metadata:
+        metadata = _get_local_package_metadata()
+        latest_version = metadata["latestVersion"]
+        assert isinstance(latest_version, str)
+    else:
+        latest_version = _get_version_info_from_pypi()
+
+    levels = 3 if check_for_point_releases else 2
+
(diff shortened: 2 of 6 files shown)


synapseclient/core/version_check.py
```diff
@@ -15,84 +15,176 @@
...
 import requests
 
 import synapseclient
 
 _VERSION_URL = "https://raw.githubusercontent.com/Sage-Bionetworks/synapsePythonClient/master/synapseclient/synapsePythonClient"  # noqa
+_PYPI_JSON_URL = "https://pypi.org/pypi/synapseclient/json"
+_RELEASE_NOTES_URL = "https://python-docs.synapse.org/news/"
 
 
 def version_check(
```
BryanFauble, line 30: When `version_check` is being called from within the client can you swap this logic over to use the `logger` to print out info level messages instead of `print` and `sys.xxx.write` messages?

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8663e459b321270b65a0305c5126b0e955648fc9/synapseclient/core/utils.py#L1044-L1049 is similar logic I've added elsewhere so that we can maintain backwards compatibility in the event a user is calling this function.

synapseclient/core/version_check.py
```diff
@@ -15,84 +15,176 @@
...
     Recommends upgrade, if a newer version exists.
 
-    Arguments:
-        current_version: The current version of the entity
-        version_url: The URL of the entity version
-        check_for_point_releases: Bool.
+    This wraps the _version_check function in a try except block.
+    The purpose of this is so that no exception caught running the version check stops the client from running.
+
+    Args:
```
BryanFauble, line 42:
```suggestion
    Arguments:
```

synapseclient/core/version_check.py
```diff
@@ -15,84 +15,176 @@
...
+        current_version (Optional[str], optional): The current version of the package.
+          Defaults to None.
+          This argument is mainly used for testing.
+        check_for_point_releases (bool, optional):
+          Defaults to False.
+          If True, The whole package versions will be compared (ie. 1.0.0)
+          If False, only the major and minor package version will be compared (ie. 1.0)
+        use_local_metadata (bool, optional):
+          Defaults to False.
+          If True, importlib.resources will be used to get the latest version fo the package
```
BryanFauble, line 52: nit: The typing in the docstring is not needed. The type is implied based on the typing in the function signature.

synapseclient/core/version_check.py
```diff
@@ -15,84 +15,176 @@
...
     return True
 
 
+def _get_version_info_from_pypi() -> str:
+    """Gets the current release version from PyPi
+
+    Returns:
+        str: The current release version
+    """
+    with urllib.request.urlopen(_PYPI_JSON_URL) as url:
```
BryanFauble, line 124: Swap this over to use the `httpx` library instead. I'd like for everything to funnel through that library.

https://www.python-httpx.org/quickstart/

synapseclient/core/version_check.py
```diff
@@ -15,84 +15,176 @@
...
 import re
 import sys
+import urllib.request
+from typing import Optional
 
 import requests
 
 import synapseclient
 
 _VERSION_URL = "https://raw.githubusercontent.com/Sage-Bionetworks/synapsePythonClient/master/synapseclient/synapsePythonClient"  # noqa
```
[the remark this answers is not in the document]
andrewelamb: > Are we able to completely get rid of needing to use the github link for everything?
@BryanFauble I'm happy to another run through and try to do so. I was only doing it for the `check_version` function.

synapseclient/core/version_check.py
andrewelamb: yep!

synapseclient/core/version_check.py
```diff
@@ -148,13 +240,32 @@ def _strip_dev_suffix(version):
     return re.sub(r"\.dev\d+", "", version)
 
 
-def _version_tuple(version, levels=2):
+def _version_tuple(version: str, levels: int = 2) -> tuple:
```
BryanFauble, line 243: Return type should be a bit more specific here:

```
from typing import Tuple

.
.
.
.
-> Tuple[x, y, z]
```

synapseclient/core/version_check.py
```diff
@@ -163,16 +274,37 @@ def _version_tuple(version, levels=2):
     return tuple(v)
 
 
-def _get_version_info(version_url=_VERSION_URL):
+def _get_version_info(version_url: Optional[str] = _VERSION_URL) -> dict:
```
BryanFauble, line 277: Is there a more specific Dictionary we can update the return type to? Like a Dict[str, str]?

tests/integration/synapseclient/core/test_version_check.py
```diff
@@ -0,0 +1,20 @@
...
+async def test_version_check():
+    """Integration checks for version_check"""
+    # Check current version against pypi version file
+    version_check()
+
+    # Should be higher than current version and return true
+    assert version_check(current_version="999.999.999")
+
+    # Test out of date version
+    assert not version_check(current_version="0.0.1")
```
BryanFauble, line 15: Could you add one additional item to the assertions to verify that an HTTP call was fired off to pypi to get the version?
It'll be similar to how these SPYs work: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8663e459b321270b65a0305c5126b0e955648fc9/tests/integration/synapseclient/core/test_download.py#L98-L99

BryanFauble (changes requested):
Thanks for the first round of changes. I threw some comments in. When your ready for another review re-request a review and I can take a look!

synapseclient/core/version_check.py
andrewelamb: The way I understand typing and tuples is that `Tuple[str, str]` implies the return type is two strings, and `Tuple[str, str, str]` is three strings. Since the function can return a `Tuple` of any amount of strings, I can't be more specific.

synapseclient/core/version_check.py
andrewelamb: Done

synapseclient/core/version_check.py
andrewelamb: Done

synapseclient/core/version_check.py
andrewelamb: Done

synapseclient/core/version_check.py
```diff
@@ -148,13 +240,32 @@ def _strip_dev_suffix(version):
...
         print(version_tuple('0.5.1.dev1', levels=2))
         ('0', '5')
+
+    First the version string is split into version levels.
+    If the number of levels is greater than the levels argument(x),
+      only x levels are returned.
+    If the number of levels is lesser than the levels argument(x),
+      "0" strings are used to pad out the return value.
+
+    Args:
```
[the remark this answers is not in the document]
andrewelamb: Done

synapseclient/core/version_check.py
andrewelamb: This function was removed

synapseclient/core/version_check.py
```diff
@@ -148,13 +240,32 @@ def _strip_dev_suffix(version):
...
+        version (str): A package version in string form such as "1.0.0"
+        levels (int, optional):
+          Defaults to 2.
+          The number of levels deep in the package version to return. "1.0.0", for example:
+            levels=1: only the major version ("1")
+            levels=2: the major and minor version ("1", "0")
+            levels=2: the major, minor, and patch version ("1", "0", "0")
```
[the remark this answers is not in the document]
andrewelamb: Dones

tests/integration/synapseclient/core/test_version_check.py
andrewelamb: Is this enough?

```
    # Assert _get_version_info_from_pypi called  when running version_check
    spy = mocker.spy(synapseclient.core.version_check, "_get_version_info_from_pypi")
    version_check()
    spy.assert_called_once()
```
The test you linked has quite a bit going on, but it mostly concerns checking the downloaded file.

BryanFauble: In this case you'd want to add a `mocker.spy` for the `httpx.get` function call. The verification would be to make sure that the HTTP URL you are expecting (The PYPI URL) was called.

BryanFauble:
@andrewelamb For any comments that have been fixed please go ahead and mark the conversation as resolved.

synapseclient/core/version_check.py
```diff
@@ -11,150 +11,274 @@
...
-                    "Python Synapse Client version %s release notes\n\n"
-                    % version_info["latestVersion"]
-                )
-                sys.stderr.write(version_info["releaseNotes"] + "\n\n")
             return False
 
     except Exception as e:
         # Don't prevent the client from running if something goes wrong
-        sys.stderr.write("Exception in version check: %s\n" % (str(e),))
+        sys.stderr.write(f"Exception in version check: {str(e)}\n")
```
BryanFauble, line 68: Should have a check for:


```
if logger:
# use `logger.exception("Message without exception since it'll automatically be included")
else:
# sys write
```

synapseclient/core/version_check.py
```diff
@@ -11,150 +11,274 @@
...
+_RELEASE_NOTES_URL = "https://python-docs.synapse.org/news/"
 
 
 def version_check(
-    current_version=None, version_url=_VERSION_URL, check_for_point_releases=False
-):
+    current_version: Optional[str] = None,
+    check_for_point_releases: bool = False,
+    use_local_metadata: bool = False,
+    logger: logging.Logger = None,
```
BryanFauble, line 33: Now that `logger` can be passed into this function, we'll need to pass the logger from where it's being called in the Synapse class

synapseclient/core/version_check.py
andrewelamb: Done

synapseclient/core/version_check.py
andrewelamb: Done

tests/integration/synapseclient/core/test_version_check.py
andrewelamb: Like:

```
    # Assert httpx.get called  when running version_check
    spy = mocker.spy(httpx, "get")
    version_check()
    spy.assert_called_once()
```

BryanFauble: Yeah. The only thing is that you'd use `assert_called_once_with`

synapseclient/client.py
```diff
@@ -821,7 +821,7 @@ def login(
 
         # Check version before logging in
         if not self.skip_checks:
-            version_check()
+            version_check(self.logger)
```
thomasyu888, line 824: What is wrong with the way you're passing in this variable? Is it being passed into logger?

tests/integration/synapseclient/core/test_version_check.py
andrewelamb: done

synapseclient/client.py
andrewelamb: fixed

andrewelamb:
@BryanFauble Could I get a re-review? Thanks!

BryanFauble (approved):
LGTM! Once all tests are passing it should be ok to merge

thomasyu888 (approved):
🔥 LGTM! Thanks for the work and review. @andrewelamb in the future, I'd recommend following Bryan's suggestion of just "resolve conversation" to indicate that you've addressed an issue.

PR #1193, opened 2025-04-21
[SYNPY-1584] Add a buffer.tell when truncating bytes during dataframe upload and drop writing header to csv

BryanFauble:
# **Problem:**

- In the loop over the Pandas DataFrame to return the bytes for upload into Synapse -- a `.tell()` was not performed after we dropped the X number of bytes from the BytesIO buffer
- When the first chunk of the DataFrame was being written the bytes in the buffer from the header of the DataFrame was causing data to be dropped (Sometimes silently) from the end of the first chunk. In one case there was around 20 missing bytes from a large-text column.
- Bytes from the header were getting skipped at the front of the very first DataFrame

# **Solution:**

- After dropping bytes from the front of the buffer call `.tell()` to re-read `number_of_bytes_in_buffer`
- Removing writing the header to the buffer, instead use the `bytes_to_prepend` argument instead
- Don't skip bytes in the offset for the header, instead let the first chunk be larger to include the column names

# **Testing:**

- I was able to run the scripts that @GiaJordan and @thomasyu888 provided in https://sagebionetworks.jira.com/browse/SYNPY-1565 to verify on an EC2 instance in the service catalog that I could insert 500 rows of data as expected and have it show up in Synapse

### DIFF
--- a/synapseclient/core/upload/upload_utils.py
+++ b/synapseclient/core/upload/upload_utils.py
@@ -89,6 +89,7 @@ def get_partial_dataframe_chunk(
             buffer.truncate(0)
             buffer.write(copy_of_data)
             total_offset = 0
+            number_of_bytes_in_buffer = buffer.tell()
 
         if number_of_bytes_in_buffer >= max_bytes_to_read:
             # Return maximum number of bytes that can be read from the buffer
--- a/synapseclient/core/upload/upload_utils.py
+++ b/synapseclient/core/upload/upload_utils.py
@@ -55,7 +55,6 @@ def get_partial_dataframe_chunk(
         (total_size_of_chunks_being_uploaded - ((part_number - 1) * part_size)),
         part_size,
     )
-    header_written = False
     # TODO: This is an area for optimization. It is possible to avoid writing the entire
     # dataframe to a buffer and then reading the buffer to get the bytes. Instead, we
     # might be able to do something like keeping markers at each 100 row increment how
@@ -68,12 +67,11 @@ def get_partial_dataframe_chunk(
         df.iloc[offset_start:end].to_csv(
             buffer,
             mode="a",
-            header=(part_number == 1 and not header_written),
+            header=False,
             index=False,
             float_format="%.12g",
             **(to_csv_kwargs or {}),
         )
-        header_written = True
         number_of_bytes_in_buffer = buffer.tell()
         # Drop data from the front of the buffer until total_offset is 0
         if total_offset > 0 and total_offset >= number_of_bytes_in_buffer:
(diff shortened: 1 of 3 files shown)


synapseclient/core/upload/upload_utils.py
```diff
@@ -68,12 +67,11 @@ def get_partial_dataframe_chunk(
         df.iloc[offset_start:end].to_csv(
             buffer,
             mode="a",
-            header=(part_number == 1 and not header_written),
+            header=False,
```
linglp, line 70: is here where we ensure that the first chunk has header?

BryanFauble: Here is where the chunk is getting the first header:

Reading the bytes of the reader:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/70f6965c5c2204213d43b7751d7909bfeb86bdff/synapseclient/models/mixins/table_components.py#L3492-L3494


Passing the line as the header:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/70f6965c5c2204213d43b7751d7909bfeb86bdff/synapseclient/models/mixins/table_components.py#L3573

Bytes that are being prepended to the file:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/70f6965c5c2204213d43b7751d7909bfeb86bdff/synapseclient/models/mixins/table_components.py#L3263

Those bytes being passed into this function call:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/a64023dbd189af4eb5f362c164b06a0d08a8ee94/synapseclient/core/upload/multipart_upload_async.py#L698

The bytes being added as the header of the file:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/a64023dbd189af4eb5f362c164b06a0d08a8ee94/synapseclient/core/upload/upload_utils.py#L47-L49

BryanFauble:
> 🔥 LGTM - do we need to mock this behavior in a test? Not sure how much time an integration test of that size would add to the overall testing time.

@thomasyu888 

The issue is that Synapse has minimum file size limits on what can be uploaded. It's either the size of the file, or 5 MB, whichever is smaller. In this case we would need to have a file produced that is larger than 5 MB (https://github.com/Sage-Bionetworks/synapsePythonClient/blob/70f6965c5c2204213d43b7751d7909bfeb86bdff/synapseclient/core/upload/multipart_upload_async.py#L128 ) and upload it in 2 or more parts to be affected by this issue. My goal in (https://sagebionetworks.jira.com/browse/SYNPY-1484 ) is that each test will use a maximum of 10 KB files. That means this test would be much larger than that maximum.

What I am doing to make sure that there is something that covers this is that I created an integration test that I ran locally to verify it works (But it will not run as a part of the CI pipeline due to size):

https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1193/commits/f206e2b9a935bbd96657626fcab770b6f4fce05b

PR #903, opened 2022-03-01
[SYNPY-1209] Support materialized views

thomasyu888:
Supports materialized views as described here: https://sagebionetworks.jira.com/wiki/spaces/PLFM/pages/2514321554/Synapse+Joins

Heres how you test this branch:

1. Create a new Python virtualenv
2. Install this specific branch
    ```
    pip install pandas
    pip install https://github.com/thomasyu888/synapsePythonClient/archive/refs/heads/synpy-1209-materializedviews.zip
    ```
3. Run code with replacements of synapse ids and join
    ```
    from synapseclient.table import MaterializedViewSchema
    import synapseclient
    syn = synapseclient.Synapse()
    # no need to specify staging endpoints anymore
    # syn.setEndpoints(**synapseclient.client.STAGING_ENDPOINTS)
    syn = synapseclient.login()
    
    temp = MaterializedViewSchema(
        name="test-material-view",
        parent="syn34234",
        definingSQL="SELECT * FROM syn111 F JOIN syn2222 P on (F.PATIENT_ID = P.PATIENT_ID)"
    )
    ent = syn.store(temp)
    
    temp = syn.tableQuery(f"select * from {ent.id}")
    temp.asDataFrame()
    # Clean up materialized view
    syn.delete(ent)
    ```
4. Check in the UI.

PR #1195, opened 2025-04-23
[SYNPY-1580] Adds `VirtualTable` OOP Model

BWMac:
# **Problem:**

We did not previously have an OOP model representing the Synapse [VirtualTable](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/VirtualTable.html). We should follow the example of the MaterializedView to implement this model, with the exception that JOIN and UNION operations are not supported in the `definingSQL` of `VirtualTable`s.

# **Solution:**

Implemented the `VirtualTable` model, all expected functionality, updated documentation, added a tutorial and examples. Followed #903 as an example because there is a lot of overlap in functionality with `MaterializedView`s.

# **Testing:**

Tested implementation by manually running the code, wrote integration tests and unit tests as needed.

# **Notes:**

There was a compatibility issue causing failures in our CI pipeline with older versions of `pip` so I added a `pip upgrade` line to the build file https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1195/commits/b5dd74657dbd7f2365c26d40ab5dbb77dbf42db4

### DIFF
--- a/synapseclient/core/exceptions.py
+++ b/synapseclient/core/exceptions.py
@@ -81,10 +81,6 @@ class SynapseUploadFailedException(SynapseError):
     """Raised when an upload failed. Should be chained to a cause Exception"""
 
 
-class SynapseQueryError(SynapseError):
-    """Raised when an invalid SQL query is passed to a Synapse Table-like object"""
-
-
 def _get_message(response: httpx.Response, logger: logging.Logger) -> Union[str, None]:
     """Extracts the message body or a response object by checking for a json response
     and returning the reason otherwise getting body.
--- a/synapseclient/core/exceptions.py
+++ b/synapseclient/core/exceptions.py
@@ -81,6 +81,10 @@ class SynapseUploadFailedException(SynapseError):
     """Raised when an upload failed. Should be chained to a cause Exception"""
 
 
+class SynapseQueryError(SynapseError):
+    """Raised when an invalid SQL query is passed to a Synapse Table-like object"""
+
+
 def _get_message(response: httpx.Response, logger: logging.Logger) -> Union[str, None]:
     """Extracts the message body or a response object by checking for a json response
     and returning the reason otherwise getting body.
--- a/synapseclient/models/virtualtable.py
+++ b/synapseclient/models/virtualtable.py
@@ -1,3 +1,4 @@
+import re
 from collections import OrderedDict
 from copy import deepcopy
 from dataclasses import dataclass, field, replace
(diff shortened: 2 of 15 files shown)


synapseclient/models/virtualtable.py
```diff
@@ -0,0 +1,627 @@
...
+                )
+                virtual_table = await virtual_table.store_async()
+                print(f"Created Virtual Table with ID: {virtual_table.id}")
+
+            asyncio.run(main())
+            ```
+        """
+        # Check for unsupported operations in defining_sql
+        if self.defining_sql:
+            if any(clause in self.defining_sql.upper() for clause in ["JOIN", "UNION"]):
```
BryanFauble, line 540: This could cause some issues if `JOIN` or `UNION` is a part of the name of a column or in the sql in an unintended way. Could we swap this around maybe to look if it's a singular word `JOIN` or `UNION`, so something like `WHERE 'my_column' = 'REUNION'` wouldn't get flagged?

BWMac: Thoughts on [this](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1195/commits/c0fa2b00538ff560a7b377ed8de6d570f5930818)?

synapseclient/models/virtualtable.py
```diff
@@ -536,8 +536,19 @@ async def main():
             ```
         """
         # Check for unsupported operations in defining_sql
+
         if self.defining_sql:
-            if any(clause in self.defining_sql.upper() for clause in ["JOIN", "UNION"]):
+            import re
```
BryanFauble, line 541: I don't think this is harmful to be imported here, but this could be moved to import at the top of the module

BryanFauble (approved):
Changes look good to me!

synapseclient/core/exceptions.py
```diff
@@ -81,6 +81,10 @@ class SynapseUploadFailedException(SynapseError):
     """Raised when an upload failed. Should be chained to a cause Exception"""
 
 
+class SynapseQueryError(SynapseError):
```
thomasyu888, line 84: Discussion: Is this just a "ValueError"?

BryanFauble: Tbh our usage of exceptions throughout the client is all over the place. I'm not opposed to this new exception, however, I think your right @thomasyu888 that the way it's being used is as a `ValueError`

synapseclient/models/virtualtable.py
thomasyu888: @BryanFauble , @BWMac 

Do we want to guard against:
```
select * from syn... where column3 = 'A join value'
```

I can't imagine this happening often and i'm ok with current solution so we don't optimize for a scenario that may never occur. Just out of interest, what would the default error from Synapse be when we try to store a virtual table with UNION and/or JOIN

BryanFauble: The error is:


```
synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error: The JOIN keyword is not supported in this context
```

You bring up a good point though about client side/server side validation. If Synapse were to support a JOIN/UNION later on, then we would have several versions of the SYNPY client that don't support the functionality.


Since we have something that works for today I'm OK with this

thomasyu888 (commented):
🔥 LGTM!  I looked at the tests and defer to Bryan's final review.

PR #1197, opened 2025-04-25
[SYNPY-1593] Enforce minimum httpcore dependency to prevent critical CVE

BryanFauble:
# **Problem:**

- CVE: https://github.com/Sage-Bionetworks/synapsePythonClient/security/dependabot/26

# **Solution:**

- Enforce minimum httpcore dependency which has already been patched

# **Testing:**

- Will review integration test run


setup.cfg
```diff
@@ -59,7 +59,8 @@ install_requires =
     opentelemetry-instrumentation-urllib>=0.48b0
     nest-asyncio~=1.6.0
     asyncio-atexit~=1.0.1
-    httpx~=0.27.0
+    httpx>=0.27.0
+    httpcore>=1.0.9
```
danlu1, line 63: LGTM. I found the changelog for httpcore that adapted h11 changes here: https://pypi.org/project/httpcore/#:~:text=Version%201.0.9%20(April,dependency%20update.%20(%231008). Just out of curiosity, how did you set the httpx version?

danlu1 (approved):

setup.cfg
BryanFauble: For the httpx version I think that I had just taken the latest version available when I was writing this.

PR #1198, opened 2025-04-25
Patch integrations tests for multi-profile login 

carmmmm:
# **Problem:**

The original multi-profile implementation introduced changes to authentication logic and CLI support for named profiles via the .synapseConfig file. While the core functionality worked as expected, a number of integration tests were failing due to one of the following:
	•	Tests hardcoded to rely on the [authentication] section only, instead of resolving from [profile <name>], [default], or [authentication].
	•	Some integration tests (test_login, test_storeTable_csv, etc.) failing due to unparsed mixed output or mismatched login arguments (e.g., unexpected profile=None in mocks).
	•	Other tests failing on staging/production due to references to nonexistent users (DPETestUser2) or certification assertions that don’t apply to external accounts.

# **Solution:**

	•	Updated config-parsing lines in tests like test_login to support resolution from [profile <name>], [default], or [authentication], matching the client behavior.
	•	Adjusted mock expectations (e.g., added profile=None) where relevant to prevent assertion mismatches.

# **Testing:**
•	Ran affected test cases individually and confirmed passing (test_login, test_storeTable_csv, test_command_line_using_paths, etc.).
	•	Used SYNAPSE_PROFILE to validate that the .synapseConfig resolution logic works as expected in tests.
	•	Ran pre-commit and fixed formatting issues using black and isort.

Notes on remaining failures:
	•	8 integration test failures related to DPETestUser2 and user certification status (is_certified) which I believe is expected as they are being run outside the internal dev environment.


synapseclient/__main__.py
```diff
@@ -531,9 +531,7 @@ def _replace_existing_config(path, auth_section, profile_name):
     if matcher:
         # we matched an existing authentication section
         new_config_text = (
-            config_text[: matcher.start()]
-            + auth_section
-            + config_text[matcher.end() :]
+            config_text[: matcher.start()] + auth_section + config_text[matcher.end() :]
```
BryanFauble, line 534: You might get hit by the pre-commit check here that runs in the CI build. I would recommend that you double check that `pre-commit` is setup where you are developing:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/CONTRIBUTING.md#set-up-pre-commit

tests/integration/synapseclient/integration_test.py
```diff
@@ -40,19 +40,31 @@ async def test_login(syn):
...
         assert syn.credentials == None
 
         # login with config file no username
-        syn.login(silent=True)
+        syn.login(profile=profile, silent=True)
         assert syn.credentials.username == username
 
         # Login with ID only from config file
-        syn.login(username, silent=True)
+        syn.login(profile=profile, silent=True)
```
BryanFauble, line 67: This is changing the test slightly by removing `username` from the `.login` call. Would it make sense to provide both the `profile` and `username` here?

tests/integration/synapseclient/test_command_line_client.py
```diff
@@ -882,9 +912,13 @@ async def test_table_query(test_state):
...
-    output_rows = output.rstrip("\n").split("\n")
-    if output_rows[0] and output_rows[0].startswith(f"[{schema1.id}]: Downloaded to"):
-        output_rows = output_rows[1:]
+    output_rows = [
+        line
+        for line in output.strip().splitlines()
+        if line.strip()
+        and not line.startswith("Welcome")
+        and not line.startswith(f"[{schema1.id}]")
+    ]
```
BryanFauble, line 921: Something to think about in these tests would be to question why it's not getting the `Welcome...` logged in message when, I assume, it was not getting them before.

Did we change when the `Welcome...` messages show, to show in cases where it shouldn't be?

tests/integration/synapseclient/integration_test.py
carmmmm: I looked into passing both `profile` and `username`, but `Synapse.login()` doesn’t currently support that combination, passing both results in a `TypeError`, since `username` isn’t an accepted keyword when `profile` is provided. Also when running syn.login("$Username"), under the hood it is actually running the default profile, so if that username doesn't match the default profile there will also be an error, because the method defaults to profile="default" unless explicitly set otherwise. Even if the .synapseConfig has valid credentials under a different profile (e.g., [profile user2]), it won't be used. 

I updated the tests so that it covers the current login methods as I understand them: 
1. Login using `profile`
2. Login using `authToken`
3. Login using `email=username` with `profile=None` to ensure profile resolution doesn't interfere

Should username-based login require explicitly disabling profile resolution like this, or is that an unintended side effect of the current logic?

tests/integration/synapseclient/test_command_line_client.py
carmmmm: I was digging into this and I think I figured out what’s going on, but would love your thoughts to confirm.

So previously, the login() command did include a log line like `self.logger.info(f"Welcome, {display_name}!\n")`
And it was only triggered when `silent=False`  which was typically the case during explicit login (synapse login) but not during implicit login, like synapse get or store. So in those cases, the message wasn’t printed and test outputs stayed clean.

In the new version, this message logic was slightly updated to:

```
if (
    not self.credentials.profile_name
    or self.credentials.profile_name.lower()
    == config_file_constants.AUTHENTICATION_SECTION_NAME
):
    self.logger.info(f"Welcome, {display_name}!\n")
else:
    self.logger.info(f"Welcome, {display_name}! You are using the '{profile}' profile.\n")
```

It still checks the silent flag, but the difference now is that it logs a message for profile-based logins even in cases where the CLI used to stay quiet. I think that’s why the welcome message is now appearing in test outputs where it wasn’t before, and that might be what was breaking the CLI integration test parsing.

Does that seem right? Should we gate this welcome message only for interactive/explicit logins?

synapseclient/__main__.py
carmmmm: Thanks for the heads up so I double checked and I do have pre-commit set up locally and ran pre-commit run --all-files. All the checks passed except black, which weirdly reformats synapseclient/__main__.py with the whitespace tweak every time. I’m curious to see if that causes any issues on your end or if I maybe have something slightly off in my pre-commit setup locally.

tests/integration/synapseclient/test_command_line_client.py
BryanFauble: Thanks for looking into this. I don't think that we need to do any additional changes here. The silent logger should be setup in https://github.com/Sage-Bionetworks/synapsePythonClient/blob/13359b9261d6519985a3400cd62605127444f9ac/tests/integration/conftest.py#L53-L79 , but maybe something is slightly incorrect there when it's being used in these specifically integration tests.

I'll get these changes merged into the feature branch shortly, and when I know I can run the integration tests

tests/integration/synapseclient/integration_test.py
BryanFauble: The way this is written seems fine. This was the case I wanted to ensure works (And I verified this):

```
syn.login(email="bfauble1", profile="testing") # Produce warning
syn.login(email="bfauble", profile="testing")
```

This will give back the expected warning:


```
[WARNING] bfauble1 was defined in the user login arguments, however, it is also defined in the `~/.synapseConfig` file. Becuase they do not match we will not use the `authtoken` in the `~/.synapseConfig` file.
  File "/home/bfauble/BryansGreatWorkspace/synapsePythonClient/junk2.py", line 7, in <module>
    syn.login(email="bfauble1", profile="testing")
  File "/home/bfauble/BryansGreatWorkspace/synapsePythonClient/synapseclient/client.py", line 844, in login
    raise SynapseNoCredentialsError(
synapseclient.core.exceptions.SynapseNoCredentialsError: No valid authentication credentials provided.
Tried profile: 'testing', email: 'bfauble1'.
Check your `.synapseConfig` or ensure the provided auth token is valid.
```

tests/integration/synapseclient/test_command_line_client.py
carmmmm: Ahh also I was looking more into how `stdout` capture works and I was thinking another reason could be since we’re capturing output directly in the run() function, instead of using a subprocess, any logger messages would still end up in the captured output if we manually attach the stream handler. That might explain why the silent logger didn’t fully suppress it?

Also did these changes already make it into add-multiple-profiles, or are they still queued for merge? Just wanted to make sure I didn’t miss anything on my side!

BryanFauble: > Ahh also I was looking more into how stdout capture works and I was thinking another reason could be since we’re capturing output directly in the run() function, instead of using a subprocess, any logger messages would still end up in the captured output if we manually attach the stream handler. That might explain why the silent logger didn’t fully suppress it?

Ah! That makes a lot of sense. I was a bit confused because when I was running the functions directly while I set the silent flag I wasn't running into this. But the fact that we grab the output and parse that makes sense.

> Also did these changes already make it into add-multiple-profiles, or are they still queued for merge? Just wanted to make sure I didn’t miss anything on my side!

Nope, nothing on your side. I have been fighting to get the integration test runs tidied up before I merged this in (https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1199)

carmmmm: Oh great! Just wanted to make sure I didn’t miss anything! hHappy to pitch in if there’s anything I can do to help

carmmmm: That sounds cool! I'd be happy to help with that if it's okay with @thomasyu888 !

thomasyu888: Thanks for offering and that'd be great if you could help with that @carmmmm 

I like that approach, in the case where those aren't provided then they default to prod or what's in the "global" config.

PR #1199, opened 2025-04-25
[SYNPY-1295] Trim down integration tests

BryanFauble:
# **Problem:**

- Integration test runs are taking longer and longer, after talking with Xa we found out:

> Bottlenecks looking at the queues are download_csv_from_table (160+), query (130+) and table_view (6K+).
> We have a lot of calls to 'GET /user/#/team'.

# **Solution:**

- Use github copilot agent mode to trim down the number of required tests and combine similar logic
- The number of tests was reduced from 598 to 440 (-158). Although some tests significantly expanded, this represents a significant reduction.
- Deleted a significant build-up of data with: https://gist.github.com/BryanFauble/49342e30f13c33e037f5cc813d4c7ab4
Untitled

# **Testing:**

- I ran the tests locally and found that these tests all seemed to pass
- Will also be reviewing the integration test run in github
- Test coverage is marginally worse (Decreasing around 4%)

```
With changes:   lines-valid="28400" lines-covered="10092" line-rate="0.3554"
Before changes: lines-valid="29975" lines-covered="11736" line-rate="0.3915"
```


BWMac:
@BryanFauble Do you have a summary of how many API requests were cut down by the reduction in tests?

tests/integration/synapseclient/models/async/test_entityview_async.py
```diff
@@ -27,257 +26,206 @@
...
+                "name": "multiple_columns",
+                "description": "EntityView with multiple columns",
+                "columns": [
+                    Column(name="test_column", column_type=ColumnType.STRING),
+                    Column(name="test_column2", column_type=ColumnType.INTEGER),
+                ],
+                "include_default_columns": False,
+                "expected_column_count": 2,
+            },
+        ]
```
BWMac, line 104: Did you do this instead of pytest parametrize so you only have to make the api call to get the default columns once?

BryanFauble:
> @BryanFauble Do you have a summary of how many total API requests were cut out of the integration testing process by the reduction in tests?

@BWMac That would be a great metric, but I don't have the numbers for this no. We could probably use snowflake to look at the averages once we get the data tomorrow.

BWMac (approved):
LGTM! Just one question

tests/integration/synapseclient/models/async/test_entityview_async.py
BryanFauble: I didn't make the changes in this case (Github copilot did). I sparsely checked the work and didn't find anything super crazy.

> so you only have to make the api call to get the default columns once?

This sounds like a good reason why copilot would have done it like this, although another approach could have been to extract the default columns to a fixture used for the module, and used pytest parametrize. 🤷

PR #1200, opened 2025-05-06
[SYNPY-1244] Add delete permissions/acl functionality

BryanFauble:
https://sagebionetworks.jira.com/browse/SYNPY-1244

# **Problem:**
- When sharing subfolders or specific files with collaborators in private projects, permissions create a complex ACL (Access Control List) structure where:
  - Sharing settings can be hidden deep in folder hierarchies
  - Files don't automatically inherit the parent container's permissions
  - When releasing data, these individual ACLs need to be removed so files properly inherit parent container sharing settings
- There was no mechanism available to delete ACLs in the OOP model, nor was there a way to recursively delete the ACL

# **Solution:**
- Adds new functions to recursively reset permissions on entities within a container
- Implements functionality to determine if an entity inherits ACLs from its parent
- Works with nested folder structures to ensure proper permission inheritance/removal

**See this in the API Reference docs:**
- https://synapsepythonclient--1200.org.readthedocs.build/en/1200/reference/experimental/sync/project/#synapseclient.models.Project.delete_permissions
- https://synapsepythonclient--1200.org.readthedocs.build/en/1200/reference/experimental/sync/folder/#synapseclient.models.Folder.delete_permissions

# **Testing:**
- Automated tests were added to the codebase to:
  - Verify that permission inheritance detection works correctly
  - Ensure recursive permission setting functions properly traverse entity hierarchies
  - Confirm that entity permissions are properly modified by the new functions
  - Test error handling and edge cases

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -2725,7 +2725,7 @@ class Synapse(object):
 
         Arguments:
             parent: An id or an object of a Synapse container or None to retrieve all projects
-            includeTypes: Must be a list of entity types (ie. ["folder","file"]) which can be found [here](http://docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityType.html)
+            includeTypes: Must be a list of entity types (ie. ["folder","file"]) which can be found [here](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/EntityType.html)
             sortBy: How results should be sorted. Can be NAME, or CREATED_ON
             sortDirection: The direction of the result sort. Can be ASC, or DESC
 
@@ -2795,12 +2795,19 @@ class Synapse(object):
 
         return entity
 
-    def _getACL(self, entity: Union[Entity, str]) -> Dict[str, Union[str, list]]:
+    def _getACL(
+        self, entity: Union[Entity, str], check_benefactor: bool = True
+    ) -> Dict[str, Union[str, list]]:
         """
         Get the effective Access Control Lists (ACL) for a Synapse Entity.
 
         Arguments:
             entity: A Synapse Entity or Synapse ID
+            check_benefactor: If True (default), check the benefactor for the entity
+                to get the ACL. If False, only check the entity itself.
+                This is useful for checking the ACL of an entity that has local sharing
+                settings, but you want to check the ACL of the entity itself and not
+                the benefactor it may inherit from.
 
         Returns:
             A dictionary of the Entity's ACL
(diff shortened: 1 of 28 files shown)


synapseclient/client.py
```diff
@@ -2795,24 +2795,48 @@ def _getBenefactor(
...
+                try:
+                    return self.restGET(uri)
+                except SynapseHTTPError as e:
+                    if (
+                        "The requested ACL does not exist. This entity inherits its permissions from:"
+                        in str(e)
+                    ):
+                        # If the entity does not have an ACL, return an empty ACL
+                        return {"resourceAccess": []}
+                    raise e
```
BryanFauble, line 2839: I needed to add this set of changes into the `client.py` because the access_control.py mixin internally still uses it. The purpose of this changes is to allow us to get the ACL for a specific entity without the code automatically getting the benefactor. This is crucial during testing to verify that a given is or is not overwriting the folders ACL that it's contained within.

synapseclient/models/mixins/access_control.py
```diff
@@ -193,3 +215,359 @@ async def main():
...
+                - 'is_target_type': Whether the entity's type is in the target types
+
+        Returns:
+            None
+
+        Raises:
+            SynapseHTTPError: If there are permission issues or if the entity already inherits permissions.
+            Exception: For any other errors that may occur during deletion.
+        """
+        if not entity_info["is_target_type"] and entity_info["entity_type"] is not None:
```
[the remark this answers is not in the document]
BryanFauble: Committed suggestion

synapseclient/client.py
thomasyu888: This is a nice addition to grab the ACL of an entity - it's confusing because _technically_ the ACL is inherited on child entities from a benefactor but it's nice to not have it return [] when there is clearly an ACL

synapseclient/models/protocols/access_control_protocol.py
```diff
@@ -145,3 +163,97 @@ def set_permissions(
...
+
+        Example: Delete permissions for a single entity
+            ```python
+            from synapseclient import Synapse
+            from synapseclient.models import File
+
+            syn = Synapse()
+            syn.login()
+
+            File(id="syn123").delete_permissions()
```
thomasyu888, line 226: All entities across synapse like Tables,Views, and etc can have permissions deleted as well.  Is this function only work for files and folders?

BryanFauble: This function will work for all entity types. What the specific logic in the function is for is dealing with traversing the tree of files and folders within a container (Project/Folder). Users will still be able to delete the permissions for other entities like Tables, Views, ect..

See this script and console log output for an example:
https://gist.github.com/BryanFauble/0b8c0ab8644751241dc7a31e8e779158

thomasyu888 (commented):
🔥 LGTM!  I'll defer to someone on the team for final review, but thanks for this work!

BWMac (approved):
🚀

PR #1201, opened 2025-05-12
[SYNPY-1437] INCOMPLETE/DRAFT - Start the process to move to use the /bundle2 rest API 

BryanFauble:
**NOTE:** This is an incomplete solution. All the changes included here are largely un-tested. Development on this jira stopped when I re-discovered that https://sagebionetworks.jira.com/browse/PLFM-8313 blocks us from being able to move to the `/bundle2` rest APIs, as this prevents our ability to forcefully increment entities as available in an API like https://rest-docs.synapse.org/rest/PUT/entity/id.html

# **Problem:**

- All resources that get saved to the Synapse rest API go through a multiple step process to store the entity data, annotations, activity, and association to the entity via several disjointed HTTP calls. APIs like https://rest-docs.synapse.org/rest/PUT/entity/id/bundle2.html and https://rest-docs.synapse.org/rest/POST/entity/bundle2/create.html exist to update or create everything (But activity, as it is a separate call) in a single HTTP request.
- By swapping to use the /bundle2 rest API we can consolidate the storing process for all OOP models and remove duplicated code across the code base.

# **Solution:**

- Starts the process to implement all of the required logic to point the storing of all entities at the `/bundle2` API calls
- Start to drop all usage of the core (And soon to be deprecated/removed) Synapse class functionality in favor or dedicated API call functions

# **Testing:**

- TODO

### DIFF
--- a/synapseclient/api/entity_bundle_services_v2.py
+++ b/synapseclient/api/entity_bundle_services_v2.py
@@ -152,3 +152,92 @@ async def put_entity_id_bundle2(
         + (f"?generatedBy={generated_by}" if generated_by else ""),
         body=json.dumps(request),
     )
+
+
+async def store_entity_with_bundle2(
+    entity: Dict[str, Any],
+    parent_id: Optional[str] = None,
+    acl: Optional[Dict[str, Any]] = None,  # TODO: Consider skipping ACL?
+    annotations: Optional[Dict[str, Any]] = None,
+    activity: Optional[Dict[str, Any]] = None,
+    new_version: bool = False,
+    force_version: bool = False,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """
+    Store an entity in Synapse using the bundle2 API endpoints to reduce HTTP calls.
+
+    This function follows a specific flow:
+    1. Determines if the operation is a create or update:
+       - If no ID is provided, searches for the ID via /entity/child
+       - If no ID is found, treats as a Create
+       - If an ID is found, treats as an Update
+
+    2. For Updates:
+       - Retrieves entity by ID and merges with existing data
+       - Updates desired fields in the retrieved object
+       - Pushes modified object with HTTP PUT if there are changes
+
+    3. For Creates:
+       - Creates a new object with desired fields
+       - Pushes the new object with HTTP POST
+
+    Arguments:
+        entity: The entity to store.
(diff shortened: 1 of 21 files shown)


synapseclient/api/entity_bundle_services_v2.py
```diff
@@ -152,3 +152,92 @@ async def put_entity_id_bundle2(
...
+        client.logger.debug("Creating new entity via bundle2 API")
+
+        # For creation, parent ID is required
+        # TODO: Projects won't have a parent in this case
+        # if parent_id:
+        #     # Add parentId to the entity if not already set
+        #     if not entity.get("parentId"):
+        #         entity["parentId"] = parent_id
+        # elif not entity.get("parentId"):
+        #     raise ValueError("Parent ID must be provided for entity creation")
```
BryanFauble, line 230: To be addressed

synapseclient/api/entity_bundle_services_v2.py
```diff
@@ -152,3 +152,92 @@ async def put_entity_id_bundle2(
...
+            request=bundle_request,
+            generated_by=activity.get("id") if activity else None,
+            synapse_client=synapse_client,
+        )
+    else:
+        # This is an update
+        client.logger.debug(f"Updating entity {entity_id} via bundle2 API")
+
+        # For updates we might need to retrieve the existing entity to merge data
+        # Only retrieve if we need
```
BryanFauble, line 243: update functionality is incomplete

synapseclient/models/services/storable_entity.py
```diff
@@ -1,103 +1,162 @@
...
+        Dataset,
+        EntityView,
+        File,
+        Folder,
+        Project,
+        Table,
+    )
 
 
 async def store_entity(
```
BryanFauble, line 28: This function is incomplete, more work to support all functionality from the existing synapse class `.store()` function is needed

synapseclient/models/services/storable_entity_components.py
```diff
@@ -120,6 +120,77 @@ async def store_entity_components(
     return re_read_required
 
 
+async def store_entity_components_file_folder_only(
```
BryanFauble, line 123: This was a hack to make sure that storing files/folders attached to the `project` and `folder` class could still be saved. See the logic for `annotations` and `activity` is commented out in this code.

PR #1202, opened 2025-05-14
[SYNPY-1244] Clarify docstrings with `delete_permissions` function and `recursive` behavior

BryanFauble:
# **Problem:**

- The docstring for the `recursive` flag implied incorrect behavior of the function when used in conjunction with `include_container_content`

# **Solution:**

- Update docstring and examples to clarify behavior
- Raise a ValueError exception when falling into a situation that will cause no actions to be taken

# **Testing:**

- Verified updated behavior and unit test

### DIFF
--- a/synapseclient/models/mixins/access_control.py
+++ b/synapseclient/models/mixins/access_control.py
@@ -243,12 +243,15 @@ class AccessControllable(AccessControllableSynchronousProtocol):
         Arguments:
             include_self: If True (default), delete the ACL of the current entity.
                 If False, skip deleting the ACL of the current entity and only process
-                children if recursive=True or include_container_content=True.
+                children if recursive=True and include_container_content=True.
             recursive: If True and the entity is a container (e.g., Project or Folder),
-                recursively delete ACLs from all child containers and their children.
+                recursively process child containers. Note that this must be used with
+                include_container_content=True to have any effect. Setting recursive=True
+                with include_container_content=False will raise a ValueError.
                 Only works on classes that support the `sync_from_synapse_async` method.
             include_container_content: If True, delete ACLs from contents directly within
-                containers (files and folders inside Projects/Folders). Defaults to False.
+                containers (files and folders inside Projects/Folders). This must be set to
+                True for recursive to have any effect. Defaults to False.
             target_entity_types: Specify which entity types to process when deleting ACLs.
                 Allowed values are "folder" and "file" (case-insensitive).
                 If None, defaults to ["folder", "file"].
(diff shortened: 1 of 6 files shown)


synapseclient/models/mixins/access_control.py
```diff
@@ -327,6 +341,12 @@ async def main():
 
         should_process_children = recursive or include_container_content
         if should_process_children and hasattr(self, "sync_from_synapse_async"):
+            if recursive and not include_container_content:
```
SageGJ, line 344: Just a consideration from a ux perspective:
Had it been clarified that setting `recurive=True` and `include_container_content=False` wouldn't have deleted any of the access controls (but that it still iterates through the project and displays the folders), as I user I would have appreciated having the functionality to do a dry run and see everything that would have had access controls removed before actually doing so. @BryanFauble What are your thoughts on not halting execution and rather adjusting the messaging displayed in the case of this combination of parameters? Granted it doesn't look like files themselves are displayed in the current messaging?


```
[WARNING] Failed to delete ACL for entity syn66520312: 403 Client Error: Cannot restore inheritance for resource which has no parent.
[syn66520312:acc_control_test]: Syncing Project from Synapse.                                                                                                                                                                                         
Syncing from Synapse:   0%|                                                                                                                                                                                                | 0.00/1.00 [00:00<?, ?B/s]
[syn66520313:temp folder]: Syncing Folder from Synapse.                                                                                                                                                                                               
Syncing from Synapse:   0%|      
```

synapseclient/models/mixins/access_control.py
```diff
@@ -327,6 +341,12 @@ async def main():
...
+                raise ValueError(
+                    "When recursive=True, include_container_content must also be True. "
+                    "Setting recursive=True with include_container_content=False has no effect."
+                )
+
```
SageGJ, line 349: Just a consideration from a ux perspective:
Had it been clarified that setting `recurive=True` and `include_container_content=False` wouldn't have deleted any of the access controls (but that it still iterates through the project and displays the folders), as I user I would have appreciated having the functionality to do a dry run and see everything that would have had access controls removed before actually doing so. @BryanFauble What are your thoughts on not halting execution and rather adjusting the messaging displayed in the case of this combination of parameters? Granted it doesn't look like files themselves are displayed in the current messaging?


```
[WARNING] Failed to delete ACL for entity syn66520312: 403 Client Error: Cannot restore inheritance for resource which has no parent.
[syn66520312:acc_control_test]: Syncing Project from Synapse.                                                                                                                                                                                         
Syncing from Synapse:   0%|                                                                                                                                                                                                | 0.00/1.00 [00:00<?, ?B/s]
[syn66520313:temp folder]: Syncing Folder from Synapse.                                                                                                                                                                                               
Syncing from Synapse:   0%|      
```

BryanFauble: > [WARNING] Failed to delete ACL for entity syn66520312: 403 Client Error: Cannot restore inheritance for resource which has no parent.

This message only shows because you cannot delete the ACL for projects.

> Had it been clarified that setting recurive=True and include_container_content=False wouldn't have deleted any of the access controls (but that it still iterates through the project and displays the folders)

This was not clarified, and in the code it would have just skipped over these entries.

> as I user I would have appreciated having the functionality to do a dry run and see everything that would have had access controls removed before actually doing so. @BryanFauble What are your thoughts on not halting execution and rather adjusting the messaging displayed in the case of this combination of parameters?

This is a great idea! I created this follow-up ticket https://sagebionetworks.jira.com/browse/SYNPY-1604 to accomplish this ask. We can do quite a bit to make this easier for folks to understand the ACL structure of their project and implications of running the delete.

SageGJ: Okay that all makes sense, thanks! @BryanFauble

thomasyu888 (approved):
🔥 LGTM - Thanks for the clarification + catching this case and dry run idea both!

PR #1204, opened 2025-05-29
[SYNPY-1606] File Upload/Download OpenTelemetry Transfer Data

BryanFauble:
# **Problem:**

- When users are downloading and uploading data we do not have a rich set of data to peer into that process. The same can be said of our hosted applications where we could access cloud watch metrics/logs, however, it lacks the additional context about what was being transferred.

# **Solution:**

- Adding more verbose OpenTelemetry tracing for the SYNPY client to capture events as data is being uploaded, downloaded, and processed for MD5 calculation. This is captured as a mix of span level attributes (Doesn't have a time component) and event level attributes (Which are time specific)

# **Testing:**

- Downloading/Uploading a variety of multi-part files (>8MB) and less both on a windows machine and a service catalog EC2 instance.


synapseclient/core/otel_config.py
```diff
@@ -0,0 +1,233 @@
...
+    # Set as global tracer provider
+    trace.set_tracer_provider(provider)
+
+    return provider
+
+
+def configure_metrics(
+    resource_attributes: Optional[Dict[str, Any]] = None,
+    include_context: bool = True,
+) -> MeterProvider:
```
BryanFauble, line 155: I am not currently collecting any metrics in this code, however, I am leaving this in so we have a good starting point if/when we want to collect metrics.

synapseclient/core/download/download_functions.py
```diff
@@ -433,7 +489,15 @@ async def download_by_file_handle(
...
             storage_location_id = file_handle.get("storageLocationId")
 
+            actual_file_size = file_handle.get("contentSize", 0)
+            actual_md5 = file_handle.get("contentMd5")
+
+            if actual_file_size:
+                span.set_attribute("synapse.file.size_bytes", actual_file_size)
+
             if concrete_type == concrete_types.EXTERNAL_OBJECT_STORE_FILE_HANDLE:
+                span.set_attribute("synapse.storage.provider", "s3")
```
thomasyu888, line 499: Just wanted to check where we would be tracking the "google" buckets

synapseclient/core/otel_config.py
thomasyu888: Can you provide an example of what a metric we could collect is in the future?

synapseclient/core/otel_config.py
```diff
@@ -0,0 +1,234 @@
...
+from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
+from opentelemetry.trace import Span, SpanContext, get_current_span
+
+# Default service name for the client
+DEFAULT_SERVICE_NAME = "synapseclient"
+CLIENT_VERSION = "unspecified"
+SYNAPSE_SERVICE_VERSION = f"synapse.{SERVICE_VERSION}"
+
+
+class AttributePropagatingSpanProcessor(SpanProcessor):
```
thomasyu888, line 30: From what I remember, we have a lot of more complicated context managers set throughout the code when we first enabled otel across our functions.  Will this improvement reduce the number of those?

synapseclient/core/download/download_functions.py
```diff
@@ -705,8 +790,12 @@ def download_from_url(
                 password=password,
                 progress_bar=progress_bar,
             )
+
+            file_size = os.path.getsize(destination)
+            span.set_attribute("synapse.file.size_bytes", file_size)
```
thomasyu888, line 795: Just for information, does setting the span attribute here add a column that can be queried within signoz?

synapseclient/core/otel_config.py
```diff
@@ -0,0 +1,234 @@
...
+    Resource,
+)
+from opentelemetry.sdk.trace import SpanProcessor, TracerProvider
+from opentelemetry.sdk.trace.export import BatchSpanProcessor, ConsoleSpanExporter
+from opentelemetry.trace import Span, SpanContext, get_current_span
+
+# Default service name for the client
+DEFAULT_SERVICE_NAME = "synapseclient"
+CLIENT_VERSION = "unspecified"
+SYNAPSE_SERVICE_VERSION = f"synapse.{SERVICE_VERSION}"
```
thomasyu888, line 27: Perhaps not for this ticket, but the values here look similar also to the user-agent headers that we set so that information is available in the warehouse per processedaccessrecord.

synapseclient/core/upload/multipart_upload_async.py
```diff
@@ -609,6 +641,18 @@ def _put_part_with_retry(
 
                 else:
                     raise
+
+        self.record_span_event(
```
thomasyu888, line 645: Might be helpful to walk through and define a span, and then events in a span, and then spans in that event and how all of that impacts the dashboarding in signoz / SQL queries in clickhouse

thomasyu888 (commented):
🔥 I've taken a quick glance and this is great. This might be a good one to walk through in codespaces if you're looking for a live code review.

I'm not sure how familiar people in general are with traces and OTEL so this could be a pretty complex review.

synapseclient/core/otel_config.py
BryanFauble: Cpu/memory usage. Connection pooling for http connections.

synapseclient/core/otel_config.py
BryanFauble: The code I added into client.py and fhe otel_config.py is meant to simplify some usage. I added to the Readme and in the docstrings to show how they can be used. That combined with the auto instrumentation libraries is meant to make it easier to use.

synapseclient/core/download/download_functions.py
BryanFauble: That's right

synapseclient/core/otel_config.py
BryanFauble: It is similar, and they serve a similar purpose. These resource level attributes let us group all requests from a particular user/service.

synapseclient/core/download/download_functions.py
BryanFauble: Good question, I'll need to revist what code path Google buckets follow. This path is specifically for user owned s3 bucket downloads.

synapseclient/core/download/download_functions.py
```diff
@@ -524,17 +524,12 @@ async def download_by_file_handle(
...
-                    if os.path.exists(downloaded_path)
-                    else 0
-                )
                 span.add_event(
                     "download_chunk_completed",
                     {
                         "chunk_number": 0,
                         "start_byte": 0,
-                        "end_byte": file_size,
+                        "end_byte": actual_file_size,
```
thomasyu888, line 532: 👍🏼

synapseclient/core/download/download_functions.py
```diff
@@ -608,6 +646,24 @@ def download_fn(
...
+                    "download_chunk_completed",
+                    {
+                        "chunk_number": 0,
+                        "start_byte": 0,
+                        "end_byte": file_size - 1 if file_size > 0 else 0,
+                        "file_handle_id": file_handle_id,
+                        "synapse_id": synapse_id,
+                        "time_to_transfer_seconds": time.time() - download_start_time,
+                    },
+                )
```
thomasyu888: It's too bad these specific chunks of code are repeated, I wonder if a way-down-the-line optimization is to have a factory class for downloading, specifically around supporting the different cloud storage buckets.

That said - that may be overkill or the wrong design choice. Thoughts?

BryanFauble: I like that idea of refactoring it eventually to follow that kind of pattern. I dislike how cognitively complex this operation is.

thomasyu888 (approved):
🔥 LGTM!  I'll defer to someone on @Sage-Bionetworks/dpe  for final review

synapseclient/core/otel_config.py
```diff
@@ -0,0 +1,234 @@
...
+    def on_end(self, span: Span) -> None:
+        """Propagates attributes from the child span back to the parent span"""
+        parent_span = get_current_span()
+        if parent_span is not None and parent_span.is_recording():
+            for attribute in self.attributes_to_propagate_to_parent:
+                child_val = span.attributes.get(attribute)
+                parent_val = parent_span.attributes.get(attribute)
+                if child_val and not parent_val:
+                    # Propagate the attribute back to parent span
+                    parent_span.set_attribute(attribute, child_val)
```
SageGJ, line 78: The attribute propagation will be really useful. Is it possible to surface information about events from child spans in the parent span event list? Just thinking that would have simplified the work with schematic's exceptions a lot.

SageGJ (commented):
I really like the additional custom attributes implemented here. Thinking about the future, if we find the need to add more attributes, or if we add functionality to the codebase that requires similar span/event attributes as defined here what are your thoughts on how that process would look like? I'd be concerned about either an attribute being accidentally left out from new functionality, or a new attribute accidentally not being added to existing spans or events given the amount of times dictionaries are reused or where attributes are set one-by-one.
If there are some attributes that are required for spans of certain operations, or events within those spans, and possibly some that are optional, what do you think about capturing that in a dictionary template or dataclass (or just written documentation) with required and optional fields defined for different operations we'd be interested in tracking?

synapseclient/core/otel_config.py
SageGJ: @BryanFauble it makes sense that we'd want to control how far back events get propagated. This might be just a schematic thing but I noticed there there typically weren't any events in the list of events for the top level span. In cases where it was desirable to propagate every back up to the top level span, perhaps adding the span id to the event attributes would allow filtering by depth in our signoz queries, but I'm also just thinking out loud here. 
Thanks for the response

PR #1205, opened 2025-06-04
[SYNPY-1599]: add json schema mixin class 

linglp:
# **Problem:**
* We currently do not have an easy interface to bind, validate, and unbind jsonschemas. 
* We currently don’t have a function that handles asynchronous POST requests with pagination.

# **Solution:**
* Add json schema mixin and have Folder, File entities to inherit from the JSON SCHEMA mixin 
* Add a function in `api_client.py` to handle asynchronous POST requests with pagination

# **Testing:**

### DIFF
--- a/synapseclient/api/api_client.py
+++ b/synapseclient/api/api_client.py
@@ -1,0 +1,54 @@
+from typing import Dict, Any, AsyncGenerator, Optional, TYPE_CHECKING
+import httpx
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+
+async def rest_post_paginated_async(
+    uri: str,
+    body: Dict[str, Any] = None,
+    endpoint: str = None,
+    headers: httpx.Headers = None,
+    retry_policy: Dict[str, Any] = {},
+    requests_session_async_synapse: httpx.AsyncClient = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+    **kwargs,
+) -> AsyncGenerator[Dict[str, str], None]:
+    """
+    Asynchronously yield items from a paginated POST endpoint.
+
+    Arguments:
+        uri: Endpoint URI for the POST request.
+        body: Request payload dictionary.
+        endpoint: Optional server endpoint override.
+        headers: Optional HTTP headers.
+        retry_policy: Optional retry settings.
+        requests_session_async_synapse: Optional async HTTPX client session.
+        kwargs: Additional keyword arguments for the request.
+
+    Yields:
+        Individual items from each page of the response.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    next_page_token = None
+    while True:
+        if next_page_token is not None:
+            body = body or {}
+            body["nextPageToken"] = next_page_token
+        response = await client.rest_post_async(
+            uri=uri,
+            body=body,
+            endpoint=endpoint,
+            headers=headers,
+            retry_policy=retry_policy,
+            requests_session_async_synapse=requests_session_async_synapse,
+            **kwargs,
+        )
+        next_page_token = response.get("nextPageToken")
+        for item in response.get("page", []):
+            yield item
+        if next_page_token is None:
+            break
--- a/synapseclient/api/api_client.py
+++ b/synapseclient/api/api_client.py
@@ -1,8 +1,10 @@
-from typing import Dict, Any, AsyncGenerator, Optional, TYPE_CHECKING
+from typing import TYPE_CHECKING, Any, AsyncGenerator, Dict, Optional
+
 import httpx
 
 if TYPE_CHECKING:
     from synapseclient import Synapse
+
 
 async def rest_post_paginated_async(
     uri: str,
--- a/synapseclient/api/json_schema_services.py
+++ b/synapseclient/api/json_schema_services.py
@@ -1,0 +1,180 @@
+import json 
+from typing import TYPE_CHECKING, Optional, AsyncGenerator
+from synapseclient.api.api_client import rest_post_paginated_async
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+
+async def bind_json_schema_to_entity(synapse_id: str, json_schema_uri: str, *, synapse_client: Optional["Synapse"] = None) -> dict[str, str|int|bool]:
+    """Bind a JSON schema to an entity
+
+    Arguments:
+        json_schema_uri: JSON schema URI
+        entity:          Synapse Entity or Synapse Id
+        synapse_client:  If not passed in and caching was not disabled by
+                         `Synapse.allow_client_caching(False)` this will use the last created
+                         instance from the Synapse class constructor
+    Returns:
+        A dictionary with the following structure:
+
+        {
+            "jsonSchemaVersionInfo": {
+                "organizationId": str,
+                "organizationName": str,
+                "schemaId": str,
+                "schemaName": str,
+                "versionId": str,
+                "$id": str (unique identifier for the schema),
+                "semanticVersion": str,
+                "jsonSHA256Hex": str,
+                "createdOn": str (ISO datetime),
+                "createdBy": str (synapse user ID),
+            },
+            "objectId": int,
+            "objectType": str,
+            "createdOn": str (ISO datetime),
+            "createdBy": str (synapse user ID),
+            "enableDerivedAnnotations": bool
+        }
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    request_body = {"entityId": synapse_id, "schema$id": json_schema_uri}
+    return await client.rest_put_async(uri=f"/entity/{synapse_id}/schema/binding", body=json.dumps(request_body))
+
+async def get_json_schema_from_entity(synapse_id: str, *, synapse_client: Optional["Synapse"] = None) -> dict[str, str|int|bool]:
+    """Get bound schema from entity
+
+    Arguments:
+        synapse_id:      Synapse Id
+        synapse_client:  If not passed in and caching was not disabled by
+                         `Synapse.allow_client_caching(False)` this will use the last created
+                         instance from the Synapse class constructor
+
+    Returns:
+        A dictionary with the following structure:
+        {
+            "jsonSchemaVersionInfo": {
+                "organizationId": str,
+                "organizationName": str,
+                "schemaId": str,
+                "schemaName": str,
+                "versionId": str,
+                "$id": str (unique identifier for the schema),
+                "semanticVersion": str,
+                "jsonSHA256Hex": str,
+                "createdOn": str (ISO datetime),
+                "createdBy": str (synapse user ID),
+            },
+            "objectId": int,
+            "objectType": str,
+            "createdOn": str (ISO datetime),
+            "createdBy": str (synapse user ID),
+            "enableDerivedAnnotations": bool
+        }
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(uri=f"/entity/{synapse_id}/schema/binding")
+
+
+async def delete_json_schema_from_entity(synapse_id: str, *, synapse_client: Optional["Synapse"] = None) -> None:
+    """Delete bound schema from entity
+    Arguments:
+        synapse_id:      Synapse Id
+        synapse_client:  If not passed in and caching was not disabled by
+                         `Synapse.allow_client_caching(False)` this will use the last created
+                         instance from the Synapse class constructor
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_delete_async(uri=f"/entity/{synapse_id}/schema/binding")
+
+async def validate_entity_with_json_schema(
+    synapse_id: str, *, synapse_client: Optional["Synapse"] = None
+) -> dict[str, str|bool]:
+    """Get validation results of an entity against bound JSON schema
+
+    Arguments:
+        synapse_id:      Synapse Id
+        synapse_client:  If not passed in and caching was not disabled by
+                         `Synapse.allow_client_caching(False)` this will use the last created
+                         instance from the Synapse class constructor
+
+    Returns:
+        {
+            "objectId": str,         # Synapse ID of the object (e.g., "syn12345678")
+            "objectType": str,       # Type of the object (e.g., "entity")
+            "objectEtag": str,       # ETag of the object at the time of validation
+            "schema$id": str,        # Full URL of the bound schema version
+            "isValid": bool,         # True if the object content conforms to the schema
+            "validatedOn": str       # ISO 8601 timestamp of when validation occurred
+        }
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(uri=f"/entity/{synapse_id}/schema/validation")
+
+
+async def get_json_schema_validation_statistics(
+    synapse_id: str, *, synapse_client: Optional["Synapse"] = None
+) -> dict[str, int|str]:
+    """Get the summary statistic of json schema validation results for
+        a container entity
+     Arguments:
+        synapse_id:      Synapse Id
+        synapse_client:  If not passed in and caching was not disabled by
+                         `Synapse.allow_client_caching(False)` this will use the last created
+                         instance from the Synapse class constructor
+
+    Returns:
+        {
+            "containerId": str,              # Synapse ID of the parent container
+            "totalNumberOfChildren": int,    # Total number of child entities
+            "numberOfValidChildren": int,    # Number of children that passed validation
+            "numberOfInvalidChildren": int,  # Number of children that failed validation
+            "numberOfUnknownChildren": int,  # Number of children with unknown validation status
+            "generatedOn": str               # ISO 8601 timestamp when this summary was generated
+        }
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(uri=f"/entity/{synapse_id}/schema/validation/statistics")
+
+async def get_invalid_json_schema_validation(synapse_id: str, *, synapse_client: Optional["Synapse"] = None) -> AsyncGenerator[dict[str, str], None]:
+    """Get a single page of invalid JSON schema validation results for a container Entity
+        (Project or Folder).
+
+        Arguments:
+            synapse_id:      Synapse Id
+            synapse_client:  If not passed in and caching was not disabled by
+                             `Synapse.allow_client_caching(False)` this will use the last created
+                             instance from the Synapse class constructor"""
+
+    request_body = {"containerId": synapse_id}
+    response = rest_post_paginated_async(f"/entity/{synapse_id}/schema/validation/invalid", body=json.dumps(request_body), synapse_client=synapse_client)
+    async for item in response:
+        yield item
+    
+async def get_json_schema_derived_keys(
+    synapse_id: str, *, synapse_client: Optional["Synapse"] = None
+) -> dict[str, list[str]]:
+    """Retrieve derived JSON schema keys for a given Synapse entity.
+
+    Args:
+        synapse_id (str): The Synapse ID of the entity for which to retrieve derived keys.
+
+    Returns:
+        dict: A dictionary containing the derived keys associated with the entity.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(uri=f"/entity/{synapse_id}/derivedKeys")
+
+
--- a/synapseclient/api/json_schema_services.py
+++ b/synapseclient/api/json_schema_services.py
@@ -154,7 +154,19 @@ async def get_invalid_json_schema_validation(synapse_id: str, *, synapse_client:
             synapse_id:      Synapse Id
             synapse_client:  If not passed in and caching was not disabled by
                              `Synapse.allow_client_caching(False)` this will use the last created
-                             instance from the Synapse class constructor"""
+                             instance from the Synapse class constructor
+
+    Example usage:
+    async def main():
+        gen = get_invalid_json_schema_validation(synapse_client=syn, synapse_id=dataset_folder)
+        try:
+            while True:
+                item = await anext(gen)
+                print(item)
+        except StopAsyncIteration:
+            print("All items processed.")
+    asyncio.run(main())
+    """
 
     request_body = {"containerId": synapse_id}
     response = rest_post_paginated_async(f"/entity/{synapse_id}/schema/validation/invalid", body=json.dumps(request_body), synapse_client=synapse_client)
--- a/synapseclient/api/json_schema_services.py
+++ b/synapseclient/api/json_schema_services.py
@@ -1,11 +1,15 @@
-import json 
-from typing import TYPE_CHECKING, Optional, AsyncGenerator
+import json
+from typing import TYPE_CHECKING, AsyncGenerator, Optional
+
 from synapseclient.api.api_client import rest_post_paginated_async
 
 if TYPE_CHECKING:
     from synapseclient import Synapse
 
-async def bind_json_schema_to_entity(synapse_id: str, json_schema_uri: str, *, synapse_client: Optional["Synapse"] = None) -> dict[str, str|int|bool]:
+
+async def bind_json_schema_to_entity(
+    synapse_id: str, json_schema_uri: str, *, synapse_client: Optional["Synapse"] = None
+) -> dict[str, str | int | bool]:
     """Bind a JSON schema to an entity
 
     Arguments:
@@ -41,9 +45,14 @@ async def bind_json_schema_to_entity(synapse_id: str, json_schema_uri: str, *, s
 
     client = Synapse.get_client(synapse_client=synapse_client)
     request_body = {"entityId": synapse_id, "schema$id": json_schema_uri}
-    return await client.rest_put_async(uri=f"/entity/{synapse_id}/schema/binding", body=json.dumps(request_body))
+    return await client.rest_put_async(
+        uri=f"/entity/{synapse_id}/schema/binding", body=json.dumps(request_body)
+    )
 
-async def get_json_schema_from_entity(synapse_id: str, *, synapse_client: Optional["Synapse"] = None) -> dict[str, str|int|bool]:
+
+async def get_json_schema_from_entity(
+    synapse_id: str, *, synapse_client: Optional["Synapse"] = None
+) -> dict[str, str | int | bool]:
     """Get bound schema from entity
 
     Arguments:
@@ -80,7 +89,9 @@ async def get_json_schema_from_entity(synapse_id: str, *, synapse_client: Option
     return await client.rest_get_async(uri=f"/entity/{synapse_id}/schema/binding")
 
 
-async def delete_json_schema_from_entity(synapse_id: str, *, synapse_client: Optional["Synapse"] = None) -> None:
+async def delete_json_schema_from_entity(
+    synapse_id: str, *, synapse_client: Optional["Synapse"] = None
+) -> None:
     """Delete bound schema from entity
     Arguments:
         synapse_id:      Synapse Id
@@ -93,9 +104,10 @@ async def delete_json_schema_from_entity(synapse_id: str, *, synapse_client: Opt
     client = Synapse.get_client(synapse_client=synapse_client)
     return await client.rest_delete_async(uri=f"/entity/{synapse_id}/schema/binding")
 
+
 async def validate_entity_with_json_schema(
     synapse_id: str, *, synapse_client: Optional["Synapse"] = None
-) -> dict[str, str|bool]:
+) -> dict[str, str | bool]:
     """Get validation results of an entity against bound JSON schema
 
     Arguments:
@@ -122,7 +134,7 @@ async def validate_entity_with_json_schema(
 
 async def get_json_schema_validation_statistics(
     synapse_id: str, *, synapse_client: Optional["Synapse"] = None
-) -> dict[str, int|str]:
+) -> dict[str, int | str]:
     """Get the summary statistic of json schema validation results for
         a container entity
      Arguments:
@@ -144,9 +156,14 @@ async def get_json_schema_validation_statistics(
     from synapseclient import Synapse
 
     client = Synapse.get_client(synapse_client=synapse_client)
-    return await client.rest_get_async(uri=f"/entity/{synapse_id}/schema/validation/statistics")
+    return await client.rest_get_async(
+        uri=f"/entity/{synapse_id}/schema/validation/statistics"
+    )
 
-async def get_invalid_json_schema_validation(synapse_id: str, *, synapse_client: Optional["Synapse"] = None) -> AsyncGenerator[dict[str, str], None]:
+
+async def get_invalid_json_schema_validation(
+    synapse_id: str, *, synapse_client: Optional["Synapse"] = None
+) -> AsyncGenerator[dict[str, str], None]:
     """Get a single page of invalid JSON schema validation results for a container Entity
         (Project or Folder).
 
@@ -169,10 +186,15 @@ async def get_invalid_json_schema_validation(synapse_id: str, *, synapse_client:
     """
 
     request_body = {"containerId": synapse_id}
-    response = rest_post_paginated_async(f"/entity/{synapse_id}/schema/validation/invalid", body=json.dumps(request_body), synapse_client=synapse_client)
+    response = rest_post_paginated_async(
+        f"/entity/{synapse_id}/schema/validation/invalid",
+        body=json.dumps(request_body),
+        synapse_client=synapse_client,
+    )
     async for item in response:
         yield item
-    
+
+
 async def get_json_schema_derived_keys(
     synapse_id: str, *, synapse_client: Optional["Synapse"] = None
 ) -> dict[str, list[str]]:
@@ -188,5 +210,3 @@ async def get_json_schema_derived_keys(
 
     client = Synapse.get_client(synapse_client=synapse_client)
     return await client.rest_get_async(uri=f"/entity/{synapse_id}/derivedKeys")
-
-
--- a/synapseclient/models/mixins/json_schema.py
+++ b/synapseclient/models/mixins/json_schema.py
@@ -1,0 +1,168 @@
+from typing import TYPE_CHECKING, AsyncGenerator, Optional
+
+from synapseclient.api.json_schema_services import (
+    bind_json_schema_to_entity,
+    delete_json_schema_from_entity,
+    get_invalid_json_schema_validation,
+    get_json_schema_derived_keys,
+    get_json_schema_from_entity,
+    get_json_schema_validation_statistics,
+    validate_entity_with_json_schema,
+)
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+
+
+class JSONSchema:
+    """
+    Mixin class to provide JSON schema functionality.
+    This class is intended to be used with classes that represent Synapse entities.
+    It provides methods to bind, delete, and validate JSON schemas associated with the entity.
+    """
+
+    id: Optional[str] = None
+
+    async def bind_json_schema_to_entity(
+        self, json_schema_uri: str, *, synapse_client: Optional["Synapse"] = None
+    ) -> dict[str, str | int | bool]:
+        """Bind a JSON schema to an entity"""
+        return await bind_json_schema_to_entity(
+            synapse_id=self.id,
+            json_schema_uri=json_schema_uri,
+            synapse_client=synapse_client,
+        )
+
+    async def get_json_schema_from_entity(
+        self, *, synapse_client: Optional["Synapse"] = None
+    ) -> dict[str, str | int | bool]:
+        """Get bound schema from entity
+
+        Arguments:
+            synapse_id:      Synapse Id
+            synapse_client:  If not passed in and caching was not disabled by
+                            `Synapse.allow_client_caching(False)` this will use the last created
+                            instance from the Synapse class constructor
+
+        Returns:
+            A dictionary with the following structure:
+            {
+                "jsonSchemaVersionInfo": {
+                    "organizationId": str,
+                    "organizationName": str,
+                    "schemaId": str,
+                    "schemaName": str,
+                    "versionId": str,
+                    "$id": str (unique identifier for the schema),
+                    "semanticVersion": str,
+                    "jsonSHA256Hex": str,
+                    "createdOn": str (ISO datetime),
+                    "createdBy": str (synapse user ID),
+                },
+                "objectId": int,
+                "objectType": str,
+                "createdOn": str (ISO datetime),
+                "createdBy": str (synapse user ID),
+                "enableDerivedAnnotations": bool
+            }
+        """
+        return await get_json_schema_from_entity(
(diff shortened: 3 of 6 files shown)


synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,170 @@
...
+        """Bind a JSON schema to an entity"""
+        return await bind_json_schema_to_entity(
+            synapse_id=self.id,
+            json_schema_uri=json_schema_uri,
+            synapse_client=synapse_client,
+        )
+
+    async def get_json_schema_from_entity_async(
+        self, *, synapse_client: Optional["Synapse"] = None
+    ) -> Dict[str, Union[str, int, bool]]:
```
BryanFauble, line 40: For any of the return types of these functions try to convert them into a python dataclass with the `snake_case` format for attributes. That way someone that calls the function would be able to do:

```
(await my_folder.get_json_schema_from_entity_async()).json_schema_version_info.organization_id
```

We want to be very clear for users of the python client what fields are possible to access when they call these methods

linglp: Thank you for the suggestion! I will make the update.

synapseclient/api/json_schema_services.py
```diff
@@ -0,0 +1,224 @@
...
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+
+
+async def bind_json_schema_to_entity(
+    synapse_id: str,
+    json_schema_uri: str,
+    *,
+    enable_derived_annos: bool = False,
```
linglp, line 14: set this as an optional parameter. This parameter should probably also be added to the sync function

tests/integration/synapseclient/models/async/test_json_schema_async.py
```diff
@@ -0,0 +1,481 @@
...
+            stored_folder = await folder.store_async(parent=project_model)
+            response = await created_folder.get_json_schema_from_entity_async(
+                synapse_client=self.syn
+            )
+            response = await created_folder.get_json_schema_from_entity_async(
+                synapse_client=self.syn
+            )
+            assert response["enableDerivedAnnotations"] == True
+
+            sleep(2)
```
linglp, line 208: I tried removing it and created another function that wraps around the async call to ensure that the async call finished. Even though the wrapper function worked as expected, the test would fail. This is something that needs to be further investigated.

BryanFauble: Instead of sleep (Which will sleep the thread this is running on), the preferred way is to use asyncio:

`await asyncio.sleep(2)`


It likely fails WITHOUT the sleep because of the server side process being eventually consistent (Meaning that clients have to wait until the server side state is updated with the changes you made). Even though the Job you submitted to Synapse completed, anything that job updated may not have been fully updated yet.

Using a wait statement here is the right call (And something we've have to do in other integration tests).

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,221 @@
...
+    schemaLocation: str
+    causingExceptions: List["CausingException"] = field(default_factory=list)
+
+
+@dataclass
+class ValidationException:
+    pointerToViolation: str
+    message: str
+    schemaLocation: str
+    causingExceptions: List[CausingException]
```
BryanFauble, line 76: These fields will eventually need to be moved to `snake_case`. This is one of the small conversion items that makes working with the rest APIs a bit annoying. Since `snake_case` is the standard in Python, but `camelCase` is the standard in Java it makes for a small bit of work on our side.

This is the purpose of these `fill_from_dict` functions:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/fa4e17c4e7beef9cd5203b5fd5497ae4a38d6e18/synapseclient/models/materializedview.py#L669-L700

It's also similar to why when we go in the other direction (To store data in Synapse) we have these `to_synapse_request` functions:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/fa4e17c4e7beef9cd5203b5fd5497ae4a38d6e18/synapseclient/models/materializedview.py#L702-L728

synapseclient/models/mixins/json_schema.py
BryanFauble: Yes that looks correct (Any the same pattern can be applied to other method calls too). The only thing to note: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1205#discussion_r2136095677


It's probably not the best example that this API only has a single field: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/annotation/v2/Keys.html - However, by returning a dataclass object we are future proofing this method call to support adding additional fields (If Synapse ever does) to the return method without breaking existing code, since it would be NEW data added to the existing object.

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,221 @@
...
+
+@dataclass
+class InvalidJSONSchemaValidationResponse:
+    JSONSchemaValidationResponse,
+    allValidationMessages: List[str]
+    validationException: ValidationException
+
+
+@dataclass
+class JSONSchemaDerivedKeysResponse:
```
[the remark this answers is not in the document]
linglp: Okay, thank you, Bryan! I could rename it.

tests/integration/synapseclient/models/async/test_json_schema_async.py
linglp: Wow thanks for the explanation. I didn't realize that the job could be completed but the update might not be finished.

synapseclient/services/json_schema.py
```diff
@@ -686,14 +686,20 @@ def json_schema_validation(self, json_schema_uri: str):
         return response
 
     @authentication_required
-    def bind_json_schema_to_entity(self, synapse_id: str, json_schema_uri: str):
+    def bind_json_schema_to_entity(
```
thomasyu888, line 689: This entire module will be replaced by the new OOP Model so you may not need to add to this unless you are using this particular function

synapseclient/api/json_schema_services.py
```diff
@@ -0,0 +1,164 @@
+import json
+from typing import TYPE_CHECKING, AsyncGenerator, Optional
+
+from synapseclient.api.api_client import rest_post_paginated_async
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+    from synapseclient.models.mixins import (
```
thomasyu888, line 8: Just checking that there aren't any circular dependencies here.

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,537 @@
...
+class BaseJsonSchema:
+    """
+    Mixin class to provide JSON schema functionality.
+    This class is intended to be used with classes that represent Synapse entities.
+    It provides methods to bind, delete, and validate JSON schemas associated with the entity.
+    """
+
+    id: Optional[str] = None
+
+    async def bind_json_schema_to_entity_async(
```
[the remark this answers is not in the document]
linglp: Thanks for the suggestion, Tom! I am thinking how about: 
```
bind_schema_async
get_schema_async
delete_schema_async
validate_schema_async
get_schema_derived_keys_async
get_schema_validation_statistics_async
get_invalid_validation_async
```

docs/scripts/object_orientated_programming_poc/oop_poc_file.py
```diff
@@ -12,16 +12,24 @@
 7. Copying a file
 8. Storing an activity to a file
 9. Retrieve an activity from a file
+10. Bind a JSON schema to files and validate its contents
```
BryanFauble, line 15: I apologize that I did not catch this earlier, but these files will eventually be removed in favor of the tutorial pages instead. The intention of these files was to give the very first iteration on what tutorial scripts could look like.

Take a look at this confluence page that covers tutorial creation: https://sagebionetworks.jira.com/wiki/spaces/SYNPY/pages/3864690727/SYNPY-1549+Python+Client+New+Service+Implementation+SOP#Tutorials

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,467 @@
...
+        """
+        response = await get_json_schema_validation_statistics(
+            synapse_id=self.id, synapse_client=synapse_client
+        )
+        return JSONSchemaValidationStatistics(
+            container_id=response.get("containerId", ""),
+            total_number_of_children=response.get("totalNumberOfChildren", ""),
+            number_of_valid_children=response.get("numberOfValidChildren", ""),
+            number_of_invalid_children=response.get("numberOfInvalidChildren", ""),
+            number_of_unknown_children=response.get("numberOfUnknownChildren", ""),
```
BryanFauble, line 404: These are `int` types, so I would expect that a `""` is not a valid default value. I would rather see these be `None` if not in the response.

docs/scripts/object_orientated_programming_poc/oop_poc_file.py
thomasyu888: @linglp we will want to be careful adding lines of code to these, because we actually link specific lines of these tutorial code in the docs.

synapseclient/models/protocols/json_schema_protocol.py
```diff
@@ -0,0 +1,135 @@
...
+        Validate the entity against the bound JSON schema.
+
+        Args:
+            synapse_client (Optional[Synapse], optional): The Synapse client instance. If not provided,
+                the last created instance from the Synapse class constructor will be used.
+
+        Returns:
+            Union[JSONSchemaValidation, InvalidJSONSchemaValidation]: The validation results.
+        """
+        return InvalidJSONSchemaValidation() or JSONSchemaValidation()
```
linglp, line 81: @BryanFauble can I do something like this?

synapseclient/models/protocols/json_schema_protocol.py
```diff
@@ -0,0 +1,135 @@
...
+
+        Args:
+            synapse_client (Optional[Synapse], optional): The Synapse client instance. If not provided,
+                the last created instance from the Synapse class constructor will be used.
+
+        Yields:
+            InvalidJSONSchemaValidation: An object containing the validation response, all validation messages,
+                                         and the validation exception details.
+        """
+        return InvalidJSONSchemaValidation(), None
```
linglp, line 135: @BryanFauble this one involves the async generator. Could you double check if this is correct?

synapseclient/models/protocols/json_schema_protocol.py
BryanFauble: Probably

docs/tutorials/json_schema.md
```diff
@@ -0,0 +1,163 @@
...
+## Prerequisites
+* You have a working installation of the Synapse Python Client
+* You have a project on Synapse ready for testing
+* You are familiar with Synapse concepts: Project, Folder, File
+* You are authenticated `(syn = synapseclient.Synapse(); syn.login())`
+
+
+##  Step 1: Define a JSON Schema
+```python
+title = "OOP Test Schema"
```
BryanFauble, line 22: Rather than duplicating the python code the preferences instead is to tell mkdocs to grab certain lines from the python script. See other markdowns for examples

docs/tutorials/json_schema.md
```diff
@@ -0,0 +1,163 @@
+In this tutorial, you will learn how to bind a JSON Schema to a folder. You can also bind a JSON Schema to a project, a file, a table, or a view. This tutorial will walk you through each step, from schema creation to validation.
```
BryanFauble, line 1: There should be a short explanation on what JSON Schema is

docs/tutorials/json_schema.md
```diff
@@ -0,0 +1,163 @@
...
+3. Bind the JSON Schema to a folder
+4. Add annotations and validate against the schema
+5. Attach and validate a file
+6. View schema validation statistics and results
+
+## Prerequisites
+* You have a working installation of the Synapse Python Client
+* You have a project on Synapse ready for testing
+* You are familiar with Synapse concepts: Project, Folder, File
+* You are authenticated `(syn = synapseclient.Synapse(); syn.login())`
```
BryanFauble, line 17: Not necessary as the first step in the tutorial should be to set up everything needed to run through the rest of the steps

docs/tutorials/python/tutorial_scripts/json_schema.py
```diff
@@ -0,0 +1,157 @@
+import os
```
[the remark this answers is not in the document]
thomasyu888: I'm not opposed to this idea, something to consider is the creation of jsonschemas will be done in a different way since we were going to refactor it to leverage the OOP model (that's another ticket)

So we would want to indicate that the tutorial is changing, or... Just sit tight for now and add in that section later when we get to implementation of oop models

docs/tutorials/json_schema.md
linglp: Thanks for the suggestion, Bryan! I borrowed the explanation from the synapse docs.

docs/tutorials/json_schema.md
```diff
@@ -0,0 +1,163 @@
...
+
+1. Log in and create a project and folder
+2. Create a JSON Schema and an organization
+3. Bind the JSON Schema to a folder
+4. Add annotations and validate against the schema
+5. Attach and validate a file
+6. View schema validation statistics and results
+
+## Prerequisites
+* You have a working installation of the Synapse Python Client
```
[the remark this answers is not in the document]
linglp: Okay. I added links to installation, project, folder, file.

docs/tutorials/json_schema.md
linglp: Okay got it. The first step will be logging in and retrieving the test project just like the virtual table tutorial.

synapseclient/services/json_schema.py
linglp: Thanks, Tom. I’ve removed the changes from this section—I believe the current diff is only related to formatting.

docs/tutorials/python/tutorial_scripts/json_schema.py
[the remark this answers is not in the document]
BryanFauble: > How about I create the organization and pre-register the schema in advance? That way, users can simply retrieve the existing schema instead of creating it themselves.


I would prefer that the tutorial take the user on the entire journey themselves. My suggestion was that instead of you creating the organization ahead of time, or have cleanup logic to delete the organization in the script is that you:

1. Use a try-catch block that will:
2. - Attempt to get the given organization
3. - If not found, create a new organization


If ^ Is not possible, then yes - Go ahead and create the org/schema in advance

docs/tutorials/python/json_schema.md
```diff
@@ -0,0 +1,124 @@
...
+6. View schema validation statistics and results
+
+## Prerequisites
+* You have a working [installation](../installation.md)of the Synapse Python Client.
+* Make sure that you have completed the [Project](./project.md) tutorial, which covers creating and managing projects in Synapse. This is a prerequisite because you need a project to organize and store the folder used in this tutorial.
+* You are familiar with Synapse concepts: [Project](./project.md), [Folder](./folder.md), [File](./file.md).
+* You are familiar with [adding annotations](./annotation.md) to synapse entity.
+
+
+## Step 1: Set up Synapse Python client and retrieve project
```
BryanFauble, line 24: `Step` can be omitted from the header in this case

docs/tutorials/python/json_schema.md
```diff
@@ -0,0 +1,124 @@
...
+{!docs/tutorials/python/tutorial_scripts/json_schema.py!lines=65-67}
+```
+
+## Step 6: Add Invalid Annotations to the Folder and Store
+Try adding invalid annotations to your folder: This step and the step below demonstrate how the system handles invalid annotations and how the schema validation process works.
+```python
+{!docs/tutorials/python/tutorial_scripts/json_schema.py!lines=70-76}
+```
+
+In the synapse web UI, you could also see your invalid annotations being marked by a yellow label similar to this:
```
BryanFauble, line 68: Is an error message or something else printed to the console when there is an issue with the annotations you set?


I also see that `Step 7: Validate Folder Against the Schema` might cover this, however, I would expect that we mention the programmatic way to verify this data first, followed by the UI way to verify this. I would also combine these into the same step.

Lastly - if there are messages that should be printing to the console that the user would be expecting to see, that should be added to the tutorial. For example in the file tutorial: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/d9998ef905aa98a40ed98a5590df4bd314c51ccb/docs/tutorials/python/file.md?plain=1#L105-L121

docs/tutorials/python/json_schema.md
```diff
@@ -0,0 +1,124 @@
...
+{!docs/tutorials/python/tutorial_scripts/json_schema.py!lines=79-81}
+```
+
+## Step 8: Create a File with Invalid Annotations and Upload It
+Try creating a test file locally and store the file in the folder that we created earlier. Then, try adding invalid annotations to a file. This step demonstrates how the files inside a folder also inherit the schema from the parent entity.
+```python
+{!docs/tutorials/python/tutorial_scripts/json_schema.py!lines=84-111}
+```
+
+In the synapse web UI, you could also see your invalid annotations being marked by a yellow label similar to this:
```
BryanFauble, line 85: Similar comments to:
https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1205/files#r2150665500

Show programmatic validation first (And console logs) and then web UI

docs/tutorials/python/json_schema.md
```diff
@@ -0,0 +1,124 @@
...
+
+![jsonschema](./tutorial_screenshots/jsonschema_file.png)
+
+## Step 9: View Validation Statistics
+This step is only relevant for container entities, such as a folder or a project. Using this function could provide you with information such as the number of children with invalid annotations inside a container.
```
BryanFauble, line 90: Show console logs the users would expect

docs/tutorials/python/tutorial_scripts/json_schema.py
```diff
@@ -0,0 +1,121 @@
...
+            "type": "integer",
+            "description": "Quantitative cognitive function score",
+        },
+        "diagnosis_stage": {
+            "type": "string",
+            "description": "Stage of Alzheimer's diagnosis (e.g., Mild, Moderate, Severe)",
+            "const": "Mild",  # Example constant for derived annotation
+        },
+    },
+}
```
BryanFauble, line 48: I think the big gap that exists in the tutorial is to show how to persist this schema to Synapse. Although it's not yet available in the OOP way, we should still show users how to do this today (And we will swap over the tutorial when the new way is available).

This was part of my motivation behind this comment: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1205/files#r2150570695

I would include a step or two showing the creation of the org, saving the schema, and perhaps making a small update to the schema.



The rest of the tutorial that you have showing off the various functionality is great!

The one addition I might suggest is showing the `delete_schema` function (Noting that it doesn't delete the schema itself, but rather, removes the schema from the entity). **Given that - What are you thoughts on naming it `unbind_schema` instead of `delete_schema`?**

synapseclient/models/protocols/json_schema_protocol.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class ContainerEntityJSONSchemaProtocol(BaseJSONSchemaProtocol):
+    """
+    Mixin class to provide JSON schema functionality for container entities.
+    This class extends BaseJSONSchemaProtocol and is intended for use with Synapse container entities.
+    It provides methods to bind, delete, and validate JSON schemas associated with the container entity.
+    """
+
+    async def get_schema_validation_statistics_async(
```
BryanFauble, line 106: The methods in this class will need to be updated to remove the async keywords

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,471 @@
...
+
+    async def delete_schema_async(
+        self, *, synapse_client: Optional["Synapse"] = None
+    ) -> None:
+        """
+        Delete the JSON schema bound to the entity.
+
+        Args:
+            synapse_client (Optional[Synapse], optional): The Synapse client instance. If not provided,
+                the last created instance from the Synapse class constructor will be used.
```
BryanFauble, line 285: nit: `Arguments:` is preferred over `Args:` (Both should be ok though as far as mkdocs is concerned).

Also, there is no need to add the typing to the docstring - It's implied from the type hint

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,471 @@
...
+                json_sha256_hex=json_schema_version.get("jsonSHA256Hex", ""),
+                created_on=json_schema_version.get("createdOn", ""),
+                created_by=json_schema_version.get("createdBy", ""),
+            ),
+            object_id=response.get("objectId", ""),
+            object_type=response.get("objectType", ""),
+            created_on=response.get("createdOn", ""),
+            created_by=response.get("createdBy", ""),
+            enable_derived_annotations=response.get("enableDerivedAnnotations", False),
+        )
```
BryanFauble, line 238: I'd expect that the default value is `None` rather than an empty string. An empty string has a different meaning to `None`, and if Synapse isn't returning a value `None` is the correct meaning

tests/integration/synapseclient/models/async/test_json_schema_async.py
```diff
@@ -0,0 +1,576 @@
...
+                    "type": "integer",
+                },
+            },
+        }
+        test_org = js.JsonSchemaOrganization(org_name)
+        created_schema = test_org.create_json_schema(schema, TEST_SCHEMA_NAME, "0.0.1")
+        yield test_org, created_schema.uri
+
+        js.delete_json_schema(created_schema.uri)
+        js.delete_organization(created_org["id"])
```
BryanFauble, line 111: Any cleanup needs to occur in a `try/finally` block

tests/integration/synapseclient/models/async/test_json_schema_async.py
```diff
@@ -0,0 +1,576 @@
...
+                validation_response = item.validation_response
+                validation_error_message = item.validation_error_message
+                validation_exception = item.validation_exception
+                causing_exceptions = validation_exception.causing_exceptions
+
+                assert validation_response.object_type == "entity"
+                assert validation_response.object_etag is not None
+                assert (
+                    validation_response.id
+                    == f"https://repo-prod.prod.sagebase.org/repo/v1/schema/type/registered/{test_org.name}-{TEST_SCHEMA_NAME}-{SCHEMA_VERSION}"
```
BryanFauble, line 553: We cannot assume what the URL of the synapse server is that we will hit as this could be run against any environment

BryanFauble (changes requested):
Great progress!

docs/tutorials/python/tutorial_scripts/json_schema.py
linglp: Thanks for the suggestion, Bryan! I added a step to create the organization and schema if they do not exist. I think currently if you make an update to the schema, you would basically just create a new schema with a newer version. Maybe I could just add that as a comment? 

I named the method `delete_schema` so that it aligns with the original method `js.delete_json_schema_from_entity`. `unbind_schema` sounds like a better name! I could make the update as long as we do not have to align the method names.

mkdocs.yml
```diff
@@ -47,7 +47,8 @@ nav:
         # - Migrate data to other storage locations: tutorials/python/migrate_data_to_other_storage_locations.md
       - Working with the Command Line Client: tutorials/command_line_client.md
       - Working with file versioning: tutorials/file_versioning.md
-      - Working with tables: tutorials/tables.md
+      - Working with tables: tutorials/python/table.md
```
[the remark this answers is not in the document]
linglp: Thanks for pointing it out. Let me fix that.

docs/tutorials/python/tutorial_scripts/json_schema.py
BryanFauble: > I could make the update as long as we do not have to align the method names.

We do not have to align the method names. It only needs to be documented in the API docs and referenced in the tutorial.

> Maybe I could just add that as a comment?

It would be nice to show an update to the schema so we are giving the users the tools that they would need to succeeded using this feature.

docs/tutorials/python/json_schema.md
```diff
@@ -0,0 +1,124 @@
+JSON Schema is a tool used to validate data. In Synapse, JSON Schemas can be used to validate the metadata applied to a project, file, folder, table, or view, including the [annotations](https://help.synapse.org/docs/Annotating-Data-With-Metadata.2667708522.html) applied to it. To learn more about JSON Schemas, check out [JSON-Schema.org](https://json-schema.org./).
```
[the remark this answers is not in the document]
linglp: I actually copied it from Synapse documentation [here](https://help.synapse.org/docs/JSON-Schemas.3107291536.html#JSONSchemas-JSONSchemasandAnnotations). But yes, I can add a link to point to a list of entities supported by Synapse [here](https://rest-docs.synapse.org/rest/index.html#org.sagebionetworks.repo.web.controller.EntityController)

tests/integration/synapseclient/models/async/test_json_schema_async.py
linglp: Got it. I added the try/finally block

synapseclient/models/protocols/json_schema_protocol.py
linglp: Got it. Thanks for catching it.

tests/integration/synapseclient/models/async/test_json_schema_async.py
BryanFauble: ok. How about we just verify that it ends in `/repo/v1/schema/type/registered/{test_org.name}-{TEST_SCHEMA_NAME}-{SCHEMA_VERSION` in case this is a bug and/or changes on the Synapse API side?

thomasyu888 (commented):
🔥 Great work and reviews here.  I'll defer to @BryanFauble as final reviewer after @linglp addresses all the comments. 

excited for this!

synapseclient/api/api_client.py
```diff
@@ -0,0 +1,56 @@
...
+
+async def rest_post_paginated_async(
+    uri: str,
+    body: Dict[str, Any] = None,
+    endpoint: str = None,
+    headers: httpx.Headers = None,
+    retry_policy: Dict[str, Any] = {},
+    requests_session_async_synapse: httpx.AsyncClient = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
```
BryanFauble, line 17: This argument is missing in the docstring

synapseclient/api/json_schema_services.py
```diff
@@ -0,0 +1,175 @@
...
+    )
+
+
+async def bind_json_schema_to_entity(
+    synapse_id: str,
+    json_schema_uri: str,
+    *,
+    enable_derived_annos: bool = False,
+    synapse_client: Optional["Synapse"] = None,
+) -> "JSONSchemaBinding":
```
BryanFauble, line 23: Is this the correct return types for these function calls? It seems like it's directly returning the call to `client.rest_*`, which is returning back a `Union[Dict[str, Any], str]` type

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,540 @@
...
+
+        Arguments:
+            json_schema_uri (str): The URI of the JSON schema to bind to the entity.
+            enable_derived_annos (bool, optional): If true, enable derived annotations. Defaults to False.
+            synapse_client (Optional[Synapse], optional): The Synapse client instance. If not provided,
+                the last created instance from the Synapse class constructor will be used.
+
+        Returns:
+            JSONSchemaBinding: An object containing details about the JSON schema binding.
+        """
```
BryanFauble, line 213: This is great! My suggestion on all of these functions (And you'll need to do the same for the the `ContainerEntityJSONSchemaProtocol` and `BaseJSONSchemaProtocol` functions too) would be to pick 2 Entity types you expect will be the most comment (Maybe a Folder, and a File?) that are likely to use these functions and create an example showing how the function would be used. This is separate from the Tutorials and the example in the docstring is what would show in in the API reference page here:

1. https://synapsepythonclient--1205.org.readthedocs.build/en/1205/reference/experimental/sync/folder/#synapseclient.models.Folder.bind_schema
2. https://synapsepythonclient--1205.org.readthedocs.build/en/1205/reference/experimental/sync/file/#synapseclient.models.File.bind_schema
3. https://synapsepythonclient--1205.org.readthedocs.build/en/1205/reference/experimental/async/folder/#synapseclient.models.Folder.bind_schema_async
4. https://synapsepythonclient--1205.org.readthedocs.build/en/1205/reference/experimental/async/file/#synapseclient.models.File.bind_schema_async

docs/reference/experimental/mixins/base_json_schema.md
```diff
@@ -0,0 +1,3 @@
+# BaseJSONSchema
+
+::: synapseclient.models.mixins.BaseJSONSchema
```
BryanFauble, line 3: The other dataclasses that can be returned or used in any of these models or mixins should also be added to this markdown page. For example:

`::: synapseclient.models.mixins.JSONSchemaBinding`


The consequence of not including this is that you cannot click on the return type when you look at the function in the API docs, and it's unclear from the doc page what fields will be available on that object:

image

BryanFauble (changes requested):
This is some seriously great work! I left the last few comments to be addressed, I didn't see anything else that needed updating.

docs/tutorials/python/tutorial_scripts/json_schema.py
```diff
@@ -0,0 +1,130 @@
...
+time.sleep(2)
+
+validation_statistics = test_folder.get_schema_validation_statistics()
+print("Validation statistics were retrieved successfully. Please see details below:")
+pprint(vars(validation_statistics))
+
+invalid_validation = test_folder.get_invalid_validation()
+for child in invalid_validation:
+    print("See details of validation results: ")
+    pprint(vars(child))
```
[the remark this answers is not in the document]
BryanFauble: This fix is just to add:

```

```

To https://github.com/Sage-Bionetworks/synapsePythonClient/blob/51ab0595ec231c6972b07eae3ca074f012d8ec7b/synapseclient/api/json_schema_services.py#L150-L158

docs/tutorials/python/json_schema.md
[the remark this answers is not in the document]
linglp: @thomasyu888 Currently, I’ve only extended the JSONSchema mixin to Project, Folder, File, EntityView, and Table, based on what's documented in the Synapse docs. Would you like me to include additional entity types?

The [Synapse documentation](https://help.synapse.org/docs/JSON-Schemas.3107291536.html#JSONSchemas-JSONSchemasandAnnotations) specifically mentions only those five, so I'm not sure whether other entity types support JSON Schema binding. (For view, do you think if they meant all three types of view: entity view, materialized view, and submission view? or just entity view?) Let me know if I should post this in the synapse channel.

synapseclient/api/json_schema_services.py
linglp: This is not! I just made some edits.

synapseclient/models/mixins/json_schema.py
linglp: Thanks Bryan! I have added the docstring.

docs/tutorials/python/tutorial_scripts/json_schema.py
linglp: Thanks for the suggestion. I added it in the tutorial.

synapseclient/models/mixins/json_schema.py
```diff
@@ -208,6 +208,76 @@ async def bind_schema_async(
...
+                ).store()
+
+                async def bind_schema_to_file():
+                    bound_schema_file = await example_file.bind_schema_async(
+                        json_schema_uri=SCHEMA_URI,
+                        enable_derived_annos=True
+                    )
+                    return bound_schema_file
+                asyncio.run(bind_schema_to_file())
+
```
BryanFauble, line 280: The examples of these is great, however, the formatting of the code block is slightly off.

image


https://github.com/Sage-Bionetworks/synapsePythonClient/blob/d9998ef905aa98a40ed98a5590df4bd314c51ccb/synapseclient/models/virtualtable.py#L79-L93

Is an example.

The corrected format for this would be:

```
        Example: Using this function
            Binding JSON schema to a folder or a file

            ```python
              import asyncio
              import synapseclient
              from synapseclient.models import File, Folder

              syn = synapseclient.Synapse()
              syn.login()

              # Define Project and JSON schema info
              PROJECT_ID = syn.findEntityId(name="test_json_schema_project") #replace with your project name
              ORG_NAME = "UniqueOrg" # replace with your organization name
              SCHEMA_NAME = "myTestSchema" # replace with your schema name
              VERSION = "0.0.1"
              SCHEMA_URI = f"{ORG_NAME}-{SCHEMA_NAME}-{VERSION}"
              # Create organization (if not already created)
              js = syn.service("json_schema")
              all_orgs = js.list_organizations()
              for org in all_orgs:
                  if org["name"] == ORG_NAME:
                      print(f"Organization {ORG_NAME} already exists.")
                      break
              else:
                  print(f"Creating organization {ORG_NAME}.")
                  js.create_organization(ORG_NAME)
              # Create the schema (if not already created)
              schema_definition = {
                  "$id": "mySchema",
                  "type": "object",
                  "properties": {
                      "foo": {"type": "string"},
                      "bar": {"type": "integer"},
                  },
                  "required": ["foo"]
              }
              my_test_org = js.JsonSchemaOrganization(ORG_NAME)
              test_schema = my_test_org.get_json_schema(SCHEMA_NAME)
              if not test_schema:
                  test_schema = my_test_org.create_json_schema(schema_definition, SCHEMA_NAME, VERSION)
              # Create a test folder
              test_folder = Folder(name="test_script_folder", parent_id=PROJECT_ID)
              test_folder.store()
              # Bind JSON schema to the folder
              async def bind_json_schema():
                  bound_schema = await test_folder.bind_schema_async(
                      json_schema_uri=SCHEMA_URI,
                      enable_derived_annos=True
                  )
                  return bound_schema
              asyncio.run(bind_json_schema())
              # Optionally, bind the same schema to a file
              example_file = File(
                  path="Sample.txt",  # Replace with your test file path
                  parent_id=test_folder.id,
              ).store()
              async def bind_schema_to_file():
                  bound_schema_file = await example_file.bind_schema_async(
                      json_schema_uri=SCHEMA_URI,
                      enable_derived_annos=True
                  )
                  return bound_schema_file
              asyncio.run(bind_schema_to_file())
            ```
```

linglp: Got it. Thank you for catching that, Bryan! Let me look at the documentation again to make sure everything is looking good.

synapseclient/api/json_schema_services.py
```diff
@@ -0,0 +1,217 @@
...
+) -> Union[Dict[str, Any], str]:
+    """
+    <https://rest-docs.synapse.org/rest/PUT/entity/id/schema/binding.html>
+
+    Bind a JSON schema to an entity
+
+    Arguments:
+        synapse_id:      Synapse Entity or Synapse Id
+        json_schema_uri: JSON schema URI
+        enable_derived_annos:  If True, derived annotations will be enabled for this entity
```
BryanFauble, line 34: Should annotations be spelled out for this kwarg instead of annos? I don't think we are using much, if any, short hand in the new models.

synapseclient/models/mixins/json_schema.py
```diff
@@ -0,0 +1,1179 @@
...
+                SCHEMA_NAME = "myTestSchema" # replace with your schema name
+                VERSION = "0.0.1"
+                SCHEMA_URI = f"{ORG_NAME}-{SCHEMA_NAME}-{VERSION}"
+
+                # Create organization (if not already created)
+                js = syn.service("json_schema")
+                all_orgs = js.list_organizations()
+                for org in all_orgs:
+                    if org["name"] == ORG_NAME:
+                        print(f"Organization {ORG_NAME} already exists.")
```
BryanFauble, line 237: On these examples it might more sense to push folks to use an info level log messages using the Synapse Client logger instead of print statement.

synapseclient/models/protocols/json_schema_protocol.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        Bind a JSON schema to the entity.
+
+        Arguments:
+            json_schema_uri (str): The URI of the JSON schema to bind to the entity.
+            enable_derived_annos (bool, optional): If true, enable derived annotations. Defaults to False.
+            synapse_client (Optional[Synapse], optional): The Synapse client instance. If not provided,
+                the last created instance from the Synapse class constructor will be used.
+
+        Returns:
```
BryanFauble, line 39: The unfortunately clunky thing we have to worry about when we create these sync function definitions is that we still want to have examples for the doc pages. The examples you have in the async method are great, they could be reuse here and swapped to remove asyncio.

synapseclient/api/json_schema_services.py
linglp: Okay got it. I can make that update.

docs/tutorials/python/json_schema.md
```diff
@@ -0,0 +1,225 @@
...
+</details>
+
+
+## Reference
+- [syn.findEntityId][synapseclient.Synapse.findEntityId]
+- [syn.store][synapseclient.Synapse.store]
+- [syn.login][synapseclient.Synapse.login]
+- [File][file-reference-sync]
+- [Folder][folder-reference-sync]
+- [Annotations][synapseclient.Annotations]
```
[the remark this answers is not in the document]
linglp: Okay. Just link it to the annotation tutorial.

synapseclient/models/protocols/json_schema_protocol.py
linglp: Got it. Thanks for the reminder. I have added the examples.

tests/integration/synapseclient/models/async/test_json_schema_async.py
```diff
@@ -0,0 +1,578 @@
...
+        try:
+            response = await created_entity.bind_schema_async(
+                json_schema_uri=test_product_schema_uri, synapse_client=self.syn
+            )
+            json_schema_version_info = response.json_schema_version_info
+            assert json_schema_version_info.organization_name == test_org.name
+            assert json_schema_version_info.id == test_product_schema_uri
+        finally:
+            # Clean up the JSON schema binding
+            await created_entity.unbind_schema_async(synapse_client=self.syn)
```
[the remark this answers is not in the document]
linglp: Good question. My experience is that if you delete an entity that has a JSON schema bound to it, the JSON Schema doesn't get automatically unbind. If you try to delete the JSON Schema, you would get an error even if the entity is in trash can: 
```
"Cannot delete a schema version that is bound to an object."
```
I tried asking [this question ](https://sagebionetworks.slack.com/archives/C0AFCTGEQ/p1749743391158219), but I don't think there's a way to delete the JSON Schema "by force" for now.

tests/integration/synapseclient/models/synchronous/test_json_schema.py
```diff
@@ -0,0 +1,572 @@
...
+    def entity_view(self, project_model: Project) -> EntityView:
+        entity_view = EntityView(
+            parent_id=project_model.id,
+            scope_ids=[project_model.id],
+            view_type_mask=ViewTypeMask.FILE | ViewTypeMask.FOLDER,
+        )
+        return entity_view
+
+    @pytest.mark.parametrize("entity_type", [Folder, Project, File, EntityView, Table])
+    def test_bind_schema(
```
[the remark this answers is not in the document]
linglp: wow interesting. Thanks! I just made the update.

tests/integration/synapseclient/models/async/test_json_schema_async.py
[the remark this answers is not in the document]
BryanFauble: Makes sense. Leave as is then

PR #1206, opened 2025-06-11
[SYNPY-1351] Implement 'Wiki2' model into OOP

danlu1:
# **Problem:**

[WikiPage Services 2](https://rest-docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.V2WikiController) is not implemented in the client. 

# **Solution:**

Add a new Wiki2 service to Python Client

# **Testing:**

(wip)

### DIFF
--- a/synapseclient/api/wiki_service.py
+++ b/synapseclient/api/wiki_service.py
@@ -1,0 +1,464 @@
+"""This module is responsible for exposing the services defined at:
+<https://rest-docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.WikiController2>
+"""
+
+import json
+from typing import TYPE_CHECKING, Any, AsyncGenerator, Dict, List, Optional
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
+
+import sys
+
+
+async def post_wiki(
+    owner_id: str,
+    request: Dict[str, Any],
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Create a new wiki page.
+    <https://rest-docs.synapse.org/rest/POST/entity/ownerId/wiki2.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        request: The wiki page to create.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The created wiki page.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_post_async(
+        uri=f"/entity/{owner_id}/wiki2",
+        body=json.dumps(request),
+    )
+
+
+async def get_wiki_page(
+    owner_id: str,
+    *,
+    wiki_id: Optional[str] = None,
+    wiki_version: Optional[int] = None,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Get a wiki page.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2.html>
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2/wikiId.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: Optional ID of the wiki. If not provided, returns the root wiki page.
+        wiki_version: Optional version of the wiki page.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The requested wiki page.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Construct the URI based on whether wiki_id is provided
+    uri = f"/entity/{owner_id}/wiki2"
+    if wiki_id is not None:
+        uri = f"{uri}/{wiki_id}"
+
+    # Add version as a query parameter if provided
+    params = {}
+    if wiki_version is not None:
+        params["wikiVersion"] = wiki_version
+
+    return await client.rest_get_async(
+        uri=uri,
+        params=params,
+    )
+
+
+async def put_wiki_page(
+    owner_id: str,
+    wiki_id: str,
+    request: Dict[str, Any],
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Update a wiki page.
+    <https://rest-docs.synapse.org/rest/PUT/entity/ownerId/wiki2/wikiId.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        request: The updated wiki page.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The updated wiki page.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_put_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}",
+        body=json.dumps(request),
+    )
+
+
+async def put_wiki_version(
+    owner_id: str,
+    wiki_id: str,
+    wiki_version: int,
+    request: Dict[str, Any],
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Update a specific version of a wiki page.
+    <https://rest-docs.synapse.org/rest/PUT/entity/ownerId/wiki2/wikiId/version.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        wiki_version: The version number to update.
+        request: The updated wiki page.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The updated wiki page version.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_put_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}/{wiki_version}",
+        body=json.dumps(request),
+    )
+
+
+async def delete_wiki_page(
+    owner_id: str,
+    wiki_id: str,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> None:
+    """Delete a wiki page.
+    <https://rest-docs.synapse.org/rest/DELETE/entity/ownerId/wiki2/wikiId.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    await client.rest_delete_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}",
+    )
+
+
+async def get_wiki_header_tree(
+    owner_id: str,
+    offset: Optional[int] = 0,
+    limit: Optional[int] = 20,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> AsyncGenerator[Dict[str, Any], None]:
+    """Get the header tree (hierarchy) of wiki pages for an entity.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wikiheadertree2.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        offset: The index of the pagination offset. For a page size of 10, the first page would be at offset = 0,
+               and the second page would be at offset = 10. Default is 0.
+        limit: Limits the size of the page returned. For example, a page size of 10 requires limit = 10.
+               Limit must be 50 or less. Default is 20.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        A generator over the wiki header tree for the entity. The tree contains the hierarchy of wiki pages
+        including their IDs, titles, and parent-child relationships.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    prev_num_results = sys.maxsize
+    while prev_num_results > 0:
+        params = {"offset": offset, "limit": limit}
+        page = await client.rest_get_async(
+            uri=f"/entity/{owner_id}/wikiheadertree2",
+            params=params,
+        )
+        results = page["results"] if "results" in page else page["children"]
+        prev_num_results = len(results)
+
+        for result in results:
+            offset += 1
+            yield result
+
+
+async def get_wiki_history(
+    owner_id: str,
+    wiki_id: str,
+    offset: Optional[int] = 0,
+    limit: Optional[int] = 20,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> AsyncGenerator[Dict[str, Any], None]:
+    """Get the history of a wiki page.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2/wikiId/wikihistory.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        offset: The index of the pagination offset. For a page size of 10, the first page would be at offset = 0,
+               and the second page would be at offset = 10. Default is 0.
+        limit: Limits the size of the page returned. For example, a page size of 10 requires limit = 10.
+               Limit must be 50 or less. Default is 20.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        A generator over the history of the wiki page. The history contains the history of the wiki page
+        including their IDs, titles, and parent-child relationships.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    params = {"offset": offset, "limit": limit}
+    return await client.rest_get_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}/wikihistory",
+        params=params,
+    )
+
+
+async def get_attachment_handles(
+    owner_id: str,
+    wiki_id: str,
+    *,
+    wiki_version: Optional[int] = None,
+    synapse_client: Optional["Synapse"] = None,
+) -> List[str, Any]:
+    """Get the file handles of all attachments on a wiki page.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2/wikiId/attachmenthandles.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        wiki_version: Optional version of the wiki page.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The the list of FileHandles for all file attachments of a specific WikiPage for a given owning Entity.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Add version as a query parameter if provided
+    params = {}
+    if wiki_version is not None:
+        params["wikiVersion"] = wiki_version
+
+    return await client.rest_get_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}/attachmenthandles",
+        params=params,
+    )
+
+
+async def get_attachment_url(
+    owner_id: str,
+    wiki_id: str,
+    file_name: str,
+    *,
+    redirect: Optional[bool] = False,
+    wiki_version: Optional[int] = None,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Get the URL of a wiki page attachment.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2/wikiId/attachment.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        file_name: The name of the file to get.
+                The file names can be found in the FileHandles from the GET /entity/{ownerId}/wiki/{wikiId}/attachmenthandles method.
+        redirect: When set to false, the URL will be returned as text/plain instead of redirecting. Default is False.
+        wiki_version: Optional version of the wiki page.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The URL that can be used to download a file for a given WikiPage file attachment.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Add version and redirect as a query parameter if provided
+    params = {}
+    params["fileName"] = file_name
+    if wiki_version is not None:
+        params["wikiVersion"] = wiki_version
+    if redirect is not None:
+        params["redirect"] = redirect
+
+    return await client.rest_get_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}/attachment",
+        params=params,
+    )
+
+
+async def get_attachment_preview_url(
+    owner_id: str,
+    wiki_id: str,
+    file_name: str,
+    redirect: Optional[bool] = False,
+    wiki_version: Optional[int] = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Get the preview of a wiki page attachment.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2/wikiId/attachmentpreview.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        file_name: The name of the file to get.
+                The file names can be found in the FileHandles from the GET /entity/{ownerId}/wiki/{wikiId}/attachmenthandles method.
+        redirect: When set to false, the URL will be returned as text/plain instead of redirecting. Default is False.
+        wiki_version: Optional version of the wiki page.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The URL that can be used to download a preview file for a given WikiPage file attachment.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Add version and redirect as a query parameter if provided
+    params = {}
+    params["fileName"] = file_name
+    if redirect is not None:
+        params["redirect"] = redirect
+    if wiki_version is not None:
+        params["wikiVersion"] = wiki_version
+
+    return await client.rest_get_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}/attachmentpreview",
+        params=params,
+    )
+
+
+async def get_markdown_url(
+    owner_id: str,
+    wiki_id: str,
+    redirect: Optional[bool] = False,
+    wiki_version: Optional[int] = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Get the markdown of a wiki page.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2/wikiId/markdown.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        redirect: When set to false, the URL will be returned as text/plain instead of redirecting. Default is False.
+        wiki_version: Optional version of the wiki page.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The URL that can be used to download the markdown file for a given WikiPage.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    # Add version and redirect as a query parameter if provided
+    params = {}
+    if redirect is not None:
+        params["redirect"] = redirect
+    if wiki_version is not None:
+        params["wikiVersion"] = wiki_version
+
+    return await client.rest_get_async(
+        uri=f"/entity/{owner_id}/wiki2/{wiki_id}/markdown",
+        params=params,
+    )
+
+
+async def get_wiki_order_hint(
+    owner_id: str,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Get the order hint of a wiki page tree.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2orderhint.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The order hint that corresponds to the given owner Entity.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_get_async(
+        uri=f"/entity/{owner_id}/wiki2orderhint",
+    )
+
+
+async def put_wiki_order_hint(
+    owner_id: str,
+    request: Dict[str, Any],
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Update the order hint of a wiki page tree.
+    <https://rest-docs.synapse.org/rest/PUT/entity/ownerId/wiki2orderhint.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        request: The updated order hint.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The updated order hint that corresponds to the given owner Entity.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+    return await client.rest_put_async(
+        uri=f"/entity/{owner_id}/wiki2orderhint",
+        body=json.dumps(request),
+    )
--- a/synapseclient/models/wiki.py
+++ b/synapseclient/models/wiki.py
@@ -1,1 +1,804 @@
-# TODO
+"""Script to work with Synapse wiki pages."""
+
+import gzip
+import os
+import shutil
+import uuid
+from dataclasses import dataclass, field
+from typing import Any, Dict, List, Optional, Union
+
+from synapseclient import Synapse
+from synapseclient.api.wiki_service import (
+    delete_wiki_page,
+    get_attachment_handles,
+    get_attachment_preview_url,
+    get_attachment_url,
+    get_markdown_url,
+    get_wiki_header_tree,
+    get_wiki_history,
+    get_wiki_order_hint,
+    get_wiki_page,
+    post_wiki,
+    put_wiki_order_hint,
+    put_wiki_page,
+    put_wiki_version,
+)
+from synapseclient.core.async_utils import async_to_sync, otel_trace_method
+from synapseclient.core.upload.upload_functions_async import upload_file_handle
+from synapseclient.core.utils import delete_none_keys
+
+
+@dataclass
+@async_to_sync
+class WikiOrderHint:
+    """
+    A WikiOrderHint contains the order hint for the root wiki that corresponds to the given owner ID and type.
+
+    Attributes:
+        owner_id: The ID of the owner object (e.g., entity, evaluation, etc.).
+        owner_object_type: The type of the owner object.
+        id_list: The list of sub wiki ids that in the order that they should be placed relative to their siblings.
+        etag: The etag of this object.
+    """
+
+    owner_id: Optional[str] = None
+    """The ID of the owner object (e.g., entity, evaluation, etc.)."""
+
+    owner_object_type: Optional[str] = None
+    """The type of the owner object."""
+
+    id_list: List[str] = field(default_factory=list)
+    """The list of sub wiki ids that in the order that they should be placed relative to their siblings."""
+
+    etag: Optional[str] = field(default=None, compare=False)
+    """The etag of this object."""
+
+    def fill_from_dict(
+        self,
+        wiki_order_hint: Dict[str, Union[str, List[str]]],
+    ) -> "WikiOrderHint":
+        """
+        Converts a response from the REST API into this dataclass.
+
+        Arguments:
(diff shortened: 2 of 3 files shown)


danlu1:
@BryanFauble @thomasyu888  I put this draft PR to show tentative interface of the wiki2 service. Protocal, tutorials and test scripts are wip. Feel free to leave comments if you envision it differently.

synapseclient/models/wiki.py
```diff
@@ -1 +1,804 @@
...
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Get_Wiki_Order_Hint: {self.owner_id}"
+    )
+    async def get_async(
+        self,
+        *,
+        synapse_client: Optional["Synapse"] = None,
+    ) -> "WikiOrderHint":
+        """
+        Get the order hint of a wiki page tree asynchronously.
```
BryanFauble, line 84:
```suggestion
        Get the order hint of a wiki page tree.
```

We don't need to mention `asynchronously` in the docstring

synapseclient/api/wiki_service.py
```diff
@@ -0,0 +1,464 @@
...
+        page = await client.rest_get_async(
+            uri=f"/entity/{owner_id}/wikiheadertree2",
+            params=params,
+        )
+        results = page["results"] if "results" in page else page["children"]
+        prev_num_results = len(results)
+
+        for result in results:
+            offset += 1
+            yield result
```
BryanFauble, line 210: See my comment here:

https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1206/files#r2141133748

It would be great to extract this paginated function that uses the `rest_get_async` function into it's own `rest_get_paginated_async` function

BryanFauble:
A tutorial should be created showing how to use the Wiki objects:
https://sagebionetworks.jira.com/wiki/spaces/SYNPY/pages/3864690727/SYNPY-1549+Python+Client+New+Service+Implementation+SOP#Tutorials

synapseclient/models/wiki.py
```diff
@@ -1 +1,804 @@
...
+        method_to_trace_name=lambda self, **kwargs: f"Get_Wiki_Header_Tree for Owner ID {kwargs['owner_id']}"
+    )
+    async def get_async(
+        cls,
+        owner_id: str,
+        *,
+        offset: int = 0,
+        limit: int = 20,
+        synapse_client: Optional["Synapse"] = None,
+    ) -> list:
```
BryanFauble, line 261: Same note: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1206/files#r2141179552

synapseclient/models/wiki.py
```diff
@@ -1 +1,804 @@
...
+
+        # Upload the gzipped file to get a file handle
+        file_handle = await upload_file_handle(
+            syn=synapse_client,
+            parent_entity_id=self.owner_id,
+            path=file_path,
+        )
+
+        # delete the temp gzip file
+        os.remove(file_path)
```
BryanFauble, line 503: This should be in a `try/finally` block to ensure that `os.remove` gets called even if an exception occurs

synapseclient/models/wiki.py
```diff
@@ -1 +1,804 @@
...
+            file_name=file_name,
+            wiki_version=self.wiki_version,
+            redirect=redirect,
+            synapse_client=synapse_client,
+        )
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Get_Attachment_Preview_URL: Owner ID {self.owner_id}, Wiki ID {self.id}, File Name {kwargs['file_name']}"
+    )
+    async def get_attachment_preview_url_async(
```
BryanFauble, line 736: Are `get_markdown_url_async`, `get_attachment_preview_url_async`, `get_attachment_url_async` methods that a user is intended to call?

It seems to me like the use-case here is that instead - The user wants to get the file by making these calls

synapseclient/models/wiki.py
```diff
@@ -1 +1,804 @@
...
+            )
+            async for result in results:
+                if result.get("title") == self.title:
+                    matching_header = result
+                    break
+                else:
+                    matching_header = None
+
+            if not matching_header:
+                raise ValueError(f"No wiki page found with title: {self.title}")
```
BryanFauble, line 577: Is this the right way to use an async generator? Wouldn't the correct syntax be:

```suggestion
            async for result in get_wiki_header_tree(
                owner_id=self.owner_id,
                synapse_client=synapse_client,
            ):
                if result.get("title") == self.title:
                    matching_header = result
                    break
                else:
                    matching_header = None

            if not matching_header:
                raise ValueError(f"No wiki page found with title: {self.title}")
```

(The code is not tested)

BryanFauble (changes requested):
This is a great start to the changes!

Feel free to reach out to me with questions, and re-request review when you'd like me to take a look at this PR again.

synapseclient/models/wiki.py
danlu1: I think you are right. I forgot to remove the `await` before `get_wiki_header_tree`. This is my test code and I didn't specify await..
```
async def main():
    results = get_wiki_header_tree(owner_id="syn3380222", limit=5)
    async for result in results:
        if result.get("title") == 'Mutations In Cis':
            print("Found match:", result['id'])
            break
        else:
            matching_header = None
        print(matching_header)

asyncio.run(main())

```

synapseclient/models/wiki.py
```diff
@@ -1 +1,804 @@
...
+            limit=limit,
+            synapse_client=synapse_client,
+        ):
+            headers.append(cls().fill_from_dict(item))
+        return headers
+
+
+@dataclass
+@async_to_sync
+class WikiPage:
```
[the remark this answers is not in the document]
danlu1: Just added the `wikipage_protocol` file.

synapseclient/api/wiki_service.py
danlu1: On it.

synapseclient/models/wiki.py
danlu1: Yes, this is a good point. I will try to address this. Thanks

synapseclient/api/wiki_service.py
```diff
@@ -0,0 +1,464 @@
+"""This module is responsible for exposing the services defined at:
+<https://rest-docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.WikiController2>
+"""
+
+import json
+from typing import TYPE_CHECKING, Any, AsyncGenerator, Dict, List, Optional
+
+if TYPE_CHECKING:
+    from synapseclient import Synapse
```
thomasyu888, line 9: @BryanFauble something to consider when working on the new functional interface, is that currently given this import of `Synapse` class, there will be a circular dependency if we switched `Synapse.getWiki` to use this service.

synapseclient/api/wiki_service.py
thomasyu888: I'd work with @linglp to add this function

synapseclient/models/protocols/wikipage_protocol.py
thomasyu888, line 1: Great addition of the protocols

synapseclient/api/wiki_service.py
BryanFauble: By putting this behind the `if TYPE_CHECKING` it won't be a circular dependency. This is why we have a bunch of the the lazy imports in the code like:

```

```

I agree that it's unfortunate to have to deal with this issue, but there are ways to work around it.

synapseclient/api/wiki_service.py
danlu1: @BryanFauble @thomasyu888 I saw @linglp add the rest_post_paginated to a new file api_client.py under api folder, do you want me to add there?

danlu1: Sounds good. Thanks for confirming.

docs/tutorials/python/wiki.md
```diff
@@ -1,2 +1,142 @@
...
+```
+
+## 6. WikiOrderHint - Managing Wiki Order
+### Get wiki order hint (No id_list returned, same result getting from direct endpoint calls)
+```python
+{!docs/tutorials/python/tutorial_scripts/wiki.py!lines=212-213}
+```
+### Update wiki order hint
+```python
+{!docs/tutorials/python/tutorial_scripts/wiki.py!lines=216-222}
```
danlu1, line 130: TODO: A screenshot of the example code will be added here.

docs/tutorials/python/wiki.md
thomasyu888, line 1: Fantastic tutorial!

thomasyu888 (commented):
🔥 This is great @danlu1! I'll defer to @BryanFauble for final review here. One question I had is whether we need integration tests, but I took a close look at the unit tests and they look good for that function.

We do seem to be missing unit and integration tests for the wiki service itself...

danlu1:
> 🔥 This is great @danlu1! I'll defer to @BryanFauble for final review here. One question I had is whether we need integration tests, but I took a close look at the unit tests and they look good for that function.
> 
> We do seem to be missing unit and integration tests for the wiki service itself...

Yes, both the unit and integration test scripts are still wip.

docs/tutorials/python/tutorial_scripts/wiki.py
```diff
@@ -0,0 +1,284 @@
+#!/usr/bin/env python3
+"""
+Tutorial script demonstrating the Synapse Wiki models functionality.
+
+This script shows how to:
+1. Create, read, and update wiki pages
+2. Work with WikiPage Markdown
```
BryanFauble, line 7: Is there another descriptive way we can work some of these actions rather than "Work with.." maybe, describing the create/update/delete/restore actions taking place during that section of code?

docs/tutorials/python/tutorial_scripts/wiki.py
```diff
@@ -0,0 +1,284 @@
...
+    id=root_wiki_page.id,
+).store()
+
+# Restore the wiki page to the original version
+wiki_page_restored = WikiPage(
+    owner_id=my_test_project.id, id=root_wiki_page.id, wiki_version="0"
+).restore()
+
+# check if the content is restored
+comparisons = [
```
BryanFauble, line 68: It is probably easier and cleaner to use assert statements here. Like


`assert condition, "Message for failure"`

That was if it does fail, it's clear which one failed the compare check

docs/tutorials/python/tutorial_scripts/wiki.py
```diff
@@ -0,0 +1,284 @@
...
+sub_wiki_2 = WikiPage(
+    owner_id=my_test_project.id,
+    parent_id=root_wiki_page.id,
+    title="Sub Page 2 created from markdown text",
+    markdown=markdown_content,
+).store()
+
+# Create a wiki page from a markdown file
+# Create a temporary markdown gzipped file from the markdown_content
+markdown_file_path = "temp_markdown_file.md.gz"
```
BryanFauble, line 133: In most of the tutorials we have been using the "~/temp` directory to write files to

docs/tutorials/python/tutorial_scripts/wiki.py
```diff
@@ -0,0 +1,284 @@
...
+# clean up the downloaded markdown file
+os.remove(markdown_file_path)
+
+# Section 3: WikiPage with Attachments
+# Create a temporary file for the attachment
+attachment_file_name = "temp_attachment.txt.gz"
+with gzip.open(attachment_file_name, "wt", encoding="utf-8") as gz:
+    gz.write("This is a sample attachment.")
+
+# reformat '.' and '_' in the attachment file name to be a valid attachment path
```
BryanFauble, line 181: It's a bit unfortunate that we have to do this client side. I wonder if there is a better way that we can do this though. Maybe a static function on WikiPage that we can point to.

Is it possible that we take the string from the `markdown` field and automatically convert these? I imagine that `.` and `_` are not the only characters that have to be escaped.

docs/tutorials/python/wiki.md
```diff
@@ -1,2 +1,152 @@
...
-![Under Construction](../../assets/under_construction.png)
+
+# Synapse Wiki Models Tutorial
+
+This tutorial demonstrates how to work with Wiki models in the Synapse Python client. Wikis in Synapse provide a way to create rich documentation and collaborative content for projects, folders, files, datasets, and other entities.
+
+## Overview
+
+The Synapse Wiki models include:
+- **WikiPage**: The main wiki page model for creating and managing wiki content
```
BryanFauble, line 10: Could these also hyperlink to the relevant api doc pages for each model?

docs/tutorials/python/wiki.md
```diff
@@ -1,2 +1,152 @@
...
+
+## Basic Setup
+```python
+{!docs/tutorials/python/tutorial_scripts/wiki.py!lines=14-29}
+```
+
+## 1. Create, read, and update wiki pages
+### Create a new wiki page for the project with plain text markdown
+```python
+{!docs/tutorials/python/tutorial_scripts/wiki.py!lines=33-37}
```
BryanFauble, line 32: Iine 32 is missing, that is where the project variable is created

docs/tutorials/python/wiki.md
```diff
@@ -1,2 +1,152 @@
...
+```
+
+### OR you can create a wiki page with an existing markdown file
+```python
+{!docs/tutorials/python/tutorial_scripts/wiki.py!lines=40-45}
+```
+
```
BryanFauble, line 39: A great addition would be to include a section that can be expanded to show what the console logs would be after running a particular step.

There are some examples on other tutorials that can be followed

synapseclient/api/wiki_service.py
```diff
@@ -0,0 +1,456 @@
...
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Update a wiki page.
+    <https://rest-docs.synapse.org/rest/PUT/entity/ownerId/wiki2/wikiId.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        request: The updated wiki page.
```
BryanFauble, line 95: Could there be a link here to the rest api doc page for the exact object/fields that `request` is supposed to be?

synapseclient/api/wiki_service.py
```diff
@@ -0,0 +1,456 @@
...
+    return await client.rest_post_async(
+        uri=f"/entity/{owner_id}/wiki2",
+        body=json.dumps(request),
+    )
+
+
+async def get_wiki_page(
+    owner_id: str,
+    *,
+    wiki_id: Optional[str] = None,
```
BryanFauble, line 43: Nit: Adding these as kwargs instead of positional args is inconsistent with the other functions you have.

synapseclient/api/wiki_service.py
```diff
@@ -0,0 +1,456 @@
...
+        offset: The index of the pagination offset. For a page size of 10, the first page would be at offset = 0,
+               and the second page would be at offset = 10. Default is 0.
+        limit: Limits the size of the page returned. For example, a page size of 10 requires limit = 10.
+               Limit must be 50 or less. Default is 20.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        A generator over the wiki header tree for the entity. The tree contains the hierarchy of wiki pages
```
BryanFauble, line 189: For all of these functions where we are returning stuff like a dictionary could you linked to the rest API doc that shows the fields that are expected?

synapseclient/api/wiki_service.py
```diff
@@ -0,0 +1,456 @@
...
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """Get the preview of a wiki page attachment.
+    <https://rest-docs.synapse.org/rest/GET/entity/ownerId/wiki2/wikiId/attachmentpreview.html>
+
+    Arguments:
+        owner_id: The ID of the owner entity.
+        wiki_id: The ID of the wiki.
+        file_name: The name of the file to get.
+                The file names can be found in the FileHandles from the GET /entity/{ownerId}/wiki/{wikiId}/attachmenthandles method.
```
BryanFauble, line 338: Rather than mention what endpoint would be used here, the better method would be to mention the function one would use. This example format https://github.com/Sage-Bionetworks/synapsePythonClient/blob/d9998ef905aa98a40ed98a5590df4bd314c51ccb/synapseclient/models/entityview.py#L140 shows how the mkdocs can automatically create a hyperlink to the function when it gets rendered on the docs site

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+        *,
+        offset: int = 0,
+        limit: int = 20,
+        synapse_client: Optional["Synapse"] = None,
+    ) -> List["WikiHeader"]:
+        """
+        Get the header tree (hierarchy) of wiki pages for an entity.
+
+        Arguments:
+            owner_id: The ID of the owner entity.
```
BryanFauble, line 291: Should this mention that this is the syn### id?

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+    title: Optional[str] = None
+    """The title of this page."""
+
+    parent_id: Optional[str] = None
+    """When set, the WikiPage is a sub-page of the indicated parent WikiPage."""
+
+    markdown: Optional[str] = None
+    """The markdown content of this page."""
+
+    attachments: List[Dict[str, Any]] = field(default_factory=list)
```
BryanFauble, line 353: Instead of returning a Dict here, could we instead return back a WikiAttachment dataclass? If one doesn't exist you'd need to create it

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+            )
+            try:
+                # Upload the gzipped file to get a file handle
+                file_handle = await upload_file_handle(
+                    syn=client,
+                    parent_entity_id=self.owner_id,
+                    path=file_path,
+                )
+
+                client.logger.info(
```
BryanFauble, line 536: If a TQDM progress bar is already showing up for the regular upload process this can be bumped down to a debug level log message.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+        Arguments:
+            synapse_client: If not passed in and caching was not disabled by
+                    `Synapse.allow_client_caching(False)` this will use the last created
+                    instance from the Synapse class constructor.
+
+        Returns:
+            The created wiki page.
+
+        Raises:
+            ValueError: If owner_id is not provided or if required fields are missing.
```
BryanFauble, line 518: Each of the method for both the async and non-async functions should have at least 1 example to show how to use and call the function, along with the required imports. Ideally one could copy/paste the code, minimally change something, and run it.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+            filehandle_dict = await get_attachment_handles(
+                owner_id=self.owner_id,
+                wiki_id=self.id,
+                wiki_version=self.wiki_version,
+                synapse_client=client,
+            )
+            # check the file_size
+            file_size = int(WikiPage._get_file_size(filehandle_dict, file_name))
+            # use single thread download if file size < 8 MiB
+            if file_size < 8388608:
```
BryanFauble, line 860: I think there is a constant you can import instead of using a specific number on this line.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+            redirect=redirect,
+            synapse_client=client,
+        )
+
+        if download_file:
+            if not download_location:
+                raise ValueError("Must provide download_location to download a file.")
+
+            # construct PresignedUrlInfo for downloading
+            presigned_url_info = PresignedUrlInfo(
```
BryanFauble, line 846: It doesn't seem like you need to construct this object because you turn around and directly use the attributes off the object to pass to subsequent function calls. You could just directly pass those values to the function calls.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+            client.logger.debug(
+                f"Downloaded file {presigned_url_info.file_name} to {download_location}"
+            )
+        else:
+            return attachment_url
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Get_Attachment_Preview_URL: Owner ID {self.owner_id}, Wiki ID {self.id}, File Name {kwargs['file_name']}"
+    )
+    async def get_attachment_preview_async(
```
BryanFauble, line 880: Same comments from the above method

BryanFauble (changes requested):
Great progress! Please re-request review when you are ready for another round.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+                    )
+                else:
+                    raise e
+            else:
+                client.logger.info(
+                    "A wiki page already exists within the owner. Update the existing wiki page."
+                )
+                # Update the existing wiki page
+                if not (self.id):
+                    raise ValueError("Must provide id to update a wiki page.")
```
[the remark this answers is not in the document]
danlu1: This `id` is required only if we want to update an existing wiki page meaning that we get results from `WikiHeader.get` and there is no `parent_id` given. Putting it at the beginning would prevent us from creating a root wikipage for the owner entity.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+
+                # retrieve the wiki page
+                existing_wiki = await get_wiki_page(
+                    owner_id=self.owner_id,
+                    wiki_id=self.id,
+                    wiki_version=self.wiki_version,
+                )
+                # Update existing_wiki with current object's attributes if they are not None
```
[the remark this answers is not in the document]
danlu1: On it.

synapseclient/models/wiki.py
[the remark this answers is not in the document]
BryanFauble: That's correct—however, my concern is that if any of these ValueError exceptions are raised after file uploads have started, the effort and resources spent uploading files to Synapse could be wasted, or could even result in duplicate uploads if this function is called again. Ideally, all necessary validation (such as checking required IDs) should occur before any file uploads take place. This way, we can ensure that we only upload files if we're certain we'll proceed to the post_wiki or put_wiki_page calls, reducing unnecessary uploads and potential inconsistencies.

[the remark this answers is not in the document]
danlu1: Gotcha. I will address this in the next commit.

synapseclient/models/wiki.py
danlu1: Done.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+        if not isinstance(wiki_content, str):
+            raise SyntaxError(f"Expected a string, got {type(wiki_content).__name__}")
+        # Get the cache directory path to save the newly created gzipped file
+        cache_dir = os.path.join(synapse_client.cache.cache_root_dir, "wiki_content")
+        if not os.path.exists(cache_dir):
+            os.makedirs(cache_dir)
+
+        # Check if markdown looks like a file path and exists
+        if os.path.isfile(wiki_content):
+            # If it's already a gzipped file, save a copy to the cache
```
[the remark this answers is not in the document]
danlu1: Ya, there should be better idea but I copied because I would like to call os.remove(file_path) after getting the file_handles. By doing this, I remove files used to get file_handles all at once without needing to distinguish which is user's file and which is the temp file.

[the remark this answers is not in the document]
danlu1: Yes, this is a lazy way to output file_path and make sure I removed all files that are used for getting file handle all at once without distinguishing file types.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1090 @@
...
+                wiki_id=self.id,
+                wiki_version=self.wiki_version,
+            )
+            # convert to dataclass
+            existing_wiki = WikiPage()
+            existing_wiki = existing_wiki.fill_from_dict(
+                synapse_wiki=existing_wiki_dict
+            )
+            # Update existing_wiki with current object's attributes if they are not None
+            updated_wiki = merge_dataclass_entities(
```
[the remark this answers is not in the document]
BryanFauble: Ah, you have `source` and `destination` flipped.

`destination` is going to be `self` and source is `existing_wiki`.

Basically we want to copy any fields the user is not setting on `self`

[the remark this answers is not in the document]
danlu1: Thanks, I will give it a try. In my original method, I was using self to update existing_wiki.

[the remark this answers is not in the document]
danlu1: @BryanFauble I re-run with source and destination flipped but run into 
```
(Pdb) existing_wiki
WikiPage(id='633249', etag='65a51e2d-8696-4c76-b153-d7a3c76e960b', title='My Root Wiki Page', parent_id=None, markdown=None, attachments=[], owner_id=None, created_on='2025-06-30T20:00:55.562Z', created_by='3434599', modified_on='2025-06-30T20:00:55.562Z', modified_by='3434599', wiki_version=None, markdown_file_handle_id='159595996', attachment_file_handle_ids=[])
(Pdb) self
WikiPage(id='633249', etag=None, title='My First Root Wiki Page NEW', parent_id=None, markdown='# Welcome to My Root Wiki NEW\n\nThis is a sample root wiki page created with the Synapse client.', attachments=[], owner_id='syn68555989', created_on=None, created_by=None, modified_on=None, modified_by=None, wiki_version=None, markdown_file_handle_id='159596059', attachment_file_handle_ids=[])
(Pdb) self
WikiPage(id='633249', etag=None, title='My First Root Wiki Page NEW', parent_id=None, markdown='# Welcome to My Root Wiki NEW\n\nThis is a sample root wiki page created with the Synapse client.', attachments=[], owner_id='syn68555989', created_on=None, created_by=None, modified_on=None, modified_by=None, wiki_version=None, markdown_file_handle_id='159596059', attachment_file_handle_ids=[])
(Pdb) updated_wiki
WikiPage(id='633249', etag=None, title='My First Root Wiki Page NEW', parent_id=None, markdown='# Welcome to My Root Wiki NEW\n\nThis is a sample root wiki page created with the Synapse client.', attachments=[], owner_id='syn68555989', created_on=None, created_by=None, modified_on=None, modified_by=None, wiki_version=None, markdown_file_handle_id='159596059', attachment_file_handle_ids=[])
(Pdb) c
  File "<stdin>", line 6, in <module>
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/core/async_utils.py", line 72, in f
    return self.fn(obj, *args, **kwds)
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/core/async_utils.py", line 129, in newmethod
    return asyncio.run(wrapper(*args, **kwargs))
  File "/Users/danlu/.pyenv/versions/3.10.0/lib/python3.10/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Users/danlu/.pyenv/versions/3.10.0/lib/python3.10/asyncio/base_events.py", line 641, in run_until_complete
    return future.result()
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/core/async_utils.py", line 116, in wrapper
    return await getattr(self, async_method_name)(*args, **kwargs)
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/core/async_utils.py", line 56, in otel_trace_method_wrapper
    return await func(self, *arg, **kwargs)
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/models/wiki.py", line 681, in store_async
    wiki_data = await put_wiki_page(
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/api/wiki_service.py", line 106, in put_wiki_page
    return await client.rest_put_async(
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/client.py", line 6663, in rest_put_async
    response = await self._rest_call_async(
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/client.py", line 6553, in _rest_call_async
    self._handle_httpx_synapse_http_error(response)
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/client.py", line 6238, in _handle_httpx_synapse_http_error
    exceptions._raise_for_status_httpx(
  File "/Users/danlu/Sage/synapsePythonClient/synapseclient/core/exceptions.py", line 296, in _raise_for_status_httpx
synapseclient.core.exceptions.SynapseHTTPError: 412 Client Error: ObjectId: syn68555989 was updated since you last fetched it, retrieve it again and reapply the update
```
I guess the reason is self.etag is None. I recalled that is the reason I used self to update existing_wiki.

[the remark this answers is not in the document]
danlu1: Right. Nice catch!

synapseclient/models/wiki.py
danlu1: Here the attachments just a list of file path. For attachment-related endpoints, only url is given. Do you still think this is needed?

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+
+        # Convert attachments to gzipped file if needed
+        if self.attachments:
+            try:
+                file_handles = []
+                for attachment in self.attachments:
+                    file_path = self._to_gzip_file(
+                        wiki_content=attachment, synapse_client=client
+                    )
+                    file_handle = await upload_file_handle(
```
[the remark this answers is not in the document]
danlu1: Added `gather` function to run tasks in parallel since I would like to get the file handle the same order as the given file path.

synapseclient/models/wiki.py
BryanFauble: Looking at the Rest API Docs for the `V2WikiPage`: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/v2/wiki/V2WikiPage.html

[On these lines](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/e56540eb29e272a551fa1a339997e642936bd53f/synapseclient/models/wiki.py#L397-L413) you are setting:

1. `markdown`
2. `attachments`
3. `owner_id`
4. `wiki_version`

But - These fields don't exist on the Rest API docs. When you call the `fill_from_dict`, these would always be wiped out and replaced with `None` or `[]`, since Synapse would never return these values.

I understand how these fields can be used during the store operations - But, how would a user read for these values?

synapseclient/models/wiki.py
```diff
@@ -1 +1,1018 @@
...
+        if not file_name:
+            raise ValueError("Must provide file_name to get attachment URL.")
+
+        client = Synapse.get_client(synapse_client=synapse_client)
+        attachment_url = await get_attachment_url(
+            owner_id=self.owner_id,
+            wiki_id=self.id,
+            file_name=file_name,
+            wiki_version=self.wiki_version,
+            redirect=redirect,
```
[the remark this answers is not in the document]
danlu1: Resolved.

synapseclient/models/wiki.py
```diff
@@ -480,6 +478,46 @@ def _to_gzip_file(
 
         return file_path
 
+    def _unzip_gzipped_file(self, file_path: str) -> str:
```
danlu1: This function is added to unzip markdown file.

synapseclient/models/wiki.py
```diff
@@ -498,6 +536,20 @@ def _get_file_size(filehandle_dict: dict, file_name: str) -> str:
             f"File {file_name} not found in filehandle_dict. Available files: {available_files}"
         )
 
+    @staticmethod
+    def _reformat_attachment_file_name(attachment_file_name: str) -> str:
```
danlu1: Add this helper function to assist embedding attachment files in markdown.

synapseclient/models/wiki.py
```diff
@@ -516,16 +568,19 @@ async def _get_markdown_file_handle(self, synapse_client: Synapse) -> "WikiPage"
...
-                    syn=synapse_client,
-                    parent_entity_id=self.owner_id,
-                    path=file_path,
-                )
-                synapse_client.logger.debug(
-                    f"Uploaded file handle {file_handle.get('id')} for wiki page markdown."
-                )
-                # Set the markdown file handle ID from the upload response
-                self.markdown_file_handle_id = file_handle.get("id")
+                async with synapse_client._get_parallel_file_transfer_semaphore(
```
danlu1: Add _get_parallel_file_transfer_semaphore to run upload_file_handle.

synapseclient/models/wiki.py
danlu1: On it

docs/tutorials/python/wiki.md
```diff
@@ -1,2 +1,152 @@
 # Wikis on Projects
-![Under Construction](../../assets/under_construction.png)
+
```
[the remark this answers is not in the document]
danlu1: I uncommented the line of wiki in mkdocs.yml. Do I need to update `reference/wiki.md`?

synapseclient/api/wiki_service.py
danlu1: Done

docs/reference/experimental/async/wiki.md
```diff
@@ -0,0 +1,36 @@
...
+        inherited_members: true
+        members:
+        - store_async
+        - restore_async
+        - get_async
+        - delete_async
+        - get_attachment_handles_async
+        - get_attachment_async
+        - get_attachment_preview_async
+        - get_markdown_file_async
```
BryanFauble: `reformat_attachment_file_name` is mentioned in the docs and should be included in this list. `unzip_gzipped_file` is not private, should that be included here?

synapseclient/client.py
```diff
@@ -9465,6 +9465,37 @@ async def rest_delete_async(
...
+            page = await self.rest_get_async(
+                uri=uri,
+                params=params,
+            )
+            results = page["results"] if "results" in page else page["children"]
+            prev_num_results = len(results)
+
+            for result in results:
+                offset += 1
+                yield result
```
BryanFauble: This function is not needed, please revert this and swap to using this call which already exists:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/be5586ea46078f4ec5c70dce1b5d10bc9312505c/synapseclient/api/api_client.py#L64

synapseclient/models/protocols/wikipage_protocol.py
```diff
@@ -0,0 +1,345 @@
...
+                WikiOrderHint,
+            )
+            syn = Synapse()
+            syn.login()
+
+            project = Project(name="My uniquely named project about Alzheimer's Disease").get()
+            # get the WikiOrderHint for the project
+            wiki_order_hint = WikiOrderHint(owner_id=project.id).get()
+
+            print(wiki_order_hint)
```
BryanFauble: The formatting of every example that you've added is a bit broken:

Image


```suggestion
        Example: Get the WikiOrderHint for a project
            This example shows how to get a WikiOrderHint for existing wiki pages in a project.
            
            ```python
            from synapseclient.models import (
                Project,
                WikiOrderHint,
            )
            syn = Synapse()
            syn.login()

            project = Project(name="My uniquely named project about Alzheimer's Disease").get()
            # get the WikiOrderHint for the project
            wiki_order_hint = WikiOrderHint(owner_id=project.id).get()

            print(wiki_order_hint)
            ```
```

The code needs to be wrapped in backticks

synapseclient/models/wiki.py
```diff
@@ -1 +1,1278 @@
...
+        else:
+
+            async def task_of_uploading_attachment(attachment: str) -> tuple[str, str]:
+                """Process a single attachment and return its file handle ID and cache directory."""
+                # if the attachment is not a gzipped file or image file, gzip it
+                if (
+                    not attachment.endswith(".gz")
+                    and not attachment.endswith(".png")
+                    and not attachment.endswith(".jpg")
+                    and not attachment.endswith(".jpeg")
```
BryanFauble: Is there a list of these somewhere, or is this derived from documentation on the rest-docs? I am curious if there are additional image formats to add here.

synapseclient/models/protocols/wikipage_protocol.py
danlu1: Will do. @BryanFauble qq. how can I render the docs locally?

synapseclient/models/wiki.py
```diff
@@ -1 +1,1278 @@
...
+            # Update self with the returned WikiPage objects that have file handle IDs set
+            self = await self._get_markdown_file_handle(synapse_client=client)
+            self = await self._get_attachment_file_handles(synapse_client=client)
+
+        # Handle root wiki page creation if parent_id is not given
+        if wiki_action == "create_root_wiki_page":
+            client.logger.info(
+                "No wiki page exists within the owner. Create a new wiki page."
+            )
+            # Create the wiki page
```
BryanFauble: nit: Lighten up on the number of in-line comments that don't expand beyond what the code is clearly completed. In this case `Create the wiki page`, doesn't add much information, but it adds a marginal maintenance burden to update it later.

IMO: Let the code speak for itself. If the code is complex enough that it needs an in-line comment, that is an indicator that the code could be structured a bit better to make it more clear.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1278 @@
...
+            wiki_data = await post_wiki_page(
+                owner_id=self.owner_id,
+                request=self.to_synapse_request(),
+                synapse_client=client,
+            )
+            client.logger.info(
+                f"Created wiki page: {wiki_data.get('title')} with ID: {wiki_data.get('id')}."
+            )
+        if wiki_action == "update_existing_wiki_page":
```
BryanFauble: nit: There can be only a single `wiki_action`, so I could place these into an `elif` statement

synapseclient/models/wiki.py
```diff
@@ -1 +1,1278 @@
...
+        if not file_name:
+            raise ValueError("Must provide file_name to get attachment URL.")
+
+        client = Synapse.get_client(synapse_client=synapse_client)
+        # check if the file_name is in gzip format or image format
+        if (
+            not file_name.endswith(".gz")
+            and not file_name.endswith(".png")
+            and not file_name.endswith(".jpg")
+            and not file_name.endswith(".jpeg")
```
BryanFauble: nit: This condition is re-used in more than 1 location. It should be extracted to a single function that is re-used in both places to prevent a possibility that they get out of sync with one another.

tests/integration/synapseclient/models/async/test_wiki_async.py
```diff
@@ -0,0 +1,751 @@
...
+    wiki_title = f"Root Wiki Page"
+    wiki_markdown = "# Root Wiki Page\n\nThis is a root wiki page."
+
+    wiki_page = WikiPage(
+        owner_id=project_model.id,
+        title=wiki_title,
+        markdown=wiki_markdown,
+    )
+    root_wiki = await wiki_page.store_async(synapse_client=syn)
+    schedule_for_cleanup(root_wiki.id)
```
BryanFauble: Does this work? I know I had to add some specific logic for a few class types to work properly with the `clean` function

tests/integration/synapseclient/models/async/test_wiki_async.py
```diff
@@ -0,0 +1,751 @@
...
+    async def test_get_attachment_preview_url_gz_file(
+        self,
+        wiki_page_with_gz_attachment: tuple[WikiPage, str],
+        schedule_for_cleanup: Callable[..., None],
+    ) -> None:
+        """Test getting attachment preview URL for a gz file."""
+        # GIVEN a wiki page with a gz attachment
+        wiki_page, attachment_name = wiki_page_with_gz_attachment
+        # Sleep for 0.5 minutes to ensure the attachment preview is created
+        time.sleep(0.5 * 60)
```
BryanFauble: When the test is in an asyncio context please use:

`await asyncio.sleep(0.5 * 60)` instead.

synapseclient/models/wiki.py
```diff
@@ -1 +1,1278 @@
...
+        snapshots = []
+        async for item in get_wiki_history(
+            owner_id=owner_id,
+            wiki_id=id,  # use id instead of wiki_id to match other classes
+            offset=offset,
+            limit=limit,
+            synapse_client=synapse_client,
+        ):
+            snapshots.append(cls().fill_from_dict(item))
+        return snapshots
```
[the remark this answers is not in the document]
danlu1: It depends on how many versions a wiki page can have. I will tweak it.

BryanFauble (changes requested):
Thank you for this round. There are a few things to patch up, but this is in a great place

synapseclient/models/wiki.py
```diff
@@ -1 +1,1278 @@
...
+            raise ValueError("Must provide owner_id to get wiki header tree.")
+        headers = []
+        async for item in get_wiki_header_tree(
+            owner_id=owner_id,
+            offset=offset,
+            limit=limit,
+            synapse_client=synapse_client,
+        ):
+            headers.append(cls().fill_from_dict(item))
+        return headers
```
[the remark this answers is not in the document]
danlu1: One benefit I see in using a list here is that the wiki header tree can be viewed as a whole, which makes the hierarchy clearer to the user. What do you think?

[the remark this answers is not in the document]
danlu1: Gotcha.

synapseclient/models/wiki.py
danlu1: Cleaned the redundant comments.

BryanFauble (approved):
Thanks for your hard work on this. It looks good to me!

PR #1209, opened 2025-06-25
[SYNPY-1618] Clean up test resources in Synapse created during integration test runs

BryanFauble:
# **Problem:**

- When integration tests run a failure in another OS or python version causes all test runs to quit out. This can cause resources to continue existing in Synapse and not be cleaned up.

# **Solution:**

- Create some simple test scripts to clean up the resources and run on a weekly cadence to clean up the resources

# **Testing:**

- This Github run was used to verify that the scripts could be run within github: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/15886384891/job/44799278492

### DIFF
--- a/.github/scripts/delete_evaluations.py
+++ b/.github/scripts/delete_evaluations.py
@@ -1,0 +1,70 @@
+import asyncio
+from typing import Set
+
+from synapseclient import Evaluation, Synapse
+
+syn = Synapse()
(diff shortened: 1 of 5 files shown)


thomasyu888 (approved):
🔥 LGTM! Thanks for creating this pipeline

PR #1210, opened 2025-07-01
[SYNPY-1607] Enum conversion from String

BryanFauble:
# **Problem:**

- When defining a ColumnType, or a FacetType you must use the Enum to set these values. This is problematic with the R-client where it won't be possible to create python style classes.

# **Solution:**

- Allow ColumnType and FacetType to be defined from strings, expanding usability

# **Testing:**

- Added integration testing around ColumnType
- I verified that I could install this python branch and hit this logic from R


thomasyu888 (approved):
🔥 LGTM!

PR #1211, opened 2025-07-01
[SYNPY-1599] Patching json schema code

BryanFauble:
# **Problem:**

- Patching a variety of issues around the JSON schema code

# **Solution:**

- Correct sync code as it was erroring out
- Remove some indentation
- Cleaning up examples to be a bit more clear what is going on

# **Testing:**

- Verified that I could run all of the examples defined in the docstring


synapseclient/models/protocols/json_schema_protocol.py
```diff
@@ -624,14 +544,94 @@ def get_invalid_validation(
...
+        )
+
+        # Validate the folder entity against the bound schema with invalid data
+        test_folder.annotations = {"foo": "test_value_new", "bar": "invalid_string"}  # 'bar' should be integer
+        test_folder.store()
+
+        # Get invalid validation results
+        invalid_items = test_folder.get_invalid_validation()
+        for item in invalid_items:
+            print(item)
```
BryanFauble, line 634: @linglp This example doesn't seem to actually work. Am I doing something wrong, or reading the API docs incorrectly?

My understanding after reading https://rest-docs.synapse.org/rest/POST/entity/id/schema/validation/invalid.html is that it won't validate the Folder itself, but rather, the content within the folder.

linglp: Your understanding is correct. That's my mistake. Let me run and fix all these examples

synapseclient/models/protocols/json_schema_protocol.py
```diff
@@ -490,99 +553,26 @@ def get_schema_validation_statistics(
...
-                # Get schema validation statistics
-                stats = test_folder.get_schema_validation_statistics()
-                print('Validation statistics:', stats)
-            ```
-        """
-        return JSONSchemaValidationStatistics()
-
-    def get_invalid_validation(
-        self, *, synapse_client: Optional["Synapse"] = None
-    ) -> Generator["InvalidJSONSchemaValidation", None, None]:
```
BryanFauble, line 560: This function was removed from `synapseclient/models/protocols/json_schema_protocol.py` because it already exists in `synapseclient/models/mixins/json_schema.py`

thomasyu888 (approved):
🔥 LGTM! Thanks for checking all the docstrings - not sure what's happening with the tests but I'll defer to you.

linglp (approved):
LGTM! I just made a minor change to make `FILEPATH` consistent.

PR #1212, opened 2025-07-08
[SYNPY-1426] Deprecate user and team services in Synapse client

BryanFauble:
# **Problem:**

- Team and user method in the Synapse core class have been replaced and need to be deprecated
- A handful of new async methods still used the core Synapse class for it's implementation.

# **Solution:**

- Added asynchronous methods for retrieving user profiles by ID and username.
- Implemented a method to check if a user is certified.
- Deprecated several synchronous methods in the Synapse class, providing migration paths to new asynchronous methods in the UserProfile model.
- Updated the Team model to use new asynchronous API functions for team management (create, delete, get members, invite).
- Use UserGroupHeader to manage user group metadata.
- Enhanced documentation with migration examples for deprecated methods.
- Updated tests to reflect changes in error message formatting for non-existent teams.

# **Testing:**

- Manually testing the suggested migrations as well a unit/integration regression testing via the existing tests

### DIFF
--- a/docs/reference/experimental/async/team.md
+++ b/docs/reference/experimental/async/team.md
@@ -18,3 +18,5 @@ at your own risk.
             - invite_async
             - open_invitations_async
 ---
+
+::: synapseclient.models.TeamMember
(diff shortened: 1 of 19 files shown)


thomasyu888 (approved):
🔥 LGTM! Thanks for adding the team and user services.  I appreciate the deprecation documentation about how to use new services.

Thanks for updating the tests and I see that all the tests passed using the new methods

PR #1213, opened 2025-07-10
[SYNPY-1599] Fix line references in JSON schema tutorial and update syn class usage

BryanFauble:
# **Problem:**

1. client = synapseclient.Synapse().get_client(synapse_client=syn) ← this has to be removed in the tutorial.  Or else it causes the large error below
2. The tempfile creation could be done with the builtin tempfile module because that fails with pathing
3. there is a number of test_folder.get_invalid_validation(synapse_client=syn )  The synapse_client=syn isn’t necessary

# **Solution:**

- Updates to tutorial and line numbers

# **Testing:**

- Verified that I could run the tutorial

### DIFF
--- a/docs/tutorials/python/tutorial_scripts/json_schema.py
+++ b/docs/tutorials/python/tutorial_scripts/json_schema.py
@@ -1,14 +1,13 @@
-import os
 import time
 from pprint import pprint
 
 import synapseclient
+from synapseclient.core.utils import make_bogus_data_file
 from synapseclient.models import File, Folder
 
 # 1. Set up Synapse Python client and retrieve project
 syn = synapseclient.Synapse()
 syn.login()
-client = synapseclient.Synapse().get_client(synapse_client=syn)
 
 # Retrieve test project
 PROJECT_ID = syn.findEntityId(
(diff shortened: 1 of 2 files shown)


thomasyu888 (approved):
🔥 thanks! Lingling is ooo today, so we may want to tag in @jaymedina for a final review. 

Looking great!

docs/tutorials/python/tutorial_scripts/json_schema.py
```diff
@@ -101,69 +100,47 @@
...
-    """Create a random file with random data.
-
-    :param path: The path to create the file at.
-    """
-    with open(path, "wb") as f:
-        f.write(os.urandom(1))
-
-
-create_random_file(path_to_file)
+path_to_file = make_bogus_data_file(n=5)
```
jaymedina, line 129: out-of-scope, but do we want to rename the parameter `n` in this helper function to something more descriptive?

BryanFauble: Users shouldn't be using this in any regular code. The argument is explained well enough that I don't see us needing to make a change: https://python-docs.synapse.org/en/stable/reference/core/#synapseclient.core.utils.make_bogus_data_file

Ideally users would be replacing this line to point to a file they want to upload.

This disclaimer is also on that doc page:
`This section is for super users / developers only. These functions are subject to change as they are internal development functions. Use at your own risk.`

PR #1215, opened 2025-07-11
[SYNPY-1426] Deprecate activity services in Synapse client

BryanFauble:
# **Problem:**

- Activity methods in the Synapse core class have been replaced and need to be deprecated
- A handful of new async methods still used the core Synapse class for it's implementation.

# **Solution:**

- Enhance Activity Protocol with Parent ID Support and Retrieval Methods
- Updated ActivitySynchronousProtocol to accept EntityView and Dataset as valid parent types.
- Added support for string IDs as parent entities in store, delete, and disassociate methods.
- Introduced a new method `get` to retrieve activities by activity ID or parent entity ID, with version support.
- Enhanced integration tests for Activity to cover new retrieval methods and string parent ID functionality.
- Improved test cases to ensure proper handling of references and activity properties during storage and retrieval.

# **Testing:**

- Manually verifying the migration suggestion works when pointing at real Synapse objects
- Integration tests


synapseclient/client.py
```diff
@@ -3468,6 +3516,34 @@ def setProvenance(
...
+            syn.login()
+
+            # Create an activity
+            activity = Activity(
+                name="Analysis",
+                description="Data processing"
+            )
+
+            # Store the activity and associate it with the entity
+            stored_activity = activity.store(parent="syn123")
```
thomasyu888, line 3544: From a product perspective, should this be an "entity" function? I'm taking a look at the API and I'm wondering if this should be 

`entity.store_activity(Activity)` instead of `Activity.store(parent="syn2222")`

My understanding is that you can create activity objects without actually storing it at all to an entity (not that that's particularly helpful.

I believe this can be a separate ticket

BryanFauble: It is already available within the entities. See the `activity` field on https://python-docs.synapse.org/en/stable/reference/experimental/sync/file/?h=activity#synapseclient.models.File . What would occur is that you'd set your Activity object on the entity, and then store the entity. No specific method required like `store_activity`. Using this `activity.store` was a way to provide a 1-to-1 mapping for functionality in the event there are folks that have code that work this way instead.

synapseclient/api/entity_services.py
```diff
@@ -542,3 +543,327 @@ async def get_entities_by_md5(
...
     )
+
+
+async def get_entity_provenance(
+    entity_id: str,
+    version_number: Optional[int] = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Any]:
+    """
```
andrewelamb, line 554: Should the dict this returns be a dataclass?

synapseclient/api/entity_services.py
```diff
@@ -542,3 +543,327 @@ async def get_entities_by_md5(
...
+        uri = f"/entity/{syn_id}/version/{syn_version}/generatedBy"
+    else:
+        uri = f"/entity/{syn_id}/generatedBy"
+
+    return await client.rest_get_async(uri=uri)
+
+
+async def set_entity_provenance(
+    entity_id: str,
+    activity: Dict[str, Any],
```
andrewelamb, line 623: Same as above

synapseclient/models/activity.py
```diff
@@ -46,6 +53,27 @@ def format_for_manifest(self) -> str:
...
+            A dictionary representation matching the Synapse REST API format.
+        """
+        return {
+            "concreteType": USED_ENTITY,
+            "reference": {
+                "targetId": self.target_id,
+                "targetVersionNumber": self.target_version_number,
+            },
+            "wasExecuted": was_executed,
+        }
```
andrewelamb, line 75: Would it make sense for this to be a `TypedDict` (to enforce the keys)?

andrewelamb (approved):

synapseclient/api/entity_services.py
BryanFauble: Not in this case. The responsibility for these API modules has been to "talk" in the same language as the Rest API. So in this case it's to accept and return back the dictionaries that the API would accept/return. Then - Where this is being used within the Dataclass (Activity) is where it is usually ported into a Dataclass.

PR #1218, opened 2025-07-14
[SYNPY-1426] Deprecate column method calls

BryanFauble:
# **Problem:**

- There are a few column related method in the Synapse class that has been migrated away and needs to be deprecated.
- Several API function calls needed to be ported over to the use the HTTPX client

# **Solution:**

- Deprecate `getTableColumns`, `getColumns`, `getColumn` methods on the Synapse class
- Implementing several new methods for the Column to ensure that all functionality is still available to use such as listing all columns and searching by prefix

# **Testing:**

- I verified that I could run the migration examples, I added new integration tests, and existing tests should be passing the logic.

### DIFF
--- a/tests/integration/synapseclient/models/synchronous/test_column.py
+++ b/tests/integration/synapseclient/models/synchronous/test_column.py
@@ -135,7 +135,9 @@ class TestColumn:
         assert "test_prefix_column_filter_col2" in column_names
         assert "other_col" not in column_names
 
-    async def test_list_columns_with_limit_and_offset(self, project_model: Project) -> None:
+    async def test_list_columns_with_limit_and_offset(
+        self, project_model: Project
+    ) -> None:
         """Test listing columns with limit and offset parameters to verify pagination behavior."""
         # GIVEN a single table with multiple columns
         table_name = str(uuid.uuid4())


synapseclient/api/table_services.py
```diff
@@ -123,6 +123,97 @@ async def get_columns(
...
+    Arguments:
+        column_id: The ID of the column to retrieve.
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The column matching <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/ColumnModel.html>
+    """
+    from synapseclient import Synapse
```
andrewelamb, line 142: Is this to avoid a circular dependency?

synapseclient/api/table_services.py
```diff
@@ -123,6 +123,97 @@ async def get_columns(
...
+
+async def list_columns(
+    prefix: Optional[str] = None,
+    limit: int = 100,
+    offset: int = 0,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> AsyncGenerator["Column", None]:
+    """List columns with optional prefix filtering.
+
```
andrewelamb, line 157: It's not clear (to me) what this function does based on the docstring. Are these all the columns from a table?

synapseclient/api/table_services.py
```diff
@@ -123,6 +123,97 @@ async def get_columns(
...
+def list_columns_sync(
+    prefix: Optional[str] = None,
+    limit: int = 100,
+    offset: int = 0,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Generator["Column", None, None]:
+    """List columns with optional prefix filtering (synchronous version).
+
```
andrewelamb, line 191: Same as above

synapseclient/api/table_services.py
BryanFauble: Not sure how you'd like me to update this. This function will list columns stored in Synapse. Synapse stores columns independently from tables. This is the same as how Activities in Synapse works. Tables then reference the IDs of columns to say "This table is using these columns"/

synapseclient/api/table_services.py
BryanFauble: Yes

PR #1219, opened 2025-07-15
[SYNPY-1426] Deprecate move method

BryanFauble:
# **Problem:**

- The move method needs to be deprecated

# **Solution:**

- There is replacement functionality for the move method available on all of the dataclasses which have a `parent_id` attribute. By setting this attribute to a new value and storing the entity - it will move the entity

# **Testing:**

- Verified that I could move all of the entities to a Folder parent

### DIFF
--- a/docs/news.md
+++ b/docs/news.md
@@ -9,6 +9,99 @@ detailing some of the changes.
 the 4.x.x versions hidden behind optional feature flags or different import paths. Any
 breaking changes will not be included until v5.0.
 
+## 4.9.0 (2025-07-09)
+
+## Highlights
+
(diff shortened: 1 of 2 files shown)


linglp (commented):
LGTM too! But I am only seeing tests like [async def test_store_move_file](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/848a7c6c7eab895065e503cca6e96060373b1437/tests/integration/synapseclient/models/async/test_file_async.py#L269) to test that the moving works. But I am not seeing other equivalent tests for testing moving other entity types, such as dataset. I am not sure if I missed anything.

andrewelamb (approved):

PR #1221, opened 2025-07-16
[SYNPY-1426] Deprecate annotations

BryanFauble:
# **Problem:**

- Annotations through the previous lass is deprecated along with the methods in the Synapse class

# **Solution:**

- Adding deprecation notices
- Adding migration suggestions to move to new compliant code

# **Testing:**

- I verified that I could run the suggested migration method examples

### DIFF
--- a/docs/news.md
+++ b/docs/news.md
@@ -9,6 +9,99 @@ detailing some of the changes.
 the 4.x.x versions hidden behind optional feature flags or different import paths. Any
 breaking changes will not be included until v5.0.
 
+## 4.9.0 (2025-07-09)
+
+## Highlights
+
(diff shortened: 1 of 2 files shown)


synapseclient/annotations.py
```diff
@@ -284,6 +286,13 @@ def set_privacy(
     raise KeyError('The key "%s" couldn\'t be found in the annotations.' % key)
 
 
+@deprecated(
+    version="4.9.0",
+    reason="Use the dataclass model attributes instead. "
```
linglp, line 291: Nit: For consistency, includes "To be removed in 5.0.0." in reason?

PR #1223, opened 2025-07-19
[SYNPY-1426] Deprecate acl and permission management methods

BryanFauble:
# **Problem:**

- The permission and ACL management methods in the Synapse class are now deprecated
- The methods within the access control mixin were still using the ACL management methods from the Synapse class

# **Solution:**

- Migrate the ACL management methods to API calls that can be used within the access control mixin
- Expand the listing and deleting ACL methods to support other entity types that can be contained with a project and within a folder

# **Testing:**

- I verified that I could run the suggested code in the migration docs
- Expanded test cases to support testing additional entity types


synapseclient/api/user_services.py
```diff
@@ -226,3 +230,107 @@ async def _get_certified_passing_record(
...
+            print(f"User ID for 'username': {user_id}")
+
+        asyncio.run(main())
+        ```
+    """
+    from synapseclient import PUBLIC, Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    if principal_id is None or principal_id == "PUBLIC":
```
thomasyu888, line 283: Nit: are we allowing the "string" here?

synapseclient/api/user_services.py
```diff
@@ -226,3 +230,107 @@ async def _get_certified_passing_record(
...
+        return PUBLIC
+    try:
+        return int(principal_id)
```
thomasyu888, line 286: I'm not sure how often this occurs, but I think usernames can also be integers, which can complicate things

thomasyu888 (approved):
🔥 LGTM!  I read through the tests and the functions make sense to me!  Thanks for doing this! I'll defer to someone else on the team for a final review

synapseclient/api/user_services.py
BryanFauble: That is right that they can be ints. This logic in the API classes was mostly copied from the `client.py` script, for this method:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/babe7857aceef55df267eeb684b6ba801202f862/synapseclient/client.py#L3669-L3707

I decided just to keep these quirks as is.

Where the user is meant to start the chain that leads them to this function clearly indicates that they should be providing the Principal ID, and not the Username:
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/babe7857aceef55df267eeb684b6ba801202f862/synapseclient/models/mixins/access_control.py#L300-L309
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/babe7857aceef55df267eeb684b6ba801202f862/synapseclient/models/mixins/access_control.py#L256-L262

synapseclient/api/entity_services.py
```diff
@@ -964,3 +1245,305 @@ async def main():
 
     async for child in response:
         yield child
+
+
+async def set_entity_permissions(
```
linglp, line 1250: Do we have test for this function?

synapseclient/api/entity_services.py
```diff
@@ -964,3 +1245,305 @@ async def main():
...
+        ):
+            effective_permission_set = effective_permission_set.union(
+                permissions["accessType"]
+            )
+    return list(effective_permission_set)
+
+
+async def update_entity_acl(
+    entity_id: str,
+    acl: Dict[str, Any],
```
linglp, line 1489: Nit: inconsistent with other ACL data types.

synapseclient/api/entity_services.py
```diff
@@ -964,3 +1245,305 @@ async def main():
...
+            {'resourceAccess': [
+                {'accessType': ['READ', 'DOWNLOAD'],
+                 'principalId': 222222}
+            ]}
+        synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+    Returns:
+        The created or updated ACL.
```
linglp, line 1508: Nit: add link to ACL object: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/AccessControlList.html

linglp (commented):
Hi @BryanFauble ! This is a fairly large PR, and I am not deeply familiar with the ACL logic, so it will probably take me about 1 story point to review it thoroughly. I’ve done an initial pass and noticed a few issues with docstrings and type hinting.

Also, since many of the functions return an ACL-like object, would it make sense to define a `@dataclass` to represent the ACL? It could help with readability and type safety across the codebase.

synapseclient/api/entity_services.py
```diff
@@ -964,3 +1245,305 @@ async def main():
...
+
+        syn = Synapse()
+        syn.login()
+
+        async def main():
+            # Grant all registered users download access
+            acl = await set_entity_permissions(
+                entity_id="syn123",
+                principal_id="273948",
+                access_type=["READ", "DOWNLOAD"]
```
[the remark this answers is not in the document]
BryanFauble: READ access must be granted in order for someone to be able to download it.

synapseclient/api/entity_services.py
BryanFauble: This is being implicitly tested by the existing integration tests since it's used in the AccessControllable mixin

synapseclient/api/entity_services.py
```diff
@@ -442,6 +442,235 @@ async def get_entity_acl(
     )
 
 
+async def get_entity_acl_with_benefactor(
+    entity_id: str,
+    check_benefactor: bool = True,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> Dict[str, Union[str, List[Dict[str, Union[int, List[str]]]]]]:
```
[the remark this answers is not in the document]
BryanFauble: This is where I am a bit conflicted. What you are suggesting makes sense and I like it, but also, on the other hand as well - If we make a class for this return type it would need to be an exact replica of the https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/AccessControlList.html class which wouldn't match our `snake_case` formatting for classes. These API functions are in this weird in-between shim layer where we are taking what Synapse is giving us and returning it and THEN we are doing a conversion into our compliant `snake_case` classes.

Here is one of those cases:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/6098eeb15bb18ec452cfb1a310796a29635430bf/synapseclient/models/mixins/access_control.py#L248-L254

We are making a call to `get_entity_permissions` and then turning it into a `Permissions` dataclass.

We've also had plans to use an OpenAPI generated Shim/API layer (That does what these `/api/*` python files are doing) - And that makes a lot of sense to have more strict classing there.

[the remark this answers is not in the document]
andrewelamb: A compromise could be to have this type as a Constant and then use that constant as a type.

```
ComplexType = Dict[str, Union[str, List[Dict[str, Union[int, List[str]]]]]]

def func() -> ComplexType:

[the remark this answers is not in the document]
Comment (linglp) [reply synapseclient/api/entity_services.py:454]: @BryanFaubleI I see what you mean. I actually had to do a similar conversion like `Permissions.from_dict` for the `JSONSchemaBinding` data class as well. In this case, I think it's fine to leave it as-is for now, since users can still refer to the docstring to understand the return object.

File: synapseclient/client.py [hunk 8]
```diff
@@ -3786,10 +3918,21 @@ def getPermissions(
 
         return self.get_acl(entity=entity, principal_id=principal_id)
 
+    @deprecated(
```
linglp, line 3921: I am a bit confused looking at the message. Based on my understanding of the old `ger_permissions` method, it seems like it is calling`f"/entity/{entity_id}/permissions"` which is the same as the new `get_entity_permission`. So if we deprecate the old get_permission method, shouldn't that the method to replace it will be `get_entity_permission`?

linglp (commented):
It looks good to me overall! Just some minor comments.

synapseclient/api/team_services.py
```diff
@@ -109,6 +109,34 @@ async def delete_team(
     await client.rest_delete_async(uri=f"/team/{id}")
 
 
+async def get_teams_for_user(
+    user_id: str,
+    *,
+    synapse_client: Optional["Synapse"] = None,
```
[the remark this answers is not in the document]
BryanFauble: No. Even if auth is not required we still need to have a Synapse class created and used. The user just would not have used a `.login()` command.

synapseclient/client.py
BryanFauble: We are pointing users to use the method on the dataclass objects instead of the API calls. If we sent users to use the API calls - Then yes, that is the direct replacement, however, the intention is that they'll use `.get_permissions` or `.get_acl` on a class object.

synapseclient/api/entity_services.py
```diff
@@ -964,3 +1262,307 @@ async def main():
...
+                and user_profile_bundle is not None
+            )
+        ):
+            effective_permission_set = effective_permission_set.union(
+                permissions["accessType"]
+            )
+    return list(effective_permission_set)
+
+
+async def update_entity_acl(
```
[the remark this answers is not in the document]
BryanFauble: Not in this case. All of the functions in these `/api/*` files follow a naming scheme around the URI it's hitting. Technically this could be `update_or_create_entity_acl`, but, just `update_entity_acl` is sufficient here.

There URI we are hitting in this case is: `uri = f"/entity/{entity_id}/acl"`

If we were hitting something like `uri = f"/entity/{entity_id}/permissions"` then I would re-name this

synapseclient/client.py
linglp: Got it. Thanks for explaining.

linglp (approved):
LGTM! Thanks for your hard work!

PR #1226, opened 2025-07-25
[SYNPY-1426] Noting all final deprecations and the tickets they will be deprecated with

BryanFauble:
# **Problem:**

- There are several areas left to be deprecated, but to made a cut off point for the SYNPY-1426 ticket, the rest of the work will be broken up into future tickets.
- 
# **Solution:**

- Adding todo comments to the sections of code where further tickets will be adding the deprecation.

# **Testing:**

- No testing needed for comments


thomasyu888 (approved):
🔥 LGTM! thanks for all the mappings to the different Jira tickets.

linglp (approved):
LGTM!! Thanks for your hard work!

PR #1233, opened 2025-08-14
[SYNPY-1632] Deprecate tables from the Synapse class and table.py module

linglp:
# **Problem:**
Deprecate the following: 
```
tableQuery
_queryTable
_queryTableNext
_uploadCsv
_check_table_transaction_response
_queryTableCsv
downloadTableColumns
_build_table_download_file_handle_list
_get_default_view_columns
_get_annotation_view_columns
```

# **Solution:**
Map the following to new methods: 
```
tableQuery -> query or query_async
_queryTable -> query_part_mask
_queryTableNext -> query_part_mask
_uploadCsv -> _chunk_and_upload_csv
_check_table_transaction_response -> internal function, no replacement
_uploadCSV -> store_rows_async
_queryTableCSV -> query_async (with downloadLocation parameter)
downloadTableColumns -> not seeing a direct replacement?
_build_table_download_file_handle_list -> internal function, no replacement
_get_default_view_columns -> internal function, no replacement
_get_annotation_view_columns -> internal function, no replacement
```

# New data classes: 
**SumFileSizes**
**QueryResultOutput** (created by using the old `QueryBundleRequest`)
**Row**
**RowSet**
**SelectColumn**
**ActionRequiredCount**
**Query**
 **QueryResult**
**QueryResultBundle**
 **QueryNextPageToken**
**QueryJob**
**QueryBundleRequest**

# **Testing:**
1. Make sure all the new data classes have unit tests
2. Added TestQueryTableRowSet to test  `_query_table_row_set`
3. Added TestQueryTableNextPage to test `_query_table_next_page`
4. Added TestQueryTableCsv to test `_query_table_csv`


synapseclient/models/table_components.py
```diff
@@ -482,6 +514,391 @@ def __repr__(self) -> str:
...
+        """Create a RowSet from a dictionary response."""
+        headers = data.get("headers", None)
+        rows = data.get("rows", None)
+
+        # Handle headers
+        headers_data = data.get("headers")
+        if headers_data and isinstance(headers_data, list):
+            headers = [SelectColumn.fill_from_dict(header) for header in headers_data]
+
+        # Handle rows
```
linglp, line 748: This section was added to ensure backward compatibility with existing functions See how casting works in the original function here: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/table.py#L1669

synapseclient/models/table_components.py
```diff
@@ -482,6 +514,391 @@ def __repr__(self) -> str:
...
+    To update any rows from a RowSet the etag must be provided with the POST."""
+
+    headers: Optional[List[SelectColumn]] = None
+    """The list of SelectColumns that describes the rows of this set."""
+
+    rows: Optional[List[Row]] = None
+    """The Rows of this set. The index of each row value aligns with the index of each header."""
+
+    @classmethod
+    def cast_row(
```
linglp, line 726: similar to the original function here: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/table.py#L336

linglp (commented):
@BryanFauble Could you look at my self-review and see if this is on the right track? There are more tests that need to be added but I want to gather some of your feedbacks before moving forward. Thank you!!

synapseclient/models/mixins/table_components.py
```diff
@@ -2295,26 +2661,43 @@ async def main():
...
+                elif column["columnType"] in LIST_COLUMN_TYPES:
+                    list_columns.append(column["name"])
+                elif column["columnType"] == "DATE" and convert_to_datetime:
+                    date_columns.append(column["name"])
+
+        return csv_to_pandas_df(
+            filepath=csv_path,
+            separator=kwargs.get("separator", DEFAULT_SEPARATOR),
+            quote_char=kwargs.get("quote_character", DEFAULT_QUOTE_CHARACTER),
+            escape_char=kwargs.get("escape_character", DEFAULT_ESCAPSE_CHAR),
```
BryanFauble, line 2697: Since these are already named kw arguments - Do we need to use `kwargs` here? I'd think we could do something like `separator=separator or DEFAULT_SEPARATOR`

synapseclient/models/mixins/table_components.py
```diff
@@ -117,6 +141,350 @@ def test_import_pandas() -> None:
...
+                escape_character=kwargs.get("escape_character", DEFAULT_ESCAPSE_CHAR),
+                line_end=kwargs.get("line_end", str(os.linesep)),
+                separator=kwargs.get("separator", DEFAULT_SEPARATOR),
+                header=kwargs.get("header", True),
+                include_row_id_and_row_version=kwargs.get(
+                    "include_row_id_and_row_version", True
+                ),
+                download_location=kwargs.get("download_location", None),
+            ),
+        )
```
BryanFauble, line 354: Once `_query_table_csv` is made async you will be able to remove the `run_in_executor` logic

synapseclient/models/mixins/table_components.py
```diff
@@ -117,6 +136,346 @@ def test_import_pandas() -> None:
...
+        synapse_id=extract_synapse_id_from_query(query),
+        entity_type="TableEntity",
+        destination=os.path.join(download_dir, filename),
+        synapse_client=synapse,
+    )
+
+    return download_from_table_result, path
+
+
+def _query_table_next_page(
```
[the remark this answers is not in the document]
BryanFauble: This would swap to be an async function once `QueryResultBundle` AsynchronousCommunicator mixin is implemented

synapseclient/models/mixins/table_components.py
```diff
@@ -2389,36 +2773,32 @@ async def main():
...
+        results = await _table_query(
+            query=query,
+            results_as="rowset",
+            part_mask=part_mask,
+            limit=limit,
+            offset=offset,
         )
 
-        # TODO: Replace method in https://sagebionetworks.jira.com/browse/SYNPY-1632
         as_df = await loop.run_in_executor(
```
BryanFauble, line 2787: To remove the run_in_executor when the underlying function is async

synapseclient/client.py
```diff
@@ -7487,9 +7605,16 @@ def _queryTable(
 
         return self._waitForAsync(uri=uri, request=query_bundle_request)
 
-    # TODO: Deprecate method in https://sagebionetworks.jira.com/browse/SYNPY-1632
+    @deprecated(
+        version="4.5.0",
+        reason="To be removed in 5.0.0. "
+        "Use the `query_part_mask` method on the `Table`, `EntityView`, `SubmissionView`, `MaterializedView`, or `Dataset` classes instead.",
+    )
     def _queryTableNext(self, nextPageToken: str, tableId: str) -> TableQueryResult:
```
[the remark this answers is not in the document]
BryanFauble: Alternatively we just slap a deprecation method on here that there isn't a replacement since it's a private method and only we should be using it.

synapseclient/client.py
```diff
@@ -7533,6 +7709,55 @@ def _uploadCsv(
...
+                )
+
+        Example: Migration to new method
+            &nbsp;
+
+            ```python
+            import asyncio
+            from synapseclient import Synapse
+            from synapseclient.models import Table
+            from synapseclient.models.mixins.table_components import CsvTableDescriptor
```
BryanFauble, line 7728: If users should be using a particular class can this be exposed at the `synapseclient.models` level as well? So that you could do:

`from synapseclient.models import Table, CsvTableDescriptor`

synapseclient/models/mixins/table_components.py
```diff
@@ -117,6 +141,353 @@ def test_import_pandas() -> None:
...
+        column_name = header.name
+        series[column_name] = pd.Series(
+            name=column_name,
+            data=[row.values[i] for row in rows],
+            index=rownames,
+        )
+
+    next_page_token = query_result.next_page_token
+
+    while next_page_token:
```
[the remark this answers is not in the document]
BryanFauble: Slack thread (No resolution at the time of posting):
https://sagebionetworks.slack.com/archives/C0AFCTGEQ/p1756313663707229

linglp (commented):
@BryanFauble Thank you for the last round of review! For this round, I implemented AsynchronousCommunicator to new dataclass `QueryJob` and `QueryBundleRequest`. Could you take a look again?

synapseclient/models/table_components.py
```diff
@@ -482,6 +518,498 @@ def __repr__(self) -> str:
...
+
+    sql: str
+    """The SQL query string"""
+
+    additional_filters: Optional[List[Dict[str, Any]]] = None
+    """Appends additional filters to the SQL query. These are applied before facets.
+    Filters within the list have an AND relationship. If a WHERE clause already exists
+    on the SQL query or facets are selected, it will also be ANDed with the query
+    generated by these additional filters."""
+    """TODO: create QueryFilter dataclass"""
```
BryanFauble, line 964: Do these TODO comments need to be completed now? Or could they be done later on?

synapseclient/models/mixins/table_components.py
```diff
@@ -117,6 +140,346 @@ def test_import_pandas() -> None:
...
+                (row.row_id, row.version_number, row.etag)
+                if row.etag
+                else (row.row_id, row.version_number)
+            )
+            for row in rows
+        ]
+    )
+
+
+async def _query_table_csv(
```
BryanFauble, line 160: For this method it's got the `_` prefix, but `query_table_row_set` does not. Should these be the same? Both private or not?

synapseclient/models/table_components.py
linglp: I can create additional tickets for these! I don't think they need to be completed now.

synapseclient/models/mixins/table_components.py
```diff
@@ -117,6 +140,346 @@ def test_import_pandas() -> None:
...
+    # first page of rows
+    offset = 0
+    rownames = construct_rownames(query_result_bundle, offset)
+    query_result = query_result_bundle.query_result
+    rowset = query_result.query_results
+
+    if not rowset:
+        raise ValueError("The provided query_result_bundle has no 'rowset' data.")
+
+    rows = rowset.rows or []
```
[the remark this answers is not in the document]
BryanFauble: > Alternatively, I could default rows to an empty list in the data class. Will that be better? For reference, the original function set rows as an empty list.


It could be worthwhile to do both what you're doing in this code `rows = rowset.rows or []` and initialize it to an empty list in the dataclass.

There are a bunch of cases where I am initializing things to an empty list. I think that it ends up being a better user experience because they have a lower chance to run into errors since it's not None.

synapseclient/models/mixins/table_components.py
linglp: That's a good catch. I think they should both be private.

synapseclient/models/mixins/table_components.py
BryanFauble: Since `escape_char` is already defined as an argument to the method call:

```
    @staticmethod
    async def query_async(
        query: str,
        include_row_id_and_row_version: bool = True,
        convert_to_datetime: bool = False,
        download_location=None,
        quote_character='"',
        escape_character="\\",
        line_end=str(os.linesep),
        separator=",",
        header=True,
        *,
        **kwargs,
    ) -> Union["DATA_FRAME_TYPE", str]:
```

synapseclient/models/mixins/table_components.py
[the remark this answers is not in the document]
linglp: Got it. Thanks for the feedback. I could do initialize `rows` as an empty list then.

synapseclient/models/mixins/table_components.py
linglp: Just fixed query_table_row_set

synapseclient/models/mixins/table_components.py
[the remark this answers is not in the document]
linglp: Just done :)

synapseclient/models/mixins/table_components.py
linglp: Mark this as resolved for now since `_query_table_csv` has been turned to async

synapseclient/models/table_components.py
```diff
@@ -92,7 +129,7 @@ def to_synapse_request(self):
             "quoteCharacter": self.quote_character,
             "escapeCharacter": self.escape_character,
             "lineEnd": self.line_end,
-            "isFirstLineHeader": self.is_file_line_header,
```
linglp, line 95: This is a typo in the original `CsvTableDescriptor` class. It should be `is_first_line_header`

synapseclient/client.py
linglp: Got it. I added them to models/__init__.py, so the import should be working as expected now.

BryanFauble:
> 🔥 LGTM! Im going to defer to @BryanFauble for final review, but thanks for doing the giant deprecation.
> 
> Will there be a tutorial page we are going to update to use the new functions?

Yes, https://sagebionetworks.jira.com/browse/SYNPY-1377 should capture the work to write up the tutorial page.

@linglp Do you mind reviewing this jira and add onto the topic for anything that we should cover in addition to whats there (if anything),

synapseclient/models/mixins/table_components.py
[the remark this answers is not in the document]
linglp: No, this has not been resolved. I will tag you in the slack thread.

linglp:
> I appreciate all your hard work that you've put into this!

I appreciate your review. Thank you again, Bryan!

PR #1235, opened 2025-08-18
[SYNPY-1633] Add membership_status to Team model and Deprecate Old Method

linglp:
# **Problem:**
This method here needs to be deprecated: [synapsePythonClient/synapseclient/client.py at 02d3b38d2a0f043811637064f61ae8b537e4253b · Sage-Bionetworks/synapsePythonClient](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/02d3b38d2a0f043811637064f61ae8b537e4253b/synapseclient/client.py#L5519-L5535)

# **Solution:**
Added a new method called `get_user_membership_status` under team

# **Testing:**
- Ensured that the example in the docstring works
- Surfaced the new function in documentation 
- Added test

### DIFF
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -6038,10 +6038,19 @@ class Synapse(object):
             dict of TeamMembershipStatus
 
         Example: Using this function (DEPRECATED)
-        Getting a user's membership status for a team
+            &nbsp;
+            Getting a user's membership status for a team
 
-            status = syn.get_membership_status(user="12345", team="67890")
+            ```python
+            from synapseclient import Synapse
+
+            # Login to Synapse
+            syn = Synapse()
+            syn.login()
+
+            status = syn.get_membership_status(userid="12345", team="67890")
             print(status)
+            ```
 
 
         Example: Migration to new method
--- a/synapseclient/client.py
+++ b/synapseclient/client.py
@@ -6023,7 +6023,7 @@ class Synapse(object):
     @deprecated(
         version="4.9.0",
         reason="To be removed in 5.0.0. "
-        "Use the `get_user_membership_status_async` method on the `Team` class from `synapseclient.models` instead. "
+        "Use the `get_user_membership_status` method on the `Team` class from `synapseclient.models` instead. "
         "Check the docstring for the replacement function example.",
     )
     def get_membership_status(self, userid, team):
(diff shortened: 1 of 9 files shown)


synapseclient/models/team.py
```diff
@@ -356,3 +431,28 @@ async def open_invitations_async(
...
+
+        Arguments:
+            user_id: Synapse user ID
+            synapse_client: If not passed in and caching was not disabled by
+                `Synapse.allow_client_caching(False)` this will use the last created
+                instance from the Synapse class constructor.
+
+        Returns:
+            TeamMembershipStatus object
+        """
```
BryanFauble, line 451: Could you add in an example of using this function (And one in the non-async version)?

synapseclient/models/team.py
```diff
@@ -56,6 +57,80 @@ def fill_from_dict(
...
+        membership_approval_required: true if and only if team admin approval is required for the user to join the team
+        has_unmet_access_requirement: true if and only if there is at least one unmet access requirement for the user on the team
+        can_send_email: true if and only if the user can send an email to the team
+    """
+
+    team_id: Optional[str] = None
+    """The ID of the team"""
+
+    user_id: Optional[str] = None
+    """The ID of the user"""
```
BryanFauble, line 83: On the https://github.com/Sage-Bionetworks/synapsePythonClient/blob/dffef121d550630c4e4161475917417178192da3/synapseclient/models/user.py#L110-L111 object the ID is an int. Is there a reason to use a string instead here, or could we keep this consistent? This would have some changes across a few lines in the PR to make it a consistent int if changing it.

synapseclient/client.py
```diff
@@ -6031,6 +6036,49 @@ def get_membership_status(self, userid, team):
...
+
+                # Get user's membership status
+                status = await team.get_user_membership_status_async(
+                    user_id="12345"
+                )
+                return status
+
+            # Run the async function
+            status = asyncio.run(get_membership_status())
+            print(status)
```
BryanFauble, line 6080: tbh I probably wouldn't even include the async version in the migration note. It is more complete to include it, but may confuse someone coming in and looking at this.

synapseclient/models/team.py
linglp: Based on the synapse documentation [here](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/TeamMembershipStatus.html), `user_id` and `team_id` are both string. 

I also looked into the `UserProfile`: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/UserProfile.html, and it looks like that `ownerId` which is basically the same as `id` in UserProfile is also a string. 

Should they all be kept as string to align with the synapse documentation? If not, I could make `team_id` and `user_id` integers.

BryanFauble: Great points. this is fine as is

PR #1243, opened 2025-09-10
[SYNPY-1623] Factory function for get and creation of link entity

BryanFauble:
**View the API docs here:**

https://synapsepythonclient--1243.org.readthedocs.build/en/1243/reference/experimental/sync/factory_operations/
https://synapsepythonclient--1243.org.readthedocs.build/en/1243/reference/experimental/sync/link_entity/

async:

https://synapsepythonclient--1243.org.readthedocs.build/en/1243/reference/experimental/async/factory_operations/
https://synapsepythonclient--1243.org.readthedocs.build/en/1243/reference/experimental/async/link_entity/

# **Problem:**

- In order to use the current OOP object models you must know ahead of time what the class is for the Synapse ID that you have is. Sometimes, but not always, is this possible.
- A generic get factory method will better support R.
- The link entity was the last synapse ID type entity to be supported, and had some implications in the generic get function.

# **Solution:**

- Create a get/get_async function that serves as the entry point for searching for an entity by its id, or name+parent and return the correct class.
- If the class is not yet implemented in the client it will return back an entity bundle
- The link entity was created and with some get/store functionality as well as following the link to the target entity

# **Testing:**

- I did a bunch of manual testing to verify the logic would get entities I have stored in a test project across a variety of types
- Integration tests to verify the logic across the 11 different entities fully supported.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,7 @@ jobs:
     needs: [pre-commit]
 
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-22.04, macos-13, windows-2022]
 


synapseclient/models/factory_operations.py
```diff
@@ -0,0 +1,1209 @@
...
+        if version_number is not None:
+            return await get_entity_id_version_bundle2(
+                entity_id=synapse_id,
+                version=version_number,
+                synapse_client=synapse_client,
+            )
+        else:
+            return await get_entity_id_bundle2(
+                entity_id=synapse_id, synapse_client=synapse_client
+            )
```
BryanFauble, line 1209: Fallback behavior so when more entity types get added to Synapse this function will still return something that folks can use.

synapseclient/models/folder.py
```diff
@@ -330,14 +330,10 @@ async def store_async(
...
             )
             delete_none_keys(synapse_folder)
-            entity = await loop.run_in_executor(
-                None,
-                lambda: Synapse.get_client(synapse_client=synapse_client).store(
-                    obj=synapse_folder,
-                    set_annotations=False,
-                    isRestricted=self.is_restricted,
-                    createOrUpdate=False,
-                ),
```
BryanFauble, line 340: This run_in_executor code is stuff I am trying to get rid it since it was the original hack used to wrap the original Synapse client.

synapseclient/models/link.py
```diff
@@ -0,0 +1,423 @@
...
+        Project,
+        SubmissionView,
+        Table,
+        VirtualTable,
+    )
+
+
+@dataclass()
+@async_to_sync
+class Link:
```
BryanFauble, line 38: This is an implementation of the Link entity. This is used to create a pointer to another Synapse entity. In the Web UI it looks like:

image

synapseclient/models/link.py
```diff
@@ -0,0 +1,423 @@
...
+            )
+
+            return await factory_get_async(
+                synapse_id=self.target_id,
+                version_number=self.target_version_number,
+                file_options=FileOptions(download_file=False),
+                synapse_client=synapse_client,
+            )
+        else:
+            return self
```
BryanFauble, line 311: If we get feedback or if reviewers have suggestions here I am more than happy to make changes.

What this `follow_link` logic does is to return the Entity that the link is actually pointing at. So if it's a file you'll get back the metadata about that FileEntity, one thing I was not certain on was if we should default behaivor around `download_file`.

Something we could possibly do here is to allow any of the options that the new `get/get_async` factory operation allows and pass it along from here.

Thoughts?

synapseclient/models/mixins/table_components.py
```diff
@@ -363,7 +363,7 @@ async def _table_query(
 
     - For `csv`, you can specify:
         - `quote_character`: Character used for quoting fields. Default is double quote (").
-        - `escape_character`: Character used for escaping special characters. Default is backslash (\).
+        - `escape_character`: Character used for escaping special characters. Default is backslash.
```
BryanFauble, line 366: I was seeing a warning about this since it's a special escape character, so I just removed it for now

synapseclient/models/factory_operations.py
```diff
@@ -0,0 +1,1209 @@
...
+        elif isinstance(entity, Project):
+            print(f"Project: {entity.name}")
+        ```
+
+    Example: Retrieving entities by name
+        Get an entity by name and parent:
+
+        ```python
+        from synapseclient import Synapse
+        from synapseclient.models import get
```
thomasyu888, line 476: For discussion: I had always thought that models should contain the classes and the "services" and/or factory methods could live elsewhere. That said, I'm not opposed to this, just wanted to be mindful of the "models" module growing a lot

tests/integration/synapseclient/models/async/test_factory_operations.py
```diff
@@ -0,0 +1,740 @@
...
+        """Test retrieving a File entity with activity information."""
+        # GIVEN a file with activity in a project
+        file = self.create_file_instance()
+        file.parent_id = project_model.id
+        file.activity = self.create_activity()
+        stored_file = await file.store_async(synapse_client=self.syn)
+        self.schedule_for_cleanup(stored_file.id)
+
+        # AND activity options to include activity
+        activity_options = ActivityOptions(include_activity=True)
```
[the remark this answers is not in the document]
BryanFauble: Not totally certain on this, and we could swap this later on based on feedback that we get.

synapseclient/models/link.py
BryanFauble: This is defaulted to following the link to get the entity the Link points to, and one can set that behavior to False in order to get the Link itself

synapseclient/models/link.py
```diff
@@ -0,0 +1,635 @@
...
+    created_by: Optional[str] = None
+    """(Read Only) The ID of the user that created this entity."""
+
+    modified_by: Optional[str] = None
+    """(Read Only) The ID of the user that last modified this entity."""
+
+    parent_id: Optional[str] = None
+    """The ID of the Entity that is the parent of this Entity."""
+
+    target_id: Optional[str] = None
```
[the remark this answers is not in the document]
BryanFauble: This was done to make the user experience easier to use the Link class. It gets a bit awkward building out nested classes (Although it would mirror the REST API closer). I've done this in a few other places too

synapseclient/models/link.py
```diff
@@ -0,0 +1,635 @@
...
+        )
+        self._set_last_persistent_instance()
+
+        if follow_link:
+            from synapseclient.operations.factory_operations import (
+                get_async as factory_get_async,
+            )
+
+            return await factory_get_async(
+                synapse_id=self.target_id,
```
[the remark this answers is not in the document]
BryanFauble: I'm not sure it's possible to get to this line of code if `target_id` is None. Basically if you provide a way to look up what the LinkEntity is in Synapse it will fill in the `target_id` even if you don't specify it.

synapseclient/models/link.py
```diff
@@ -0,0 +1,635 @@
...
+            raise ValueError("The link must have a name.")
+        if not self.parent_id and not self.id:
+            raise ValueError("The link must have a parent_id.")
+        if not self.target_id and not self.id:
+            raise ValueError("The link must have a target_id.")
+
+        if existing_entity := await self._find_existing_entity(
+            synapse_client=synapse_client
+        ):
+            merge_dataclass_entities(
```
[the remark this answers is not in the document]
BryanFauble: Good catch. This was actually an issue since we changed the default behavior to follow the link to the target entity. I corrected this

synapseclient/models/link.py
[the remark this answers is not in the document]
linglp: Got it. Thanks for explaining.

synapseclient/models/link.py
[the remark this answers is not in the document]
linglp: I ran into a usability issue while experimenting: I accidentally passed a folder ID to Link(...) and then called .get(...):
```
link_entity = Link(id="my-folder-id").get(
                    follow_link=True,
                    file_options=FileOptions(
                    download_file=True,
                    download_location="/Users/lpeng/code/synapsePythonClient",
                    if_collision="overwrite.local"
                ))
```
I got this error: 
```
  File "/Users/lpeng/code/synapsePythonClient/synapseclient/operations/factory_operations.py", line 1072, in get_async
    raise ValueError(
ValueError: Must specify either synapse_id or entity_name. Use synapse_id for direct lookup or entity_name with optional parent_id for name-based lookup.
```
Could we validate the ID/entity type up front and give a friendlier message (e.g., “expected a Link ID, not a Folder ID”)?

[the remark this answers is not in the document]
BryanFauble: There is a more general ticket to apply this change across the board @linglp : https://sagebionetworks.jira.com/browse/SYNPY-1572 - I'll make a note that Link is also one of these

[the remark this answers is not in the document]
linglp: Got it. Thanks for noting in the ticket.

synapseclient/api/entity_services.py
```diff
@@ -1264,6 +1264,99 @@ async def main():
         yield child
 
 
+async def get_child(
```
linglp, line 1267: Do we have tests for `get_child`?

synapseclient/models/link.py
```diff
@@ -0,0 +1,636 @@
...
+            await self.get_async(
+                synapse_client=synapse_client,
+            )
+
+        self._set_last_persistent_instance()
+        return self
+
+    async def _find_existing_entity(
+        self, *, synapse_client: Optional[Synapse] = None
+    ) -> Union["File", None]:
```
linglp, line 600: Here it should return Link and not File?

linglp (approved):
It looks good to me overall! Just two small comments.

PR #1244, opened 2025-09-10
[SYNPY-1589] Implement "Evaluation" OOP model

jaymedina:
# **Problem:**

This work is part of the ongoing effort to refactor the Synapse Python Client using object-oriented principles, following the pattern established in https://sagebionetworks.jira.com/browse/SYNPY-1418.

We currently to not have an OOP model for communications to the [Evaluation API services](https://rest-docs.synapse.org/rest/index.html#org.sagebionetworks.repo.web.controller.EvaluationController) provided by the Synapse platform.

# **Solution:**

- [ ] A new `DataClass` is introduced following the official attributes of an [Evaluation entity](https://rest-docs.synapse.org/rest/org/sagebionetworks/evaluation/model/Evaluation.html) on Synapse.
- [ ] Communications to the following services are supported:

Create evaluation: https://rest-docs.synapse.org/rest/POST/evaluation.html 
Get evaluation by ID: https://rest-docs.synapse.org/rest/GET/evaluation/evalId.html 
Get evaluations by entity: https://rest-docs.synapse.org/rest/GET/entity/id/evaluation.html 
Get all evaluations: https://rest-docs.synapse.org/rest/GET/evaluation.html 
Get available evaluations: https://rest-docs.synapse.org/rest/GET/evaluation/available.html 
Find evaluation by name: https://rest-docs.synapse.org/rest/GET/evaluation/name/name.html 
Update evaluation: https://rest-docs.synapse.org/rest/PUT/evaluation/evalId.html 
Delete evaluation: https://rest-docs.synapse.org/rest/DELETE/evaluation/evalId.html 
Get ACL: https://rest-docs.synapse.org/rest/GET/evaluation/evalId/acl.html 
Update ACL: https://rest-docs.synapse.org/rest/PUT/evaluation/acl.html 
Get permissions: https://rest-docs.synapse.org/rest/GET/evaluation/evalId/permissions.html 

- [ ] async/sync methods are introduced to the `DataClass`
- [ ] async/sync Integration tests are introduced
- [ ] Our Synapse Python Client documentation (using mkdocs) is updated for this model
- [ ] A tutorial is made for popular use-cases of this model

# **Testing:**

- [ ] Passing integration tests
- [ ] Passing documentation build
- [ ] Running through the tutorial presents no issues

PR #1245, opened 2025-09-11
[SYNPY-1589] Implement core functionality of Evaluation OOP model

jaymedina:
See #1244

### DIFF
--- a/synapseclient/core/utils.py
+++ b/synapseclient/core/utils.py
@@ -1508,17 +1508,19 @@ def merge_dataclass_entities(
 
     # Override any specified fields in destination with values from source
     if fields_to_preserve_from_source:
+        # Use provided logger or fall back to default
+        log = logger
+        if not log:
+            import logging
+
+            log = logging.getLogger(LOGGER_NAME)
+
+        # Warn if any preserved fields were modified in destination
         for field_name in fields_to_preserve_from_source:
             if hasattr(source, field_name) and hasattr(destination, field_name):
                 source_value = getattr(source, field_name)
                 destination_value = getattr(destination, field_name)
                 if destination_value != source_value:
-                    # Use provided logger or fall back to default
-                    log = logger
-                    if not log:
-                        import logging
-
-                        log = logging.getLogger(LOGGER_NAME)
                     log.warning(
                         f"Field '{field_name}' cannot be modified. Changes will be ignored."
                     )


synapseclient/models/evaluation.py
```diff
@@ -1 +1,459 @@
...
+
+        if not self.id:
+            raise ValueError("id must be set to delete an evaluation")
+
+        await delete_evaluation_async(
+            evaluation_id=self.id,
+            synapse_client=synapse_client,
+        )
+
+    async def get_acl_async(
```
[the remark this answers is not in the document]
thomasyu888: I'll defer to Bryan to add more, but evaluations are not entities.  They have their own set of ACLs: https://rest-docs.synapse.org/rest/PUT/evaluation/acl.html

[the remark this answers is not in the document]
jaymedina: Sounds good!

synapseclient/api/evaluation_services.py
```diff
@@ -0,0 +1,465 @@
...
+        request_body["description"] = description
+    if content_source is not None:
+        request_body["contentSource"] = content_source
+    if submission_instructions_message is not None:
+        request_body["submissionInstructionsMessage"] = submission_instructions_message
+    if submission_receipt_message is not None:
+        request_body["submissionReceiptMessage"] = submission_receipt_message
+
+    uri = f"/evaluation/{evaluation_id}"
+    response = await client.rest_put_async(uri, body=json.dumps(request_body))
```
jaymedina, line 326: I'm wondering if someone can take a look at this - I have all the integration tests working for the async methods except the ones testing the PUT requests to `/evaluation/{id}`. For some reason, this particular line of code will just hang forever. Not sure if it's related to the OCC logic for preventing conflicting updates to the same object on Synapse.

Tomorrow I'll take a look at what the current Evaluation model is doing, but throwing it out here in case someone already knows what's going on.

image

jaymedina: Sharing the solution I discussed with @thomasyu888 here for record-keeping:

Ah yeah, the problem was that I was building a new request body from scratch, with just the ID and eTag, and conditionally populating the other attributes, leading to an incomplete request:
# Build request body from scratch
    request_body = {"id": evaluation_id, "etag": etag}

    # Update request body with only provided fields
    if name is not None:
        request_body["name"] = name
    if description is not None:
        request_body["description"] = description
    if content_source is not None:
        request_body["contentSource"] = content_source
    if submission_instructions_message is not None:
        request_body["submissionInstructionsMessage"] = submission_instructions_message
    if submission_receipt_message is not None:
        request_body["submissionReceiptMessage"] = submission_receipt_message
I should be passing in the original Evaluation with all of its attributes and updating when necessary.
This hackfix seemed to work:
# Build request body with required fields
    request_body = {"id": evaluation_id,
                    "etag": etag,
                    "name": "ah",
                    "description": "ah",
                    "contentSource": "69835628",
                    "submissionInstructionsMessage": "ah",
                    "submissionReceiptMessage": "ah"}

    # Update request body with only provided fields
    if name is not None:
        request_body["name"] = name
    if description is not None:
        request_body["description"] = description
    if content_source is not None:
        request_body["contentSource"] = content_source
    if submission_instructions_message is not None:
        request_body["submissionInstructionsMessage"] = submission_instructions_message
    if submission_receipt_message is not None:
        request_body["submissionReceiptMessage"] = submission_receipt_message
Thanks!

synapseclient/models/evaluation.py
```diff
@@ -1 +1,492 @@
...
+        self.submission_instructions_message = evaluation.get(
+            "submissionInstructionsMessage", None
+        )
+        self.submission_receipt_message = evaluation.get(
+            "submissionReceiptMessage", None
+        )
+
+        return self
+
+    def to_synapse_request(self, request_type: str):
```
jaymedina, line 99: I'd like people's thoughts on this re-design:

Following the bug fix for `update_async`, I realized the way I was having users update/create their evaluations didn't match, which was confusing.

The original logic looked like this:

```
# Creating my first evaluation
evaluation = Evaluation(name="eval_name", description=...)
await evaluation.store_async(synapse_client=syn)

# Updating my evaluation
await evaluation.update_async(name="new_name", synapse_client=syn)
```

So for `store_async`, the logic relied on the object attributes to build and submit the request. For `update_async`, the logic relied on the attributes being passed through as **parameters** in the method call.

I decided to consolidate this inconsistency so both `update` and `store` functionality relies on the user updating their evaluation object accordingly. 

This means neither `update_async` and `store_async` require any parameters get passed to the methods themselves, rather the object itself needs to be updated. This will also help the user know what exactly their evaluation is going to look like when they make an update or store a new evaluation.

This is what it would look like irl:

```
In [27]: await new_evaluation.store_async(synapse_client=syn)
Out[27]: Evaluation(id='9616331', etag='8a8dbe71-5ce9-45d2-95f2-7f80987db2f7', name='ah2', description='ah', owner_id='3489628', created_on='2025-09-16T15:09:58.819Z', content_source='syn69835628', submission_instructions_message='ah', submission_receipt_message='ah')

In [28]: new_evaluation.name = 'ah2_updated'

In [29]: await new_evaluation.update_async(synapse_client=syn)
Out[29]: Evaluation(id='9616331', etag='8c027b5b-6447-4009-ba12-4a33a4d0b363', name='ah2_updated', description='ah', owner_id='3489628', created_on='2025-09-16T15:09:58.819Z', content_source='syn69835628', submission_instructions_message='ah', submission_receipt_message='ah')
```

The downside of this is a user would have to manually update every attribute e.g:

```
test_evaluation.name = new_name
test_evaluation.description = new_description
test_evaluation.submission_instructions_message = new_instructions
```

LMK your thoughts @BryanFauble @thomasyu888

BryanFauble: I don't have a strong dislike how you've split apart the create and update logic, however - If it ends up the case that these are split methods a user needs to call I would suggest that they are renamed to `create_async` and `update_async`.


However, here is something to consider. There is an alternative way to go about this which would allow you to continue using `store_async` for create OR update operations. If this is the route you go this is how I accomplished this for other classes.

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/da36c84dc1810ac9facd1d916701d09eca85dfa8/synapseclient/models/project.py#L335-L349

To explain what is going on here:

1. The `create_or_update` is optional here, and is probably fine not to have for Evaluation.
2. The `_last_persistent_instance` is a bit of a complex concept the purpose behind this was to track if the class instance had already been read from Synapse, or stored to Synapse before during the current session. In most cases it would end up being None, but it would be filled in during the `get_async` operation since it was the last state that the client application knew the data to be.
3. There is a search to find the existing entity id (Evaluation ID in your case). This would either get the ID attribute if passed in OR search for it by name.
4. If an ID is found then a new `get_async` operation is performed to get the current state that Synapse has. After this is performed then we continue to the merge:


The merge is the important piece here, and what I think is the core problem that you're trying to solve. This merge operation allows you to perform non-destructive updates, if you implement it this would mean you could do something like:

```
my_eval = Evaluation(id='1234')
my_eval.description = new_description
my_eval.store()
```

The logic I described above would mean that ONLY the description gets updated here in a non-destructive update.


An edge case and something you might be thinking here, is "How do I delete fields/set them to None?" and that would be a great question.

The check for `_last_persistent_instance` means that after a resource has been read from Synapse (Or previously stored) any NEW `store()` operations are total replacements of the object.

That means this behaves the same as the small example I gave above:


```
my_eval = Evaluation(id='1234').get()
my_eval.description = new_description
my_eval.store()
```

But then you could also do something like:


```
my_eval = Evaluation(id='1234')
my_eval.description = None # Delete description
my_eval.store()
```

Hopefully this makes sense.

jaymedina: @BryanFauble thanks for the feedback - since this is the logic already in place for existing models, it makes sense for me to implement it this way too for consistency. I'm working on the refactor now and will tag you when it's finished.

jaymedina: I've implemented the refactor in this commit: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1245/commits/ecfc2d296cfec7bf065564116569ba164e1d31fd

Notable details:
* I refactored and implemented `merge_dataclass_entities` to also support Evaluation objects, not just Synapse entities. Another major change was adding a new argument that passes a list of fields that SHOULD NOT change from the source object if a user changes them, such as things like `owner_id`. I'm open to discussing if this is worth it or not. The motivation behind this was to prevent users from "reusing" the Evaluation objects to interact with other Evaluation objects.
* I implemented those attributes you leverage for when/when not to trigger `merge_dataclass_entities`, such as `_last_persistent_instance` and `has_changed`.
* One area where I deviated from your design in the Project model is I didn't implement `create_or_update`. Instead, I relied on the `id` and `_last_persistent_instance` and `has_changed` models for when and when not to make a non-destructive change to the Synapse object. Because this is a choice that's happening "behind the curtain", I opted for adding a stdout message notifying the user when their object is going down the "create" or "update" pipeline, or not being updated at all.

Let me know your thoughts on this implementation, and if you have any other suggestions for the `store_async` portion of the model!

synapseclient/models/evaluation.py
```diff
@@ -1 +1,719 @@
...
+        if not self.id:
+            raise ValueError("id must be set to get evaluation permissions")
+
+        return await get_evaluation_permissions_async(
+            evaluation_id=self.id,
+            synapse_client=synapse_client,
+        )
+
+    @staticmethod
+    async def get_all_evaluations_async(
```
BryanFauble, line 521: Would it make sense for this and the other staticmethods that are getting a bunch of Evaluations to be Generators instead of returning the full populated list? I am thinking of cases where there could be many Evaluations to return back and it might take a while.

Generators don't automatically work with that `@async_to_sync` decorator I wrote, so this is what I did to handle for async to sync generator methods:

https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1246/files#diff-cf7e968a89b2d58593efc3cc3af519a60ddde8c34d6ee3e9d3dd9a4737cb3335

Then on the `_async` generator method include this skip: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f9f23a96897b33606a78beba719580e320f859ae/synapseclient/models/curation.py#L528

Then just lightly wrap the async call like:

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f9f23a96897b33606a78beba719580e320f859ae/synapseclient/models/curation.py#L232-L255


This code is only in a feature branch at the moment, but once reviewed will be available in develop

BryanFauble (changes requested):
This is great progress! There are a few items to patch up, but this is just about there.

synapseclient/models/evaluation.py
```diff
@@ -1 +1,719 @@
...
+                        f"Synapse will automatically assign an ID upon creation.",
+                        response=e.response,
+                    ) from e
+                else:
+                    raise
+            except Exception:
+                raise
+
+        # CASE 3: We have an ID from previous Synapse interaction
+        elif self.id and self._last_persistent_instance:
```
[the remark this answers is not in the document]
jaymedina: Yes, the `store()` method supports the following:

1. Storing a new Evaluation by name and contentSource, e.g.:
```
evaluation = await Evaluation(
                    name="My Challenge Evaluation",
                    description="Evaluation for my data challenge",
                    content_source="syn123456",
                    submission_instructions_message="Submit CSV files only",
                    submission_receipt_message="Thank you for your submission!"
                ).store_async()
```
2. Updating an Evaluation by name or ID

I just did another [refactor](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1245/commits/130b475f45c153be3442a4cf525e99e18f19470a) that simplifies the `store()` logic even further by just using `_last_persistant_instance` for all decisions, instead of going down that rabbit hole in `CASE 2` where I was retrieving evaluations under the hood.

Now the logic goes like this:

```
if not self._last_persistent_instance:
--> attempt to create an Evaluation

elif self._last_persistent_instance:
--> attempt to update an Evaluation
```

This offloads error-handling to the API, and if there's any tinkering with the class attributes that should NOT be changed (ID, etag, created_on, modified_by), it'll quietly replace them with the original values from the last call to Synapse.

synapseclient/models/evaluation.py
jaymedina: Update: Verena gave feedback that a generator object would be preferred!

Let's continue to iterate on the other threads in this PR, and feel free to add me as a reviewer to your PR when it's ready.

synapseclient/models/evaluation.py
```diff
@@ -1 +1,719 @@
...
+            evaluation_id=self.id,
+            synapse_client=synapse_client,
+        )
+
+        # Clear the persistent instance since this object has been deleted
+        self._last_persistent_instance = None
+
+    async def get_acl_async(
+        self, *, synapse_client: Optional["Synapse"] = None
+    ) -> dict:
```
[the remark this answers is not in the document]
linglp: How about creating a new dataclass called AccessControlList based on the [Synapse documentation](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/AccessControlList.html), and then converting the result from get_evaluation_acl_async into an AccessControlList object?

synapseclient/models/evaluation.py
```diff
@@ -1 +1,838 @@
...
+        """
+        Get the access control list (ACL) governing this evaluation.
+
+        Arguments:
+            synapse_client: If not passed in and caching was not disabled by `Synapse.allow_client_caching(False)` this will use the last created
+                            instance from the Synapse class constructor.
+
+        Returns:
+            AccessControlList: The ACL for this Evaluation.
```
linglp, line 440: Right now it looks like this returns a dictionary representing the `AccessControlList` rather than an `AccessControlList` object. Once you define an AccessControlList class, you could return that instead.

synapseclient/api/evaluation_services.py
```diff
@@ -0,0 +1,400 @@
...
+        params.append(f"offset={offset}")
+    if limit is not None:
+        params.append(f"limit={limit}")
+
+    # Build URI with query parameters
+    uri = f"/entity/{project_id}/evaluation"
+    if params:
+        uri += "?" + "&".join(params)
+
+    evaluation_list = await client.rest_get_async(uri)
```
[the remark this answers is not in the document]
jaymedina: This would be one of the returns that becomes a generator return, rather than a batch of evaluations. For the generator, I plan to have it yield the evaluations one at a time, so I don't think pagination support would be necessary, but LMK your thoughts!

synapseclient/models/evaluation.py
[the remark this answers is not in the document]
BryanFauble: > LMK if you want me to fold this function into the mixin the way it's done for the entity ACLs, i.e. Evaluation(..., AccessControllable).get_acl instead becomes a method that returns a list of permissions based on a given principal ID.

I'd say that folding the evaluation specific way into AccessControllable mixin might be more pain than anything we'd get from it. Since it's handled differently, let's keep this as is.

> Was introducing an ACL OOP model in the cards, or were the AccessControllable mixin, EntityAcl, & AclList ([here](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/core/models/acl.py)) designed to take its place?

Yes, the intention of what you linked to was meant to be the replacement for the previous permissions related methods that you'd call. It also combined with this `Permissions` dataclass: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/da36c84dc1810ac9facd1d916701d09eca85dfa8/synapseclient/core/models/permission.py#L9-L41 This `Permissions` concept is close to what @linglp is suggesting. The issue though with `Permissions` is that it was only built as a Read only concept.

synapseclient/api/evaluation_services.py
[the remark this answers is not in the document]
BryanFauble: Agree with you @jaymedina

synapseclient/models/evaluation.py
```diff
@@ -1 +1,970 @@
...
+        if not self.id:
+            raise ValueError("id must be set to get evaluation ACL")
+
+        return await get_evaluation_acl(
+            evaluation_id=self.id,
+            synapse_client=synapse_client,
+        )
+
+    async def update_acl_async(
+        self,
```
andrewelamb, line 528: @BryanFauble I wonder if there's an opportunity to create some utils for dealing with ACLs that some/all the model classes can use?

synapseclient/api/evaluation_services.py
```diff
@@ -0,0 +1,401 @@
...
+
+    if not request_body.get("id"):
+        uri = "/evaluation"
+        response = await client.rest_post_async(uri, body=json.dumps(request_body))
+
+        logger.info(
+            f"Evaluation '{request_body.get('name')}' has been created with ID: {response.get('id')}"
+        )
+
+    elif request_body.get("id"):
```
andrewelamb, line 50: nit: Could this just be an else statement? `id` either is or isn't in the `request_body` right?

synapseclient/models/evaluation.py
```diff
@@ -1 +1,970 @@
...
+                return acl
+
+        # Add new principal
+        acl["resourceAccess"].append(
+            {"principalId": principal_id_int, "accessType": access_type}
+        )
+
+        return acl
+
+    def to_synapse_request(self, request_type: str):
```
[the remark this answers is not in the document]
jaymedina: Yes, there are only 2 possible values here so an `Enum` type would make sense. Thanks!

BryanFauble (approved):
Great work here! I didn't see anything else that popped out at me. Thanks for implementing this.

synapseclient/core/utils.py
```diff
@@ -1494,6 +1506,24 @@ def merge_dataclass_entities(
...
+                if destination_value != source_value:
+                    # Use provided logger or fall back to default
+                    log = logger
+                    if not log:
+                        import logging
+
+                        log = logging.getLogger(LOGGER_NAME)
+                    log.warning(
+                        f"Field '{field_name}' cannot be modified. Changes will be ignored."
+                    )
```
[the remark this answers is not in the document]
jaymedina: @thomasyu888 did you want me to implement this before merging (into the "main" feature branch)?

cc @BryanFauble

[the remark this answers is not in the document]
BryanFauble: Yeah ideally a single logger is created and reused thoughout the lifecycle

PR #1246, opened 2025-09-20
[SYNPY-1653] Create RecordSet, Grid, and CurationTask classes

BryanFauble:
Actually covers SYNPY-1653 and SYNPY-1655

# **Problem:**

- RecordSet and CurationTask are new concepts introduced to the Synapse Rest APIs and need to be implemented in Python.

# **Solution:**

- Implement these new concepts, new APIs, and some supporting classes

# **Testing:**

- [ ] Unit tests
- [ ] Integration tests

- Did manual testing during development, but needs more comprehensive testing

### DIFF
--- a/synapseclient/api/__init__.py
+++ b/synapseclient/api/__init__.py
@@ -16,6 +16,13 @@ from .configuration_services import (
     get_config_section_dict,
     get_transfer_config,
 )
+from .curation_services import (
+    create_curation_task,
+    delete_curation_task,
+    get_curation_task,
+    list_curation_tasks,
+    update_curation_task,
+)
 from .entity_bundle_services_v2 import (
     get_entity_id_bundle2,
     get_entity_id_version_bundle2,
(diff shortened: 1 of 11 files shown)


synapseclient/core/async_utils.py
```diff
@@ -93,6 +93,65 @@ def wrap_async_to_sync(coroutine: Coroutine[Any, Any, Any], syn: "Synapse") -> A
...
+        # Create the async generator
+        async_gen = async_gen_func(*args, **kwargs)
+
+        # Yield items from the async generator synchronously
+        try:
+            while True:
+                try:
+                    item = loop.run_until_complete(async_gen.__anext__())
+                    yield item
+                except StopAsyncIteration:
```
jaymedina, line 127: Could this try/except loop break for other reasons except `StopAsyncIteration`? Does it matter that there's no cleanup after the break here?

BryanFauble:
As a note tests are patched in https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1256

PR #1249, opened 2025-09-29
[SYNPY-1613]allow download from presigned url

danlu1:
# **Problem:**

download_from_url and download_from_url_multi_threaded are not supporting download directly via presigned url.

# **Solution:**

Update download_from_url and download_from_url_multi_threaded functions to allow file download directly via presigned url.

# **Testing:**

Unit test and integration test cases are added.

### DIFF
--- a/tests/unit/synapseclient/core/unit_test_download.py
+++ b/tests/unit/synapseclient/core/unit_test_download.py
@@ -685,7 +685,7 @@ class TestDownloadFromUrlMultiThreaded:
                 client=self.syn, download_request=request
             )
             mock_os_remove.assert_not_called()
-            mock_move.assert_called_once_with("/temp/path", "/myfakepath")
+            mock_move.assert_called_once_with("/temp/path", path)
 
     async def test_download_with_presigned_url_with_mismatch_md5(self) -> None:
         """Test downloading a file when presigned_url is provided."""


tests/integration/synapseclient/core/test_download.py
```diff
@@ -343,6 +348,175 @@ async def test_download_expired_url(
...
 
+    async def test_download_expired_presigned_url(
+        self,
+        syn: Synapse,
+        mocker: MockerFixture,
+        project_model: Project,
+    ) -> None:
+        """Tests that a file download is retried with a presigned URL but it is expired."""
+        # GIVEN a file stored in synapse
+        original_file = utils.make_bogus_data_file(40000)
```
BryanFauble, line 359: Could this be made smaller (Does the default of 100 work?). The reason is that I want to ensure the tests need to transfer as little data as possible so they run quicker

tests/integration/synapseclient/core/test_download.py
```diff
@@ -343,6 +348,175 @@ async def test_download_expired_url(
...
+
+        path = download_from_url(
+            url=url,
+            destination=tempfile.gettempdir(),
+            entity_id=None,
+            file_handle_associate_type=None,
+            file_handle_id=None,
+            expected_md5=file.file_handle.content_md5,
+            url_is_presigned=True,
+        )
```
BryanFauble, line 428: Is the intention that these function calls be made for FileEntities stored on Synapse (Or is this just that your testing this case?)

The downside of directly calling this function for Synapse FileEntities is that it bypasses the cache mechanism set up.

If the intention is that "Give a random presigned URL, just download the thing", then this is a fine solution

danlu1:
@BryanFauble My integration tests failed on a module that I didn't touch: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/18144188608/job/51642151462. What do we usually do about this type of issue?

BryanFauble:
> @BryanFauble My integration tests failed on a module that I didn't touch: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/18144188608/job/51642151462. What do we usually do about this type of issue?

@danlu1 I reran it. Just do that in the future. There are some flaky tests because of the eventually consistent nature of Synapse: https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/18144188608?pr=1249

PR #1250, opened 2025-09-30
[SYNPY-1590] Implement Submission(+Status, +Bundle) OOP model 

jaymedina:
# **Problem:**

- Describe the specific technical problem that this PR solves.
- Include  details such as:
    - How you reproduced the problem
    - What code is affected
    - Any relevant background information

# **Solution:**

- Describe the specific technical solution that this PR provides.
- Include details such as:
    - How you debugged the problem
    - Why you chose this solution
    - Does the solution make any impact beyond the immediate problem
    - Any necessary technical debt incurred
    - Any important design decisions/links to design documentation (if applicable)

# **Testing:**

- Describe any testing that was performed to validate the solution.
- Include details such as:
    - How you tested the solution
    - Any relevant test results (Do not include any test results that may contain sensitive information)
    - Any automated tests that were added to the codebase


synapseclient/models/submission.py
```diff
@@ -0,0 +1,939 @@
...
+        """
+        if not self.entity_id:
+            raise ValueError("entity_id must be set to fetch entity information")
+
+        from synapseclient import Synapse
+
+        client = Synapse.get_client(synapse_client=synapse_client)
+
+        try:
+            entity_info = await client.rest_get_async(f"/entity/{self.entity_id}")
```
BryanFauble: This could be replaced by this function call https://github.com/Sage-Bionetworks/synapsePythonClient/blob/436125780767c4cecb4fc0810d3d1fe6aa72cd75/synapseclient/api/entity_services.py#L151 which will make this API call.

synapseclient/models/submission.py
```diff
@@ -0,0 +1,939 @@
...
+
+            # If this is a DockerRepository, fetch docker image tag & digest, and add it to the entity_info dict
+            if (
+                entity_info.get("concreteType")
+                == "org.sagebionetworks.repo.model.docker.DockerRepository"
+            ):
+                docker_tag_response = await client.rest_get_async(
+                    f"/entity/{self.entity_id}/dockerTag"
+                )
```
BryanFauble: Ideally all these calls can be put into a file (Or create a new one) here: https://github.com/Sage-Bionetworks/synapsePythonClient/tree/develop/synapseclient/api

synapseclient/api/evaluation_services.py
```diff
@@ -386,3 +388,468 @@ async def get_evaluation_permissions(
...
+    Returns:
+        A response JSON containing the created Submission model's attributes.
+    """
+    from synapseclient import Synapse
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    uri = "/evaluation/submission"
+
+    # Add etag as query parameter if provided
```
linglp: since `etag` is a required parameter, I think this needs to be reframed?

tests/unit/synapseclient/models/synchronous/unit_test_submission.py
```diff
@@ -0,0 +1,801 @@
...
+                evaluation_id=evaluation_id,
+                status=status,
+                synapse_client=self.syn,
+            )
+            assert len(submissions) == 1
+            assert isinstance(submissions[0], Submission)
+            assert submissions[0].id == SUBMISSION_ID
+
+    @pytest.mark.asyncio
+    async def test_get_user_submissions_async(self) -> None:
```
linglp: same comment like above.

tests/unit/synapseclient/models/synchronous/unit_test_submission.py
```diff
@@ -0,0 +1,801 @@
...
+                evaluation_id=evaluation_id,
+                user_id=user_id,
+                synapse_client=self.syn,
+            )
+            assert len(submissions) == 1
+            assert isinstance(submissions[0], Submission)
+            assert submissions[0].id == SUBMISSION_ID
+
+    @pytest.mark.asyncio
+    async def test_get_submission_count_async(self) -> None:
```
linglp: same comment like above.

synapseclient/models/submission.py
```diff
@@ -0,0 +1,939 @@
...
+        """
+        yield from wrap_async_generator_to_sync_generator(
+            async_gen_func=cls.get_evaluation_submissions_async,
+            evaluation_id=evaluation_id,
+            status=status,
+            synapse_client=synapse_client,
+        )
+
+    @classmethod
+    def get_user_submissions(
```
linglp: I am not exactly sure what's going on, but when I tested it out the code on my end, I observed different outputs from sync v.s. async method: 

```
def get_user_submission():
    for submission in Submission.get_user_submissions(
        evaluation_id=evaluation.id,
        user_id="3443707"
    ):
        print("user submission", submission)
get_user_submission()

async def get_user_submission_async():
    async for submission in Submission.get_user_submissions_async(
        evaluation_id=evaluation.id,
        user_id="3443707"
    ):
        print("user submission async", submission)

asyncio.run(get_user_submission_async())
```
The async version is able to print out many submission records, while the sync function gets stuck.

synapseclient/models/submission.py
```diff
@@ -0,0 +1,939 @@
...
+            user_id=user_id,
+            synapse_client=synapse_client,
+        ):
+            submission_object = cls().fill_from_dict(synapse_submission=submission_data)
+            yield submission_object
+
+    @staticmethod
+    async def get_submission_count_async(
+        evaluation_id: str,
+        status: Optional[str] = None,
```
linglp: Same comment as above: I think we should add validation to ensure that `status` is one of the allowed enum values.

linglp (dismissed):
Hi @jaymedina ! I spent some time today testing different functions in this script. For some reason, a few of them hang for me, and I’m curious whether you’ve encountered similar issues.
In addition, there are two things you might want to consider:
*  I’m not currently seeing any code that validates whether status is one of the allowed values listed [here](https://restdocs.synapse.org/rest/org/sagebionetworks/evaluation/model/SubmissionStatusEnum.html). It may be worth adding validation to ensure status is valid.
* In the unit tests under models/synchronous/unit_test_submission_bundle.py and models/synchronous/unit_test_submission.py, many tests still appear to be calling async methods instead of their sync counterparts (for example, cancel_async instead of cancel).

synapseclient/models/submission.py
jaymedina: Sounds good! Initially was going to fold it into `entity_services.py`, but I think this would be incorrect since technically this service falls under handling the docker commits (/tags) and not the Docker Repository entities themselves ([API reference](https://rest-docs.synapse.org/rest/#org.sagebionetworks.repo.web.controller.DockerCommitController)). I've made a new file called `docker_commit_services.py` and stored it there.

synapseclient/models/submission.py
jaymedina: I'm not able to re-produce the error described. For my attempt, I'm using the `nf-dynamic-challenge-test` evaluation queue in our **DPE-testing** Project on Synapse:

```
In [29]: for i in Submission.get_user_submissions(nf_dynamic_challenge):
    ...:     print(i.id)
    ...: 
9751432
9751433
9751434
9752613
9752619

In [30]: async def get_user_submission_async():
    ...:     async for submission in Submission.get_user_submissions_async(
    ...:         nf_dynamic_challenge
    ...:     ):
    ...:         print("user submission async", submission.id)
    ...: 

In [31]: import asyncio

In [32]: asyncio.run(get_user_submission_async())
user submission async 9751432
user submission async 9751433
user submission async 9751434
user submission async 9752613
user submission async 9752619
```

jaymedina:
@BryanFauble I've addressed the remaining threads from Lingling and left them open so you can take a look at their resolution while she's out on leave. Please LMK if you think this is ready for a merge, or if there's anything else I can do to get this ready to ship before the end of this sprint 🚢

PR #1315, opened 2026-02-04
[DPE-1577] Supporting storage locations and migration OOP models

BryanFauble:
# **Problem:**

Storage locations and migration functions were not ported over into the OOP models — users had to use the legacy `synapseclient` and `synapseutils` threading-based API to configure storage locations or migrate files between them.

# **Solution:**
1. Added StorageLocation model with `store_async` / `get_async` support, covering all location types: Synapse S3, External S3, GCS, SFTP, HTTPS, External Object Store, and Proxy servers
2. Added StorageLocationConfigurableMixin on Project and Folder so storage locations can be set directly from those entities
3. Added async migration.py service with SQLite-backed indexing and `multipart_copy_async` for migrating files between storage locations, replacing the threading-based synapseutils.migrate_functions
4. Fixed StorageLocationType disambiguation: SFTP and HTTPS share the same concreteType, now resolved via (concreteType, uploadType) tuple lookup

# **Testing:**

1. Unit tests added for StorageLocation, migration, migration_types, and storage_location_services
2. Integration tests added covering SYNAPSE_S3, External S3 (requires AWS credentials), GCS (requires google-cloud-storage), and project/folder storage location assignment
3. https://synapsepythonclient--1315.org.readthedocs.build/en/1315/


docs/explanations/storage_location_architecture.md
```diff
@@ -0,0 +1,819 @@
...
+
+Use this decision tree to select the appropriate storage type for your use case:
+
+```mermaid
+flowchart TB
+    Start{Need custom storage?}
+    Start -->|No| DEFAULT[Use default Synapse storage]
+    Start -->|Yes| Q1{Want Synapse to<br/>manage storage?}
+
+    Q1 -->|Yes| SYNAPSE_S3[Use SYNAPSE_S3]
```
thomasyu888: Should this point to "use default synapse storage node"?

thomasyu888 (commented):
Thanks for this

1. I see that the tests are failing - is this due to what's been added here?
2. The documentation isn't building on readthedocs

here's the error

```
ERROR   -  mkdocstrings: synapseclient.models.mixins.manifest.ManifestGeneratable could not be found
ERROR   -  Error reading page 'reference/experimental/mixins/manifest_generatable.md':
ERROR   -  Could not collect 'synapseclient.models.mixins.manifest.ManifestGeneratable'

Aborted with a BuildError!
```

synapseclient/models/protocols/storable_container_protocol.py
thomasyu888: This looks out of scope for this PR and it seems like a harmless addition. (both can be true). Thoughts?

synapseclient/models/mixins/storage_location_mixin.py
```diff
@@ -0,0 +1,451 @@
...
+            }
+            return await create_project_setting(
+                request=project_destination,
+                synapse_client=synapse_client,
+            )
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Entity_GetProjectSetting: {self.id}"
+    )
+    async def get_project_setting_async(
```
BryanFauble: This should not live in a storage specific mixin because ProjectSettings will be used outside of storage locations. Instead there could be ProjectSettingsMixin that has these project-specific methods. That ProjectSettingsMixin can extend this `StorageLocationConfigurable` mixin so that it inherits what exists here.

synapseclient/models/mixins/storage_location_mixin.py
```diff
@@ -0,0 +1,451 @@
...
+            existing_setting["locations"] = locations
+            await update_project_setting(
+                request=existing_setting,
+                synapse_client=synapse_client,
+            )
+            return await get_project_setting(
+                project_id=self.id,
+                setting_type="upload",
+                synapse_client=synapse_client,
+            )
```
BryanFauble: It needs to be made clear in the docstring that calling this method is a destructive update. Setting this will replace all storage locations that may have previously been set. An example to show a partial update should be included.

synapseclient/models/mixins/storage_location_mixin.py
```diff
@@ -0,0 +1,451 @@
...
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Entity_SetStorageLocation: {self.id}"
+    )
+    async def set_storage_location_async(
+        self,
+        storage_location_id: Optional[Union[int, List[int]]] = None,
+        *,
+        synapse_client: Optional[Synapse] = None,
+    ) -> Dict[str, Any]:
```
BryanFauble: Should this return StorageLocation instead?

synapseclient/models/services/migration_types.py
```diff
@@ -0,0 +1,391 @@
...
+    def __eq__(self, other: object) -> bool:
+        if not isinstance(other, MigrationKey):
+            return False
+        return (
+            self.id == other.id
+            and self.type == other.type
+            and self.version == other.version
+            and self.row_id == other.row_id
+            and self.col_id == other.col_id
+        )
```
BryanFauble: I thought that these are default dunder methods added as a result of this being an `@dataclass`

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1530 @@
...
+# https://github.com/aws/aws-sdk-java/blob/57ed2e4bd57e08f316bf5c6c71f6fd82a27fa240/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManagerConfiguration.java#L46
+DEFAULT_PART_SIZE = 100 * utils.MB
+
+# Batch size for database operations so the batch operations are chunked.
+BATCH_SIZE = 500
+
+# Maximum concurrent file copy.
+MAX_CONCURRENT_FILE_COPIES = max(int(Synapse().max_threads / 2), 1)
+
+logger = logging.getLogger(__name__)
```
BryanFauble: This should swap to using the logger from the `Synapse` class instead of creating a new logger like this.

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1530 @@
...
+        )
+        return user_input.strip().lower() == "y"
+    else:
+        logger.info(
+            "%s items for migration. "
+            "force option not used, and console input not available to confirm migration, aborting. "
+            "Use the force option or run from an interactive shell to proceed with migration.",
+            count,
+        )
+        return False
```
BryanFauble: This is among the (only?) patterns to ask for user input. We need to ensure that any doc strings that are user facing should have the clear expectation that there is a user input step. Additionally, there could also be a boolean that users could set in order to skip user input and assume a "Y" to this question.

synapseclient/models/protocols/download_list_protocol.py
```diff
@@ -0,0 +1,97 @@
...
+generated at runtime."""
+
+from typing import Any, Dict, Optional, Protocol
+
+from typing_extensions import Self
+
+from synapseclient import Synapse
+
+
+class DownloadListManifestRequestSynchronousProtocol(Protocol):
```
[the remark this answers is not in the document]
BryanFauble: Thanks

docs/js/mermaid-init.js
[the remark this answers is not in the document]
danlu1: This was introduced by the first draft version of code . I will double check this.

docs/reference/experimental/mixins/storage_location_configurable.md
```diff
@@ -0,0 +1,54 @@
...
+    bucket="my-bucket",
+    sts_enabled=True,
+).store()
+
+# Set storage location on a folder
+folder = Folder(id="syn123").get()
+folder.set_storage_location(storage_location_id=storage.storage_location_id)
+
+# Get STS credentials
+credentials = folder.get_sts_storage_token(
```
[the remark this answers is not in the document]
danlu1: I think so otherwise it will error out with `synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error: 
Entity must have an STS-enabled storage location` and it's useful when we do sts tranfer. Here is another example. https://github.com/Sage-Bionetworks/synapsePythonClient/blob/e06c715e645adfa49d6a2265c92246f4d1ad0418/synapseclient/core/upload/upload_functions.py#L94-L118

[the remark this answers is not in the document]
danlu1: I updated the instruction and docstrings.

docs/tutorials/python/tutorial_screenshots/migration_results.png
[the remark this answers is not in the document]
danlu1: Yes, this is intentional. This is attached to show detailed trackback can be checked in the csv. I can share a successful record as well.

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1530 @@
...
+
+This module provides native async implementations of the indexing and migration functionality
+"""
+
+import asyncio
+import collections.abc
+import json
+import logging
+import os
+import sqlite3
```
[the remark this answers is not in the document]
danlu1: Added a test_import_sqlite function as what we did for pandas.

synapseclient/models/services/migration_types.py
danlu1: This was used to make the MigrationKey hashable. I removed these sections and use frozen=True instead.

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1530 @@
...
+# with upload failures of the actual bytes.
+# this value aligns with what some AWS client libraries use e.g.
+# https://github.com/aws/aws-sdk-java/blob/57ed2e4bd57e08f316bf5c6c71f6fd82a27fa240/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManagerConfiguration.java#L46
+DEFAULT_PART_SIZE = 100 * utils.MB
+
+# Batch size for database operations so the batch operations are chunked.
+BATCH_SIZE = 500
+
+# Maximum concurrent file copy.
+MAX_CONCURRENT_FILE_COPIES = max(int(Synapse().max_threads / 2), 1)
```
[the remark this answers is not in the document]
danlu1: I replaced the semaphore creation with `_get_parallel_file_transfer_semaphore`. One thing to note is that `MAX_CONCURRENT_FILE_COPIES` (defined as `max(int(Synapse().max_threads / 2), 1)`) was carried over from older code, and I couldn’t find a clear reference for it.
Using `_get_parallel_file_transfer_semaphore` will change the concurrency to `max(self.max_threads * 2, 1)`, which increases the number of concurrent file transfers.

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1530 @@
...
+            active_tasks.add(task)
+
+        if row_count == 0 and not pending_file_handles:
+            # we've run out of migratable sqlite rows, we have nothing else
+            # to submit, so we break out and wait for all remaining
+            # tasks to conclude.
+            break
+
+        # Wait for tasks if at capacity or end of batch
+        if len(active_tasks) >= MAX_CONCURRENT_FILE_COPIES or len(batch) < BATCH_SIZE:
```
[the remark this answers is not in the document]
danlu1: This was carried over from the old code. The semaphore controls execution concurrency, not task creation.
It’s acquired inside `_migrate_item_async` when the task actually starts running, so `active_tasks` can exceed the semaphore’s file limit since it’s really tracking queued tasks as well.
Since our goal is to maximize throughput within the while loop, I think using `asyncio.FIRST_COMPLETED` should do the work so the if condition is redundant. I went ahead and remove it.

docs/explanations/storage_location_architecture.md
```diff
@@ -0,0 +1,819 @@
...
+    Q1 -->|No| Q2{What storage<br/>backend?}
+
+    Q2 -->|AWS S3| Q3{Synapse accesses<br/>bucket directly?}
```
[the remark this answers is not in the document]
danlu1: Yes, as this [instruction](https://docs.synapse.org/synapse-docs/custom-storage-locations), you should update bucket policy and enable cross-origin resource sharing for bucket like storage location.

thomasyu888 (commented):
Thanks Dan for your work here - I see that the tests are failing - do the tests pass locally? If so, does this just need to be re-run?

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,268 @@
...
+    Folder,
+    Project,
+    StorageLocation,
+    StorageLocationType,
+)
+
+syn = synapseclient.login()
+
+# Step 1: Retrieve the project
+my_project = Project(name="My uniquely named project about Alzheimer's Disease").get()
```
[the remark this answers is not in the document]
thomasyu888: This is ok

docs/explanations/storage_location_architecture.md
```diff
@@ -0,0 +1,820 @@
...
+
+    }
+    StorageLocation --> StorageLocationType : storage_type
+    StorageLocation --> UploadType : upload_type
+    StorageLocationConfigurable <|-- Project : implements
+    StorageLocationConfigurable <|-- Folder : implements
+    StorageLocationConfigurable ..> ProjectSetting : returns
+    StorageLocationConfigurable ..> UploadDestinationListSetting : uses
+
+```
```
BryanFauble, line 161: Are there updates that are needed to align these models and concepts to what is in the code?

synapseclient/models/project_setting.py
```diff
@@ -0,0 +1,328 @@
...
+                from synapseclient.models import ProjectSetting
+
+                syn = Synapse()
+                syn.login()
+
+                async def main():
+                    setting = await ProjectSetting(
+                        project_id="syn123", settings_type="upload"
+                    ).get_async()
+                    setting.locations = [12345, 67890]
```
BryanFauble, line 289: This would be great to point to documentation on how to create a new one of these locations, or read what the available values are that can be used.

synapseclient/models/mixins/storage_location_mixin.py
```diff
@@ -0,0 +1,461 @@
...
+        synapse_client: Optional[Synapse] = None,
+    ) -> "ProjectSetting":
+        """Set the upload storage location for this entity. This configures where
+        files uploaded to this entity will be stored.
+
+        **This is a destructive update.** The provided `storage_location_id` value(s)
+        will **replace** any storage locations previously configured on this entity.
+        To add a storage location without removing existing ones, first retrieve the
+        current setting via `get_project_setting_async`, append to its `locations`
+        list, and call `store_async` on the returned `ProjectSetting` directly.
```
BryanFauble, line 295: It would be great to define in our documentation how the "default" upload location works with the `locations` list in this file here. Last is default? First is default?

synapseclient/models/storage_location.py
```diff
@@ -0,0 +1,542 @@
...
+# Subset of _STORAGE_TYPE_SPECIFIC_FIELDS that are required (no default value).
+# Fields with defaults (e.g. base_key=None, sts_enabled=False) are omitted.
+_REQUIRED_STORAGE_TYPE_SPECIFIC_FIELDS: Dict[str, set] = {
+    StorageLocationType.EXTERNAL_S3: {"bucket", "endpoint_url"},
+    StorageLocationType.EXTERNAL_GOOGLE_CLOUD: {"bucket"},
+    StorageLocationType.EXTERNAL_OBJECT_STORE: {"bucket", "endpoint_url"},
+    StorageLocationType.EXTERNAL_SFTP: {"url"},
+    StorageLocationType.EXTERNAL_HTTPS: {"url"},
+    StorageLocationType.PROXY: {"proxy_url", "secret_key", "benefactor_id"},
+}
```
BryanFauble: Not referenced in any code in this branch

synapseclient/models/storage_location.py
```diff
@@ -0,0 +1,542 @@
...
+            the Synapse REST API.
+        banner: The banner text to display to a user every time a file is uploaded.
+            This field is optional.
+        description: A description of the storage location. This description is
+            shown when a user has to choose which upload destination to use.
+
+    Attributes:
+        bucket: The name of the S3 or Google Cloud Storage bucket. Applicable to
+            SYNAPSE_S3, EXTERNAL_S3, EXTERNAL_GOOGLE_CLOUD, and
+            EXTERNAL_OBJECT_STORE types.
```
BryanFauble: I believe a format like this across the docstring would make this the most readable for users.

```suggestion
        bucket: The name of the S3 or Google Cloud Storage bucket. Applicable to
            `SYNAPSE_S3`, `EXTERNAL_S3`, `EXTERNAL_GOOGLE_CLOUD`, and
            `EXTERNAL_OBJECT_STORE` types.
```

synapseclient/models/storage_location.py
```diff
@@ -0,0 +1,542 @@
...
+
+    # REQUIRED fields
+    _REQUIRED_FIELDS = {
+        StorageLocationType.EXTERNAL_S3: {"bucket"},
+        StorageLocationType.EXTERNAL_GOOGLE_CLOUD: {"bucket"},
+        StorageLocationType.EXTERNAL_OBJECT_STORE: {"bucket", "endpoint_url"},
+        StorageLocationType.EXTERNAL_SFTP: {"url"},
+        StorageLocationType.EXTERNAL_HTTPS: {"url"},
+        StorageLocationType.PROXY: {"proxy_url", "secret_key", "benefactor_id"},
+    }
```
BryanFauble, line 232: If a user printers out an instance of `StorageLocation` do these show up? Since there are internal concepts I'd rather we don't expose it out. Perhaps we can use the dataclass `field` for these with a `repr=False`

synapseclient/models/storage_location.py
```diff
@@ -0,0 +1,542 @@
...
+        self.created_on = (
+            synapse_response.get("createdOn", None)
+            if synapse_response.get("createdOn", None) is not None
+            else None
+        )
+        self.created_by = (
+            synapse_response.get("createdBy", None)
+            if synapse_response.get("createdBy", None) is not None
+            else None
+        )
```
BryanFauble: The extra wrapping of these `.get("xxx", None)` shouldn't be needed.

`synapse_response.get("createdBy")` for example should get us the same functional result.

synapseclient/models/storage_location.py
```diff
@@ -0,0 +1,542 @@
...
+            key = (type_suffix, self.upload_type)
+            if key in _CONCRETE_UPLOAD_TO_STORAGE_TYPE:
+                self.storage_type = _CONCRETE_UPLOAD_TO_STORAGE_TYPE[key]
+        # Type-specific fields — only populate attributes relevant to this storage type
+        if self.storage_type:
+            for field_name, api_key in _STORAGE_TYPE_SPECIFIC_FIELDS.get(
+                self.storage_type, {}
+            ).items():
+                setattr(self, field_name, synapse_response.get(api_key, None))
+        return self
```
BryanFauble, line 365: This should clean up some of the logic and add in a warning if for some reason the storage location concrete_type is not known:

```
            if key in _CONCRETE_UPLOAD_TO_STORAGE_TYPE:
                self.storage_type = _CONCRETE_UPLOAD_TO_STORAGE_TYPE[key]
                # Type-specific fields — only populate attributes relevant to this storage type
                for field_name, api_key in _STORAGE_TYPE_SPECIFIC_FIELDS.get(
                    self.storage_type, {}
                ).items():
                    setattr(self, field_name, synapse_response.get(api_key, None))
            else:
                # Log a warning if we encounter an unknown concreteType, rather than failing to parse all fields.
                pass
        return self
```

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,268 @@
...
+external_s3_folder = external_s3_folder.store()
+
+# Set the storage location for the folder
+external_s3_folder.set_storage_location(
+    storage_location_id=external_s3_storage_location.storage_location_id
+)
+external_s3_folder_storage_location = external_s3_folder.get_project_setting()
+# Verify the storage location is set correctly
+assert (
+    external_s3_folder_storage_location["locations"][0]
```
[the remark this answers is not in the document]
BryanFauble: @danlu1 To review

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,311 @@
...
+sftp_folder.set_storage_location(storage_location_id=sftp_storage.storage_location_id)
+sftp_folder_storage_location = sftp_folder.get_project_setting()
+# Verify the storage location is set correctly
+assert (
+    sftp_folder_storage_location["locations"][0] == sftp_storage.storage_location_id
+), "Folder storage location does not match the storage location"
+
+# Add a file to the sftp folder, need 'pysftp' package installed.
+file = File(path="/path/to/your/file.csv", parent_id=sftp_folder.id)
+file = file.store()
```
BryanFauble: Consider dropping this single file upload example out of the tutorial script.

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,311 @@
...
+
+# Step 6: Create an HTTPS storage location
+# EXTERNAL_HTTPS shares the same underlying API type as EXTERNAL_SFTP but is used
+# when the external server is accessed over HTTPS rather than SFTP.
+my_https_folder = Folder(name="my-folder-for-https", parent_id=my_project.id)
+my_https_folder = my_https_folder.store()
+
+my_https_url = "https://my-https-server.example.com"
```
BryanFauble: nit: Inconsistent caps compared to `MY_SFTP_URL = "sftp://your-sftp-server.example.com/upload"` above.

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,311 @@
...
+print(f"storage location type: {https_storage.storage_type}")
+
+my_https_folder.set_storage_location(
+    storage_location_id=https_storage.storage_location_id
+)
+my_https_folder_storage_location = my_https_folder.get_project_setting()
+assert (
+    my_https_folder_storage_location["locations"][0]
+    == https_storage.storage_location_id
+), "Folder storage location does not match the storage location"
```
BryanFauble: Adding these lines in is great for testing, but requires that a user following this script reads more to understand what is going on. I'd consider removing all of these from the tutorial script and assume out testing efforts verify this functionality is correct.

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,311 @@
...
+
+# Note: The Python client does not support uploading files directly to HTTPS
+# storage locations. To add files, use the Synapse web UI or REST API directly.
+
+# Step 7: Create an External Object Store storage location
+# Use this for S3-compatible stores (e.g. OpenStack Swift) not accessed by Synapse.
```
BryanFauble: Is there a reason to keep a mention of "OpenStack Swift" in this example?

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,311 @@
...
+
+
+proxy_file_handle_id = asyncio.run(create_proxy_file_handle())
+# Associate the ProxyFileHandle with a Synapse File entity
+proxy_file = File(
+    parent_id=my_proxy_folder.id,
+    name="test.csv",
+    data_file_handle_id=proxy_file_handle_id,
+).store()
+print(f"Synapse entity: {proxy_file.id}")
```
BryanFauble: It might sense to organize this into a separate script for proxy storage locations specifically. Including this here overloads the core concepts of this tutorial script.

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -0,0 +1,311 @@
...
+migrate_result = my_migration_folder.migrate_indexed_files(
+    db_path="/path/to/your/migration.db",
+    continue_on_error=True,
+    force=True,  # Skip interactive confirmation for tutorial purposes
+)
+migrate_result.as_csv("/path/to/your/migrate_results.csv")
+if migrate_result is not None:
+    print(f"Migrated counts by status: {migrate_result.counts_by_status}")
+else:
+    print("Migration was aborted (confirmation declined).")
```
BryanFauble: Move migration pieces to it's own tutorial so that storage location setup is focused on that piece only.

docs/explanations/storage_location_architecture.md
```diff
@@ -0,0 +1,820 @@
...
+
+### Key Components
+
+| Component | Description |
+|-----------|-------------|
+| [synapseclient.models.StorageLocation] | The model representing a storage location setting in Synapse |
+| [synapseclient.models.StorageLocationType] | Enumeration defining the supported storage backend types |
+| [synapseclient.models.UploadType] | Enumeration defining the upload protocol for each storage type |
+| [synapseclient.models.mixins.StorageLocationConfigurable] | Mixin providing storage management methods to entities |
+| [synapseclient.models.mixins.UploadDestinationListSetting] | Dataclass defining the upload destination list setting containing storage location IDs |
```
BryanFauble: This does not exist

docs/explanations/storage_location_architecture.md
```diff
@@ -0,0 +1,820 @@
...
+
+### STS Token Retrieval
+
+STS (AWS Security Token Service) enables direct S3 access using temporary credentials.
+
+When a Synapse client is constructed (`Synapse.__init__`), it creates an in-memory token cache:
+
+- `self._sts_token_store = sts_transfer.StsTokenStore()` (see `synapseclient/client.py`)
+
+The store caches STS tokens per entity and permission so repeated access to the same storage location can reuse credentials without a round-trip to the REST API.
```
BryanFauble: I would be hesitant to use the word "cache" here.

docs/reference/experimental/mixins/storage_location.md
```diff
@@ -0,0 +1,127 @@
...
+| `index_files_for_migration` | Index files for migration to a new storage location |
+| `migrate_indexed_files` | Migrate previously indexed files |
+
+### `ProjectSettingsMixin` (extends `StorageLocationConfigurable`)
+
+| Method | Description |
+|--------|-------------|
+| `set_storage_location` | Set the upload storage location for this entity (destructive replace) |
+| `get_project_setting` | Get project settings (upload, external_sync, etc.) |
+| `delete_project_setting` | Delete a project setting |
```
BryanFauble: Rather than duplicating information like function names/descriptions - these should be sourced from the code itself with the `:::`  syntax that mkdocs has

docs/reference/experimental/mixins/storage_location.md
```diff
@@ -0,0 +1,127 @@
...
+
+    # Step 2: migrate (force=True for non-interactive scripts)
+    result = await project.migrate_indexed_files_async(
+        db_path=index_result.db_path,
+        force=True,
+    )
+    print(result.counts_by_status)
+
+asyncio.run(main())
+```
```
BryanFauble: These are good example to show, but rather these should exist within the docstrings of the methods/functions themselves instead of independently represented here.

docs/tutorials/python/storage_location.md
```diff
@@ -0,0 +1,266 @@
...
+Synapse is never involved in the data transfer, only in storing the metadata.
+
+You can add a profile to work with s3 in `~/.synapseConfig`
+
+Add a section matching your endpoint+bucket URL:
+```
+[https://s3.us-east-1.amazonaws.com/test-external-object-store]
+profile_name = my-s3-profile
+```
+Then ensure my-s3-profile exists in `~/.aws/config` with valid keys:
```
BryanFauble: Could we point to the correct AWS documents instead of telling folks to to set their AWS credentials.

synapseclient/core/remote_file_storage_wrappers.py
```diff
@@ -316,8 +316,9 @@ def progress_callback(*args, **kwargs) -> None:
             progress_bar.update(args[0] - progress_bar.n)
 
         parsedURL = SFTPWrapper._parse_for_sftp(url)
+        port_kwargs = {"port": parsedURL.port} if parsedURL.port else {}
```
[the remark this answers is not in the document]
BryanFauble: +1

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1561 @@
...
+
+    Returns:
+        True if the entity is already indexed.
+    """
+    indexed_row = cursor.execute(
+        "select 1 from migrations where id = ?", (entity_id,)
+    ).fetchone()
+
+    if indexed_row:
+        synapse_client.logger.debug(f"{entity_id} already indexed, skipping")
```
BryanFauble, line 190: The synapse_client is `Optional` in the arguments, which means it could be `None`. This would error if not passed in.

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1561 @@
...
+
+            # Create table snapshot if needed using the async API
+            if (
+                key.type == MigrationType.TABLE_ATTACHED_FILE.value
+                and create_table_snapshots
+                and last_key.id != key.id
+            ):
+                await asyncio.to_thread(
+                    Table(id=key.id).snapshot_async, synapse_client=synapse_client
+                )
```
BryanFauble: Is there a reason why this is using `.to_thread` instead of `await .snapshot_async`?

synapseclient/models/services/migration.py
```diff
@@ -0,0 +1,1561 @@
...
+                # if it the record shares a file handle it will be picked up later
+                # when its file handle is completed.
+                continue
+
+            pending_keys.add(key)
+
+            # Check for existing copy — run in a thread to avoid blocking the event loop.
+            to_file_handle_id = await asyncio.to_thread(
+                _check_file_handle_exists, cursor, from_file_handle_id
+            )
```
BryanFauble: I don't think that these `.to_thread` calls are getting us anything and might cause more problems that it could possibly solve.

mkdocs.yml
```diff
@@ -140,15 +142,18 @@ nav:
               - JSONSchema: reference/experimental/async/json_schema.md
               - Wiki: reference/experimental/async/wiki.md
               - FormGroup and Form: reference/experimental/async/form.md
+              - StorageLocation: reference/experimental/async/storage_location.md
           - Mixins:
               - AccessControllable: reference/experimental/mixins/access_controllable.md
               - StorableContainer: reference/experimental/mixins/storable_container.md
+              - ManifestGeneratable: reference/experimental/mixins/manifest_generatable.md
```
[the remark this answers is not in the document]
BryanFauble: +1

BryanFauble (commented):
There are a few things to consider, but nothing that would change the logic from how this is written and consumed. Thanks for this!

synapseclient/core/remote_file_storage_wrappers.py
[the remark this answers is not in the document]
danlu1: This is added since I run into `pysftp host key verification failure when connecting to `127.0.0.1:2222`.` The actual error was: `"paramiko.ssh_exception.SSHException: No hostkey for host 127.0.0.1 found."` in `_retry_pysftp_connection` . This is added to check the correct port number when it's not on the default port.

synapseclient/models/services/migration.py
danlu1: The synapse_client would be passed in `index_files_for_migration_async` which is called before calling this function. So synapse_client should not be None.

synapseclient/models/mixins/asynchronous_job.py
```diff
@@ -29,6 +30,7 @@
 ASYNC_JOB_URIS = {
     AGENT_CHAT_REQUEST: "/agent/chat/async",
     CREATE_GRID_REQUEST: "/grid/session/async",
+    DOWNLOAD_LIST_MANIFEST_REQUEST: "/download/list/manifest/async",
```
danlu1: This was added by AI from the first draft and is removed.

synapseclient/models/mixins/enum_coercion.py
```diff
@@ -0,0 +1,32 @@
+"""Mixin for automatic enum coercion in dataclasses."""
+
+from typing import Any, ClassVar, Dict
+
+
+class EnumCoercionMixin:
```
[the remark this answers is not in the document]
danlu1: Added "This allows strings or enums to be used interchangeably for the same field." to the docstring

synapseclient/models/storage_location.py
danlu1: This can be mitigate in the custom _repr__ functionl

danlu1 (approved):

PR #1335, opened 2026-03-05
[SYNPY-1786] Added examples for getting data into a grid

andrewelamb:
# **Problem:**
Curator Extension users asked for simple example scripts for adding data into a grid for both file-based and record-based curation.

# **Solution:**
Two simple examples were added.

https://synapsepythonclient--1335.org.readthedocs.build/en/1335/


BryanFauble (changes requested):
Thanks for adding these examples, they'll be really helpful for folks getting started with Grid curation. There are a few small things to clean up before we merge.

_Note: This comment was drafted with AI assistance and reviewed by me for accuracy._

docs/guides/extensions/curator/metadata_curation.md
```diff
@@ -272,6 +272,116 @@ else:
...
+
+# 6. Create EntityView and CurationTask
+view_id, task_id = create_file_based_metadata_task(
+    folder_id=folder.id,
+    curation_task_name=CURATION_TASK_NAME,
+    instructions=INSTRUCTIONS,
+    schema_uri=JSON_SCHEMA_URI,
+    synapse_client=syn,
+)
+
```
linglp: Before you clean up all the test resources, what's the expected behavior? I think currently the code can run, but as a user, I don't know what the expected behaviors are.

docs/guides/extensions/curator/metadata_curation.md
```diff
@@ -272,6 +272,116 @@ else:
...
+# 5. Store the record set with the path to the CSV file as an annotation.
+# TODO: https://sagebionetworks.jira.com/browse/SYNPY-1781
+# Once SYNPY-1781 is finished add code here for uploading data from a CSV file into a grid session.
+record_set.get(synapse_client=syn)
+record_set.path = CSV_PATH
+record_set = record_set.store(synapse_client=syn)
+
+# 6. Cleanup all Synapse entities created
+folder.delete(synapse_client=syn)
+record_set.delete(synapse_client=syn, delete_source=True)
```
linglp: I think this is incorrect. The `delete_source` keyword was added to `task` not `record_set`

linglp (changes requested):
@andrewelamb Thanks for putting together the documentation. In your example, you are reading the data model CSV and trying to use its rows as annotation data, and even when I tested with our own example CSVs, headers like "Valid Values" (it should be `ValidValues`) will trigger an error. 

Also, based on my understanding, Bulk RNA-seq Assay or Biospecimen would be better examples for file-based tasks since they describe actual data files. Patient records is tabular data records rather than file-based in my opinions. 

After you finish all the changes, could you run both scripts on your end to make sure that everything is working smoothly? I had some issues when I tested it out on my end.

linglp (approved):
LGTM! Thanks

docs/guides/extensions/curator/metadata_curation.md
thomasyu888, line 1: I'm concerned about the existing link: https://python-docs.synapse.org/en/stable/guides/extensions/curator/metadata_curation/ that would break

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,179 @@
...
+    source_id = curation_task.task_properties.file_view_id
+    grid_sessions = list(Grid.list(source_id=source_id))
+    if grid_sessions:
+        latest_grid = grid_sessions[0]
+        print(f"Attaching to existing grid session: {latest_grid.session_id}")
+    else:
+        print(f"No active grid session for {source_id}; creating one.")
+        latest_grid = Grid(
+            initial_query=Query(sql=f"SELECT * FROM {source_id}")
+        ).create()
```
thomasyu888: We should get some user input here, but this is probably speaking to the fact that `source_id` could be a better source of creation of the grid instead of having users having a different interface of creating a grid

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,179 @@
...
+
+Each `CurationTask` carries the information you need to decide which workflow to follow — most importantly, `task_properties` tells you whether it's a record-based task (you'll see `record_set_id`) or a file-based task (you'll see `file_view_id`).
+
+```python
+from pprint import pprint
+from synapseclient.models import CurationTask
+
+PROJECT_ID = "syn123456789"  # The Synapse ID of the project to list tasks from
+
+for curation_task in CurationTask.list(project_id=PROJECT_ID):
```
thomasyu888: Can we create a ticket to extend this, there is a "assignToMe" boolean that we're missing here now: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/curation/ListCurationTaskRequest.html

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,179 @@
...
+    ```python
+    latest_grid.export_to_record_set()
+    print(f"Exported to RecordSet version: {latest_grid.record_set_version_number}")
+    ```
+
+- For **file-based** tasks, synchronize the grid against the file view to push annotation changes back to the underlying files.
+
+    ```python
+    latest_grid.synchronize()
+    ```
```
thomasyu888, line 148: Should synchronize be run before and after? This applies the changes

docs/guides/extensions/curator/curation.md
```diff
@@ -272,6 +271,154 @@ else:
     print("No validation results available. The Grid session must be exported to generate validation results.")
 ```
 
+### Example: Getting data into a Grid for a file-based workflow
```
thomasyu888: If we focussed this section on setting up Curator, then we can put the validation and getting data into the grid into the contribution section

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: https://sagebionetworks.jira.com/browse/SYNPY-1841

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,181 @@
...
+`import_csv` upserts rows into the grid based on the `upsert_keys` the administrator configured on the curation task. Existing rows matching on those keys are updated; new rows are inserted.
+
+```python
+latest_grid = latest_grid.import_csv(path=edited_path)
+print(
+    "Upserted edits into grid session: "
+    f"https://www.synapse.org/Grid:default?sessionId={latest_grid.session_id}"
+)
+```
+
```
thomasyu888, line 132: Is there some sort of validation that we can call here using the API before the changes are applied in the grid

andrewelamb:
Done as part of https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1383

PR #1347, opened 2026-04-02
[SYNPY-1802] Add DownloadList functions

andrewelamb:
# **Problem:**
Download list functionality was part of the main client class. 

# **Solution:**
Create `DownloadList` functions in an operations file
Deprecate the old methods

# **Testing:**
Unit and Integration tests for the new functions


tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+from synapseclient.core.exceptions import SynapseError
+from synapseclient.models.download_list import (
+    DownloadList,
+    DownloadListItem,
+    DownloadListManifestRequest,
+)
+
+
+@pytest.fixture(scope="module")
+def syn():
```
linglp: I think here you should just use the `syn` in `conftest.py`. I commented out this part and verified that all the tests can pass without problems. There's no need to use a module level `syn` here.

andrewelamb: I removed this and the tests are now failing  [see here](https://github.com/Sage-Bionetworks/synapsePythonClient/actions/runs/24095892421/job/70294352345?pr=1347).

linglp: I think the problem is related to your test, not the `syn` fixture. See comment below

tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+    """
+    Create a Synapse for this modules tests.
+    When using the syn fixture from .conftest, the patches here were causing issues in other tests.
+    """
+    syn = Synapse(debug=False, skip_checks=True, cache_client=False)
+    syn.logger = logging.getLogger()
```
linglp: the `conftest.py` uses: 
```
syn.logger = logging.getLogger(SILENT_LOGGER_NAME)
```
which by default is the `synapseclient` logger, and here you use the root logger. That's the only difference. I think we should  aovid sending logs directly to the root.

tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+
+    async def test_remove_from_download_list_async_returns_count(self):
+        """remove_from_download_list_async extracts numberOfFilesRemoved from the response."""
+        files = [DownloadListItem(file_entity_id="syn123", version_number=1)]
+        with patch(
+            "synapseclient.Synapse.get_client",
+            return_value=self.syn,
+        ):
+            self.syn.rest_post_async = AsyncMock(
+                return_value={"numberOfFilesRemoved": 1}
```
linglp: If you have `self.syn` refers to the `syn` in `conftest.py`, then this line here pollutes other tests because it overwrites the session-scoped `syn`. 

we should change it to something like: 
```
patch.object(
            self.syn,
            "rest_post_async",
            new_callable=AsyncMock,
            return_value={"numberOfFilesRemoved": 1},
        )
```

tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+        self.syn = syn
+
+    async def test_add_to_download_list_async_returns_count(self):
+        """add_to_download_list_async extracts numberOfFilesAdded from the response."""
+        files = [DownloadListItem(file_entity_id="syn123", version_number=1)]
+        with patch(
+            "synapseclient.Synapse.get_client",
+            return_value=self.syn,
+        ):
+            self.syn.rest_post_async = AsyncMock(return_value={"numberOfFilesAdded": 1})
```
linglp: similar to comment below, this overwrites the session-scoped `syn` with `self.syn.rest_post_async = AsyncMock(return_value={"numberOfFilesAdded": 1})`

tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+            assert os.path.exists(result_path)
+            os.remove(result_path)
+
+    async def test_download_row_success(self):
+        """_download_row sets path on the row and returns a DownloadListItem."""
+        fake_entity = MagicMock()
+        fake_entity.path = "/tmp/file.txt"
+
+        row = {"ID": "syn111", "versionNumber": "5", "name": "file.txt"}
+        self.syn.get_async = AsyncMock(return_value=fake_entity)
```
linglp: similarly, this should be: 
```
patch.object(
            self.syn,
            "get_async",
            new_callable=AsyncMock,
            return_value=fake_entity,
        )
```

tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+        assert item is not None
+        assert item.file_entity_id == "syn111"
+        assert item.version_number == 5
+        assert row["path"] == "/tmp/file.txt"
+        assert row["error"] == ""
+
+    async def test_download_row_failure(self):
+        """_download_row sets error on the row and returns None on failure."""
+        row = {"ID": "syn222", "versionNumber": "1", "name": "secret.txt"}
+        self.syn.get_async = AsyncMock(side_effect=SynapseError("Forbidden"))
```
linglp: similarly, this should be fixed with: 
```
patch.object(
            self.syn,
            "get_async",
            new_callable=AsyncMock,
            side_effect=SynapseError("Forbidden"),
        )
```

tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+            item = await DownloadList._download_row(row, synapse_client=self.syn)
+
+        assert item is None
+        assert row["path"] == ""
+        assert "Forbidden" in row["error"]
+
+    async def test_download_row_failure_unexpected_exception(self):
+        """_download_row catches non-SynapseError exceptions so one failure does not abort the run."""
+        row = {"ID": "syn222", "versionNumber": "1", "name": "secret.txt"}
+        self.syn.get_async = AsyncMock(side_effect=AttributeError("path"))
```
linglp: similarly, this should be fixed like above

tests/unit/synapseclient/models/async/unit_test_download_list_async.py
```diff
@@ -0,0 +1,798 @@
...
+
+        assert item is None
+        assert row["path"] == ""
+        assert "versionNumber" in row["error"]
+
+    async def test_download_row_no_download_location(self):
+        """_download_row does not pass downloadLocation when it is None."""
+        fake_entity = MagicMock()
+        fake_entity.path = "/tmp/file.txt"
+        self.syn.get_async = AsyncMock(return_value=fake_entity)
```
linglp: similarly, this should be fixed like above.

linglp (changes requested):
Hi @andrewelamb ! I focused on reading the tests for this round of review. I will take a look at the core functions later. Thanks for your hard work!

synapseclient/api/download_list_services.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+async def add_to_download_list_async(
+    files: list["DownloadListItem"],
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> int:
+    """Add a batch of specific file versions to the user's Synapse download list.
+
+    [Endpoint](https://rest-docs.synapse.org/rest/POST/download/list/add.html)
```
linglp: Nit: I think the convention is to format the link like `get_entity_id_by_repository_name` (see [here](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/api/docker_services.py#L20))

synapseclient/api/download_list_services.py
```diff
@@ -0,0 +1,93 @@
...
+    client = Synapse.get_client(synapse_client=synapse_client)
+    batch = [
+        {"fileEntityId": item.file_entity_id, "versionNumber": item.version_number}
+        for item in files
+    ]
+    request_body = {"batchToAdd": batch}
+    result = await client.rest_post_async(
+        "/download/list/add", body=json.dumps(request_body)
+    )
+    return result.get("numberOfFilesAdded", 0)
```
linglp: If the API response doesn't contain `"numberOfFilesAdded"` at all, `.get("numberOfFilesAdded", 0)` silently returns 0. Should we raise an error here instead?

synapseclient/api/download_list_services.py
```diff
@@ -0,0 +1,93 @@
...
+    client = Synapse.get_client(synapse_client=synapse_client)
+    batch = [
+        {"fileEntityId": item.file_entity_id, "versionNumber": item.version_number}
+        for item in files
+    ]
+    request_body = {"batchToRemove": batch}
+    result = await client.rest_post_async(
+        "/download/list/remove", body=json.dumps(request_body)
+    )
+    return result.get("numberOfFilesRemoved", 0)
```
linglp: same issue as above. Should we raise an error here because the response is unexpected instead of letting it silently return 0

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+from synapseclient.models.protocols.download_list_protocol import (
+    DownloadListSynchronousProtocol,
+)
+
+_PATH_COLUMN = "path"
+_ERROR_COLUMN = "error"
+
+
+@dataclass
+class DownloadListItem:
```
linglp, line 34: Here is missing docstring to the synapse documentation: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/download/DownloadListItem.html

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+    """Identifies a specific file version in the Synapse download list.
+
+    Attributes:
```
linglp, line 39: Can we use the definition in the synapse documentation: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/download/DownloadListItem.html

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+    manifest_path: Optional[str] = field(init=False, default=None)
+    """Absolute local path of the downloaded manifest CSV. Populated by
+    ``_post_exchange_async()`` after the file is downloaded. ``None`` until
+    ``send_job_and_wait_async()`` returns."""
+
+    def to_synapse_request(self) -> dict[str, Any]:
+        """Build the request body for the manifest async job."""
+        return {
+            "concreteType": self.concrete_type,
+            "csvTableDescriptor": {
```
linglp: Here we should use the existing class  `CsvTableDescriptor()` instead of duplicating the code?

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+        Exchanges the file handle ID for a pre-signed S3 URL via
+        ``get_file_handle_for_download_async()``, then streams the CSV to disk
+        via ``download_from_url()`` (run in a thread pool via
+        ``asyncio.to_thread`` since it is a blocking sync method). Stores the
+        local path in ``self.manifest_path``.
+
+    After ``send_job_and_wait_async()`` returns, ``manifest_path`` holds the
+    local path to the downloaded CSV and is ready to use.
+
+    [Request](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/download/DownloadListManifestRequest.html)
```
linglp: Similar to other comments, the format of this docstring is incorrect.  We use something like: 
```                                                                          
  <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/MaterializedView.html>        
```
in the code base

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+
+    file_entity_id: str
+    """Synapse ID of the file entity (e.g. ``"syn123"``)."""
+
+    version_number: Optional[int] = None
+    """Version of the file to target."""
+
+
+@dataclass
+class DownloadListManifestRequest(AsynchronousCommunicator):
```
linglp, line 52: This class is called `DownloadListManifestRequest`, but based on the docstring, it seems like this class is also the response class? I think the naming here is confusing.

synapseclient/models/download_list.py
linglp: Also, based on the documentation, `csvTableDescriptor` should be an actual field, not hidden in to_synapse_request()

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+            writer = csv.DictWriter(f, fieldnames=columns, extrasaction="ignore")
+            writer.writeheader()
+            writer.writerows(rows)
+
+    @staticmethod
+    async def _download_all_rows(
+        rows: list[dict[str, Any]],
+        download_location: Optional[str],
+        parallel: bool,
+        max_concurrent: int = 10,
```
linglp, line 375: Nit: From a design perspective, users can technically enter any numbers here for `max_concurrent` right? Should an error be raised if max_concurrent is zero or negative?

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+                async with sem:
+                    return await DownloadList._download_row(
+                        row,
+                        download_location=download_location,
+                        synapse_client=synapse_client,
+                    )
+
+            items = await asyncio.gather(*[bounded_download(row) for row in rows])
+            return [item for item in items if item is not None]
+        else:
```
linglp, line 425: Nit: `max_concurrent` is only effective if parallel is set to True. What if users set parallel to False and also with max_concurrent greater than 0? Should a warning be raised?

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,552 @@
...
+        finally:
+            os.remove(manifest_path)
+
+        if columns is None:
+            raise SynapseError(
+                "Manifest job succeeded but the downloaded CSV has no headers. "
+                "This is unexpected — the Synapse server may have returned an empty file."
+            )
+
+        if _PATH_COLUMN in columns or _ERROR_COLUMN in columns:
```
linglp: this is an interesting defensive check. The manifest content is entirely determined by the Synapse server by `DownloadList.get_manifest_async`. Why this manifest would have naming collision?

linglp (changes requested):
One main thing that I noticed is that some parts of the code would fail silently. For example: 
```
num_files = DownloadList().add_files(
    files=[
        DownloadListItem(
            file_entity_id="syn24181472",
            version_number=1
        ),
        DownloadListItem(
            file_entity_id="syn24246013",
            version_number=100
        )
    ]
)
```
Unless users check `num_files`, there's no way of knowing if all the files have successfully been added to the download list or only some of them have been added. 

Similarly, when you remove a file from download list, it is hard to know if all the files have successfully been removed or only some of them have been removed. 


In addition, how can we download only a subset of items in the cart? My download cart has over a thousand files but I only want a subset of them to be downloaded. Currently, the only way to do this is to clear the cart, and then re-add specific files. But after clearing all the items in download cart, there's no way to restore them. It seems like there's an endpoint [here](https://rest-docs.synapse.org/rest/POST/download/list/query/async/start.html) that could let you query the download cart, but I didn't see it get implemented here. Please let me know if this is not part of the ticket.

andrewelamb:
> One main thing that I noticed is that some parts of the code would fail silently. For example:
> 
> ```
> num_files = DownloadList().add_files(
>     files=[
>         DownloadListItem(
>             file_entity_id="syn24181472",
>             version_number=1
>         ),
>         DownloadListItem(
>             file_entity_id="syn24246013",
>             version_number=100
>         )
>     ]
> )
> ```
> 
> Unless users check `num_files`, there's no way of knowing if all the files have successfully been added to the download list or only some of them have been added.
> 
> Similarly, when you remove a file from download list, it is hard to know if all the files have successfully been removed or only some of them have been removed.
> 
> In addition, how can we download only a subset of items in the cart? My download cart has over a thousand files but I only want a subset of them to be downloaded. Currently, the only way to do this is to clear the cart, and then re-add specific files. But after clearing all the items in download cart, there's no way to restore them. It seems like there's an endpoint [here](https://rest-docs.synapse.org/rest/POST/download/list/query/async/start.html) that could let you query the download cart, but I didn't see it get implemented here. Please let me know if this is not part of the ticket.

You are correct! This ticket was only for replicating the functionality in `synapseclient/client.py` that has been deprecated. I'll create a ticket to expand on this one.

synapseclient/api/download_list_services.py
andrewelamb: Yeah, I like that better.

synapseclient/models/download_list.py
andrewelamb: It's the name of the Endpoint: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/download/DownloadListManifestRequest.html, I'm happy to change it to something else!

synapseclient/models/download_list.py
andrewelamb: Right, this is more of a guard against those being added in the future.

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+    """Represents a user's Synapse Download List (cart).
+
+    The download list is a user-scoped cart of files queued for bulk download.
+    Files can be added via the Synapse web UI or API and downloaded in batch.
+
+    Note: Files are not packaged into a zip because download lists can exceed
+    100 GB. Instead, files are downloaded individually and removed from the list
+    after successful download, so interrupted runs are safely resumable.
+
+    Example: Download all files in the cart::
```
linglp: The docstring format is wrong here. the convention everywhere else (see evaluation.py) is markdown fenced code blocks.

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+
+        from synapseclient.models import DownloadList
+        manifest_path = await DownloadList().download_files_async(download_location="./data")
+
+    Example: Sync usage::
```
linglp: The pattern is to just show the sync method directly in examples. See `class Evaluation(EvaluationSynchronousProtocol)` You should just use the sync method without treating it as a special case.

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+    Files can be added via the Synapse web UI or API and downloaded in batch.
+
+    Note: Files are not packaged into a zip because download lists can exceed
+    100 GB. Instead, files are downloaded individually and removed from the list
+    after successful download, so interrupted runs are safely resumable.
+
+    Example: Download all files in the cart::
+
+        from synapseclient.models import DownloadList
+        manifest_path = await DownloadList().download_files_async(download_location="./data")
```
linglp: Here the example is wrong right? You call await without async context.

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+        """
+        Attempt to download a single file from the manifest.
+
+        Mutates ``row`` in place, setting ``"path"`` and ``"error"`` columns.
+        Each call must receive exclusive ownership of its ``row`` dict — do not
+        pass the same row to multiple concurrent calls.
+        Files in the cart that the user cannot access are caught here so that a
+        single failure does not abort the entire run.
+
+        Arguments:
```
linglp, line 344: Nit: missing docstring of `download_location`

synapseclient/models/protocols/download_list_protocol.py
```diff
@@ -0,0 +1,208 @@
...
+            manifest_path = DownloadList().download_files(
+                download_location="./downloads",
+                parallel=True,
+                max_concurrent=5,
+            )
+            ```
+        """
+        return ""
+
+    def get_manifest(
```
linglp, line 86: `get_manifest` loses parameters in comparison with the old `_generate_manifest_from_download_list`: 
```
    def _generate_manifest_from_download_list(
        self,
        quoteCharacter: str = '"',
        escapeCharacter: str = "\\",
        lineEnd: str = os.linesep,
        separator: str = ",",
        header: bool = True,
    ):
```
Is this intentional?

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+        """
+        Write the annotated result rows to the output manifest CSV.
+        Intended to be called via ``asyncio.to_thread`` to avoid blocking the
+        event loop on synchronous file I/O.
+
+        Arguments:
+            path: Destination path for the output manifest CSV.
+            columns: Field names for the CSV header, including ``"path"`` and
+                ``"error"``.
+            rows: List of row dicts, each mutated by :meth:`_download_row` to
```
linglp: what is this `:meth:` in the docstring?

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+    @staticmethod
+    async def _save_result_manifest(
+        rows: list[dict[str, Any]],
+        columns: list[str],
+        download_location: Optional[str],
+    ) -> str:
+        """Write the annotated rows to a new result manifest CSV and return its path.
+
+        Arguments:
+            rows: List of row dicts, each mutated by :meth:`_download_row` to
```
linglp: what is this :meth:

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+        parallel: bool = False,
+        max_concurrent: Optional[int] = None,
+        *,
+        synapse_client: Optional["Synapse"] = None,
+    ) -> list[DownloadListItem]:
+        """Download all rows from the manifest, either sequentially or concurrently.
+
+        Arguments:
+            rows: List of row dicts from the manifest. Each row is mutated in
+                place by :meth:`_download_row` to include ``"path"`` and
```
linglp: fix docstring here

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+                ``"error"`` values.
+            download_location: Directory to download files to.
+            parallel: If ``True``, rows are downloaded concurrently (bounded by
+                ``max_concurrent``) via ``asyncio.gather``. If ``False``, rows are
+                downloaded one at a time.
+            max_concurrent: Maximum number of concurrent downloads when
+                ``parallel=True``. Defaults to ``None``, which is treated as
+                10. Must be at least 1. Has no effect when ``parallel=False``;
+                a ``UserWarning`` is emitted if set explicitly in that case.
```
linglp: The double backtick is RST syntax for inline code. But this codebase uses markdown in docstrings.

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,574 @@
...
+        Raises:
+            ValueError: If ``max_concurrent`` is less than 1.
+
+        Warns:
+            UserWarning: If ``max_concurrent`` is set but ``parallel=False``.
+
+        Returns:
+            List of :class:`DownloadListItem` for each successfully downloaded file.
+        """
+        if max_concurrent is not None and not parallel:
```
linglp: hmm after thinking about this more, I think the warning should be removed. The default case for user is: 
```
manifest_file = DownloadList().download_files(download_location="./downloads")
```
Since max_concurrent is default to 10, this would always trigger the warning. I think users might get confused and wonder what they did wrong. It will be better to just remove

linglp (changes requested):
I flag some issues with the docstring. Could you fix them? Also, I noticed a problem with `get_manifest` losing parameters in comparison with the old function. I am not sure if that's intentional

synapseclient/models/protocols/download_list_protocol.py
andrewelamb: No you are correct, this got left out. `get_manifest` should have had the option to take a `CsvTableDescriptor` and pass it on to the `DownloadListManifestRequest`.

synapseclient/models/download_list.py
andrewelamb: Yes, it's Sphynx markup. Claude "helpfully" added these without me noticing. I'll remove them.

andrewelamb:
Reverting back to draft, I found some issues that need to be adressed.

CLAUDE.md
```diff
@@ -36,6 +36,19 @@ pip install -e ".[docs]" && mkdocs serve
 
 ## Conventions
 
+### Docstring style
+Do not use backticks (single or double) in docstrings. Write parameter names, method names, and values as plain text.
```
linglp, line 40: Is this true? We do use backticks in some cases. For example, when we provided an example in the docstring.

tests/integration/synapseclient/models/async/test_download_list_async.py
```diff
@@ -0,0 +1,576 @@
...
+import csv
+import os
+import tempfile
+import uuid
+from typing import Callable
+
+import pytest
+
+import synapseclient.core.utils as utils
+from synapseclient import Project as Synapse_Project
```
linglp, line 12: Nit: I am curious why you did: `from synapseclient import Project as Synapse_Project` instead of just `from synapseclient import Project as Project`?

tests/integration/synapseclient/models/async/test_download_list_async.py
```diff
@@ -0,0 +1,576 @@
...
+
+    async def test_add_files_multiple_files_and_versions(
+        self,
+        project: Synapse_Project,
+        syn: Synapse,
+        schedule_for_cleanup: Callable[..., None],
+    ) -> None:
+        """add_files_async() adds multiple files with multiple versions in a single call."""
+        # GIVEN an empty cart and two files, each with two versions
+        await DownloadList.clear_async(synapse_client=syn)
```
linglp, line 83: I am a bit worried about all these blanket-level of `await DownloadList.clear_async` in the beginning of all the integration tests. For example, after this PR gets merged, and let's say there are people working on two separate PRs and running integration tests. If one user has one test runs `await DownloadList.clear_async` in GH actions, then does that mean another user will potentially run into problems with their tests?

linglp: Maybe a better approach is just to remove all files that being added to download list in a given test using `await DownloadList.remove_files_async`

tests/integration/synapseclient/models/async/test_download_list_async.py
```diff
@@ -0,0 +1,576 @@
...
+
+        # AND only the added versions are in the manifest
+        manifest_path = await DownloadList.get_manifest_async(synapse_client=syn)
+        schedule_for_cleanup(manifest_path)
+        with open(manifest_path, newline="") as f:
+            reader = csv.DictReader(f)
+            cart_entries = {(row["ID"], int(row["versionNumber"])) for row in reader}
+        await DownloadList.clear_async(synapse_client=syn)
+
+        assert (
```
linglp, line 121: Nit: I think this might be overly complicated? Will the following be sufficient? 
```
assert count == 2
assert (file_a.id, file_a_v1) in cart_entries
assert (file_b.id, file_b_v2) in cart_entries
```

tests/integration/synapseclient/models/async/test_download_list_async.py
```diff
@@ -0,0 +1,576 @@
...
+        lines = [line for line in lines if line]
+
+        # AND there is no header row so the file ID appears in the raw content
+        assert len(lines) >= 1, "Expected at least one row in the manifest"
+        assert any(
+            file.id in line for line in lines
+        ), f"Expected {file.id} in manifest content"
+
+        # AND the single-quote character is used for quoting
+        assert (
```
linglp, line 574: I found this pretty strange. The goal is to assert the single-quote character is used for quoting, but here you are asserting that we are not using double quotes? Why not just assert that single quote was used?

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,653 @@
...
+                    row,
+                    download_location=download_location,
+                    synapse_client=synapse_client,
+                )
+                if item is not None:
+                    downloaded.append(item)
+            return downloaded
+
+    @staticmethod
+    async def _download_manifest_file(
```
linglp, line 438: will it be worth adding a unit test for _download_manifest_file

synapseclient/models/download_list.py
```diff
@@ -0,0 +1,653 @@
...
+        """
+        from synapseclient import Synapse
+        from synapseclient.models.file import File
+
+        client = Synapse.get_client(synapse_client=synapse_client)
+        entity_id = row[_ID_COLUMN]
+        version_str = row.get(_VERSION_COLUMN)
+        version_number = int(version_str) if version_str else None
+
+        if version_number is None:
```
linglp, line 469: Nit: I am a bit confused when reading this, can we add a comment here? This is to prevent cases where a malformed manifest gets generated by the server? Should this fail loudly since this would mean the return from server is wrong?

linglp (commented):
I am mostly worried about all these `await DownloadList.clear_async` in every test in `TestDownloadList` at the start to ensure a clean cart. This works within a single CI run. However, the download list is a global, per-user resource on Synapse. What If PR A's test calls clear_async while PR B's test has already added files and before assertion, PR B's cart is wiped?

tests/integration/synapseclient/models/async/test_download_list_async.py
andrewelamb: I fixed this by not using the `clear` method any longer and just removing the specific files added by each test.

synapseclient/models/download_list.py
andrewelamb: After thinking about this, you are right. When the version number is missing, it should be treated as the user requesting the latest version of the file.

tests/integration/synapseclient/models/async/test_download_list_async.py
andrewelamb: Claude added it to avoid confusion between the `project` fixture and the `Project` class. I agree it's not needed.

tests/integration/synapseclient/models/async/test_download_list_async.py
andrewelamb: Agreed, and they have been removed. `clear-async` isn't used in integration tests now.

tests/integration/synapseclient/operations/async/test_download_list_operations_async.py
```diff
@@ -0,0 +1,604 @@
...
+    schedule_for_cleanup(new_path)
+    file.path = new_path
+    await file.store_async(synapse_client=syn)
+    return file.version_number
+
+
+async def _add_to_cart(
+    file: File,
+    syn: Synapse,
+    scheduled_for_cart_removal: list,
```
linglp: is this correct? shouldn't this be: `list[DownloadListItem]`

tests/integration/synapseclient/operations/async/test_download_list_operations_async.py
```diff
@@ -0,0 +1,604 @@
...
+    download_list_remove_async,
+)
+
+
+@pytest_asyncio.fixture
+async def scheduled_for_cart_removal(syn: Synapse):
+    """Track items a test adds to the cart and remove only those items on teardown."""
+    scheduled: list[DownloadListItem] = []
+    yield scheduled
+    if scheduled:
```
linglp, line 39: Just in case something goes wrong, can we add a try-except block here: 
```
        try:
            await download_list_remove_async(files=scheduled, synapse_client=syn)
        except Exception as e:
            pytest.fail(
                f"Cart teardown failed — {len(scheduled)} item(s) may remain in "
                f"the cart and affect subsequent tests: {e}"
            )
```

CLAUDE.md
```diff
@@ -36,6 +36,16 @@ pip install -e ".[docs]" && mkdocs serve
 
 ## Conventions
 
+### Docstring examples format
+Use mkdocstrings admonition syntax. Start with a description line, then wrap code in a fenced python block. Examples should use sync methods (not async) and include login boilerplate. See evaluation.py for reference.
+
```
linglp, line 41: I don't think this is accurate. If you look at `models/activity.py`, we definitely have async examples there.

tests/CLAUDE.md
```diff
@@ -19,6 +19,7 @@ Use `pytest.mark.parametrize` when possible to merge similar tests into one test
 - Use `AsyncMock` for async method mocking, `create_autospec` for type-safe mocks
 - Class-based test organization with `@pytest.fixture(scope="function", autouse=True)` for setup
 - Test file naming: `unit_test_*.py` (legacy) or `test_*.py` (newer) — both patterns are discovered by pytest
+- Mock isolation: when mocking fixture/instance attributes (e.g., `self.syn.rest_post_async`), always wrap in `patch.object()` context managers instead of direct assignment. This prevents the mock from leaking to other tests: `with patch.object(self.syn, "method_name", new_callable=AsyncMock, return_value=...):`. Direct assignment leaves the mock in place after the test, polluting subsequent tests in the class.
```
linglp, line 22: this is accurate but only because that syn in unit tests is scope="session"

linglp (approved):
I still have some minor comments. Please see them below. I am going to approve now. Thank you for all your work!

thomasyu888 (approved):
🔥 LGTM! Thanks for the work and thorough review here, just approving as well.

PR #1353, opened 2026-04-06
[SYNPY-1800] Added new sync_to_synapse method

andrewelamb:
# **Problem:**
The `syncToSynapse` function doesn't fit the new OOP model paradigm.

# **Solution:**

- A new `sync_to_synapse` method was added to the `StorableContainer` class
- `syncToSynapse` was deprecated

# **Testing:**

- Unit tests for the new helper functions
- Integration tests for the new `sync_to_synapse` method


docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py
```diff
@@ -5,59 +5,63 @@
...
-    syn=syn,
-    directory_path=DIRECTORY_FOR_MY_PROJECT,
-    parent_id=my_project_id,
-    manifest_path=PATH_TO_MANIFEST_FILE,
-)
+# Step 2: Create a manifest CSV file to upload data in bulk
+# Walk the local directory tree and build a manifest with the required "path" and
+# "parentId" columns.  Folders that do not yet exist in Synapse are created
+# automatically by sync_to_synapse, so we set parentId to the project for every file.
+# NOTE: In a future release, Project.sync_from_synapse will support writing a manifest
```
linglp: This "note" sounds like a "TODO" to me. Can we link it to a ticket?

synapseutils/sync.py
```diff
@@ -719,6 +733,12 @@ def _get_file_entity_provenance_dict(syn, entity):
             raise  # unexpected error so we re-raise the exception
 
 
+@deprecated(
+    version="4.13.0",
+    reason=(
+        "To be removed in 5.0.0. " "Moved to synapseclient.models.services.manifest.py"
```
linglp: I can't find the equivalent of `generate_manifest` in `synapseclient.models.services.manifest.py`. Does this really have a replacement?

synapseclient/models/mixins/storable_container.py
```diff
@@ -496,6 +504,127 @@ async def _sync_from_synapse_async(
...
         return self
 
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"{self.__class__.__name__}_sync_to_synapse: {self.id}"
+    )
+    async def sync_to_synapse_async(
+        self: Self,
+        manifest_path: str,
+        dry_run: bool = False,
+        send_messages: bool = False,
```
linglp: In the old `syncToSynapse`, the default of `send_messages` is `True`. I am curious why this gets set to `False`?

tests/integration/synapseclient/models/async/test_storable_container_async.py
```diff
@@ -0,0 +1,385 @@
...
+    with open(path, "w", newline="", encoding="utf-8") as f:
+        writer = csv.DictWriter(
+            f, fieldnames=fieldnames, restval="", extrasaction="ignore"
+        )
+        writer.writeheader()
+        writer.writerows(rows)
+    return path
+
+
+def _create_local_test_file(content: str = "test content") -> Path:
```
linglp: can file clean up be baked into `_create_local_test_file` and not leave it to the caller?

linglp: Can the temporary files be created like `_write_manifest`?

linglp (changes requested):
Old `syncToSynapse` defaults to `sendMessages=True`, and new `sync_to_synapse` defaults to send_messages=False. Users migrating by following the deprecation example will silently stop receiving upload notifications with no warning. Is this behavior intentional? 

Also, `_create_local_test_file` leaves cleanup to the caller. I think it will be better if clean up can happen within `_create_local_test_file` so that we could avoid all the scattered `schedule_for_cleanup` calls.

synapseclient/models/mixins/storable_container.py
andrewelamb: No good reason, I can change it

synapseutils/sync.py
andrewelamb: You are correct, I think this got mistakenly added.

tests/integration/synapseclient/models/async/test_storable_container_async.py
```diff
@@ -0,0 +1,385 @@
...
+                    "name": unique_name,
+                    "my_string": "hello",
+                    "my_number": "42",
+                },
+            ],
+            tmp_path,
+        )
+
+        # WHEN I sync to Synapse
+        await project_model.sync_to_synapse_async(
```
linglp: For all the `sync_to_synapse_async` in the tests, `send_messages` should be set to `False` since now you turn `True` to as default.

synapseclient/models/protocols/storable_container_protocol.py
```diff
@@ -224,3 +230,71 @@ def sync_from_synapse(
 
         """
         return self
+
+    def sync_to_synapse(
+        self: Self,
+        manifest_path: str,
+        dry_run: bool = False,
+        send_messages: bool = False,
```
linglp: here you would also need to update this as `True`

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,121 @@
...
+
+!!! note "Replaces legacy TSV manifest"
+    This CSV manifest replaces the legacy TSV manifest (`SYNAPSE_METADATA_MANIFEST.tsv`) produced by [synapseutils.syncFromSynapse][]. The `syncFromSynapse` and `syncToSynapse` utility functions are deprecated and will be removed in v5.0.0. Use `Project.sync_from_synapse` / `Folder.sync_from_synapse` and `Project.sync_to_synapse` / `Folder.sync_to_synapse` instead. See the [legacy TSV manifest documentation](manifest_tsv.md) for details on the old format.
+
+## Manifest file format
+
+The format of the manifest file is a comma-separated values (CSV) file with one row per file and columns describing the file. The minimum required columns for uploading are **path** and **parentId**, where `path` is the local file path and `parentId` is the Synapse ID of the project or folder where the file is uploaded to.
```
linglp: The legacy format uses "parent" instead of "parentId". If we switch to use "parentid", I think this will be a breaking change that should be called out. See `DEFAULT_GENERATED_MANIFEST_KEYS` [here](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseutils/sync.py#L64)

linglp (changes requested):
Thanks for the fixes! Just some small comments. Also, could you make sure that the documentation format is correct? 

For example, the format [here](https://synapsepythonclient--1353.org.readthedocs.build/en/1353/reference/experimental/mixins/storable_container/?h=sync_to_synapse_async#synapseclient.models.mixins.StorableContainer.sync_to_synapse_async) is slightly off. Also, the links in reference are not working (See [here](https://synapsepythonclient--1353.org.readthedocs.build/en/1353/explanations/manifest_csv/?h=manifest#references))

docs/explanations/manifest_csv.md
andrewelamb: I'd quibble that it's a breaking change since sync_to_synapse is a new method, and sync_from_synapse will use this format as well. That said I'll add a note here.

docs/tutorials/python/upload_data_in_bulk.md
```diff
@@ -19,93 +20,100 @@ With a project that has this example layout:
...
-/home/user_name/my_ad_project/single_cell_RNAseq_batch_2/SRR12345678_R2.fastq.gz  syn60109537
-/home/user_name/my_ad_project/single_cell_RNAseq_batch_2/SRR12345678_R1.fastq.gz  syn60109537
-/home/user_name/my_ad_project/biospecimen_experiment_2/fileD.txt  syn60109543
-/home/user_name/my_ad_project/biospecimen_experiment_2/fileC.txt  syn60109543
-/home/user_name/my_ad_project/single_cell_RNAseq_batch_1/SRR12345678_R2.fastq.gz  syn60109534
-/home/user_name/my_ad_project/single_cell_RNAseq_batch_1/SRR12345678_R1.fastq.gz  syn60109534
-/home/user_name/my_ad_project/biospecimen_experiment_1/fileA.txt  syn60109540
-/home/user_name/my_ad_project/biospecimen_experiment_1/fileB.txt  syn60109540
+path,parentId
+/home/user_name/my_ad_project/single_cell_RNAseq_batch_2/SRR12345678_R2.fastq.gz,syn60109500
```
linglp: Take a look at the old documentation:  the example manifest output changed from:
  * Before — different folders have different IDs
  * After — all files point to the same parent  

Is this intentional?

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,124 @@
...
+
+Values that contain commas are automatically quoted (e.g., `"hello, world"`). This is handled by the standard CSV format using `QUOTE_MINIMAL` quoting.
```
linglp: why do we have to mention about  QUOTE_MINIMAL quoting here? I think this is confusing for the users

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,124 @@
...
+
+### Example manifest file
+
+| path            | parentId | ID      | name      | annot1 | annot2 | collection_date           | used                     | executed                     |
+|-----------------|----------|---------|-----------|--------|--------|---------------------------|--------------------------|------------------------------|
+| /path/file1.txt | syn1243  | syn5001 | file1.txt | bar    | 3.1415 | 2023-12-04T07:00:00Z      | syn124;/path/file2.txt   | https://github.org/foo/bar   |
+| /path/file2.txt | syn12433 | syn5002 | file2.txt | baz    | 2.71   | 2001-01-01T08:00:00Z      |                          | https://github.org/foo/baz   |
+| /path/file3.txt | syn12455 | syn5003 | file3.txt | zzz    | 3.52   | 2023-12-04T07:00:00Z      |                          | https://github.org/foo/zzz   |
+
+## References
```
linglp, line 112: The links are still not rendering correctly. See [here](https://synapsepythonclient--1353.org.readthedocs.build/en/1353/explanations/manifest_csv/#references)

docs/tutorials/python/upload_data_in_bulk.md
```diff
@@ -19,93 +20,100 @@ With a project that has this example layout:
...
+/home/user_name/my_ad_project/single_cell_RNAseq_batch_1/SRR12345678_R1.fastq.gz,syn60109500
+/home/user_name/my_ad_project/biospecimen_experiment_1/fileA.txt,syn60109500
+/home/user_name/my_ad_project/biospecimen_experiment_1/fileB.txt,syn60109500
 ```
 </details>
 
 ## 3. Upload the data in bulk
 ```python
-{!docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py!lines=32-36}
+{!docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py!lines=46-48}
```
linglp: This doesn't seem to be the right code block: https://synapsepythonclient--1353.org.readthedocs.build/en/1353/tutorials/python/upload_data_in_bulk/#3-upload-the-data-in-bulk

docs/tutorials/python/upload_data_in_bulk.md
```diff
@@ -19,93 +20,100 @@ With a project that has this example layout:
...
 ## 4. Add an annotation to our manifest file
-At this point in the tutorial we will start to use pandas to manipulate a TSV file. If
+At this point in the tutorial we will use pandas to manipulate the CSV manifest. If
 you are not comfortable with pandas you may use any tool that can open and manipulate
-TSV such as excel or google sheets.
+CSV files such as Excel or Google Sheets.
 
 ```python
-{!docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py!lines=37-57}
+{!docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py!lines=50-63}
```
linglp: is this the right code block? I am seeing a mix of step 4 and step 5 here: 
https://synapsepythonclient--1353.org.readthedocs.build/en/1353/tutorials/python/upload_data_in_bulk/#3-upload-the-data-in-bulk

Image

docs/tutorials/python/upload_data_in_bulk.md
```diff
@@ -120,14 +128,14 @@ Let's create an [Activity/Provenance](../../explanations/domain_models_of_synaps
...
-In this code we are finding a row in our TSV file and pointing to the file path of
+In this code we are finding a row in our CSV file and pointing to the file path of
 another file within our manifest. By doing this we are creating a relationship between
 the two files. This is a simple example of how you can create a provenance record in
 Synapse. Additionally we'll link off to a sample URL that describes a process that we
 may have executed to generate the file.
 
 ```python
-{!docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py!lines=58-90}
+{!docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py!lines=68-92}
```
linglp: This also doesn't seem right: 
https://synapsepythonclient--1353.org.readthedocs.build/en/1353/tutorials/python/upload_data_in_bulk/#5-create-an-activityprovenance

See here: 

Image

docs/tutorials/python/upload_data_in_bulk.md
```diff
@@ -154,6 +162,6 @@ navigated to the Files tab and selected the file that we added a Provenance reco
 
 - [syn.login][synapseclient.Synapse.login]
 - [syn.findEntityId][synapseclient.Synapse.findEntityId]
-- [synapseutils.generate_sync_manifest][]
-- [synapseutils.syncToSynapse][]
+- [Project.sync_to_synapse][synapseclient.models.mixins.StorableContainer.sync_to_synapse]
```
linglp, line 167: also broken link: https://synapsepythonclient--1353.org.readthedocs.build/en/1353/tutorials/python/upload_data_in_bulk/#references-used-in-this-tutorial

linglp (changes requested):
Hi @andrewelamb ! I found more problems with the documentation. Please see comments below

docs/explanations/manifest_csv.md
andrewelamb: Yep I needed to add them to the reference/experimental docs.

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Manifest sources
+
+The CSV manifest format is shared across multiple tools:
+
+| Source                                                   | Filename                        |
+|----------------------------------------------------------|---------------------------------|
+| `Project.sync_from_synapse` / `Folder.sync_from_synapse` | manifest.csv                    |
+| Synapse UI download cart                                 | manifest.csv                    |
+| CLI `synapse get-download-list`                          | manifest\_\<timestamp\>.csv     |
```
linglp: Nit: this doesn't seem to be rendered correctly: https://synapsepythonclient--1353.org.readthedocs.build/en/1353/explanations/manifest_csv/#manifest-sources

Image

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,118 @@
...
+
+A manifest generated by any of these sources can be used as input to `sync_to_synapse`, provided the `path` column is present with valid local file paths. Manifests from the Synapse UI and CLI do not include a `path` column by default, so users must add it before uploading.
```
linglp: Is this statement correct? I tried `synapse get-download-list`, and the manifest returned has a `path` column

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,118 @@
...
+| /path/file1.txt | syn1243 | bar | 3.1415 | "aaaa, bbbb" | "[14,27,30]" | "Annotation, with a comma" |
+| /path/file2.txt | syn12433 | baz | 2.71 | value_1 | "[1,2,3]" | test 123 |
+| /path/file3.txt | syn12455 | zzz | 3.52 | value_3 | "[42,56,77]" | a single annotation |
+
+#### Multiple values of annotations per key
+
+Using multiple values for a single annotation should be used sparingly as it makes it more
+difficult for you to manage the data. However, it is supported.
+
+**Annotations can be comma `,` separated lists surrounded by brackets `[]`.**
```
linglp, line 74: Take a look at the old documentation here: https://python-docs.synapse.org/en/stable/explanations/manifest_tsv/?h=manifest#multiple-values-of-annotations-per-key, the examples here are much shorter. Is this epxected?

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,118 @@
...
+These columns are recognized by `sync_to_synapse` and have specific meaning. Any of these columns may be present in the manifest but only `path` and `parentId` are required for upload.
+
+| Field               | Meaning                                    | Example                                      |
+|---------------------|--------------------------------------------|----------------------------------------------|
+| path                | local file path or URL                     | /path/to/local/file.txt                      |
+| parentId            | Synapse ID of parent container             | syn1235                                      |
+| ID                  | Synapse entity ID                          | syn2345                                      |
+| name                | name of file in Synapse                    | Example_file                                 |
+| synapseStore        | whether to upload the file                 | True                                         |
+| contentType         | content type of file to overload defaults  | text/html                                    |
```
linglp: Nit: fix typo. `overwrite` instead of `overload`

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,118 @@
...
+| forceVersion        | whether to update version                  | False                                        |
+| activityName        | name of activity in provenance             | Ran normalization                            |
+| activityDescription | text description of what was done          | Ran algorithm xyz with parameters...         |
+| used                | list of items used to generate file        | syn1235;/path/to_local/file.txt              |
+| executed            | list of items executed                     | https://github.org/;/path/to_local/code.py   |
```
linglp, line 39: I like that the old documentation has an explanation of what `syn1235;/path/to_local/file.txt` means. Can we add that too? For reference: https://python-docs.synapse.org/en/stable/explanations/manifest_tsv/?h=manifest#activityprovenance-fields

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,954 @@
...
+    SynapseProvenanceError,
+)
+from synapseclient.core.utils import (
+    bool_or_none,
+    datetime_or_none,
+    get_synid_and_version,
+    is_synapse_id_str,
+    is_url,
+    test_import_pandas,
+    topolgical_sort,
```
linglp, line 26: probably this needs to be fixed in a different PR. Is there a typo in the function name here? should this be `topological_sort` instead of `topolgical_sort`?

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
```diff
@@ -0,0 +1,1716 @@
...
+            used=[],
+            executed=[],
+            activity_name=None,
+            activity_description=None,
+            dependent_futures=[],
+        )
+
+        assert result is mock_file
+        mock_file.store_async.assert_awaited_once_with(synapse_client=self.uploader.syn)
+        assert (
```
linglp: I think this is always true. I tried: 
```
used=["syn999"],
```
And this still return True. 

This is because Magic mock does not raise an AttributeError when trying to access an attribute that does not exist so `mock_file.activity == mock_file.activity` will always return True and `not hasattr(mock_file, "activity")` always return False.

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
```diff
@@ -0,0 +1,1716 @@
...
+        assert mock_file.activity.description == "my description"
+        assert len(mock_file.activity.used) == 1
+        assert len(mock_file.activity.executed) == 1
+
+    async def test_dependent_futures_resolved(self) -> None:
+        """Dependent futures are awaited and their results populate resolved_file_ids."""
+        dep_file = _make_file_mock("/dep.txt", "syn_dep")
+
+        # Create a real future that resolves to dep_file
+        dep_future = asyncio.get_event_loop().create_future()
```
linglp: should we be using `asyncio.get_event_loop()` here? Based on the [documentation](https://docs.python.org/3.11/library/asyncio-eventloop.html#asyncio.get_event_loop), for Python 3.10+, we should consider using `asyncio.run`

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
```diff
@@ -0,0 +1,1716 @@
...
+
+        with patch(
+            "synapseclient.models.services.manifest._check_provenance",
+            new=AsyncMock(return_value=existing_file),
+        ) as mock_check:
+            result = await _sort_and_fix_provenance(self.syn, df)
+
+        # _check_provenance must be called with the File object, not item.strip()
+        mock_check.assert_awaited_once_with(
+            existing_file, str(f), self.syn, mock_check.call_args[0][3]
```
linglp: this is weird. This will always pass right?

linglp (changes requested):
@andrewelamb Thanks for the last round of fixes. there are still some problems with tests and documentation. Could you check again?

docs/explanations/manifest_csv.md
andrewelamb: You're right it's just the UI

docs/explanations/manifest_csv.md
andrewelamb: Yes, I wasn't sure all the examples needed to be included. What do you think?

synapseclient/models/services/manifest.py
andrewelamb: That is correct. We would need to decide if this is a public function, and if so it would need to be deprecated first.

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
andrewelamb: I fixed this and added a couple more unit tests

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
andrewelamb: So Claude says on this :

```
In this context — an async test function — neither asyncio.get_event_loop() nor asyncio.run() is the right choice. 
Since the test is already running inside an async event loop (managed by pytest-asyncio), the correct approach is:                                      
                                                                                                                                                                                                                                        
  dep_future = asyncio.get_running_loop().create_future()                                                                                                                                                                               
                                                                                                                                                                                                                                        
  - get_event_loop() is deprecated for this use case in 3.10+ and emits a deprecation warning when no running loop exists.                                                                                                              
  - asyncio.run() would try to create a new event loop, which conflicts with the already-running one.
  - get_running_loop() is the correct call from within an async function — it returns the loop that's already executing the test.  
```

synapseclient/models/mixins/storable_container.py
```diff
@@ -496,6 +504,123 @@ async def _sync_from_synapse_async(
...
+                created instance from the Synapse class constructor.
+
+        Returns:
+            List of File entities that were created or updated. Returns an
+            empty list if dry_run=True or if no rows were eligible for
+            upload.
+
+        Example: Using this function
+
+            ```python
```
linglp, line 559: the rendering still has a bit of issue? see here: https://synapsepythonclient--1353.org.readthedocs.build/en/1353/reference/experimental/async/folder/#synapseclient.models.Folder.sync_to_synapse_async

and here: 

Image

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
```diff
@@ -0,0 +1,2151 @@
...
+        """A manifest without a 'parentId' column raises ValueError."""
+        f = tmp_path / "file.txt"
+        f.write_text("content")
+        csv = f"path\n{f}\n"
+        manifest = tmp_path / "manifest.csv"
+        manifest.write_text(csv)
+        with pytest.raises(ValueError, match="'parentId'"):
+            await read_manifest_for_upload(str(manifest), self.syn, True, False)
+
+    # -- error column handling ---------------------------------------------
```
linglp: this needs to be deleted?

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
```diff
@@ -0,0 +1,2151 @@
...
+        """A manifest row pointing to a zero-byte file raises ValueError."""
+        f = tmp_path / "empty.txt"
+        f.write_text("")  # 0 bytes
+        csv = f"path,parentId\n{f},syn1\n"
+        manifest = tmp_path / "manifest.csv"
+        manifest.write_text(csv)
+        with pytest.raises(ValueError, match="empty"):
+            await read_manifest_for_upload(str(manifest), self.syn, True, False)
+
+    # -- successful upload items -------------------------------------------
```
linglp: what is this for?

tests/unit/synapseclient/models/async/unit_test_manifest_async.py
andrewelamb: Sorry I'm still getting used to Claude Code, it keeps inserting these things

tests/integration/synapseclient/models/async/test_storable_container_async.py
```diff
@@ -0,0 +1,398 @@
...
+            manifest_path=str(manifest_path),
+            send_messages=False,
+            synapse_client=self.syn,
+        )
+
+        # THEN both files should exist in Synapse
```
linglp: Nit: I think here we shouldn't rely on `sync_from_synapse_async` for clean up. A better way to do this would be something like this: 
```
uploaded_files = await project_model.sync_to_synapse_async(
            manifest_path=str(manifest_path),
            send_messages=False,
            synapse_client=self.syn,
        )
for f in uploaded_files:
       self.schedule_for_cleanup(f.id)
```

tests/unit/synapseclient/models/async/unit_test_storable_container_async.py
```diff
@@ -0,0 +1,82 @@
...
+import uuid
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from synapseclient import Synapse
+from synapseclient.models import Project
+
+_TEST_PROJECT = Project(id="test", name="test")
```
linglp: why are we using a test project like this instead of using a fixture?

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,119 @@
...
+
+The CSV manifest format is shared across multiple tools:
+
+| Source                                                   | Filename                        |
+|----------------------------------------------------------|---------------------------------|
+| `Project.sync_from_synapse` / `Folder.sync_from_synapse` | manifest.csv                    |
+| Synapse UI download cart                                 | manifest.csv                    |
+| CLI `synapse get-download-list`                          | `manifest_<timestamp>.csv`      |
+
+A manifest generated by any of these sources can be used as input to `sync_to_synapse`, provided the `path` column is present with valid local file paths. Manifests from the Synapse UI  not include a `path` column by default, so users must add it before uploading.
```
linglp: Typo here needs to be fixed. `Manifests from the Synapse UI do not include a `path` column by default`

docs/explanations/manifest_csv.md
```diff
@@ -0,0 +1,119 @@
...
+| path     | local file path or URL     | /path/to/local/file.txt |
+| parentId | Synapse ID of parent       | syn1235                 |
+
+!!! note
+    The legacy TSV manifest used a column named `parent`. The CSV manifest uses `parentId` instead, which is consistent with the Synapse REST API field name. If you are migrating an existing TSV manifest to CSV, rename the `parent` column to `parentId`.
+
+### Standard fields
+
+These columns are recognized by `sync_to_synapse` and have specific meaning. Any of these columns may be present in the manifest but only `path` and `parentId` are required for upload.
+Each of these are individual examples and is what you would find in a row in each of these columns. To clarify, "syn1235;/path/to_local/file.txt" below states that you would like both "syn1234" and "/path/to_local/file.txt" added as items used to generate a file. You can also specify one item by specifying "syn1234"
```
linglp: Here you stated: To clarify, "syn1235;/path/to_local/file.txt" below states that you would like both "syn1234" and "/path/to_local/file.txt" added as items used to generate a file. 

The example uses `syn1235` and the explanation uses `syn1234`

linglp: Fix typo: Each of these are individual examples -> Each of these is individual examples

linglp (approved):
There are still some problems with tests and documentation. One concern that I have is that a lot of tests are relying on `sync_from_synapse_async` for clean up which means if the assertion failed after `sync_from_synapse_async` or the function itself failed, the test files won't be cleaned up at all. But overall it looks good!

tests/unit/synapseclient/models/async/unit_test_storable_container_async.py
andrewelamb: There's not much reason to use a fixture since it isn't being stored. The Folder is just created locally, and for all the tests in this file, an error happens before it's ever stored. That said, I removed it since it is a bit confusing.

PR #1354, opened 2026-04-08
[SYNPY-1799, SYNPY-1804] Add manifest generation to sync_from_synapse

danlu1:
# **Problem:**

StorableContainer.sync_from_synapse() does not generate a manifest file after downloading content from Synapse. We want to add manifest generation similar to synapseutils.syncFromSynapse() so users can use that manifest for uploading data via sync_to_synapse.

# **Solution:**

1. Added manifest CSV generation to StorableContainer.sync_from_synapse()
2. The manifest is written in CSV format using parentId and ID column names, consistent with the Synapse UI download cart and synapse get-download-list CLI output. Annotations and provenance are included as additional columns.
3. The implementation lives in three new helpers in storable_container.py: 

- `_extract_entity_metadata_for_manifest_csv()` — extracts per-file rows and discovers annotation column keys dynamically

- `_write_manifest_data_csv()` — writes the CSV using DictWriter
- `generate_manifest_csv()` — public entry point, skips gracefully if path or files are absent

4. The protocol definition in `storable_container_protocol.py` was updated to keep the sync and async signatures in alignment.
5. Documents and tutorials have been updated accordingly.

# **Testing:**

Unit tests and integration tests were added.


synapseclient/models/mixins/storable_container.py
```diff
@@ -412,6 +434,7 @@ async def _sync_from_synapse_async(
         link_hops: int = 1,
         queue: asyncio.Queue = None,
         include_types: Optional[List[str]] = None,
+        manifest: str = "all",
```
andrewelamb, line 437: This type should be `Literal["all", "suppress", "root"]

synapseclient/models/mixins/storable_container.py
```diff
@@ -426,6 +449,12 @@ async def _sync_from_synapse_async(
         syn.logger.info(
             f"[{self.id}:{self.name}]: Syncing {self.__class__.__name__} from Synapse."
         )
+
+        if manifest not in ("all", "root", "suppress"):
+            syn.logger.error(
+                f"[{self.id}:{self.name}]: Invalid manifest value: {manifest}"
+            )
+
```
andrewelamb, line 457: Shouldn't an exception be raised here if it's not one of these values?

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+    usable by csv.DictWriter
+
+    Arguments:
+        all_files: an iterable that provides File entities
+
+    Returns:
+        keys: a list column headers
+        data: a list of dicts containing data from each row
+    """
+    from synapseutils.sync import _convert_manifest_data_items_to_string_list
```
andrewelamb, line 66: We want to remove `synapseutils.sync` in v5.0.0, so anythign you need from there should be copied over to this file used here.

synapseutils/sync.py
```diff
@@ -78,6 +78,13 @@
 COMMAS_OUTSIDE_DOUBLE_QUOTES_PATTERN = re.compile(r",(?=(?:[^\"]*\"[^\"]*\")*[^\"]*$)")
 
 
+@deprecated(
+    version="4.11.0",
```
andrewelamb, line 82: This should be 4.12.0

tests/unit/synapseclient/models/async/unit_test_folder_async.py
```diff
@@ -1,17 +1,26 @@
...
 from synapseclient import Synapse
 from synapseclient.core.constants import concrete_types
-from synapseclient.core.constants.concrete_types import FILE_ENTITY
+from synapseclient.core.constants.concrete_types import FILE_ENTITY, FOLDER_ENTITY
 from synapseclient.core.exceptions import SynapseNotFoundError
-from synapseclient.models import FailureStrategy, File, Folder
+from synapseclient.models import Activity, FailureStrategy, File, Folder
+from synapseclient.models.services.manifest import (
+    _extract_entity_metadata_for_manifest_csv,
+    _write_manifest_data_csv,
```
andrewelamb, line 21: This doesn't seem to be used

tests/unit/synapseclient/models/async/unit_test_folder_async.py
```diff
@@ -821,3 +795,341 @@ async def mock_get_children(*args, **kwargs):
...
+        # WHEN metadata is extracted
+        keys, data = _extract_entity_metadata_for_manifest_csv([f])
+
+        # THEN annotation keys are included and single-item lists are serialized as scalars
+        assert "tissue" in keys
+        assert "count" in keys
+        assert data[0]["tissue"] == "brain"
+        assert data[0]["count"] == "42"
+
+    def test_write_manifest_data_csv_produces_comma_separated_output(self) -> None:
```
andrewelamb, line 1043: `write_manifest_data_csv` doesn't seem to be called by this test?

synapseclient/models/mixins/storable_container.py
```diff
@@ -494,6 +523,22 @@ async def _sync_from_synapse_async(
                 for task in worker_tasks:
                     task.cancel()
 
+        if create_workers and path and manifest != "suppress":
```
andrewelamb, line 526: I think a comment is needed here describing the `create_workers` logic.

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+
+def _manifest_csv_filename(path: str) -> str:
+    return os.path.join(os.path.expanduser(path), MANIFEST_CSV_FILENAME)
+
+
+def _get_entity_provenance_dict_for_manifest(entity: "File") -> Dict[str, str]:
+    """
+    Get the provenance metadata for the entity.
+    Arguments:
+        entity: Entity object
```
andrewelamb, line 34: This is a bit vague, it should at least be "File entity object"

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+
+def _extract_entity_metadata_for_manifest_csv(
+    all_files: List["File"],
+) -> Tuple[List[str], List[Dict]]:
+    """
+    Extracts metadata from the list of File Entities and returns them in a form
+    usable by csv.DictWriter
+
+    Arguments:
+        all_files: an iterable that provides File entities
```
andrewelamb, line 60: The type hint is a list not iterable

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+        "used": ";".join(used),
+        "executed": ";".join(executed),
+        "activityName": entity.activity.name or "",
+        "activityDescription": entity.activity.description or "",
+    }
+
+
+def _extract_entity_metadata_for_manifest_csv(
+    all_files: List["File"],
+) -> Tuple[List[str], List[Dict]]:
```
andrewelamb, line 54: `List[Dict]]`: could be `list[dict[str, Any]]`

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Extracts metadata from the list of File Entities and returns them in a form
+    usable by csv.DictWriter
+
+    Arguments:
+        all_files: an iterable that provides File entities
+
+    Returns:
+        keys: a list column headers
```
andrewelamb, line 63: missing "of" here

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+        "activityDescription": entity.activity.description or "",
+    }
+
+
+def _extract_entity_metadata_for_manifest_csv(
+    all_files: List["File"],
+) -> Tuple[List[str], List[Dict]]:
+    """
+    Extracts metadata from the list of File Entities and returns them in a form
+    usable by csv.DictWriter
```
andrewelamb, line 57: This function seems to be doing a lot more than just extracting metadata. Could you elaborate more on what this si doing here?

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+        data.append(row)
+    keys.extend(annotation_keys)
+    return keys, data
+
+
+def _write_manifest_data_csv(filename: str, keys: List[str], data: List[Dict]) -> None:
+    """
+    Write the manifest data to a CSV file.
+    Arguments:
+        filename: The name of the file to write to.
```
andrewelamb, line 98: Calling this a path might be more descriptive than filename.

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+        keys: The keys of the manifest.
```
andrewelamb, line 99: Could you elaborate more on what this is?

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
...
+        data: The data to write to the manifest. This should be a list of dicts where
+            each dict represents a row of data.
+    Returns:
+        None
```
andrewelamb, line 103: nit: don't need to have a return section if nothing is returned

synapseclient/models/services/manifest.py
```diff
@@ -0,0 +1,134 @@
+"""Functions for generating manifest CSV files from File entities."""
+
+import csv
+import io
+import os
+from typing import TYPE_CHECKING, Dict, List, Tuple
```
andrewelamb, line 6: you don't need to import these any more, instead use the lower case verisons below `dict`, `list`, `tuple`

andrewelamb (changes requested):

andrewelamb:
Thanks @danlu1! I left some comments. You'll also need to update the docs folder:
  
- docs/tutorials/python/download_data_in_bulk.md: Use the new method                                                                                                                                                                                                                                                                                                                                                                              
- docs/explanations/manifest_tsv.md: Don't make any changes here, this will be done in a different ticket

synapseclient/models/mixins/storable_container.py
danlu1: Added the comments. I also decoupled the `create_workers` and manifest generation cuz we would allow manifest generation when queue is passed.

synapseclient/models/services/manifest.py
```diff
@@ -77,6 +98,215 @@
...
+
+    Arguments:
+        all_files: A list of File model objects to extract metadata from.
+
+    Returns:
+        A tuple of (keys, data) where keys is the ordered list of column headers and
+        data is a list of row dicts, one per file.
+    """
+    keys = list(DEFAULT_GENERATED_MANIFEST_CSV_KEYS)
+    annotation_keys: set = set()
```
andrewelamb, line 225: Is the annotation order important? `set` is unordered.

synapseclient/models/services/manifest.py
```diff
@@ -77,6 +98,215 @@
...
+            "path": entity.path,
+            "parentId": entity.parent_id,
+            "name": entity.name,
+            "ID": entity.id,
+            "synapseStore": entity.synapse_store,
+            "contentType": entity.content_type,
+        }
+        if entity.annotations:
+            for key, val in entity.annotations.items():
+                annotation_keys.add(key)
```
andrewelamb, line 238: Can you remind me, are the keys in the `dict` right above restricted in Synapse as annotation keys? For example, if a file had an annotation of `name` this would overwrite the name value in the above dict.

synapseclient/models/protocols/storable_container_protocol.py
```diff
@@ -35,6 +35,7 @@ def sync_from_synapse(
         link_hops: int = 1,
         queue: asyncio.Queue = None,
         include_types: Optional[List[str]] = None,
+        manifest: str = "all",
```
andrewelamb: this should match the async version:

```
manifest: Literal["all", "suppress", "root"] = "all",
```

synapseclient/models/mixins/storable_container.py
```diff
@@ -434,6 +488,13 @@ async def _sync_from_synapse_async(
         syn.logger.info(
             f"[{self.id}:{self.name}]: Syncing {self.__class__.__name__} from Synapse."
         )
+
+        if manifest not in ("all", "root", "suppress"):
```
andrewelamb: It seems like this check should be at the top of method so it fails fast( the error mesage doesn't need `self.id`, `self.name`)

synapseclient/models/services/manifest.py
```diff
@@ -77,6 +98,215 @@
...
+
+    The generated file uses CSV format with comma delimiter and is interoperable
+    with the Synapse UI download cart. Column names follow the new convention:
+    `parentId` (instead of `parent`) and `ID` (instead of `id`).
+
+    Arguments:
+        all_files: A list of File model objects.
+        path: The directory path where manifest.csv will be written.
+    """
+    if path and all_files:
```
andrewelamb: This doesn't generate a manifest if `all_files == []`. I would do one of:

1. Document this and raise a warning when `all_files` is empty
2. Create a manifest with just a header

synapseclient/models/mixins/storable_container.py
```diff
@@ -1232,6 +1312,7 @@ async def _wrap_recursive_get_children(
             synapse_client=synapse_client,
             queue=queue,
             include_types=include_types,
+            manifest="suppress",  # suppress manifest generation for recursive calls since the root covers the path already in map_directory_to_all_contained_files
```
andrewelamb: This comment isn't super clear, could you clarify here?

synapseclient/models/services/manifest.py
```diff
@@ -77,6 +98,215 @@
...
+
+    When working with datetime fields we are printing the ISO 8601 UTC representation of
+    the datetime.
+
+    When working with non strings we are printing the non-quoted version of the object.
+
+    Example: Examples
+        Several examples of how this function works.
+
+            >>> _convert_manifest_data_items_to_string_list(["a", "b", "c"])
```
andrewelamb: Do these examples work? They are not in the style most of our examples are in like:

```
  Example: Multi-element string list                                                                                                                                                        
      Wraps in brackets and joins with commas.                                                                                                                                              
      ```python                                                                                                                                                                             
      _convert_manifest_data_items_to_string_list(["a", "b", "c"])
      # '[a,b,c]'                                                                                                                                                                           
      ```                                                                                                                                                                                   
  
  Example: Strings containing commas                                                                                                                                                        
      Comma-containing strings are quoted, others are left as-is.                                                                                                                         
      ```python                                                                                                                                                                             
      _convert_manifest_data_items_to_string_list(["hello,world", "plain text"])
      # '["hello,world",plain text]'                                                                                                                                                        
      ```   
```

synapseclient/models/mixins/storable_container.py
```diff
@@ -71,6 +72,8 @@
         VirtualTable,
     )
 
+from synapseclient.models.services.manifest import generate_manifest_csv
+
```
andrewelamb: nit: There are already imports from `synapseclient.models.services.manifest`, this could be added to on line 51

andrewelamb (changes requested):

synapseclient/models/services/manifest.py
danlu1: This function was copied here and examples were tested in the [unit test](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/0210ecf5f56d1214351838784463d018be42a9e1/tests/unit/synapseclient/services/unit_test_manifest.py#L380-L412). Can you clarify what style do you mean? I think they are the same.

synapseclient/models/services/manifest.py
danlu1: I don’t think the order of the annotations matters.

synapseclient/models/services/manifest.py
danlu1: Good point. Since the CSV file itself does not forbid you from having multiple columns with the same name, I don't think it will overwrite the default `name` column. Instead, it just has another `name` column with the annotation value. Do we want to preserve the default columns? 

Image

synapseclient/models/mixins/storable_container.py
danlu1: Nice catch!

synapseclient/models/mixins/storable_container.py
danlu1: The manifest is suppressed for child folders because they’re already accounted for when iterating through their parent folder. This is handled in the `map_directory_to_all_contained_files` function, which returns all files in the directory, including those in its child directories.

synapseclient/models/services/manifest.py
andrewelamb: Sorry, I meant do the examples work in that they look right in the docs that get created

danlu1: They should. I will double check.

synapseclient/models/services/manifest.py
danlu1: I went with the second option and updated both the code and the unit tests.

synapseclient/models/services/manifest.py
danlu1: Yes, examples here are correct 

Image

danlu1: @andrewelamb I reorganized the examples. Please feel free to take a look.

synapseclient/models/services/manifest.py
andrewelamb: I think that Synapse doesn't allow these to be annotation keys in the first place. Pandas definitely doesn't allow multiple columns with the same name. At some point Pandas would change the name of the second name column (name1?) I think. Either way, I'm wondering if we shoudl raise an exception here if that happens.

synapseclient/models/mixins/storable_container.py
andrewelamb: Could you replace the comment with this explanation?

synapseclient/models/services/manifest.py
danlu1: Dan will check the reserved annotation keys.

danlu1: @andrewelamb We do have property keys setup: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8cd7c056722ddcb4586ca8893bce2a9ce7cea639/synapseclient/entity.py#L88-L105 and `name` is one of them. As this [doc](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8cd7c056722ddcb4586ca8893bce2a9ce7cea639/docs/explanations/properties_vs_annotations.md), properties always take precedence over annotations in lookups. So I think we don't have to worry about the edge case here. 

Image

synapseclient/models/services/manifest.py
```diff
@@ -77,6 +98,232 @@
...
+        path: The directory path where manifest.csv will be written.
+
+    raises:
+        ValueError: If path is None.
+    """
+    if not path:
+        raise ValueError(
+            "The path argument is required to generate a manifest.csv file."
+        )
+    if path:
```
andrewelamb: nit: The code in the `if` block doesn't need to be  

```
    if not path:
        raise ValueError(
            "The path argument is required to generate a manifest.csv file."
        )
    filename = _manifest_csv_filename(path=path)
    keys, data = _extract_entity_metadata_for_manifest_csv(all_files=all_files)
    _write_manifest_data_csv(filename, keys, data)

File: synapseclient/models/services/manifest.py [hunk 26]
```diff
@@ -77,6 +98,232 @@
...
+        "used": ";".join(used),
+        "executed": ";".join(executed),
+        "activityName": entity.activity.name or "",
+        "activityDescription": entity.activity.description or "",
+    }
+
+
+def _convert_manifest_data_items_to_string_list(
+    items: list[Union[str, datetime.datetime, bool, int, float]],
+) -> Union[str, list[str]]:
```
andrewelamb: It seems like this function only returns a `str`, not `list[str]`. The dostring description and return description should be updated as well.

synapseclient/models/services/manifest.py
```diff
@@ -77,6 +98,232 @@
...
+            restval="",
+            extrasaction="ignore",
+            quoting=csv.QUOTE_MINIMAL,
+        )
+        writer.writeheader()
+        for row in data:
+            writer.writerow(_convert_manifest_data_row_to_dict(row, keys))
+
+        # add a log message to the console
+        print(f"Manifest file {path} has been generated.")
```
andrewelamb: Should `logging.logger` be used here instead of `print`?

synapseclient/models/protocols/storable_container_protocol.py
```diff
@@ -35,6 +35,7 @@ def sync_from_synapse(
         link_hops: int = 1,
         queue: asyncio.Queue = None,
         include_types: Optional[List[str]] = None,
+        manifest: Literal["all", "root", "suppress"] = "all",
```
andrewelamb: This doesn't match the order of the literal in storable container.py. I would suggest making the literal a global in storable_container.py and then importing it here.

andrewelamb (approved):

andrewelamb:
@danlu1 I just had a few suggestions, but it looks good otherwise, so I approved!

synapseclient/models/protocols/storable_container_protocol.py
danlu1: I did the other way around: define ManifestSetting in the protocol and then import it in the `storable_container.py`  to fix circular import issue.

synapseclient/models/services/manifest.py
danlu1: nice catch!

PR #1358, opened 2026-04-15
Remove sorting functionality

cconrad8:
# **Problem:**

This pull request makes a minor change to how JSON files are exported in the `schema_generation.py` module. The `sort_keys` parameter in `json.dump` is set to `False`, so the keys in the output JSON files will retain their original order rather than being sorted alphabetically.

- JSON export behavior:
  * Updated both `export_schema` and `export_json` functions to use `sort_keys=False` in `json.dump`, preserving the original key order in exported JSON files. [[1]](diffhunk://#diff-cd4e9e05c7bc72c1ef38afad795296b9a0e6c8306a4058e4a33e9baccfe86908L4117-R4117) [[2]](diffhunk://#diff-cd4e9e05c7bc72c1ef38afad795296b9a0e6c8306a4058e4a33e9baccfe86908L5525-R5525)l problem that this PR solves.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -552,12 +552,18 @@ jobs:
           exit 1
 
   # containerize the package and upload to the GHCR upon new release (whether pre-release or not)
-  ghcr-build-and-push-on-release:
+  # Step 1: Build the Docker image and save as tar for scanning
+  ghcr-build-on-release:
     needs: deploy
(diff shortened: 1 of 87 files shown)


thomasyu888:
I'm going to merge this in and work on this in the feature branch.  Thanks for the contribution @cconrad8 !

PR #1359, opened 2026-04-15
[SYNPY-1769] Sort jsonschema by order of dependsOn column

thomasyu888:
# **Problem:**

- JSON schema generation was not respecting the order of columns as defined in the `dependsOn` column of the data model CSV.
- The graph traversal code in `schema_generation.py` called `get_node_dependencies()` with `schema_ordered=False` and then wrapped the result in `sorted()`, which alphabetized dependencies regardless of their defined order.
- Similarly, `json.dump()` was called with `sort_keys=True` in both `export_schema()` and `export_json()`, alphabetizing all JSON keys at serialization time.
- The underlying graph *does* preserve CSV order via edge weights (assigned as `relationships[csv_header].index(node)` at graph construction time), but the traversal layer was discarding this information.

# **Solution:**

- In `TraversalNode.__post_init__` and `GraphTraversalState.__post_init__`, switched `get_node_dependencies()` from `schema_ordered=False` to `schema_ordered=True` and removed the `sorted()` wrappers. This causes `get_ordered_entry()` to be used, which sorts nodes by their edge weights to reconstruct the original CSV order.
- In `GraphTraversalState.move_to_next_node()`, removed the `sorted()` wrapper around `self.current_node.dependencies` when enqueuing nodes to process, so traversal order also respects the `dependsOn` column.
- In `export_schema()` and `export_json()`, changed `sort_keys=True` to `sort_keys=False` in `json.dump()` so that the insertion order of keys in the schema dict is preserved on output.
- Note: `valid_values` ordering (from `get_node_range()`) is unaffected by this PR — `find_node_range()` internally uses `list(set(...))` which loses ordering, making that a separate issue.

# **Testing:**

- Ran pre-commit checks (`ruff`, `black`, `isort`, `bandit`) — all passed.
- Manual validation: generated a JSON schema from a data model CSV and confirmed that the property order in the output matches the `dependsOn` column order rather than alphabetical order.

- [ ] Needs to be tested
- [ ] Needs to confirm ordering of conditional dependency properties that are added

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -552,12 +552,18 @@ jobs:
           exit 1
 
   # containerize the package and upload to the GHCR upon new release (whether pre-release or not)
-  ghcr-build-and-push-on-release:
+  # Step 1: Build the Docker image and save as tar for scanning
+  ghcr-build-on-release:
     needs: deploy
     runs-on: ubuntu-latest
     permissions:
       contents: read
       packages: write
+    outputs:
+      image-tags: ${{ steps.set-tags.outputs.tags }}
+      image-name: synapsepythonclient-release
+    env:
+      TARFILE_NAME: synapsepythonclient-release.tar
 
     steps:
       - name: Check out the repo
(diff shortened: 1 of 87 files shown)


thomasyu888:
Closing for now to reduce clutter

PR #1360, opened 2026-04-15
[SYNPY-1781] Implement CSV Import to Grid Session

linglp:
# Problem:                                                                                                  
There is no way to programmatically populate an existing grid session with data from a CSV file. Users who want to import CSV data must do so manually through the Synapse UI.        
                                                                                                     
  # Solution:                                                                                                 
  - Adds Grid.import_csv_async (and sync wrapper import_csv) to the Grid model, which imports a CSV file into 
  an existing grid session via two sequential async jobs:                                                     
        a.    UploadToTablePreviewRequest (POST /table/upload/csv/preview/async/start) — infers the column schema    
  from the uploaded file.                                                                                     
        b.    GridCsvImportRequest (POST /grid/import/csv/async/start) — performs the actual CSV import using the    
  inferred schema.                                                                                                                    
                                                                                                              
  # Testing:                                                                                                  
  - Unit tests cover: missing session_id raises ValueError, happy-path verifies the log message contains      
  total/created/updated counts, to_synapse_request / fill_from_dict for both new dataclasses.                 
  - Integration test verifies end-to-end: create grid session → upload CSV file → import_csv_async →
  export_to_record_set_async → assert expected row count.


tests/integration/synapseclient/models/async/test_grid_async.py
```diff
@@ -183,3 +183,63 @@ async def test_delete_grid_session_validation_error_async(self) -> None:
...
         ):
             await grid.delete_async(synapse_client=self.syn)
+
+    async def test_import_csv_to_grid_session_async(
+        self,
+        project_model,
+        record_set_fixture: RecordSet,
+        schedule_for_cleanup: Callable[..., None],
+    ) -> None:
+        """Test importing a CSV file into a grid session."""
```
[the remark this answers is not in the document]
linglp: fixed

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,86 @@ async def main():
...
+
+        preview_response = await upload_to_table_preview.send_job_and_wait_async(
+            timeout=timeout, synapse_client=synapse_client
+        )
+
+        all_columns = preview_response.suggested_columns
+        import_request = GridCsvImportRequest(
+            session_id=self.session_id,
+            file_handle_id=file_handle_id,
+            schema=all_columns,
```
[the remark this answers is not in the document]
linglp: if users try to import an empty file, when I tested it, the user would get an error: 
```
  File "/Users/lpeng/code/synapsePythonClient/synapseclient/core/exceptions.py", line 296, in _raise_for_status_httpx
synapseclient.core.exceptions.SynapseHTTPError: 400 Client Error: The upsert key cannot have null or empty values.

```
Which is sufficient to alert them that they are inserting empty files.

synapseclient/models/curation.py
```diff
@@ -6,7 +6,7 @@
 """
 
 from dataclasses import dataclass, field, replace
-from typing import Any, AsyncGenerator, Dict, Generator, Optional, Protocol, Union
+from typing import Any, AsyncGenerator, Dict, Generator, List, Optional, Protocol, Union
```
andrewelamb: nit: Don't import `List` here, use `list` instead

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+    session_id: Optional[str] = None
+    """The grid session ID."""
+
+    file_handle_id: Optional[str] = None
+    """The id of the file handle that contains the CSV data."""
+
+    csv_descriptor: CsvTableDescriptor = field(default_factory=CsvTableDescriptor)
+    """The description of a csv for upload or download."""
+
+    schema: Optional[List[Column]] = None
```
andrewelamb: nit: use `list` here

synapseclient/models/curation.py
```diff
@@ -1427,6 +1581,53 @@ def list(
             ```
         """
 
+    def import_csv(
```
andrewelamb, line 1584: Should this have OTEL tracing?

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+
+        Returns:
+            A dictionary representation of this object for API requests.
+        """
+        request_dict: Dict[str, Any] = {"concreteType": self.concrete_type}
+        if self.session_id is not None:
+            request_dict["sessionId"] = self.session_id
+        if self.file_handle_id is not None:
+            request_dict["fileHandleId"] = self.file_handle_id
+        if self.csv_descriptor is not None:
```
andrewelamb: nit: this isn't needed since `self.csv_descriptor` can't be None

tests/integration/synapseclient/models/async/test_grid_async.py
```diff
@@ -183,3 +183,62 @@ async def test_delete_grid_session_validation_error_async(self) -> None:
...
+    async def test_import_csv_to_grid_session_async(
+        self,
+        project_model,
+        record_set_fixture: RecordSet,
+    ) -> None:
+        """Test importing a CSV file into a grid session."""
+
+        # GIVEN: Create a grid session first
+        grid = Grid(record_set_id=record_set_fixture.id)
+        created_grid = await grid.create_async(
```
andrewelamb: Should this grid be cleaned up, using `schedule_for_cleanup`?

tests/integration/synapseclient/models/async/test_grid_async.py
```diff
@@ -183,3 +183,62 @@ async def test_delete_grid_session_validation_error_async(self) -> None:
...
+            timeout=ASYNC_JOB_TIMEOUT_SEC, synapse_client=self.syn
+        )
+
+        assert created_grid.session_id is not None
```
andrewelamb: what is the purpose of this check?

andrewelamb:
Should the items you added to `synapseclient/models/curation.py` be in their own file like `synapseclient/models/grid.py` to match how things are organized at the [API Docs](https://rest-docs.synapse.org/rest/index.html#org.sagebionetworks.repo.web.controller.GridController)?

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
 
+@dataclass
+class GridCsvImportRequest(AsynchronousCommunicator):
+    """
+    A request to import a CSV file into a grid. Currently supports only grid
+    created from a record set.
+
+    This result is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/GridCsvImportRequest.html>
+    """
+
```
andrewelamb, line 1004: You may want to link to [the response object](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/GridCsvImportResponse.html) as well since this class uses it.

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+    def to_synapse_request(self) -> Dict[str, Any]:
+        """
+        Converts this dataclass to a dictionary suitable for a Synapse REST API request.
+
+        Returns:
+            A dictionary representation of this object for API requests.
+        """
+        request_dict: Dict[str, Any] = {"concreteType": self.concrete_type}
+        if self.session_id is not None:
+            request_dict["sessionId"] = self.session_id
```
andrewelamb: You might consider using the `delete_none_keys` function here instead of doing the `None` check for each field.

OTOH, are these all required for the [request](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/GridCsvImportRequest.html)? Perhaps this should raise an exception if any of these fields are None?

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+    created from a record set.
+
+    This result is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/GridCsvImportRequest.html>
+    """
+
+    concrete_type: str = GRID_CSV_IMPORT_REQUEST
+    """The concrete type for this request."""
+
+    session_id: Optional[str] = None
+    """The grid session ID."""
```
andrewelamb, line 1006: Shouldn't this, `file_handle_id` and `schema` be required?

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+
+    file_handle_id: Optional[str] = None
+    """The id of the file handle that contains the CSV data."""
+
+    csv_descriptor: CsvTableDescriptor = field(default_factory=CsvTableDescriptor)
```
andrewelamb, line 1017: Why isn't this setable?

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+            request_dict["fileHandleId"] = self.file_handle_id
+        if self.csv_descriptor is not None:
+            request_dict["csvDescriptor"] = self.csv_descriptor.to_synapse_request()
+        if self.schema is not None:
+            request_dict["schema"] = [col.to_synapse_request() for col in self.schema]
+        return request_dict
+
+
+@dataclass
+class UploadToTablePreviewRequest(AsynchronousCommunicator):
```
andrewelamb, line 1067: Most of my questions about `GridCsvImportRequest` apply here as well so I won't repeat them.

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+    """
+    Request for a preview of an upload to a Table.
+
+    This result is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/table/UploadToTablePreviewRequest.html>
```
andrewelamb: nit: should "result" be "request"?

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,88 @@ async def main():
         await delete_grid_session(
             session_id=self.session_id, synapse_client=synapse_client
         )
+
+    async def import_csv_async(
+        self,
+        file_handle_id: str,
+        csv_table_descriptor: Optional[CsvTableDescriptor] = None,
```
andrewelamb: I think this should either:

- Not be optional
- Have a factory constructor default
- In the function body:

```
 if not csv_table_descriptor:
    csv_table_descriptor = CsvTableDescriptor()
```
So None doesn't get passed to `UploadToTablePreviewRequest`

andrewelamb (changes requested):

andrewelamb:
Thanks @linglp! I mostly just have some questions for the moment. I'll do a more thorough review afterwards.

synapseclient/models/curation.py
linglp: this field is still settable. Default_factory only controls what happens when you omit the argument.

synapseclient/models/curation.py
linglp: Yeah I will fix it

synapseclient/models/curation.py
linglp: the codebase still predominantly uses List. I could change it to list since we are now supporting python 3.10+

tests/integration/synapseclient/models/async/test_grid_async.py
linglp: The grid session itself is likely low-cost to leave orphaned, and grids in the other tests are also not cleaned up. That said, I could add a try-except block to clean it up.

tests/integration/synapseclient/models/async/test_grid_async.py
linglp: It's a sanity check that create_async actually did something — if the API call silently failed or fill_from_dict didn't populate session_id, this catches it before the rest of the test proceeds.

linglp:
@andrewelamb Thanks for your feedbacks. I saw that `class Grid(GridSynchronousProtocol)`, `class GridSession`, `class GridRecordSetExportRequest` are still in `curation.py`, so I put `class GridCsvImportRequest` there as well. I am happy to create a separate ticket to create`grid.py`and move them there in a separate PR. But also, the `curator.py` should also probably be named as a different file

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,93 @@ async def main():
         await delete_grid_session(
             session_id=self.session_id, synapse_client=synapse_client
         )
+
+    async def import_csv_async(
+        self,
+        file_handle_id: str,
+        csv_table_descriptor: Optional[CsvTableDescriptor] = CsvTableDescriptor(),
```
andrewelamb, line 2046: Python evaluates default argument values once, at the time the function (or class) is defined — not on each call. So there is exactly one CsvTableDescriptor instance that lives forever and is handed to every caller who doesn't pass their own. This probably isn't a big deal they way it's written, but this should probably be changed to default to None, and then in the function body, if None is passed as the parameter, a CsvTableDescriptor() is created there.

andrewelamb (changes requested):

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,93 @@ async def main():
...
+        upload_to_table_preview = UploadToTablePreviewRequest(
+            csv_table_descriptor=csv_table_descriptor,
+            upload_file_handle_id=file_handle_id,
+        )
+
+        preview_response = await upload_to_table_preview.send_job_and_wait_async(
+            timeout=timeout, synapse_client=synapse_client
+        )
+
+        all_columns = preview_response.suggested_columns
```
andrewelamb, line 2113: Should there be a check here that there are columns in `all_collumns`?

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+            The UploadToTablePreviewRequest object.
+        """
+        suggested_columns_data = synapse_response.get("suggestedColumns", None)
+        if suggested_columns_data:
+            self.suggested_columns = [
+                Column().fill_from_dict(col) for col in suggested_columns_data
+            ]
+
+        sample_rows_data = synapse_response.get("sampleRows", None)
+        if sample_rows_data:
```
andrewelamb, line 1122: Should this be `if sample_rows_data is not None`? That way, if this gets an empty list `self.sample_rows` is still set?

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+        Converts a response from the REST API into this dataclass.
+
+        Arguments:
+            synapse_response: The response from the REST API.
+
+        Returns:
+            The UploadToTablePreviewRequest object.
+        """
+        suggested_columns_data = synapse_response.get("suggestedColumns", None)
+        if suggested_columns_data:
```
andrewelamb, line 1116: Same as right below

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,93 @@ async def main():
...
+            csv_table_descriptor=csv_table_descriptor,
+            upload_file_handle_id=file_handle_id,
+        )
+
+        preview_response = await upload_to_table_preview.send_job_and_wait_async(
+            timeout=timeout, synapse_client=synapse_client
+        )
+
+        all_columns = preview_response.suggested_columns
+        import_request = GridCsvImportRequest(
```
andrewelamb, line 2114: `lines_to_skip` and `do_full_file_scan` could be set here, but aren't since this method doesn't take them as arguments. Should it?

andrewelamb (changes requested):

synapseclient/models/curation.py
linglp: I feel pretty conflicted about this. I checked the source code and front-end of Synapse: 

Image

There's no option for line skip or turning off the whole file scan

and for the source code: 
* it calls `CsvPreviewDialog` [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/02a89b7f0e863ab56a7fb0bf17ae4dfb93f10a94/packages/synapse-react-client/src/components/DataGrid/components/UploadCsvToGridDialog.tsx#L52)
* and then if you check out `CsvPreviewDialog` [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx#L51), and then you could see how `CsvPreview` [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx#L91-L96) only has two parameters: `fileHandleId` and `csvTableDescriptor`

Like the UI, there's no option for line skip and whole file scan. 

I then decided it might be best not to expose these parameters directly. From the user's perspective, it's also confusing to not scan the whole file for inferring column types while still importing the whole file.

The problem with exposing linesToSkip in import_csv_async is:                                            
It only affects type inference in the preview step — the skipped lines are excluded when guessing columns types. But those lines would still get imported in step 2, because the import API has no concept of linesToSkip. I think it is just confusing for users if we expose that and they expect that option to work for both `import` and `preview`

synapseclient/models/curation.py
linglp: Thanks for the suggestion. I will fix it using a oneliner.

synapseclient/models/curation.py
linglp: I can fix that.

synapseclient/models/curation.py
linglp: Just fixed that. Thanks for the suggestion.

synapseclient/models/curation.py
```diff
@@ -1427,6 +1581,53 @@ def list(
             ```
         """
 
+    def import_csv(
+        self,
+        file_handle_id: str,
+        *,
+        timeout: int = 120,
+        csv_table_descriptor: Optional[CsvTableDescriptor] = None,
```
andrewelamb, line 1593: The arguments are not in the same order as the async version. I think they need to be?

synapseclient/models/curation.py
andrewelamb: Don't forget the OTEL tracing here. :)

tests/integration/synapseclient/models/async/test_grid_async.py
```diff
@@ -183,3 +183,68 @@ async def test_delete_grid_session_validation_error_async(self) -> None:
...
+                    "id": [6, 7, 8, 9, 10],
+                    "name": ["Alpha", "Beta", "Gamma", "Delta", "Epsilon"],
+                    "value": [10.5, 20.3, 30.7, 40.1, 50.9],
+                    "category": ["A", "B", "A", "C", "B"],
+                    "active": [True, False, False, True, True],
+                }
+            )
+
+            # Create a temporary CSV file
+            with tempfile.NamedTemporaryFile(
```
andrewelamb, line 215: Claude found this issue:

```
What's happening at the OS level                                                                                                                                                                                
                                                                                                                                                                                                                  
  1. tempfile.NamedTemporaryFile("w", ...) creates a new file on disk and returns it already open for writing. temp_csv is a live file handle.                                                                    
  2. Inside the with block, test_data.to_csv(temp_csv.name, ...) tells pandas "write to this path." Pandas opens its own second handle to the same path, writes the CSV, and closes it.                         
  3. Still inside the with, File(path=temp_csv.name, ...).store_async() reads the file to upload it — potentially via yet another handle.                                                                         
                                                                                                                                                                                                                
  So at step 2 there are two concurrent handles to the same path: the tempfile's (untouched) and pandas's. At step 3 the tempfile handle is still open while the upload reads the file. The with block doesn't    
  exit until after the upload.                                                                                                                                                                                    
                                                                                                                                                                                                                  
  Why it matters                                                                                                                                                                                                  
                                                                                                                                                                                                                  
  - Windows: NamedTemporaryFile on Windows opens with a deny-sharing flag. A second open — even from the same process — fails with PermissionError ([WinError 32] The process cannot access the file because it is
   being used by another process). On Windows this test would blow up at the to_csv line. Python's own tempfile docs call this out: "Whether the name can be used to open the file a second time, while the named
  temporary file is still open, varies across platforms."                                                                                                                                                         
  - Linux/macOS: POSIX semantics allow many handles per inode, so the test passes — but only by coincidence. The tempfile handle was opened in "w" mode but nothing is ever written through it, so there's no
  buffer-flush race. If a future edit ever writes through temp_csv itself in addition to pandas, the order of flushes on close decides what ends up on disk.                                                      
  - Subtler risk: reading the file while the tempfile handle is still open is fine on Linux, but intermittent on Windows with strict AV scanners or on network filesystems.                                     
                                                                                                                                                                                                                  
  The CI currently runs integration tests on Linux, so this doesn't break CI today. It breaks the moment anyone runs the suite on Windows — and Synapse does publish a Windows-compatible client.
```

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,97 @@ async def main():
...
+            raise ValueError(
+                "No columns were detected in the CSV file when previewing for import. Please check that the file handle id is correct and that the file contains valid CSV data."
+            )
+
+        all_columns = preview_response.suggested_columns
+        import_request = GridCsvImportRequest(
+            session_id=self.session_id,
+            file_handle_id=file_handle_id,
+            schema=all_columns,
+            csv_descriptor=csv_table_descriptor or CsvTableDescriptor(),
```
andrewelamb: I don't know if this is a bug, but there's a weird issue. Sometimes, the `CsvTableDescriptor` object is shared between `GridCsvImportRequest` here and `UploadToTablePreviewRequest` and sometimes it isn't. 

Two scenarios:
1. User inputs a `CsvTableDescriptor` object: the same object gets passed to both `GridCsvImportRequest` and `UploadToTablePreviewRequest`.
2. Use doesn't input `CsvTableDescriptor` object: Two different `CsvTableDescriptor` objects are created in the `or` statements.

I would get rid of the `or` statements. At the top of this mehtod ~line 2093 I would instead put:

```
if not csv_table_descriptor:
   csv_table_descriptor = CsvTableDescriptor()
```

Then this same `CsvTableDescriptor` is consistently shared between the two calls. If they are not syupposed ot be shared (due to mutation of it), then use `csv_table_descriptor.copy` on the second call.

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,97 @@ async def main():
...
+            asyncio.run(main())
+            ```
+        """
+
+        if not self.session_id:
+            raise ValueError(
+                "session_id is required to import a CSV into a GridSession"
+            )
+
+        trace.get_current_span().set_attributes(
```
andrewelamb, line 2110: Does this work inside the method?

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,97 @@ async def main():
...
+
+        preview_response = await upload_to_table_preview.send_job_and_wait_async(
+            timeout=timeout, synapse_client=synapse_client
+        )
+        if not preview_response.suggested_columns:
+            raise ValueError(
+                "No columns were detected in the CSV file when previewing for import. Please check that the file handle id is correct and that the file contains valid CSV data."
+            )
+
+        all_columns = preview_response.suggested_columns
```
andrewelamb: nit: all_columns isn't used again so this line isn't needed.

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+    """The total number of rows in the CSV."""
+
+    created_count: Optional[int] = field(default=None, compare=False)
+    """The number of rows that were created."""
+
+    updated_count: Optional[int] = field(default=None, compare=False)
+    """The number of rows that were updated."""
+
+    def fill_from_dict(
+        self, synapse_response: Union[Dict[str, Any], Any]
```
andrewelamb: This is a weird type. It should be always be a `dict` based on the body of the method, using `.get`. Alternatively, if it's not always a `dict` then the method should handle those cases.

synapseclient/models/curation.py
```diff
@@ -989,6 +991,158 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+    sample_rows: Optional[list[list[Optional[str]]]] = field(
+        default=None, compare=False
+    )
+    """A sample of the rows in the file."""
+
+    rows_scanned: Optional[int] = field(default=None, compare=False)
+    """The number of rows scanned from the file."""
+
+    def fill_from_dict(
+        self, synapse_response: Union[Dict[str, Any], Any]
```
andrewelamb: This is a weird type. It should be always be a dict based on the body of the method, using .get. Alternatively, if it's not always a dict then the method should handle those cases.

synapseclient/models/curation.py
```diff
@@ -1838,3 +2039,97 @@ async def main():
...
+            csv_table_descriptor=csv_table_descriptor or CsvTableDescriptor(),
+            upload_file_handle_id=file_handle_id,
+        )
+
+        preview_response = await upload_to_table_preview.send_job_and_wait_async(
+            timeout=timeout, synapse_client=synapse_client
+        )
+        if not preview_response.suggested_columns:
+            raise ValueError(
+                "No columns were detected in the CSV file when previewing for import. Please check that the file handle id is correct and that the file contains valid CSV data."
```
andrewelamb: nit: Claude made this suggestion:

```
The second sentence guesses at why the preview came back empty:                                                                                                                                                 
                                                                       
  ▎ "Please check that the file handle id is correct and that the file contains valid CSV data."                                                                                                                  
                                                                       
  Three concrete issues:

  1. "Check that the file handle id is correct" is misleading. If the file handle id pointed at a nonexistent or inaccessible file, the preview async job would have failed (raising SynapseError /               
  SynapseHTTPError) long before returning a structured response with suggestedColumns: []. So by the time we hit this branch, we already know the file handle resolved to something readable. Pointing the user at
   the file handle id sends them hunting in the wrong direction.                                                                                                                                                  
  2. "Contains valid CSV data" is fuzzy. An empty CSV, a CSV with only a header row and no data, or a CSV whose separator disagrees with csv_table_descriptor can all legitimately produce zero suggested columns.
   The message doesn't help the user distinguish between them. "Valid CSV" is subjective — by whose grammar?                                                                                                      
  3. The actionable context is missing. The user doesn't see their file_handle_id, the csv_table_descriptor values, or the rows_scanned/sample_rows the preview did return. All of that is already in scope inside
   this method and would make the error diagnosable.                                                                                                                                                              
                                                                       
  Suggested fix                                                                                                                                                                                                   
                                                                                                                                                                                                                
  Something that states the facts and drops the guesswork:                                                                                                                                                        
  
  if not preview_response.suggested_columns:                                                                                                                                                                      
      raise ValueError(                                                                                                                                                                                         
          f"CSV preview for file handle {file_handle_id} returned no suggested "                                                                                                                                  
          f"columns (rows scanned: {preview_response.rows_scanned}). The file may "                                                                                                                               
          f"be empty, contain only a header row, or use a separator different "                                                                                                                                   
          f"from the configured csv_table_descriptor "                                                                                                                                                            
          f"(separator={effective_descriptor.separator!r})."                                                                                                                                                      
      )                                                                                                                                                                                                           
                                                                                                                                                                                                                  
  Benefits:                                                                                                                                                                                                       
                                                                                                                                                                                                                
  - Names the file_handle_id so the user can paste it into the Synapse UI or another script without hunting.                                                                                                      
  - Surfaces rows_scanned — "scanned 0 rows" vs "scanned 500 rows and still inferred nothing" are very different debugging stories.
  - Enumerates the realistic causes instead of the wrong one (bad file handle id).                                                                                                                                
  - Echoes the descriptor's separator, which is usually the actual culprit when a CSV "has data but no columns are detected."  
 ```

**Review by andrewelamb — CHANGES_REQUESTED** [review]

File: synapseclient/models/curation.py (hunk 23 above)
Comment (linglp) [reply synapseclient/models/curation.py]: I think this is not a bug because CsvTableDescriptor() will always have the same defaults. But I could make it cleaner like you suggested.

File: tests/integration/synapseclient/models/async/test_grid_async.py (hunk 22 above)
Comment (linglp) [reply tests/integration/synapseclient/models/async/test_grid_async.py:215]: I could fix this.

File: synapseclient/models/curation.py (hunk 23 above)
Comment (andrewelamb) [reply synapseclient/models/curation.py]: You are probably correct!

**Comment by andrewelamb** [comment]
LGTM!

**Review by andrewelamb — APPROVED** [review]

File: synapseclient/models/curation.py [hunk 29]
```diff
@@ -1427,6 +1582,53 @@ def list(
...
+            ```python
+            from synapseclient import Synapse
+            from synapseclient.models import Grid
+
+            syn = Synapse()
+            syn.login()
+
+            grid = Grid(session_id="abc-123-def")
+            grid = grid.import_csv(file_handle_id="123456")
+            print(f"Import complete for session: {grid.session_id}")
```
thomasyu888, line 1629: @linglp quick question - what is the exact workflow for when someone wants to upload a CSV into the grid?

linglp: The import_csv function only takes `file_handle_id` instead of a local path, and someone would first need to upload the file to Synapse and then get the file handle id likes this: 
```
file = File(id="syn1234", download_file=False).get()
file_handle_id = file.file_handle.id
```
Is this okay? I was thinking of: 
```
grid.import_csv(local_path="file/to/local/path")
```
But then we would still need to upload the file to Synapse somehow and get `file_handle_id`

thomasyu888: Thanks - I think the building blocks of this can be a file_handle_id, and I think we'll get push back if the parameter isn't "path".

I think there is a "upload_file_handle" function: https://python-docs.synapse.org/en/stable/reference/core/?h=upload_file_handle#synapseclient.core.upload.upload_functions.upload_file_handle. could we look into that?

linglp: Sure. I could look into that

thomasyu888: Thanks - I should have caught that sooner. It could be helpful to figure out what the workflow is on the web client side. That upload_file_handle function requires a "parent_entity" parameter which is a bit odd when you're just uploading a file into the "grid"

linglp: I looked into the web client, and it seems like that they would upload the file to a synapse S3 bucket: take a look at the chain of code here: 
* when uploading csv, triggers `uploadStepContent`. See [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/table/CsvPreview/CsvPreviewDialog.tsx#L136)
* and then `BasicFileHandleUpload` receives `fileHandleId` during a callback. 
* and I could see [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/file/upload/BasicFileHandleUpload.tsx#L35) that `BasicFileHandleUpload` is for uploading a file to the default Synapse S3 bucket

linglp: I could try: upload_synapse_s3 [here](https://synapsepythonclient--1360.org.readthedocs.build/en/1360/reference/core/?h=upload_synapse_s3#synapseclient.core.upload.upload_functions_async.upload_synapse_s3)

thomasyu888 (approved):
🔥 LGTM!  This is great!

PR #1361, opened 2026-04-17
Add validate-release workflow

xschildw:
# **Problem:**

- We used to run a Jenkins job to validate the python client against the staging repo and prevent releasing
   breaking changes. The Jenkins server has been sunset, I ran integration tests from my laptop but need
   a permanent solution.

# **Solution:**

- I would like to create dispatch workflow in the SynapsePython client repo to run a single pass integration test.

# **Testing:**

- None, I just trimmed an existing workflow. Considered testing in my fork but that means adding secrets etc., since
  this is a dispatch workflow it should have no effect on existing operations and can be debugged in the SageBio repo.

### DIFF
--- a/.github/workflows/validate-release.yml
+++ b/.github/workflows/validate-release.yml
@@ -1,0 +1,39 @@
+# GitHub Action workflow for validating releases against the Synapse backendname: validate-release
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  validate-release:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.14'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[boto3,pandas,pysftp,tests,curator]"
+      - name: Run unit tests
+        run: pytest -sv tests/unit
+      - name: Check for Secret availability
+        id: secret-check
+        run: |
+          if [ -z "${{ secrets.encrypted_d17283647768_key }}" ]  || [ -z "${{ secrets.encrypted_d17283647768_iv }}" ]; then
+            echo "secrets_available=false" >> $GITHUB_OUTPUT;
+          else
+            echo "secrets_available=true" >> $GITHUB_OUTPUT;
+          fi
+      - name: Run integration tests (if secrets available)
+        if: steps.secret-check.outputs.secrets_available == 'true'
+        run: |
+          # decrypt the encrypted test synapse configuration
+          openssl aes-256-cbc -K ${{ secrets.encrypted_d17283647768_key }} -iv ${{ secrets.encrypted_d17283647768_iv }} -in test.synapseConfig.enc -out test.synapseConfig -d
+          mv test.synapseConfig ~/.synapseConfig
+          pytest -sv tests/integration
--- a/.github/workflows/validate-release.yml
+++ b/.github/workflows/validate-release.yml
@@ -27,12 +27,11 @@ jobs:
           if [ -z "${{ secrets.encrypted_d17283647768_key }}" ]  || [ -z "${{ secrets.encrypted_d17283647768_iv }}" ]; then
             echo "secrets_available=false" >> $GITHUB_OUTPUT;
             echo "Secrets for integration tests are not available. Cancelling integration tests.";
-            exit(2);
+            exit 1;
           else
             echo "secrets_available=true" >> $GITHUB_OUTPUT;
           fi
       - name: Run integration tests (if secrets available)
-        if: steps.secret-check.outputs.secrets_available == 'true'
         run: |
           # decrypt the encrypted test synapse configuration
           openssl aes-256-cbc -K ${{ secrets.encrypted_d17283647768_key }} -iv ${{ secrets.encrypted_d17283647768_iv }} -in test.synapseConfig.enc -out test.synapseConfig -d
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-release.yml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.14'
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[boto3,pandas,pysftp,tests,curator]"
+      - name: Run unit tests
```
linglp, line 23: I think the step of running unit tests is not necessary because all PRs should have unit tests passed before merging.

.github/workflows/validate-release.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: pytest -sv tests/unit
+      - name: Check for Secret availability
```
linglp, line 25: if the secrets are missing, should the workflow just fail? it seems like the workflow will still be able to complete here with missing secrets

.github/workflows/validate-release.yml
```diff
@@ -0,0 +1,39 @@
...
+  workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  validate-release:
+    runs-on: ubuntu-latest
+    timeout-minutes: 60
```
BryanFauble, line 13:
```suggestion
    timeout-minutes: 180
```


Tests can take a bit longer than 60 minutes in some cases

.github/workflows/validate-release.yml
```diff
@@ -0,0 +1,39 @@
+# GitHub Action workflow for validating releases against the Synapse backendname: validate-release
+
+on:
+  workflow_dispatch:
+
+concurrency:
+  group: ci-${{ github.ref }}
```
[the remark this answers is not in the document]
BryanFauble: +1 @xschildw This should be done

.github/workflows/validate-release.yml
```diff
@@ -0,0 +1,39 @@
...
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e ".[boto3,pandas,pysftp,tests,curator]"
+      - name: Run unit tests
+        run: pytest -sv tests/unit
+      - name: Check for Secret availability
+        id: secret-check
+        run: |
+          if [ -z "${{ secrets.encrypted_d17283647768_key }}" ]  || [ -z "${{ secrets.encrypted_d17283647768_iv }}" ]; then
+            echo "secrets_available=false" >> $GITHUB_OUTPUT;
```
BryanFauble, line 29: Fail this run on this conditional

.github/workflows/validate-release.yml
```diff
@@ -0,0 +1,39 @@
...
+          else
+            echo "secrets_available=true" >> $GITHUB_OUTPUT;
+          fi
+      - name: Run integration tests (if secrets available)
+        if: steps.secret-check.outputs.secrets_available == 'true'
+        run: |
+          # decrypt the encrypted test synapse configuration
+          openssl aes-256-cbc -K ${{ secrets.encrypted_d17283647768_key }} -iv ${{ secrets.encrypted_d17283647768_iv }} -in test.synapseConfig.enc -out test.synapseConfig -d
+          mv test.synapseConfig ~/.synapseConfig
+          pytest -sv tests/integration
```
BryanFauble, line 39:
```suggestion
          pytest -sv --reruns 3 tests/integration -n 8 --ignore=tests/integration/synapseclient/test_command_line_client.py --dist loadscope
```

BryanFauble (changes requested):
Some changes to make

.github/workflows/validate-release.yml
xschildw: Agree, removed

.github/workflows/validate-release.yml
xschildw: The tests would fail to connect to backend, rather exit early.

.github/workflows/validate-release.yml
xschildw: Added msg + exit

.github/workflows/validate-release.yml
```diff
@@ -0,0 +1,39 @@
...
+            exit 1;
+          else
+            echo "secrets_available=true" >> $GITHUB_OUTPUT;
+          fi
+      - name: Run integration tests (if secrets available)
+        run: |
+          # decrypt the encrypted test synapse configuration
+          openssl aes-256-cbc -K ${{ secrets.encrypted_d17283647768_key }} -iv ${{ secrets.encrypted_d17283647768_iv }} -in test.synapseConfig.enc -out test.synapseConfig -d
+          mv test.synapseConfig ~/.synapseConfig
+          pytest -sv --reruns 3 tests/integration -n 8 --ignore=tests/integration/synapseclient/test_command_line_client.py --dist loadscope
```
BryanFauble, line 39: Something that I am realizing here is that this is going to point at synapse DEV, not synapse STAGING.

In order for us to point this to synapse STAGING we will need to create an encrypted synapseConfig file with the endpoints pointing at STAGING. I can help out with the creation of this file if you can get me an access token for the account we want these integration tests to run as.

BryanFauble (changes requested):
The last thing is to get an account set up for staging to run the tests as. Once I have that I can help get that file into source so we can use it in this workflow.

.github/workflows/validate-release.yml
xschildw: Running against DEV is fine, the dev stack is running the same code as the staging stack (we created a long lasting dev stack to avoid running against staging, which is really a production stack in waiting).

BryanFauble (approved):

PR #1366, opened 2026-04-23
[SYNPY-1782] Download CSV from Grid session 

linglp:
# **Problem:**
Couldn't download csv from grid. 

# **Solution:**
* Calls `get_file_handle_presigned_url` to get a pre-signed URL 
* When file_name is not provided, name the file as `grid_<session_id>_timestamp`. This is to align with the front-end [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/DataGrid/hooks/useExportDataGridToCsv.ts#L31)
* Downloads via asyncio.to_thread(download_from_url, ...) to get MD5 validation, retries, and progress tracking — consistent with how `download_list_manifest_async` handles presigned URL downloads. This is to align with front-end behavior when exporting a csv. See [here](https://github.com/Sage-Bionetworks/synapse-web-monorepo/blob/main/packages/synapse-react-client/src/components/DataGrid/hooks/useExportDataGridToCsv.ts#L19-L22)

### DIFF
--- a/synapseclient/models/curation.py
+++ b/synapseclient/models/curation.py
@@ -1467,7 +1467,7 @@ class GridSynchronousProtocol(Protocol):
 
         Arguments:
             destination: Local directory path where the CSV will be saved.
-                If not provided, defaults to the current working directory.
+                If not provided, defaults to the current working directory. The directory must already exist.
             write_header: Whether the first line should contain column names
                 as a header. Defaults to True.
             include_row_id_and_row_version: Whether the first two columns
@@ -1997,7 +1997,7 @@ class Grid(GridSynchronousProtocol):
         then downloads the resulting CSV to the local filesystem.
 
         Arguments:
-            destination: Local directory path where the CSV will be saved.
+            destination: Local directory path where the CSV will be saved. The directory must already exist.
                 If not provided, defaults to the current working directory.
             write_header: Whether the first line should contain column names
                 as a header. Defaults to True.
--- a/synapseclient/models/curation.py
+++ b/synapseclient/models/curation.py
@@ -1078,7 +1078,7 @@ class DownloadFromGridRequest(AsynchronousCommunicator):
 
     This request is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/DownloadFromGridRequest.html>
 
-    The response is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/DownloadFromGridResponse.html>
+    The response is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/DownloadFromGridResult.html>
     """
 
     session_id: str
--- a/synapseclient/models/curation.py
+++ b/synapseclient/models/curation.py
@@ -5,6 +5,7 @@ Curation tasks are used to guide data contributors through the process of contri
 data or metadata in Synapse.
 """
 
+import os
 from dataclasses import dataclass, field, replace
 from typing import Any, AsyncGenerator, Dict, Generator, Optional, Protocol, Union
 
@@ -1449,7 +1450,7 @@ class GridSynchronousProtocol(Protocol):
     def download_csv(
         self,
         *,
-        destination: str = ".",
+        destination: Optional[str] = None,
         write_header: bool = True,
         include_row_id_and_row_version: bool = True,
         include_etag: bool = True,
@@ -1466,7 +1467,7 @@ class GridSynchronousProtocol(Protocol):
 
         Arguments:
             destination: Local directory path where the CSV will be saved.
-                Defaults to the current working directory.
+                If not provided, defaults to the current working directory.
             write_header: Whether the first line should contain column names
                 as a header. Defaults to True.
             include_row_id_and_row_version: Whether the first two columns
@@ -1979,8 +1980,8 @@ class Grid(GridSynchronousProtocol):
     )
     async def download_csv_async(
         self,
-        destination: str = ".",
         *,
+        destination: Optional[str] = None,
         write_header: bool = True,
         include_row_id_and_row_version: bool = True,
         include_etag: bool = True,
@@ -1997,7 +1998,7 @@ class Grid(GridSynchronousProtocol):
 
         Arguments:
             destination: Local directory path where the CSV will be saved.
-                Defaults to the current working directory.
+                If not provided, defaults to the current working directory.
             write_header: Whether the first line should contain column names
                 as a header. Defaults to True.
             include_row_id_and_row_version: Whether the first two columns
@@ -2044,9 +2045,13 @@ class Grid(GridSynchronousProtocol):
         if not self.session_id:
             raise ValueError("session_id is required to download a GridSession as CSV")
 
-        trace.get_current_span().set_attributes(
-            {"synapse.session_id": self.session_id or ""}
-        )
+        if not destination:
+            destination = os.getcwd()
+
+        if not os.path.isdir(destination):
+            raise ValueError(f"Destination {destination} is not a valid directory.")
+
+        trace.get_current_span().set_attributes({"synapse.session_id": self.session_id})
 
         effective_descriptor = csv_table_descriptor or CsvTableDescriptor()
         request = DownloadFromGridRequest(
--- a/synapseclient/models/curation.py
+++ b/synapseclient/models/curation.py
@@ -2334,9 +2334,9 @@ class Grid(GridSynchronousProtocol):
             write_header: Whether the first line should contain column names
                 as a header. Defaults to True.
             include_row_id_and_row_version: Whether the first two columns
-                should contain row ID and version. Defaults to True.
+                should contain row ID and version. Defaults to False.
             include_etag: Whether a column should contain the row etag.
-                Defaults to True.
+                Defaults to False.
             csv_table_descriptor: The description of the CSV format (delimiter,
                 quote character, etc.). If not provided, the default CSV format
                 will be used.
--- a/synapseclient/models/curation.py
+++ b/synapseclient/models/curation.py
@@ -2067,7 +2067,9 @@ class Grid(GridSynchronousProtocol):
         )
         if not download_response.results_file_handle_id:
             raise ValueError(
-                "Download job did not return a file handle ID for the CSV result"
+                f"Download job for grid session '{self.session_id}' completed but "
+                "did not return a file handle ID. The CSV result may be empty or "
+                "the job may have failed silently."
             )
 
         return await download_by_file_handle(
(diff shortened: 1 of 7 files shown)


synapseclient/models/curation.py
```diff
@@ -2153,3 +2287,108 @@ async def main():
...
+        if not download_response.results_file_handle_id:
+            raise ValueError(
+                f"Download job for grid session '{self.session_id}' completed but "
+                "did not return a file handle ID. The CSV result may be empty or "
+                "the job may have failed silently."
+            )
+
+        return await download_by_file_handle(
+            file_handle_id=download_response.results_file_handle_id,
+            synapse_id=download_response.results_file_handle_id,
```
andrewelamb: Does this work? The `file_handle_id` isn't a Synapse id right?

tests/integration/synapseclient/models/async/test_grid_async.py
```diff
@@ -245,3 +245,34 @@ async def test_import_csv_to_grid_session_async(
         finally:
             if created_grid is not None and created_grid.session_id:
                 await created_grid.delete_async(synapse_client=self.syn)
+
+    async def test_download_csv_async(self, record_set_fixture: RecordSet) -> None:
+        # GIVEN: Create a grid session first
+        grid = Grid(record_set_id=record_set_fixture.id)
```
andrewelamb, line 251: I'm not seeing any teardown for this Grid

tests/unit/synapseclient/models/async/unit_test_curation_async.py
```diff
@@ -1270,3 +1271,154 @@ def test_to_synapse_request(self) -> None:
...
+            # AND the result should be the file handle ID
+            assert result == "test.csv"
+
+    async def test_download_csv_async_with_invalid_dir(self):
+        # GIVEN a Grid with a session_id
+        grid = Grid(session_id=SESSION_ID)
+
+        # WHEN I call download_csv_async with an invalid destination
+        with pytest.raises(
+            ValueError, match=f"Destination ./nonexistent_dir is not a valid directory."
```
andrewelamb: nit: uneeded f-string

synapseclient/models/curation.py
```diff
@@ -1066,6 +1068,76 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+
+    This request is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/DownloadFromGridRequest.html>
+
+    The response is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/DownloadFromGridResponse.html>
+    """
+
+    session_id: str
+    """The grid session ID."""
+
+    concrete_type: str = field(default=DOWNLOAD_FROM_GRID_REQUEST)
```
andrewelamb: nit: doesn't need to be a field

synapseclient/models/curation.py
```diff
@@ -1066,6 +1068,76 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+    """The concrete type for this request."""
+
+    write_header: Optional[bool] = True
```
andrewelamb: Is this really optional? It seems like it should be either True or False?

BryanFauble (approved):
LGTM!

synapseclient/models/curation.py
```diff
@@ -1066,6 +1068,76 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+        """
+        Converts a response from the REST API into this dataclass.
+
+        Arguments:
+            synapse_response: The response from the REST API.
+
+        Returns:
+            The DownloadFromGridRequest object.
+        """
+        self.results_file_handle_id = synapse_response.get("resultsFileHandleId")
```
andrewelamb, line 1122: This returns None here where there is no `resultsFileHandleId`. It seems like you may want to raise an execption her instead of returning None?

synapseclient/models/curation.py
```diff
@@ -2153,3 +2287,108 @@ async def main():
...
+                f"Download job for grid session '{self.session_id}' completed but "
+                "did not return a file handle ID. The CSV result may be empty or "
+                "the job may have failed silently."
+            )
+
+        return await download_by_file_handle(
+            file_handle_id=download_response.results_file_handle_id,
+            synapse_id=download_response.results_file_handle_id,
+            entity_type="FileEntity",
+            destination=destination,
```
andrewelamb: The destination in this method is for a directory, but the function called here is expecting a file path to write to.

andrewelamb (changes requested):

synapseclient/models/curation.py
linglp: I think a missing file handle ID would only occur if the server returns a malformed response.

synapseclient/models/curation.py
linglp: I switched to download pre-signed URL instead of using `download_by_file_handle` to better align with the front-end.

tests/integration/synapseclient/models/async/test_grid_async.py
linglp: Just added tear down

synapseclient/models/curation.py
andrewelamb: Agreed, should we raise an exception in that case?

linglp: I didn't raise it in `fill_from_dict`, but I have it here. Is this sufficient? 
```
        download_response = await request.send_job_and_wait_async(
            timeout=timeout, synapse_client=synapse_client
        )
        if not download_response.results_file_handle_id:
            raise ValueError(
                f"Download job for grid session '{self.session_id}' completed but "
                "did not return a file handle ID. The CSV result may be empty or "
                "the job may have failed silently."
            )
```

synapseclient/models/curation.py
linglp: I removed `Optional`

synapseclient/models/curation.py
```diff
@@ -2153,3 +2306,143 @@ async def main():
...
+
+        Arguments:
+            destination: Local directory path where the CSV will be saved. The directory must already exist.
+                If not provided, defaults to the current working directory.
+            write_header: Whether the first line should contain column names
+                as a header. Defaults to True.
+            include_row_id_and_row_version: Whether the first two columns
+                should contain row ID and version. Defaults to True.
+            include_etag: Whether a column should contain the row etag.
+                Defaults to True.
```
andrewelamb: should be False

synapseclient/models/curation.py
```diff
@@ -2153,3 +2306,143 @@ async def main():
...
+        Submits a DownloadFromGridRequest async job, waits for it to complete,
+        then downloads the resulting CSV to the local filesystem.
+
+        Arguments:
+            destination: Local directory path where the CSV will be saved. The directory must already exist.
+                If not provided, defaults to the current working directory.
+            write_header: Whether the first line should contain column names
+                as a header. Defaults to True.
+            include_row_id_and_row_version: Whether the first two columns
+                should contain row ID and version. Defaults to True.
```
andrewelamb: should be False

synapseclient/models/curation.py
```diff
@@ -1530,6 +1606,83 @@ def delete(self, *, synapse_client: Optional[Synapse] = None) -> None:
...
+        then downloads the resulting CSV to the local filesystem.
+
+        Arguments:
+            destination: Local directory path where the CSV will be saved.
+                If not provided, defaults to the current working directory. The directory must already exist.
+            write_header: Whether the first line should contain column names
+                as a header. Defaults to True.
+            include_row_id_and_row_version: Whether the first two columns
+                should contain row ID and version. Defaults to True.
+            include_etag: Whether a column should contain the row etag.
```
andrewelamb, line 1634: should be false

synapseclient/models/curation.py
```diff
@@ -1530,6 +1606,83 @@ def delete(self, *, synapse_client: Optional[Synapse] = None) -> None:
...
+        Submits a DownloadFromGridRequest async job, waits for it to complete,
+        then downloads the resulting CSV to the local filesystem.
+
+        Arguments:
+            destination: Local directory path where the CSV will be saved.
+                If not provided, defaults to the current working directory. The directory must already exist.
+            write_header: Whether the first line should contain column names
+                as a header. Defaults to True.
+            include_row_id_and_row_version: Whether the first two columns
+                should contain row ID and version. Defaults to True.
```
andrewelamb: should be false

synapseclient/models/curation.py
```diff
@@ -2153,3 +2306,143 @@ async def main():
...
+            include_etag: Whether a column should contain the row etag.
+                Defaults to True.
+            csv_table_descriptor: The description of the CSV format (delimiter,
+                quote character, etc.). If not provided, the default CSV format
+                will be used.
+            file_name: The optional name for the downloaded file. If not
+                provided, Synapse will generate a name.
```
andrewelamb: I don't think this is quite right, we generate the name locally, not from Synapse correct?

synapseclient/models/curation.py
```diff
@@ -2153,3 +2306,143 @@ async def main():
...
+            )
+
+        file_handle = await get_file_handle(
+            file_handle_id=download_response.results_file_handle_id,
+            synapse_client=synapse_client,
+        )
+        presigned_url = await get_file_handle_presigned_url(
+            file_handle_id=download_response.results_file_handle_id,
+            synapse_client=synapse_client,
+        )
```
andrewelamb, line 2436: You could do these two in parallel:

```
                                                                                                                                                                                                             
  file_handle, presigned_url = await asyncio.gather(
      get_file_handle(                                                                                                                                                                                            
          file_handle_id=download_response.results_file_handle_id,
      ),                     
      get_file_handle_presigned_url(                                                                                                                                                                              
          file_handle_id=download_response.results_file_handle_id,
      ),                     
  )
```

tests/unit/synapseclient/models/async/unit_test_curation_async.py
```diff
@@ -1270,3 +1271,249 @@ def test_to_synapse_request(self) -> None:
...
+
+        mock_file_handle = {
+            "id": "172705303",
+            "etag": "mock-etag",
+            "createdBy": "3443707",
+            "createdOn": "2026-04-30T21:21:40.000Z",
+            "modifiedOn": "2026-04-30T21:21:40.000Z",
+            "concreteType": "org.sagebionetworks.repo.model.file.S3FileHandle",
+            "contentType": "text/csv",
+            "contentMd5": "mock-med5",
```
andrewelamb: nit: `mock-md5`

andrewelamb (approved):

andrewelamb:
@linglp I had a few minor suggestions, but LGTM, so I went ahead and approved.

synapseclient/models/curation.py
linglp: Sure. I could add `await asyncio.gather`

PR #1372, opened 2026-04-27
[SYNPY-1826] fix integration test: test_download_list_manifest_with_custom_csv_descriptor

linglp:
# **Problem:**
Integration test failure

# **Solution:**
* Since workers are working concurrently, other tests could add items to the cart. Code that tests the first line of the downloaded manifest can be pretty fragile. Changed it to search for the whole manifest instead of just the first line

### DIFF
--- a/tests/integration/synapseclient/operations/async/test_download_list_operations_async.py
+++ b/tests/integration/synapseclient/operations/async/test_download_list_operations_async.py
@@ -20,7 +20,7 @@ import pytest_asyncio
 import synapseclient.core.utils as utils
 from synapseclient import Project, Synapse
 from synapseclient.core.exceptions import SynapseHTTPError
-from synapseclient.models import File, file
+from synapseclient.models import File
 from synapseclient.models.table_components import CsvTableDescriptor
 from synapseclient.operations import (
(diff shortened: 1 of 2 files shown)


tests/integration/synapseclient/operations/async/test_download_list_operations_async.py
```diff
@@ -582,29 +582,37 @@ async def test_download_list_manifest_with_custom_csv_descriptor(
         assert "\t" in content, "Expected tab separators in manifest"
 
         # AND the escape character was used for the embedded quote in the file name
-        assert "/'" in content, (
```
linglp, line 585: After thinking about this more, this maybe a server side error. If Per the Synapse docs, escapeCharacter is "for escaping a separator or quote". So the spec says it should produce:'it/'s_test.csv' when the file name is: 'it's_test.csv'

linglp: See: https://sagebionetworks.jira.com/browse/PLFM-9598

danlu1 (approved):
LGTM. Thanks for working on this.

PR #1373, opened 2026-05-04
[SYNPY-1809] Added generate_sync_manifest method

andrewelamb:
# **Problem:**
The `generate_sync_manifest` method in `sync.py` worked with the old-style manifest.

# **Solution:**
A new version was created as a `StorableContainer` method that creates a new-style manifest.
The old version was deprecated.

# **Testing:**
Integration tests for the new method
Unit tests for the various helper functions


synapseclient/models/services/manifest.py
```diff
@@ -1528,3 +1530,375 @@ async def _resolve_local_file_provenance(
...
+    # input-order guarantee.
+    # Pair-tagged results remain correct under any completion order.
+    async def _store(dirname: str) -> tuple[str, Folder]:
+        folder = await Folder(name=dirname, parent_id=parent_id).store_async(
+            synapse_client=client
+        )
+        return dirname, folder
+
+    pairs = await asyncio.gather(*[_store(d) for d in dirnames])
+    return dict(pairs)
```
BryanFauble, line 1836: `asyncio.gather(*[_store(d) for d in dirnames])` fires every sibling folder store concurrently with no cap. For a directory with hundreds of sibling subfolders that's hundreds of parallel `Folder.store_async` calls in one go. Elsewhere in this mixin we bound fan-out with `max(syn.max_threads * 2, 1)` (see `synapseclient/models/mixins/storable_container.py` around line 510). Suggestion below wraps the store call in an `asyncio.Semaphore` using the same sizing:

```suggestion
    semaphore = asyncio.Semaphore(max(client.max_threads * 2, 1))

    async def _store(dirname: str) -> tuple[str, Folder]:
        async with semaphore:
            folder = await Folder(name=dirname, parent_id=parent_id).store_async(
                synapse_client=client
            )
            return dirname, folder

    pairs = await asyncio.gather(*[_store(d) for d in dirnames])
    return dict(pairs)
```

_Note: This comment was drafted with AI assistance and reviewed by me for accuracy._

BryanFauble (changes requested):
I have a concern around concurrency, otherwise the rest looks good to me

synapseclient/models/protocols/storable_container_protocol.py
```diff
@@ -300,3 +300,79 @@ def sync_to_synapse(
...
         return []
+
+    def generate_sync_manifest(
+        self: Self,
+        directory_path: str,
+        manifest_path: str,
+        *,
+        synapse_client: Synapse | None = None,
+    ) -> None:
+        """Walk a local directory, mirror its folder hierarchy under this
```
linglp, line 311: I think the docstring here is too verbose. The caller doesn't need to know so much about implementation details to call the function. Can we simplify the docstring here?

linglp (commented):
Hi @andrewelamb ! I think overall it looks good. 

While testing this function, I noticed that local files are always re-uploaded to Synapse even when unchanged based on the log. This may not be ideal for users with large manifests who only want to re-upload a subset. For example, users who have problems with only a subset of files might only want to reupload those and ignore all the others. I am curious if the current implementation would let us skip files that have the same MD5 checksums? If not, is it possible to implement a different solution that let us pull the files from Synapse first -> compare MD5 checksums -> upload if MD5 checksums are different

andrewelamb:
> Hi @andrewelamb ! I think overall it looks good.
> 
> While testing this function, I noticed that local files are always re-uploaded to Synapse even when unchanged based on the log. This may not be ideal for users with large manifests who only want to re-upload a subset. For example, users who have problems with only a subset of files might only want to reupload those and ignore all the others. I am curious if the current implementation would let us skip files that have the same MD5 checksums? If not, is it possible to implement a different solution that let us pull the files from Synapse first -> compare MD5 checksums -> upload if MD5 checksums are different

Thanks for the suggestion! I think this is out of scope for this ticket as I'm just replicating the old function, but using the new manifest style. I'll create a ticket though!

synapseutils/sync.py
```diff
@@ -1559,11 +1559,28 @@ def _check_size_each_file(df):
...
+        "compatible with the OOP Project.sync_to_synapse / "
+        "Folder.sync_to_synapse methods."
+    ),
+)
 def generate_sync_manifest(syn, directory_path, parent_id, manifest_path) -> None:
     """Generate manifest for [syncToSynapse][synapseutils.sync.syncToSynapse] from a local directory.
 
     [Read more about the manifest file format](../../explanations/manifest_tsv/)
 
+    .. deprecated:: 4.13.0
```
linglp: I don't think this can be rendered in the documentation

Image

linglp: I saw `syncToSynapse_async` is using the same docstring. Could you remove that too?

docs/tutorials/python/upload_data_in_bulk.md
```diff
@@ -156,14 +156,14 @@ navigated to the Files tab and selected the file that we added a Provenance reco
...
-{!docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py!}
+--8<-- "docs/tutorials/python/tutorial_scripts/upload_data_in_bulk.py"
 ```
 </details>
 
 ## References used in this tutorial
 
-- [synapseutils.generate_sync_manifest][]
+- [Project.generate_sync_manifest][synapseclient.models.mixins.StorableContainer.generate_sync_manifest]
```
linglp: Here, the link resolves to Folder -> `generate_sync_manifest` rather than project. Could you fix this?

docs/tutorials/python/upload_data_in_bulk.md
```diff
@@ -156,14 +156,14 @@ navigated to the Files tab and selected the file that we added a Provenance reco
...
 - [Project.sync_to_synapse][synapseclient.models.mixins.StorableContainer.sync_to_synapse]
```
linglp: same here. This points to folder too.

tests/integration/synapseclient/models/async/test_storable_container_async.py
```diff
@@ -396,3 +401,170 @@ async def test_error_column_rows_skipped(
...
+        """A fresh Folder under the worker-scoped project per test, so
+        assertions can reference the folder's full child state without
+        interference from sibling tests. A Folder is cheaper to create than
+        a Project while providing equivalent isolation for these tests.
+        """
+        folder = await Folder(
+            name=f"{request.node.name}_{uuid.uuid4()}",
+            parent_id=project_model.id,
+        ).store_async(synapse_client=syn)
+        schedule_for_cleanup(folder)
```
linglp: Is this correct? I am seeing: 
```
Don't know how to clean: Folder(id='syn74929220', name='test_flat_directory_uses_parent_id_2234b491-2608-4a78-80cb-c1d52f4526a8', parent_id='syn74929219', description=None, etag='fa9d1140-bd47-467c-8e68-3c590f3045a4', created_on='2026-05-13T16:31:42.955Z', modified_on='2026-05-13T16:31:42.955Z', created_by='3443707', modified_by='3443707', files=[], folders=[], tables=[], entityviews=[], submissionviews=[], datasets=[], datasetcollections=[], materializedviews=[], virtualtables=[], annotations={}) (type: Folder)
```

synapseclient/models/services/manifest.py
```diff
@@ -1528,3 +1530,378 @@ async def _resolve_local_file_provenance(
...
+
+    if not rows:
+        client.logger.warning(
+            f"No uploadable files found under {directory_path};"
+            " generated manifest contains only the header row."
+        )
+    _write_manifest_csv(manifest_path, rows)
+
+
+async def _resolve_and_validate_directory_path_async(directory_path: str) -> str:
```
linglp: should this function be async? since this function is not doing anything async?

linglp (changes requested):
@andrewelamb there are some small issues that still need to be fixed. I will approve after that!

synapseclient/models/services/manifest.py
andrewelamb: You're right!

linglp (approved):
LGTM! Thanks for all the fixes

BryanFauble (approved):
LGTM!

PR #1375, opened 2026-05-06
Add missing MY_PROXY_SECRET_KEY variable in tutorials

rxu17:
# **Problem:**
The [proxy storage location tutorial](https://python-docs.synapse.org/en/latest/tutorials/python/proxy_storage_location/#2-create-a-proxy-storage-location) and [storage location tutorials (proxy storage section)](https://python-docs.synapse.org/en/latest/tutorials/python/storage_location/#8-create-a-proxy-storage-location) were missing the definition for the `MY_PROXY_SECRET_KEY` variable prior to calling it later in `StorageLocation`

# **Solution:**
This PR adds the missing variable to the tutorials

# **Testing:**
`mkdocs serve` result

Proxy storage location tutorial
image

Storage locations tutorial
image

### DIFF
--- a/docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
+++ b/docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
@@ -24,14 +24,11 @@ my_project = Project(name="My uniquely named project about Alzheimer's Disease")
 # Replace with your proxy server URL and provide the shared secret key via the
 # MY_PROXY_SECRET_KEY environment variable.
 MY_PROXY_URL = "https://my-proxy-server.example.com"
-MY_PROXY_SECRET_KEY = os.environ.get(
-    "MY_PROXY_SECRET_KEY", "REPLACE_WITH_YOUR_PROXY_SECRET_KEY"
-)
+MY_PROXY_SECRET_KEY = os.environ.get("MY_PROXY_SECRET_KEY")
 
 # Replace with the path to a local file to register via the proxy
 FILE_PATH = "/path/to/your/file.csv"
 
-# --8<-- [start:create_proxy_storage_location]
 # Use this when a proxy server controls access to the underlying storage.
 my_proxy_folder = Folder(name="my-folder-for-proxy", parent_id=my_project.id)
 my_proxy_folder = my_proxy_folder.store()
--- a/docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
+++ b/docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
@@ -23,6 +23,7 @@ my_project = Project(name="My uniquely named project about Alzheimer's Disease")
 # --8<-- [start:create_proxy_storage_location]
 # Replace with your proxy server URL and shared secret key
 MY_PROXY_URL = "https://my-proxy-server.example.com"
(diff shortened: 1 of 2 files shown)


linglp (approved):
LGTM!!

docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
```diff
@@ -21,8 +21,12 @@
 # --8<-- [end:setup]
 
 # --8<-- [start:create_proxy_storage_location]
-# Replace with your proxy server URL and shared secret key
+# Replace with your proxy server URL and provide the shared secret key via the
+# MY_PROXY_SECRET_KEY environment variable.
 MY_PROXY_URL = "https://my-proxy-server.example.com"
+MY_PROXY_SECRET_KEY = os.environ.get(
+    "MY_PROXY_SECRET_KEY", "REPLACE_WITH_YOUR_PROXY_SECRET_KEY"
+)
```
andrewelamb: I'm not sure I'm a fan of this. If a user forgets to supply their key, the script will proceed with it as "REPLACE_WITH_YOUR_PROXY_SECRET_KEY".

docs/tutorials/python/tutorial_scripts/storage_location.py
```diff
@@ -138,11 +140,16 @@
 # Use this when a proxy server controls access to the underlying storage.
 # --8<-- [start:create_proxy_storage_location]
 
+# Replace with your proxy server URL and provide the shared secret key via the
+# MY_PROXY_SECRET_KEY environment variable.
 MY_PROXY_URL = "https://my-proxy-server.example.com"
+MY_PROXY_SECRET_KEY = os.environ.get(
+    "MY_PROXY_SECRET_KEY", "REPLACE_WITH_YOUR_PROXY_SECRET_KEY"
```
andrewelamb: see above comment.

docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
```diff
@@ -21,8 +21,12 @@
 # --8<-- [end:setup]
 
 # --8<-- [start:create_proxy_storage_location]
```
andrewelamb, line 23: I know this isn't part of your PR but this start marker doesn't have an end marker.

docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
rxu17: It would then fail since it wouldn't match the secret key that's from your actual proxy server? Similar thing to the default string for `MY_PROXY_URL` but I see what you mean. I think if we don't put anything, it defaults to `None` if it can't find the value.

andrewelamb: Right! I'm thinking it should fail faster if the user doesn't supply it.

docs/tutorials/python/tutorial_scripts/proxy_storage_location.py
rxu17: Nice catch, I just removed the extra start marker in the middle of this section

PR #1380, opened 2026-05-11
[SYNPY-1838] prepare function for removal of Grid creation

andrewelamb:
# **Problem:**
Creating the Grid class shouldn't be part of the `create_record_based_metadata_task` function.
The Grid is returned as part of a tuple, and removing this would be a breaking change.

# **Solution:**
- A warning about the upcoming breaking change was added.
- TODOs were added for v5.0.0


synapseclient/extensions/curator/record_based_metadata_task.py
```diff
@@ -212,6 +216,14 @@ def create_record_based_metadata_task(
             "The project ID will be inferred from the folder ID provided."
         )
 
+    # TODO: https://sagebionetworks.jira.com/browse/SYNPY-1838
+    # remove this warning
+    synapse_client.logger.warning(
+        "The Grid object will no longer be created starting in v5.0.0. "
```
thomasyu888: This is great v1 and let me suggest an edit from a user perspective. 

From a pure user perspective- it's not clear to me what the grid enables so the impact isn't clear.

synapseclient/extensions/curator/record_based_metadata_task.py
```diff
@@ -212,6 +216,16 @@ def create_record_based_metadata_task(
             "The project ID will be inferred from the folder ID provided."
         )
 
+    # TODO: https://sagebionetworks.jira.com/browse/SYNPY-1838
+    # remove this warning
+    synapse_client.logger.warning(
+        "A Grid object will no longer be created by this function starting in v5.0.0. "
+        "A Grid is used for curation workflows where data can be edited in a grid format and then exported back to record sets. "
+        "The Grid should be created from the RecordSet created by this function like: `grid = Grid(record_set_id=record_set.id).create()`. "
```
linglp: Follow up to @thomasyu888 's comment, I think here the warning message should mention something like: `The return type will change from Tuple[RecordSet, CurationTask, Grid] to Tuple[RecordSet, CurationTask]. `?

synapseclient/extensions/curator/record_based_metadata_task.py
```diff
@@ -114,12 +114,14 @@ def create_record_based_metadata_task(
...
     project_id: Optional[str] = None,  # Deprecated, will be removed in v5.0.0
+    # TODO: https://sagebionetworks.jira.com/browse/SYNPY-1838
+    # remove Grid here
 ) -> Tuple[RecordSet, CurationTask, Grid]:
     """
     This function:
         - Generates and uploads CSV templates as a RecordSet for record-based metadata
         - Creates a CurationTask
-        - Creates a Grid
+        - Creates a Grid (will be removed in v5.0.0)
```
linglp: A lot of changes will be part of 5.0.0.. and there's no release date for 5.0.0. Should we move this to an earlier release?

linglp (approved):
LGTM! See my comments below. I am a bit worried that this fix will be too late for users if this is part of v5.0.0 release. I am wondering if you have talked to Christina and if we should move it sooner.

andrewelamb:
Converting to draft, we may not actually do this.

synapseclient/extensions/curator/record_based_metadata_task.py
andrewelamb: Unfortunately, this is a breaking change to this function so it needs to be in a major release.

synapseclient/extensions/curator/record_based_metadata_task.py
```diff
@@ -114,12 +114,14 @@ def create_record_based_metadata_task(
     *,
     synapse_client: Optional[Synapse] = None,
     project_id: Optional[str] = None,  # Deprecated, will be removed in v5.0.0
```
BryanFauble, line 116: Would it make more sense to provide a boolean flag for users to opt into the new behavior (ie: Do NOT create a grid) so they can get ahead of the change?

```suggestion
    project_id: Optional[str] = None,  # Deprecated, will be removed in v5.0.0
    create_grid[bool] = True, # Deprecated, will be removed in v5.0.0
```

BryanFauble (changes requested):
Small change to update the function args

BryanFauble (approved):

PR #1383, opened 2026-05-18
[SYNPY-1760] Add ability to link Grid to CurationTask

andrewelamb:
# **Problem:**
- Python Client users  could not link a Grid session to a CurationTask
- Curator Extension Docs were out of date

# **Solution:**
Added these methods to `CurationTask`:
- `get_status`
- `update_status`
- `set_active_grid_session`
- `set_task_state_async`
- `create_grid_session_async`

Added `owner_principal_id` to `Grid`

Separated Curator docs into one for data managers and one for contributors

# **Testing:**
All new methods have unit and integration tests


docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
+# How to Enter and Update Metadata for a Curation Task
+
+This guide is for **data contributors** — someone who has been assigned a `CurationTask` and needs to fill in metadata against it (see [How to Set Up Metadata Curation Workflows](metadata_curation.md)). Your job is to open the work assigned to you, make your edits, and apply them so they can be validated.
```
cconrad8:
```suggestion
This guide shows how to programmatically complete file-based and record-based metadata curation tasks, including adding, editing, and validating metadata.
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+
+## What you'll accomplish
```
cconrad8:
```suggestion
## Overview
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+
+By following this guide, you will:
+
+- List the curation tasks in your project to see what's been assigned
```
cconrad8:
```suggestion
- List curation tasks in a Synapse project
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+- Attach to an existing Grid session for a task, or create one if none exists
+- Pull the grid contents down to a local CSV so you can edit in bulk
+- Upsert your edits back into the grid (record-based tasks)
+- Apply your changes so the administrator can validate them
+- Review the validation report for your submission so you can fix issues before handing the task back (record-based tasks)
+
+## Prerequisites
+
+- A Synapse account with access to the project containing the curation task
```
cconrad8:
```suggestion
- A Synapse account
- Completion of the certification quiz
- A minimum of **view** access on the Synapse project
- A minimum of **edit** access on the folder containing the record-set entity. 
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+df = pd.DataFrame(
+    np.random.randint(0, 100, size=(4, len(df.columns))),
+    columns=df.columns,
+)
+
+edited_path = "./grid_edited.csv"
+df.to_csv(edited_path, index=False)
+```
+
+### Step 5: Upsert your edits back into the grid
```
cconrad8: @andrewelamb the instructions for programmatically downloading and uploading should only be for record-sets right now.  For file-based metadata, is there a way to programmatically update the file view you are considering?

cconrad8 (changes requested):
adding some changes -- still reviewing

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
+# How to Enter and Update Metadata for a Curation Task
+
+This guide is for **data contributors** — someone who has been assigned a `CurationTask` and needs to fill in metadata against it (see [How to Set Up Metadata Curation Workflows](metadata_curation.md)). Your job is to open the work assigned to you, make your edits, and apply them so they can be validated.
+
+## What you'll accomplish
+
+By following this guide, you will:
+
+- List the curation tasks in your project to see what's been assigned
+- Attach to an existing Grid session for a task, or create one if none exists
```
cconrad8: I think the "Attach to an existing grid session or create one if none exists should be baked into the code and not surfaced.

cconrad8:
```suggestion
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+- Pull the grid contents down to a local CSV so you can edit in bulk
```
cconrad8:
```suggestion
- Download record-based metadata locally as a csv
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+- Upsert your edits back into the grid (record-based tasks)
```
cconrad8:
```suggestion
- Upload the record-based metadata into Synapse
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+- Apply your changes so the administrator can validate them
```
cconrad8:
```suggestion
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+- Review the validation report for your submission so you can fix issues before handing the task back (record-based tasks)
```
cconrad8:
```suggestion
- Generate a validation report on the updated metadata
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
+# How to Enter and Update Metadata for a Curation Task
+
+This guide shows how to programmatically complete file-based and record-based metadata curation tasks, including adding, editing, and validating metadata.
```
cconrad8:
```suggestion
This guide shows how to programmatically complete record-based metadata curation tasks, including adding, editing, and validating metadata.
```

cconrad8: After reading below, I think this guide should only focus on record based metadata.  Right now we don't have functionality to re-upload metadata to file based tasks so it would be helpful to distinguish between the two

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+By following this guide, you will:
+
+- List curation tasks in a Synapse project
+- Attach to an existing Grid session for a task, or create one if none exists
+- Pull the grid contents down to a local CSV so you can edit in bulk
+- Upsert your edits back into the grid (record-based tasks)
+- Apply your changes so the administrator can validate them
+- Review the validation report for your submission so you can fix issues before handing the task back (record-based tasks)
+
+## Prerequisites
```
cconrad8:
```suggestion
## Requirements
```

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: Could you elaborate?

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: I think for now we're not going to add EntityView instructions.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+        print(f"Active grid sessions: {len(grid_sessions)}. Getting latest grid session: {latest_grid.session_id}")
+    else:
+        print(f"No active grid session for {source_id}; creating one.")
+        latest_grid = Grid(initial_query=Query(sql=f"SELECT * FROM {source_id}"))
+        latest_grid.create()
+```
+
+> **Why prefer an existing session?** Grid sessions can be owned by a team (when the administrator set `assignee_principal_id` to a team ID), so teammates can share progress on the same session instead of forking divergent copies.
+
+### Step 4: Pull the grid down as a CSV
```
cconrad8:
```suggestion
### Step 4: Download record-based metadata as a local CSV
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+
+Open the CSV, make your edits, and save it back to a local path. For example, with pandas:
+
+```python
+import pandas as pd
+import numpy as np
+
+df = pd.read_csv(csv_path)
+print(df)
+# The grid you downloaded may be empty — the columns shown here will match
```
cconrad8:
```suggestion
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+# whatever the curation administrator configured for the task's schema.
```
cconrad8:
```suggestion

```

docs/guides/extensions/curator/metadata_contribution.md
cconrad8: I think we can move any Q&A about schemas to a different section.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Add 4 rows of placeholder numerical values using whatever columns the grid has.
```
cconrad8: I think this particular example of editing a grid should be moved to a different place.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+df = pd.DataFrame(
+    np.random.randint(0, 100, size=(4, len(df.columns))),
+    columns=df.columns,
+)
+
+edited_path = "./grid_edited.csv"
+df.to_csv(edited_path, index=False)
+```
+
+### Step 5: Upsert your edits back into the grid
```
cconrad8:
```suggestion
### Step 5: Import edited record-based metadata to Synapse
```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+
+`import_csv` upserts rows into the grid based on the `upsert_keys` the administrator configured on the curation task. Existing rows matching on those keys are updated; new rows are inserted.
```
cconrad8: When downloading a record-based metadata. It would be helpful to tell the user what the upsert key and required columns are.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+Applying the changes is what triggers schema validation and makes your edits visible to administrators and other contributors.
+
+- For **record-based** tasks, export the grid back to the RecordSet. This creates a new version of the RecordSet and generates the validation report that administrators can retrieve via [RecordSet.get_detailed_validation_results][synapseclient.models.RecordSet.get_detailed_validation_results].
+
+    ```python
+    latest_grid.export_to_record_set()
+    print(f"Exported to RecordSet version: {latest_grid.record_set_version_number}")
+    ```
+
+- For **file-based** tasks, synchronize the grid against the file view to push annotation changes back to the underlying files.
```
cconrad8: I think file-based should be it's own section

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+
+    ```python
+    latest_grid.synchronize()
+    ```
+
+Once your changes are applied and you're done with the session, you can clean it up:
```
cconrad8:
```suggestion

```

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+
+```python
+latest_grid.delete()
```
cconrad8:
```suggestion
```

docs/guides/extensions/curator/metadata_contribution.md
cconrad8: Move the delete to a new section like an appendix.  The delete functionality isn't a step that would fall right after this (the data manager for example might want to access the same session again)

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+```python
+latest_grid = latest_grid.import_csv(path=edited_path)
+print(f"Upserted edits into grid session: https://www.synapse.org/Grid:default?sessionId={latest_grid.session_id}")
+```
+
+### Step 6: Apply the changes to the source entity
+
+Applying the changes is what triggers schema validation and makes your edits visible to administrators and other contributors.
+
+- For **record-based** tasks, export the grid back to the RecordSet. This creates a new version of the RecordSet and generates the validation report that administrators can retrieve via [RecordSet.get_detailed_validation_results][synapseclient.models.RecordSet.get_detailed_validation_results].
```
cconrad8: Is the validation report only run after export_to_record_set()? 

So is there any way to get the validation report when it's still a draft?

@andrewelamb

cconrad8 (changes requested):

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: I think this would need to be it's own feature.

synapseclient/models/curation.py
```diff
@@ -848,6 +1289,313 @@ async def main():
...
+        return await self.update_status_async(
+            curation_task_status=status, synapse_client=synapse_client
+        )
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: (
+            f"CurationTask_CreateGridSession: ID: {self.task_id}"
+        )
+    )
+    async def create_grid_session_async(
```
thomasyu888, line 1638: From a product perspective (and I'd defer to @cconrad8 to step in on overall flow), having this be a "create only" function means we will definitely need a function to list all grid sessions associated to a task. 

Although, I wasn't 100% certain about the future of grid sessions and whether we are trying to trend towards a world where there is one grid session per task ever

thomasyu888 (commented):
At a very high level, this makes sense, and I'd like to defer to @linglp and @BryanFauble for the thorough review.

If needed, I encourage getting together to talk through the additions so everyone is aligned.

synapseclient/extensions/curator/readme.md
```diff
@@ -1,6 +1,6 @@
 # Synapse Curator Extension - Developer Guide
 
-This document describes the design principles and architecture of the `synapseclient.extensions.curator` module. For user-facing documentation, see `metadata_curation.md` in the docs folder.
+This document describes the design principles and architecture of the `synapseclient.extensions.curator` module. For user-facing documentation, see `docs/guides/extensions/curator/curation.md` (administrators) and `docs/guides/extensions/curator/contribution.md` (contributors).
```
linglp: The file paths here are not accurate anymore.

tests/integration/synapseclient/models/async/test_curation_async.py
```diff
@@ -13,14 +13,44 @@
...
     Project,
     RecordBasedMetadataTaskProperties,
     RecordSet,
     ViewTypeMask,
 )
+from synapseclient.models.table_components import Query
+from tests.integration import ASYNC_JOB_TIMEOUT_SEC
+
+
+@pytest.fixture(scope="function")
```
linglp, line 32: When I reviewed the test, I noticed that there are many `folder_with_view` fixtures in this file. If we are going to have a function-level fixture at the top of the file like this, at the very least, the same `folder_with_view` fixture under class `TestCurationTaskDeleteAsync` should be deleted.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,245 @@
...
+If the Grid was never exported (Step 6), there is nothing to review yet.
+
+#### Retrieve and inspect the results
+
+Validation results live on the RecordSet itself, so you can retrieve them whether or not the Grid session is still open. Use the record_set_id from your CurationTask, re-fetch the RecordSet to pick up the latest validation_file_handle_id, and pull the detailed report as a pandas DataFrame:
+
+```python
+from synapseclient.models import RecordBasedMetadataTaskProperties, RecordSet
+
+if isinstance(curation_task.task_properties, RecordBasedMetadataTaskProperties):
```
linglp, line 182: I wonder if there should be a notice for people who set up their tasks using legacy code. When I tried out this part, my `curation_task` was set up maybe a while ago and `curation_task.task_properties` returned None

docs/reference/experimental/sync/curator.md
```diff
@@ -15,6 +15,8 @@ at your own risk.
             - delete
             - store
             - list
+            - create_grid_session
```
linglp, line 21: Missing `get_status`, `update_status`, `set_active_grid_session`. Same for async

synapseclient/models/curation.py
```diff
@@ -848,6 +1289,313 @@ async def main():
...
+        self,
+        *,
+        owner_principal_id: int | None = None,
+        timeout: int = 120,
+        synapse_client: Synapse | None = None,
+    ) -> "Grid":
+        """
```
linglp, line 1645: I think the docstring here is a bit misleading. It should be:  Create a new Grid session for this CurationTask and set it as the active session?

synapseclient/models/curation.py
```diff
@@ -931,6 +1682,10 @@ class CreateGridRequest(AsynchronousCommunicator):
     """Initialize a grid session from an EntityView.
     Mutually exclusive with record_set_id."""
 
+    owner_principal_id: int | None = None
```
linglp, line 1979: Not sure if this is intentional, but the description of the docstring here does not match synapse documentation: https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/CreateGridRequest.html

synapseclient/models/curation.py
```diff
@@ -168,6 +183,161 @@ def _create_task_properties_from_dict(
...
+
+    _ENUM_FIELDS: ClassVar[dict[str, type]] = {"state": CurationTaskState}
+
+    task_id: int | None = None
+    """The unique identifier of the associated curation task."""
+
+    state: str | CurationTaskState | None = None
+    """The state of a curation task in its lifecycle."""
+
+    execution_details: GridExecutionDetails | None = None
```
linglp: The API defines this as [TaskExecutionDetails](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/curation/TaskExecutionDetails.html)

linglp (changes requested):
Hi @andrewelamb ! Thanks for your hard work. I understand most of the changes. I am curious do we want to in general direct users to use `CurationTask.create_grid_session()` instead of `Grid.create() `? If so, can we also clarify that in the docstring? I think that currently, the docstring doesn't explain why we should use this over `Grid.create() `

synapseclient/models/curation.py
andrewelamb: Yeah I was trying to shorten it, but you're right its lost a lot of detail.

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: I added an exception when the Synapse response doesn't have TaskProperties, and a note to the dcos here about what to do when this happens.

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: @cconrad8 Could you elaborate on this?

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: I'll do this. 

Whose job should it be to delete the grid? Should it be part of cleaning up the CurationTask?
Should the contributor do it when they set the status of the Task to be completed/cancelled?

linglp (approved):
LGTM! Thanks for the fixes!

synapseclient/models/curation.py
```diff
@@ -957,6 +1433,341 @@ async def main():
...
+            ```
+        """
+        if not self.task_id:
+            raise ValueError(
+                "task_id is required to create a CurationTask grid session"
+            )
+
+        if not self.task_properties:
+            await self.get_async(synapse_client=synapse_client)
+
```
andrewelamb, line 1705: Should this always do `get_async` to ensure the CurationTask exists?

synapseclient/models/curation.py
andrewelamb, line 2937: Should we add a note that most users should use CurationTask.create_grid_session to create Grids instead of this class?

docs/guides/extensions/curator/metadata_contribution.md
thomasyu888, line 1: For christina to review

docs/guides/extensions/curator/metadata_curation.md
thomasyu888, line 1: For christina to review

synapseclient/models/curation.py
```diff
@@ -660,10 +1084,13 @@ def fill_from_dict(
...
 
         task_properties_dict = synapse_response.get("taskProperties", None)
-        if task_properties_dict:
-            self.task_properties = _create_task_properties_from_dict(
-                task_properties_dict
+        if task_properties_dict is None:
+            raise ValueError(
+                "taskProperties was not found in the Synapse response for this CurationTask. "
+                "This means it is likely an older CurationTask from before taskProperties was added. "
+                "It is recommended that this task be deleted: task.delete(delete_source=False)"
```
thomasyu888: Add to the message to recreate the task with the existing source entity

synapseclient/models/curation.py
thomasyu888: always get task

synapseclient/models/curation.py
```diff
@@ -957,6 +1433,341 @@ async def main():
...
+        grid = await grid.create_async(
+            timeout=timeout,
+            synapse_client=synapse_client,
+        )
+
+        try:
+            await self.set_active_grid_session_async(
+                active_session_id=grid.session_id, synapse_client=synapse_client
+            )
+        except Exception:
```
thomasyu888, line 1763: This should probably be the one that catches the 412 error.  Should we have a "finally" clause to delete the grid sesison for all other errors on this active grid session command?

synapseclient/models/curation.py
```diff
@@ -957,6 +1433,341 @@ async def main():
...
+            try:
+                await grid.delete_async(synapse_client=synapse_client)
+            except Exception:
+                Synapse.get_client(synapse_client=synapse_client).logger.warning(
+                    "Failed to delete orphan grid session %s after status "
+                    "update failure; manual cleanup may be required.",
+                    grid.session_id,
+                )
+            raise
```
thomasyu888, line 1772: Add a comment

1. There can only be one grid session that is active, but there can be multiple grid sessions.  So multiple users running this command may run into the 412 error.

thomasyu888 (approved):
🔥 LGTM! just some minor comments, and will defer to @BryanFauble for a final pass next sprint for the code and @cconrad8 on the documentation.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,295 @@
...
+curation_task = CurationTask(task_id=12345).get()
+```
+
+### Step 3: Create a Grid session for the task
+
+Ask the task to create a new Grid session — it picks the `record_set_id` from the task properties automatically and links the session back to the task. If the task already has an active session linked, calling this replaces the link with the new session.
+
+```python
+# Use whichever task you found in Step 2, for example:
+# curation_task = all_tasks[0]               # from Option A (list all)
```
cconrad8: is this using 0 because the user should be selecting the value from the list?

cconrad8: I would suggest being more consistent where regardless of how the user acquires the task id, that is the input to the next step.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -0,0 +1,295 @@
...
+
+```python
+from synapseclient.models import CurationTask
+
+curation_task = CurationTask(task_id=12345).get()
+```
+
+### Step 3: Create a Grid session for the task
+
+Ask the task to create a new Grid session — it picks the `record_set_id` from the task properties automatically and links the session back to the task. If the task already has an active session linked, calling this replaces the link with the new session.
```
cconrad8: The wording "Ask the task to create a new Grid session" is confusing.  

I would rephrase to "Open" or "Start" a grid session.

docs/guides/extensions/curator/metadata_contribution.md
cconrad8: I think the task has the upsert key as a property?

cconrad8 (changes requested):

synapseclient/models/curation.py
```diff
@@ -193,6 +208,179 @@ def _create_task_properties_from_dict(
         )
 
 
+@dataclass
+class TaskExecutionDetails(ABC):
```
BryanFauble, line 212: Although Synapse is representing this with inheritance and a base class, we do not have to follow this pattern exactly. In fact this would be the first usage of an abstract base class in the python code. Ideally - If possible I would rather this not exist.

BryanFauble: If there is a case to be said that there are going to be many more of these, then that would be a reason to keep this. If this is all we will have, then I do not think we need to have this extra layer.

synapseclient/models/curation.py
```diff
@@ -193,6 +208,179 @@ def _create_task_properties_from_dict(
...
+        """
+        Converts a response from the REST API into this dataclass.
+
+        Arguments:
+            synapse_response: The response from the REST API.
+
+        Returns:
+            The CurationTaskStatus object.
+        """
+        task_id_value = synapse_response.get("taskId", None)
```
BryanFauble: nit. `.get()` on a Dict defaults to None for the second argument. No need to set `None` there.

synapseclient/models/curation.py
```diff
@@ -957,6 +1434,349 @@ async def main():
...
+        Example: Update the status of a curation task asynchronously
+            &nbsp;
+
+            ```python
+            import asyncio
+            from synapseclient import Synapse
+            from synapseclient.models import (
+                CurationTask,
+                TaskState,
+                CurationTaskStatus,
```
BryanFauble: nit: Imported but not used.

synapseclient/models/curation.py
```diff
@@ -957,6 +1434,349 @@ async def main():
...
+                )
+            from synapseclient.models import EntityView
+
+            # raises SynapseHTTPError if EntityView does not exist
+            await EntityView(id=self.task_properties.file_view_id).get_async(
+                synapse_client=synapse_client
+            )
+            grid = Grid(
+                initial_query=Query(
+                    sql=f"SELECT * FROM {self.task_properties.file_view_id}"
```
BryanFauble, line 1736: Are there cases where this should be user controlled to set the query to something like, like `SELECT col_1, col_2, col_3 ...`?

thomasyu888: From a product perspective, this is not supported within the UI flow when people "Open Curator" from a task.  opening curator from a task will always `select *` on the underlying source id (recordset and/or fileview)

synapseclient/models/curation.py
```diff
@@ -957,6 +1434,349 @@ async def main():
...
+        # receive a 412 (precondition failed). In that case — or if recording the
+        # active session fails for any other reason — delete the session we just
+        # created so it doesn't become an orphan. If the delete also fails, log a
+        # warning so the caller knows manual cleanup is needed, then re-raise the
+        # original exception in all cases.
+        try:
+            await self.set_active_grid_session_async(
+                active_session_id=grid.session_id, synapse_client=synapse_client
+            )
+        except Exception:
```
BryanFauble, line 1763: Exception is way too wide to catch here for this logic. You need to catch for the specific exception (I assume an HTTP type exception), and inspect the message for the specific one that you care about. That should be the only case where you clean this up automatically.

BryanFauble (approved):
Thanks for the changes.

Approving, however please look to at least address https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1383#discussion_r3350771224 before merge.

docs/guides/extensions/curator/metadata_curation.md
```diff
@@ -163,10 +167,11 @@ record_set, curation_task, data_grid = create_record_based_metadata_task(
     instructions="Complete metadata for all study animals using StudyKey to link records to data files.",
     schema_uri=schema_uri,
     bind_schema_to_record_set=True,
-    assignee_principal_id="123456"  # Optional: Assign to a user or team
+    create_grid=False,
```
thomasyu888, line 170: I know we've gone back and forth a couple of times: did we want this to auto create a grid?

docs/guides/extensions/curator/metadata_curation.md
```diff
@@ -330,6 +335,7 @@ record_set, curation_task, data_grid = create_record_based_metadata_task(
     instructions="Enter metadata for each animal. All required fields must be completed.",
     schema_uri=schema_uri,
     bind_schema_to_record_set=True,
+    create_grid=False,
```
thomasyu888, line 338: Same question here as above

synapseclient/models/curation.py
andrewelamb: John has mentioned that they plan on having more of these, but I don't have any firm timeline. I'm happy to keep or change this.

synapseclient/models/curation.py
andrewelamb: Wouldn't we want to clean up this grid everytime Synapse was unable to set it as the active grid on the CurationTask?

BryanFauble: does it make any sense for the Grid created during this process to stay under any circumstances? If no, then keep as is. If yes, change the behavior.

andrewelamb: I don't think so. In my opinion, the user would be trying to create a grid and attach it with one method. If the attachment didn't work, they would probably try to call this method again, which would leave the grid from the first try.

docs/guides/extensions/curator/metadata_curation.md
andrewelamb: No, the grid would have the permissions of the Data Manager, not the Data Contributor.

cconrad8 (approved):

PR #1384, opened 2026-05-21
[SYNPY-1841] Added fields for CurationTask search

andrewelamb:
# **Problem:**
New fields were added to the list curation tasks endpoint:
- task state
- assignee id
- assigned to me

# **Solution:**
These are now arguments for `CurationTask.list` method

# **Testing:**
Integration and unit tests added

### DIFF
--- a/synapseclient/api/curation_services.py
+++ b/synapseclient/api/curation_services.py
@@ -123,6 +123,9 @@ async def delete_curation_task(
 async def list_curation_tasks(
     project_id: str,
     *,
+    assigned_to_me: Optional[bool] = None,
+    assignee_ids: Optional[list[str]] = None,
+    state_filter: Optional[list[str]] = None,
     synapse_client: Optional["Synapse"] = None,
 ) -> AsyncGenerator[Dict[str, Any], None]:
     """
@@ -132,6 +135,14 @@ async def list_curation_tasks(
 
     Arguments:
         project_id: The synId of the project.
+        assigned_to_me: When True, only return tasks that are assigned to the
+            current user. When False or None, the filter is not applied.
+            Cannot be combined with assignee_ids when True. Defaults to None.
+        assignee_ids: Optional list of principal IDs (users or teams) to filter
+            tasks by assignee. Cannot be combined with assigned_to_me=True.
+            Defaults to None.
+        state_filter: Optional list of TaskState string values to filter tasks by
+            their current state. Defaults to None (all states returned).
         synapse_client: If not passed in and caching was not disabled by
             `Synapse.allow_client_caching(False)` this will use the last created
             instance from the Synapse class constructor.
@@ -143,7 +154,14 @@ async def list_curation_tasks(
 
     client = Synapse.get_client(synapse_client=synapse_client)
 
-    request_body = {"projectId": project_id}
+    request_body: Dict[str, Any] = {"projectId": project_id}
+    # Only send the flag when True; False/None both mean "no filter"
+    if assigned_to_me is True:
+        request_body["assignedToMe"] = True
+    if assignee_ids is not None:
+        request_body["assigneeIds"] = assignee_ids
+    if state_filter is not None:
+        request_body["stateFilter"] = state_filter
 
     async for item in rest_post_paginated_async(
         "/curation/task/list", body=request_body, synapse_client=client
--- a/synapseclient/api/curation_services.py
+++ b/synapseclient/api/curation_services.py
@@ -127,7 +127,7 @@ async def list_curation_tasks(
     assignee_ids: Optional[list[str]] = None,
     state_filter: Optional[list[str]] = None,
     synapse_client: Optional["Synapse"] = None,
-) -> AsyncGenerator[Dict[str, Any], None]:
+) -> AsyncGenerator[dict[str, Any], None]:
     """
     Generator to get a list of CurationTasks for a project.
 
@@ -155,7 +155,7 @@ async def list_curation_tasks(
     client = Synapse.get_client(synapse_client=synapse_client)
 
     request_body: Dict[str, Any] = {"projectId": project_id}
-    if assigned_to_me is True:
+    if assigned_to_me:
         request_body["assignedToMe"] = True
     if assignee_ids is not None:
         request_body["assigneeIds"] = assignee_ids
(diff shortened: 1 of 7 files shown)


synapseclient/api/curation_services.py
```diff
@@ -143,7 +154,13 @@ async def list_curation_tasks(
 
     client = Synapse.get_client(synapse_client=synapse_client)
 
-    request_body = {"projectId": project_id}
+    request_body: Dict[str, Any] = {"projectId": project_id}
+    if assigned_to_me is True:
```
thomasyu888: Nit: `if assigned_to_me`

synapseclient/models/curation.py
```diff
@@ -398,14 +445,91 @@ def list(
...
+                f"Invalid state_filter value {s!r}. Expected a TaskState or str."
+            )
+        try:
+            coerced.append(TaskState(s.upper()).value)
+        except ValueError as exc:
+            valid = [e.value for e in TaskState]
+            raise ValueError(
+                f"Invalid state_filter value {s!r}. Valid values are: {valid}"
+            ) from exc
+    return coerced
```
BryanFauble: Could we move over to the `EnumCoercionMixin` rather than re-rolling out own here in this curation.py script? The tradeoff is that we expect users to pass in the correct casing.

synapseclient/models/curation.py
```diff
@@ -889,15 +1036,81 @@ async def main():
...
+            syn.login()
+
+            async def main():
+                async for task in CurationTask.list_async(
+                    project_id="syn9876543",
+                    state_filter=[TaskState.IN_PROGRESS],
+                ):
+                    print(f"Task ID: {task.task_id}")
+                    print(f"Data Type: {task.data_type}")
+                    print("---")
```
BryanFauble, line 1064: Could we add in an example of passing in a string value for state_filter please?

And to the async/sync function docstrings.

synapseclient/models/curation.py
```diff
@@ -889,15 +1036,81 @@ async def main():
...
+            )
+        if assignee_ids == []:
+            raise ValueError(
+                "assignee_ids must not be empty. Pass None to return tasks for any assignee."
+            )
+        if assigned_to_me is True and assignee_ids is not None:
+            raise ValueError(
+                "assigned_to_me and assignee_ids are mutually exclusive "
+                "and cannot be used together."
+            )
```
BryanFauble, line 1106: Would be good to include the value of `assignee_ids` in the message so it is clear what the user did actually pass in.

synapseclient/models/curation.py
```diff
@@ -889,15 +1036,81 @@ async def main():
...
+
+        if state_filter is not None:
+            state_filter = _coerce_state_filter(state_filter)
```
BryanFauble: This would end up going away

tests/integration/synapseclient/models/async/test_curation_async.py
```diff
@@ -626,3 +627,58 @@ async def test_list_empty_project_async(self) -> None:
...
+
+        # WHEN I list tasks filtered to NOT_STARTED
+        listed_task_ids = [
+            t.task_id
+            async for t in CurationTask.list_async(
+                project_id=project_model.id,
+                state_filter=[TaskState.NOT_STARTED],
+                synapse_client=self.syn,
+            )
+        ]
```
BryanFauble, line 665: Would be good to validate the actual length of listed_task_ids matches expectation.

tests/integration/synapseclient/models/async/test_curation_async.py
```diff
@@ -626,3 +627,58 @@ async def test_list_empty_project_async(self) -> None:
...
+            t.task_id
+            async for t in CurationTask.list_async(
+                project_id=project_model.id,
+                state_filter=["not_started"],
+                synapse_client=self.syn,
+            )
+        ]
+
+        # THEN the task should appear (string was coerced to TaskState.NOT_STARTED)
+        assert task.task_id in listed_task_ids
```
BryanFauble: This would go away since we are not doing a case change for the user.

tests/unit/synapseclient/models/async/unit_test_curation_async.py
```diff
@@ -577,6 +579,145 @@ async def mock_list(*args, **kwargs):
                 results[1].task_properties, RecordBasedMetadataTaskProperties
             )
 
+    async def test_list_async_assigned_to_me_and_assignee_ids_raises(self) -> None:
+        # GIVEN both assigned_to_me and assignee_ids are provided
+        # WHEN I call list_async
+        # THEN it should raise ValueError on first iteration (generators are lazy)
+        with pytest.raises(ValueError, match="mutually exclusive"):
```
BryanFauble: Since we are going to include the value in the message we should assert in the match include it.

tests/unit/synapseclient/models/async/unit_test_curation_async.py
```diff
@@ -577,6 +579,145 @@ async def mock_list(*args, **kwargs):
...
+                assignee_ids=None,
+                state_filter=["NOT_STARTED", "IN_PROGRESS"],
+                synapse_client=self.syn,
+            )
+
+    @pytest.mark.parametrize(
+        "input_value",
+        ["NOT_STARTED", "not_started", "Not_Started"],
+    )
+    async def test_list_async_state_filter_coerces_string(
```
BryanFauble: Likely do not need this test.

BryanFauble (approved):
A few changes otherwise LGTM!

synapseclient/models/curation.py
andrewelamb: @BryanFauble  After looking at it, it seems EnumCoercionMixin works as a mixin to something like a dataclass, where the things to be coerced are fields. Here, they are just parameters to the list method.

BryanFauble: Got it. Then could you extract this functionality out so it can be re-used for other cases like this?

andrewelamb: Sure! Where's a good spot for it to live?

BryanFauble: Let's throw it in `synapseclient/core/utils.py`

BryanFauble (approved):

PR #1385, opened 2026-05-21
set higher requests version min

andrewelamb:
# **Problem:**

Versions of the `requests` package have security issues:

**CVE-2026-25645: Path Traversal (Temporary Files)**: This vulnerability affects the extract_zipped_paths() utility. It uses predictable filenames in system temporary directories, allowing local attackers to pre-create malicious files that the library might then load without validation.

Fix: Upgrade to version 2.33.0 or higher.

**CVE-2024-35195: Persistent Certificate Verification Bypass**: A logic flaw in Session objects where if the first request disabled SSL verification (verify=False), all subsequent requests to the same host would also ignore verification, even if explicitly re-enabled. This could leave connections vulnerable to man-in-the-middle attacks.

Fix: Upgrade to version 2.32.2 or higher.

**Credential Leakage (.netrc)**: Due to a URL parsing issue, certain maliciously crafted URLs could trick the library into sending .netrc credentials to a third-party server.

Fix: Upgrade to version 2.32.4 or higher.

**CVE-2023-32681**: Proxy-Authorization Header Leak: Requests would unintentionally forward Proxy-Authorization headers to destination servers during HTTPS redirects, potentially exposing sensitive proxy credentials to malicious actors.

Fix: Upgrade to version 2.31.0 or higher.

# **Solution:**
Requests version is locked to 2.34.0 or higher

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "1b0f62e133c9783da2ebfd109ba8a24a7ba1961c49743568cc139a02f98bf0ed"
+            "sha256": "c041ba44bdc1f862cf8d17232e8e23a6a478907c3d046dcace249137ad2c3bec"
         },
         "pipfile-spec": 6,
         "requires": {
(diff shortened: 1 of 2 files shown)


BryanFauble (approved):
LGTM!

PR #1387, opened 2026-05-22
[SYNPY-1638] Update the fact that the config is not generated upon installation

thomasyu888:
# **Problem:**

Documentation was wrong about ~/.synapseConfig files being created automatically

# **Solution:**
Fix documentation

# **Testing:**

I tested this behavior by using a docker image

```
docker run -ti python:3.14 bash

>>>
pip3 install synapseclient
ls -la
drwx------ 1 root root 4096 May 22 06:28 .
drwxr-xr-x 1 root root 4096 May 22 06:26 ..
-rw-r--r-- 1 root root  607 May  8 16:10 .bashrc
drwxr-xr-x 3 root root 4096 May 22 06:27 .cache
-rw-r--r-- 1 root root  132 May  8 16:10 .profile
-rw-r--r-- 1 root root  169 May 20 02:47 .wget-hsts
```


andrewelamb (approved):

BryanFauble (approved):
LGTM!

PR #1390, opened 2026-05-28
[SYNPY-1835] Fix query() failure when a list column is all-NA

BryanFauble:
## Problem

`query("select * from synXXXXXXXX")` raises:

```
TypeError: Invalid value '[]' for dtype 'Int64'
```

when the result CSV has a list column whose every value is empty/NA. The traceback bottoms out in `csv_to_pandas_df` (`synapseclient/models/mixins/table_components.py`):

```python
df.fillna({col: "[]"}, inplace=True)   # ← raises
df[col] = df[col].apply(json.loads)
```

`pandas.read_csv(...).convert_dtypes()` infers an all-empty CSV column as the nullable `Int64` dtype. `fillna({col: "[]"})` then tries to write the string `"[]"` into that typed column and pandas refuses.

This affects any user querying a table or view where, for the rows returned, a list column is entirely null — not just the originally reported `syn74257215`.

## Solution

Replace the fragile `fillna(..., "[]")` + `apply(json.loads)` two-step with a single `apply` that branches on the cell's actual type:

```python
df[col] = df[col].apply(
    lambda x: json.loads(x) if isinstance(x, str) else []
)
```

A CSV cell for a list column is either a JSON string (`"[1, 2]"`) or NA — by checking `isinstance(x, str)` we parse the string when present and substitute `[]` for everything else, sidestepping the dtype mismatch entirely. The downstream `INTEGER_LIST` / `BOOLEAN_LIST` / `STRING_LIST` / `USERID_LIST` / `ENTITYID_LIST` coercion blocks already guard with `isinstance(x, list)`, so an empty `[]` flows through them untouched.

## How the test covers the scenario

`tests/unit/synapseclient/mixins/unit_test_table_components.py` — added two tests in `TestCsvToPandasDf`:

1. **`test_csv_to_pandas_df_all_na_list_column`** — feeds a CSV with one list column whose value is empty in every row (the exact reproducer for the bug). Parameterized across every list column type (`INTEGER_LIST`, `BOOLEAN_LIST`, `STRING_LIST`, `USERID_LIST`, `ENTITYID_LIST`) plus the "no `list_column_types`" case. Asserts the column comes back as `[[], [], []]`.
2. **`test_csv_to_pandas_df_mixed_all_na_and_populated_list_columns`** — two list columns, one all-NA and one populated, to ensure the fix doesn't disturb the populated column.

The new tests reproduce the original `TypeError: Invalid value '[]' for dtype 'Int64'` against the unfixed code and pass after the fix; the rest of the file's 132 tests continue to pass.

## Test plan

- [x] New parameterized reproducer fails with the original `TypeError` against unfixed code
- [x] New tests pass after the fix
- [x] All 132 tests in `unit_test_table_components.py` pass

### DIFF
--- a/synapseclient/models/mixins/table_components.py
+++ b/synapseclient/models/mixins/table_components.py
@@ -4581,10 +4581,14 @@ def csv_to_pandas_df(
     # Turn list columns into lists and convert items to their proper types
     if list_columns:
         for col in list_columns:
-            # Fill NA values with empty lists, it must be a string for json.loads to work
-            # json.loads will convert null values in boolean list, string list to None.
-            df.fillna({col: "[]"}, inplace=True)
-            df[col] = df[col].apply(json.loads)
+            # A CSV cell for a list column is either a JSON string like "[1, 2]"
+            # or NA. When every value is NA, convert_dtypes() infers a typed
+            # dtype (e.g. Int64) into which the string "[]" cannot be written,
+            # so fillna({col: "[]"}) raises. Parse strings and substitute []
+            # for NA in a single pass.
+            df[col] = df[col].apply(
+                lambda x: json.loads(x) if isinstance(x, str) else []
+            )
             # Convert list items to their proper types based on column type
             if list_column_types and col in list_column_types:
                 column_type = list_column_types[col]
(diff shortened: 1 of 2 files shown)


andrewelamb (approved):

andrewelamb:
LGTM!

thomasyu888 (approved):
🔥 LGTM!

PR #1392, opened 2026-05-29
[SYNPY-1489] Added ability to set grid authorization type field to CurationTask

andrewelamb:
# **Problem:**

Curation grid sessions could only be created with the default authorization behavior, where the user who creates the session becomes its owner (`SESSION_OWNER`). There was no way to:

- Configure a `CurationTask` to recommend a particular authorization mode for the grid sessions opened against it.
- Create a `Grid` session with `SOURCE_BENEFACTOR` authorization, which lets anyone with EDIT permission on the source data collaborate in the same session without being added to a shared ownership team.

The affected code is the curation/grid model layer (`synapseclient/models/curation.py`) and the curator extension task helpers, which previously had no concept of an authorization mode.

# **Solution:**

Added support for the grid `authorizationMode` field throughout the curation/grid models:

- Introduced a new `AuthorizationMode` enum (`SESSION_OWNER`, `SOURCE_BENEFACTOR`) mirroring the Synapse REST `AuthorizationMode` type, exported from `synapseclient.models`.
- Added a `suggested_authorization_mode` field (plus the not-yet-active `collaborator_principal_ids`) to `FileBasedMetadataTaskProperties` and `RecordBasedMetadataTaskProperties`. 
- Added an `authorization_mode` field to `Grid` and `CreateGridRequest`, and wired it through `create_async` / `to_synapse_request` so it is sent at session-creation time and populated back from the response.
- When a `CurationTask` creates a new grid session, it now forwards `task_properties.suggested_authorization_mode` to the `Grid`.
- The curator extension task helpers (`file_based_metadata_task.py`, `record_based_metadata_task.py`) accept and pass through the suggested authorization mode.

Design notes / side effects:

- `to_synapse_request` for the task properties was switched to build a full dict and then strip `None` keys via `delete_none_keys`, instead of conditionally adding keys.
- `owner_principal_id` is now coerced to `int` when populated from API responses, fixing a type inconsistency.
- Documentation was updated in the metadata curation guide and the curator API reference pages.

# **Testing:**

- Added unit tests verifying that `CurationTask` passes the correct authorization parameters when creating a grid session, and that the task properties / grid serialize and deserialize `authorizationMode` correctly (`tests/unit/synapseclient/models/async/unit_test_curation_async.py`, `tests/unit/synapseclient/extensions/unit_test_curator.py`).
- Added integration tests covering the new authorization mode behavior for grids and curation tasks (`tests/integration/.../models/async/test_grid_async.py`, `test_curation_async.py`).

### DIFF
--- a/docs/guides/extensions/curator/metadata_curation.md
+++ b/docs/guides/extensions/curator/metadata_curation.md
@@ -139,7 +139,6 @@ Both [create_record_based_metadata_task][synapseclient.extensions.curator.create
 - `SOURCE_BENEFACTOR` extends access to anyone with `EDIT` rights on the source entity. Use it when curation should be open to all editors of the source.
 
 ```python
-from synapseclient.models import AuthorizationMode
 
 entity_view_id, task_id = create_file_based_metadata_task(
     synapse_client=syn,
--- a/docs/guides/extensions/curator/metadata_curation.md
+++ b/docs/guides/extensions/curator/metadata_curation.md
@@ -135,8 +135,8 @@ print(f"Created CurationTask: {task_id}")
 
 Both `create_record_based_metadata_task` and `create_file_based_metadata_task` accept an optional `suggested_authorization_mode` that tells clients how to scope access when a grid session is created for the task:
 
-- `AuthorizationMode.SESSION_OWNER` (the server default applied when the mode is omitted) limits access to the session owner and their team. Use it when curation should be restricted to a specific user or team.
-- `AuthorizationMode.SOURCE_BENEFACTOR` extends access to anyone with `EDIT` rights on the source entity. Use it when curation should be open to all editors of the source.
+- `SESSION_OWNER` (the server default applied when the mode is omitted) limits access to the session owner and their team. Use it when curation should be restricted to a specific user or team.
+- `SOURCE_BENEFACTOR` extends access to anyone with `EDIT` rights on the source entity. Use it when curation should be open to all editors of the source.
 
 ```python
 from synapseclient.models import AuthorizationMode
@@ -148,7 +148,7 @@ entity_view_id, task_id = create_file_based_metadata_task(
     instructions="Annotate each file with metadata according to the schema requirements.",
     entity_view_name="Animal Study Files View",
     schema_uri=schema_uri,
-    suggested_authorization_mode=AuthorizationMode.SOURCE_BENEFACTOR,
+    suggested_authorization_mode="SOURCE_BENEFACTOR",
 )
 ```
 
--- a/docs/guides/extensions/curator/metadata_curation.md
+++ b/docs/guides/extensions/curator/metadata_curation.md
@@ -133,7 +133,7 @@ print(f"Created CurationTask: {task_id}")
 
 ### Controlling who can access the grid session
 
-Both [create_record_based_metadata_task][synapseclient.extensions.curator.create_record_based_metadata_task] and [create_file_based_metadata_task][synapseclient.extensions.curator.create_file_based_metadata_task] accept an optional `suggested_authorization_mode` that tells clients how to scope access when a grid session is created for the task:
+Both [create_record_based_metadata_task][synapseclient.extensions.curator.create_record_based_metadata_task] and [create_file_based_metadata_task][synapseclient.extensions.curator.create_file_based_metadata_task] accept an optional `authorization_mode` that tells clients how to scope access when a grid session is created for the task:
 
 - `SESSION_OWNER` (the server default applied when the mode is omitted) limits access to the session owner and their team. Use it when curation should be restricted to a specific user or team.
 - `SOURCE_BENEFACTOR` extends access to anyone with `EDIT` rights on the source entity. Use it when curation should be open to all editors of the source.
@@ -148,14 +148,14 @@ entity_view_id, task_id = create_file_based_metadata_task(
     instructions="Annotate each file with metadata according to the schema requirements.",
     entity_view_name="Animal Study Files View",
     schema_uri=schema_uri,
-    suggested_authorization_mode="SOURCE_BENEFACTOR",
+    authorization_mode="SOURCE_BENEFACTOR",
 )
 ```
 
 When [CurationTask.create_grid_session][synapseclient.models.CurationTask.create_grid_session] is later called, it forwards this mode to the new grid session and the server uses it to determine access: `SESSION_OWNER` limits access to the session owner (the caller, or an explicit `owner_principal_id`) and their team, while `SOURCE_BENEFACTOR` lets the session inherit access from the source entity's benefactor.
 
 !!! warning "Changing the mode invalidates the active session"
-    Updating `suggested_authorization_mode` on an existing task causes the server to automatically clear `activeSessionId` on the task status. The previously active grid session is no longer linked to the task, so you must create a new grid session before curation can continue.
+    Updating `authorization_mode` on an existing task causes the server to automatically clear `activeSessionId` on the task status. The previously active grid session is no longer linked to the task, so you must create a new grid session before curation can continue.
 
 ## Complete example script
 
--- a/docs/guides/extensions/curator/metadata_curation.md
+++ b/docs/guides/extensions/curator/metadata_curation.md
@@ -133,7 +133,7 @@ print(f"Created CurationTask: {task_id}")
 
 ### Controlling who can access the grid session
 
-Both `create_record_based_metadata_task` and `create_file_based_metadata_task` accept an optional `suggested_authorization_mode` that tells clients how to scope access when a grid session is created for the task:
+Both [create_record_based_metadata_task][synapseclient.extensions.curator.create_record_based_metadata_task] and [create_file_based_metadata_task][synapseclient.extensions.curator.create_file_based_metadata_task] accept an optional `suggested_authorization_mode` that tells clients how to scope access when a grid session is created for the task:
 
 - `SESSION_OWNER` (the server default applied when the mode is omitted) limits access to the session owner and their team. Use it when curation should be restricted to a specific user or team.
 - `SOURCE_BENEFACTOR` extends access to anyone with `EDIT` rights on the source entity. Use it when curation should be open to all editors of the source.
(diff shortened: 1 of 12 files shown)


docs/guides/extensions/curator/metadata_curation.md
```diff
@@ -131,6 +131,32 @@ print(f"Created CurationTask: {task_id}")
 - Automatic schema binding to the folder for validation
 - Optional wiki attached to the folder
 
+### Controlling who can access the grid session
+
+Both `create_record_based_metadata_task` and `create_file_based_metadata_task` accept an optional `suggested_authorization_mode` that tells clients how to scope access when a grid session is created for the task:
```
BryanFauble: could both references to functions get hyperlinks to the API reference page?

synapseclient/extensions/curator/file_based_metadata_task.py
```diff
@@ -377,6 +380,19 @@ def create_file_based_metadata_task(
             ViewTypeMask.FILE. Additional types can be added using bitwise OR
             (e.g., ViewTypeMask.FILE | ViewTypeMask.DOCKER). Accepts either a
             ViewTypeMask enum member or its raw integer value.
+        suggested_authorization_mode: Recommends who is allowed to access the curation
```
BryanFauble: It seems really strange to me to add "suggested_" prefix to this. Suggestion indicates to me that it may or may not respect what I were to put here, where the rest API talked about this in absolutes https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/AuthorizationMode.html . I would suggest dropping the prefix.

andrewelamb: I agree with everything you say here. The reason for the prefix is that it's the name of the endpoint [parameter](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/SYNPY-1849/synapseclient/models/curation.py#L191).

But since this is a user-facing argument, I'm happy to drop it here, while keeping the prefix on the underlying classes to mirror the endpoint names.

docs/guides/extensions/curator/metadata_curation.md
```diff
@@ -131,6 +131,32 @@ print(f"Created CurationTask: {task_id}")
...
 
+### Controlling who can access the grid session
+
+Both [create_record_based_metadata_task][synapseclient.extensions.curator.create_record_based_metadata_task] and [create_file_based_metadata_task][synapseclient.extensions.curator.create_file_based_metadata_task] accept an optional `authorization_mode` that tells clients how to scope access when a grid session is created for the task:
+
+- `SESSION_OWNER` (the server default applied when the mode is omitted) limits access to the session owner and their team. Use it when curation should be restricted to a specific user or team.
+- `SOURCE_BENEFACTOR` extends access to anyone with `EDIT` rights on the source entity. Use it when curation should be open to all editors of the source.
+
+```python
+from synapseclient.models import AuthorizationMode
```
linglp: You don't need to import AuthorizationMode

synapseclient/models/curation.py
```diff
@@ -133,16 +184,23 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
-        request_dict = {"concreteType": FILE_BASED_METADATA_TASK_PROPERTIES}
-        if self.upload_folder_id is not None:
-            request_dict["uploadFolderId"] = self.upload_folder_id
-        if self.file_view_id is not None:
-            request_dict["fileViewId"] = self.file_view_id
+        request_dict = {
+            "concreteType": FILE_BASED_METADATA_TASK_PROPERTIES,
+            "uploadFolderId": self.upload_folder_id,
+            "fileViewId": self.file_view_id,
+            "suggestedAuthorizationMode": (
```
linglp, line 201: figure out if this part is actually needed.

synapseclient/models/curation.py
```diff
@@ -2051,6 +2179,11 @@ def to_synapse_request(self) -> Dict[str, Any]:
             self.initial_query.to_synapse_request() if self.initial_query else None
         )
         request_dict["ownerPrincipalId"] = self.owner_principal_id
+        request_dict["authorizationMode"] = (
```
linglp, line 2202: figure it out if the `if...None` is needed

tests/integration/synapseclient/models/async/test_grid_async.py
```diff
@@ -139,6 +141,33 @@ async def test_create_and_list_grid_sessions_async(
...
+            authorization_mode=AuthorizationMode.SOURCE_BENEFACTOR,
+        )
+
+        # WHEN: Creating a grid session
+        created_grid = await grid.create_async(
+            timeout=ASYNC_JOB_TIMEOUT_SEC, synapse_client=self.syn
+        )
+
+        # AND: The grid session is scheduled for cleanup
+        request.addfinalizer(
```
linglp: try to use schedule_for_cleanup for the grid

BryanFauble (approved):
Thanks!

synapseclient/models/curation.py
linglp, line 123: I think the docstring here also needs to be updated.

docs/guides/extensions/curator/metadata_curation.md
```diff
@@ -131,6 +131,32 @@ print(f"Created CurationTask: {task_id}")
...
+
+entity_view_id, task_id = create_file_based_metadata_task(
```
linglp, line 143: Based on the PR [here](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1410), we should be returning entity view and task object, and not id?

synapseclient/extensions/curator/file_based_metadata_task.py
```diff
@@ -351,8 +353,9 @@ def create_file_based_metadata_task(
             attach_wiki=False,
             entity_view_name="Biospecimen Metadata View",
             schema_uri="sage.schemas.v2571-amp.Biospecimen.schema-0.0.1",
-            assignee_principal_id=123456,  # Optional: Assign to a user or team (can be str or int)
-            view_type_mask=ViewTypeMask.FILE | ViewTypeMask.DOCKER,  # Optional: include additional entity types in the view
+            assignee_principal_id=123456, # Optional: Assign to a user or team (can be str or int)
+            view_type_mask=ViewTypeMask.FILE | ViewTypeMask.DOCKER, # Optional: include additional entity types in the view
+            authorization_mode=AuthorizationMode.SOURCE_BENEFACTOR,
```
linglp: Nit: other optional fields have comments `Optional:`. Should we also add a comment here and say this is optional?

synapseclient/models/curation.py
```diff
@@ -152,9 +210,33 @@ class RecordBasedMetadataTaskProperties:
         record_set_id: The synId of the RecordSet that will contain all record-based metadata
     """
 
+    _ENUM_FIELDS: ClassVar[dict[str, type]] = {
```
linglp, line 233: I see how it works now. If users pass in `authorization_mode="SOURCE_BENEFACTOR"` then, this would get converted to: `AuthorizationMode.SOURCE_BENEFACTOR`

synapseclient/models/curation.py
linglp: I took a closer look, and I think this part is not necessary. We could just do something like: `self.collaborator_principal_ids`

synapseclient/models/curation.py
```diff
@@ -104,12 +125,36 @@ class FileBasedMetadataTaskProperties:
...
+        "suggested_authorization_mode": AuthorizationMode
+    }
+
     upload_folder_id: Optional[str] = None
     """The synId of the folder where data files of this type are to be uploaded"""
 
     file_view_id: Optional[str] = None
     """The synId of the FileView that shows all data of this type"""
 
+    suggested_authorization_mode: Optional[Union[AuthorizationMode, str]] = None
```
linglp, line 148: Since `EnumCoercionMixin.__setattr__` coerces string to AuthorizationMode, I think here `suggested_authorization_mode` will just be AuthorizationMode

synapseclient/models/curation.py
```diff
@@ -152,9 +210,33 @@ class RecordBasedMetadataTaskProperties:
...
     """
 
+    _ENUM_FIELDS: ClassVar[dict[str, type]] = {
+        "suggested_authorization_mode": AuthorizationMode
+    }
+
     record_set_id: Optional[str] = None
     """The synId of the RecordSet that will contain all record-based metadata"""
 
+    suggested_authorization_mode: Optional[Union[AuthorizationMode, str]] = None
```
linglp, line 240: same as comment above. Here `suggested_authorization_mode` should just be `AuthorizationMode`.

synapseclient/models/curation.py
```diff
@@ -177,9 +265,17 @@ def to_synapse_request(self) -> Dict[str, Any]:
         Returns:
             A dictionary representation of this object for API requests.
         """
-        request_dict = {"concreteType": RECORD_BASED_METADATA_TASK_PROPERTIES}
-        if self.record_set_id is not None:
-            request_dict["recordSetId"] = self.record_set_id
+        request_dict = {
+            "concreteType": RECORD_BASED_METADATA_TASK_PROPERTIES,
+            "recordSetId": self.record_set_id,
+            "suggestedAuthorizationMode": (
```
linglp, line 291: same as above. I think here it can just be `self.suggested_authorization_mode`

linglp (approved):
LGTM! I could approve. I took another look after the review, and there are just some small comments.

synapseclient/models/curation.py
andrewelamb: Turns out this is needed, the mixin converts it from a string to an enum value when getting a Synapse response, but it DOES NOT convert it from an enum to a string when making a Synapse request.

synapseclient/models/curation.py
andrewelamb: It is needed, see above comment.

docs/guides/extensions/curator/metadata_curation.md
andrewelamb: I assume you mean [this PR](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1411)? This doc was updated as well in that PR.

synapseclient/models/curation.py
andrewelamb: This can be created by doing `FileBasedMetadataTaskProperties(suggested_authorization_mode="SESSION_OWNER")`, which you wouldn't be able to do if the type hint was changed to  `Optional[AuthorizationMode]`.

synapseclient/models/curation.py
andrewelamb: See reply above.

synapseclient/models/curation.py
linglp: Oh I see. Got it. Thanks.

PR #1395, opened 2026-06-05
Fix bad import of pandas

andrewelamb:
An import of pandas was added at the top lvel of a module. This doesn't break any tests, but it causes an import error when the check-build workflow does `synapse --version` with the test-pypi package.

This is fixed by removing the module-level import and moving it to the function.

### DIFF
--- a/synapseclient/models/mixins/table_components.py
+++ b/synapseclient/models/mixins/table_components.py
@@ -13,7 +13,6 @@ from datetime import datetime
 from io import BytesIO
 from typing import Any, Dict, List, Optional, Protocol, Tuple, Union
 
-import pandas as pd
 from tqdm import tqdm
 from tqdm.contrib.logging import logging_redirect_tqdm
 from typing_extensions import Self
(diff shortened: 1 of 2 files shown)


synapseclient/models/mixins/table_components.py
```diff
@@ -201,6 +200,7 @@ def convert_dtypes_to_json_serializable(df) -> pd.DataFrame:
         df = convert_dtypes_to_json_serializable(df)
         print(df)
     """
+    import pandas as pd
```
thomasyu888, line 204: There is a standard function that we are using to catch when pandas isn't installed.  @BryanFauble can you help identify said function?

andrewelamb: I added it

linglp (approved):
LGTM! Thanks for the fix

thomasyu888 (approved):
🔥 LGTM! I'll defer to someone on the team for final review, and lets make sure this makes it into `dev` branch at some point.

PR #1401, opened 2026-06-11
[SYNPY-1861] remove submissionInstructionsMessage and submissionReceiptMessage from required attributes

jaymedina:
# **Problem:**

The Evaluation OOP model's `store()` method incorrectly required `submission_instructions_message` and
`submission_receipt_message` to be set before creating or updating an evaluation. These are optional display messages
shown to submitters. The Synapse REST API accepts PUT requests without them.

# **Solution:**

- [x] Remove `submission_instructions_message` and `submission_receipt_message` from the required attributes validated in `to_synapse_request()`.
- [x] Updated field and class-level docstrings to clearly mark name, description, and content_source as the true required-to-store attributes.
- [x] Added unit test coverage verifying the request body is built correctly without the message fields
- [x] Added integration tests for creating and updating evaluations without them, including a test confirming that submissions can be stored and retrieved against an evaluation that was created without these fields.

# **Testing:**

Tests are passing locally:

image


jaymedina (commented):
_Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Reviewed `evaluation.py`, `unit_test_evaluation.py`, `test_evaluation_async.py`, and `test_submission_async.py`. The core logic change is correct — the two message fields are properly removed from the required list and are now conditionally included in the request body only when non-`None`. Docstrings and examples are updated consistently. Two minor findings left as inline comments.

tests/integration/synapseclient/models/async/test_evaluation_async.py
```diff
@@ -294,6 +316,31 @@ async def test_update_multiple_fields(self, test_evaluation: Evaluation):
...
+        created_evaluation.description = new_description
+        updated_evaluation = await created_evaluation.store_async(
+            synapse_client=self.syn
+        )
+
+        # THEN the update should succeed
+        assert updated_evaluation.description == new_description
+        assert updated_evaluation.id == created_evaluation.id
+        assert updated_evaluation.submission_instructions_message is None
+        assert updated_evaluation.submission_receipt_message is None
```
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

This asserts the Synapse API returns `null`/absent for these fields, but the API could return `""` (empty string) instead, which would fail the `is None` check. Since the test's goal is verifying the update succeeds, consider dropping these two assertions — they're testing API behavior we don't control.

tests/unit/synapseclient/models/unit_test_evaluation.py
```diff
@@ -149,6 +125,36 @@ def test_to_synapse_request_missing_required_fields(self):
             ):
                 evaluation.to_synapse_request(RequestType.CREATE)
 
+    def test_to_synapse_request_without_optional_message_fields(self):
+        """Test that submission_instructions_message and submission_receipt_message
+        are not required — the request body omits them when absent."""
```
jaymedina, line 130: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Two-line docstring; all other test methods in this class use one line. Condense to: `"""Test that the request body omits optional message fields when absent."""`

jaymedina: less descriptive docstring with no added benefit

tests/integration/synapseclient/models/async/test_evaluation_async.py
```diff
@@ -46,6 +46,28 @@ async def test_create_evaluation(self):
         assert created_evaluation.owner_id is not None  # Check that owner_id is set
         assert created_evaluation.created_on is not None  # Check that created_on is set
 
+    async def test_create_evaluation_without_optional_message_fields(self):
+        # GIVEN a project to work with
+        project = await Project(name=f"test_project_{uuid.uuid4()}").store_async(
```
linglp: should we also create a class level fixture here just like other evaluation tests? `test_create_evaluation_without_optional_message_fields` and `test_create_evaluation` could then share a fixture.

tests/integration/synapseclient/models/async/test_submission_async.py
```diff
@@ -98,6 +98,40 @@ async def test_store_submission_successfully_async(
         assert created_submission.created_on is not None
         assert created_submission.version_number is not None
 
+    async def test_store_submission_to_evaluation_without_message_fields_async(
```
linglp, line 101: Nit: the submission tests shouldn't care whether `submission_instructions_message` or `submission_receipt_message` are set. What about just removing the message fields from `test_evaluation`, delete `test_store_submission_to_evaluation_without_message_fields_async`

synapseclient/models/evaluation.py
```diff
@@ -33,11 +33,11 @@ class Evaluation(EvaluationSynchronousProtocol):
         etag: Synapse employs an Optimistic Concurrency Control (OCC) scheme to handle concurrent updates.
               The eTag changes every time an Evaluation is updated; it is used to detect when a client's copy
               of an Evaluation is out-of-date.
-        name: The name of this Evaluation.
-        description: A text description of this Evaluation.
+        name: **Required to store.** The name of this Evaluation.
```
linglp, line 36: I think the docstring of `async def store_async` also needs to be updated. Currently, I am seeing: 
```
        Example: Creating a new evaluation
            &nbsp;
            Create a new evaluation on Synapse by storing an evaluation object with the required fields. If there are any fields missing, an error will be raised.
            ```python
            from synapseclient.models import Evaluation
            import asyncio

            syn = Synapse()
            syn.login()

            async def create_evaluation():
                evaluation = await Evaluation(
                    name="My Challenge Evaluation",
                    description="Evaluation for my data challenge",
                    content_source="syn123456",
                    submission_instructions_message="Submit CSV files only",
                    submission_receipt_message="Thank you for your submission!"
                ).store_async()

                return evaluation

            created_evaluation = asyncio.run(create_evaluation())
            ```

```
We should either remove the optional fields or update the description.

tests/integration/synapseclient/models/async/test_evaluation_async.py
```diff
@@ -46,6 +46,28 @@ async def test_create_evaluation(self):
         assert created_evaluation.owner_id is not None  # Check that owner_id is set
```
linglp, line 40: I think the entire `TestEvaluationValidation` class should also probably be moved to unit test

linglp, line 40: Also, `test_create_evaluation_missing_required_fields` only covers when the description is missing. But missing `name` and missing `content_source` aren't tested.

linglp, line 40: I think `def _build_evaluation_with_all_fields(self) -> Evaluation` and probably some other tests related to `test_to_synapse_request_create` should also be updated.

synapseclient/models/evaluation.py
linglp, line 273: I saw that under `to_synapse_request`, we have: 
```
        for attribute in required_attributes:
            if not getattr(self, attribute):
                raise ValueError(
                    f"Your evaluation object is missing the '{attribute}' attribute. This attribute is required to {request_type.value} an evaluation"
                )
```
But this would only report the first missing field. I think the consistent pattern in the code base is to check per field: 
```
if not self.x: raise ValueError(...)
```

linglp (changes requested):
@jaymedina Hi Jenny! It looks good overall. I found some issues in the tests and docstring that should probably be fixed before merging.

synapseclient/models/evaluation.py
```diff
@@ -123,10 +121,10 @@ class Evaluation(EvaluationSynchronousProtocol):
     of an Evaluation is out-of-date."""
 
     name: Optional[str] = None
-    """The name of this Evaluation."""
+    """**Required to store.** The name of this Evaluation."""
```
thomasyu888, line 124: For discussion, should there be an `Optional[str]` here if it is "required"?  Is **Required to store** a new semantic within the client?

linglp: I think the idea is that "name" is required when calling `evaluation.store()`. But I think we don't have the convention of doing something like `**Required to store.**` in the code base. My understanding is that we will just raise an error like [here](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/synapseclient/models/project.py#L335)

andrewelamb: I agree with Lingling here on both counts.

synapseclient/models/evaluation.py
```diff
@@ -274,9 +270,13 @@ def to_synapse_request(self, request_type: RequestType):
...
             "contentSource": self.content_source,
-            "submissionInstructionsMessage": self.submission_instructions_message,
-            "submissionReceiptMessage": self.submission_receipt_message,
         }
+        if self.submission_instructions_message is not None:
+            request_body["submissionInstructionsMessage"] = (
+                self.submission_instructions_message
+            )
+        if self.submission_receipt_message is not None:
+            request_body["submissionReceiptMessage"] = self.submission_receipt_message
```
andrewelamb, line 282: Shouldn't `delete_none_keys(request_body)` be used here?

synapseclient/models/evaluation.py
thomasyu888: Should type hinting be...

```
name: None
    """The name of this Evaluation."""
```

instead of 

```
name: Optional[str] = None
```

tests/integration/synapseclient/models/async/test_evaluation_async.py
jaymedina: This assertion is scoped to the evaluation that was scored in this test, which is not fed a submission instructions/receipt message, and I can confirm that the response body of a GET request to an evaluation like this returns `None` for those attributes:

image

Skipping

tests/integration/synapseclient/models/async/test_evaluation_async.py
jaymedina: Good catch! Implementing now :)

jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Good point — added a class-level `test_project` fixture and updated both tests to use it.

tests/integration/synapseclient/models/async/test_submission_async.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

The intent of `test_store_submission_to_evaluation_without_message_fields_async` is to ensure that the submission service doesn't require `submission_instructions_message` or `submission_receipt_message` to be set on the evaluation — since those attributes surface to submitters, it's worth explicitly verifying that submissions can be created against an evaluation that omits them. Happy to keep the test and clarify the comment if that makes the intent clearer.

synapseclient/models/evaluation.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Good catch — removed the optional fields from the "Creating a new evaluation" example and updated the description accordingly.

synapseclient/models/evaluation.py
BryanFauble: I do not think that we can remove the Optional here because then logic like:


`evaluation = Evaluation(id="9999999").get()`

Would not work, and you would have to do something like:

`evaluation = Evaluation(id="9999999", name="").get()`

tests/integration/synapseclient/models/async/test_evaluation_async.py
```diff
@@ -17,18 +17,23 @@ def init(self, syn: Synapse, schedule_for_cleanup: Callable[..., None]) -> None:
         self.syn = syn
         self.schedule_for_cleanup = schedule_for_cleanup
 
-    async def test_create_evaluation(self):
-        # GIVEN a project to work with
+    @pytest.fixture(scope="class")
+    async def test_project(
```
BryanFauble: We could consider using the shared Project set up in conftest.py if this does not poison other tests, or other tests cause this test to become unstable.

tests/integration/synapseclient/models/async/test_evaluation_async.py
```diff
@@ -42,10 +47,28 @@ async def test_create_evaluation(self):
...
+            content_source=test_project.id,
+        )
+        created_evaluation = await evaluation.store_async(synapse_client=self.syn)
+        self.schedule_for_cleanup(created_evaluation.id)
+
+        # THEN the evaluation should be created successfully
+        assert created_evaluation.id is not None
+        assert created_evaluation.etag is not None
+        assert created_evaluation.name == evaluation.name
+        assert created_evaluation.content_source == test_project.id
```
BryanFauble: assert that the `submission_instructions_message or submission_receipt_message` fields are None

tests/integration/synapseclient/models/async/test_evaluation_async.py
```diff
@@ -294,6 +317,31 @@ async def test_update_multiple_fields(self, test_evaluation: Evaluation):
...
+    async def test_update_evaluation_without_optional_message_fields(
+        self, test_project: Project
+    ):
+        # GIVEN an evaluation created without submission message fields
+        evaluation = Evaluation(
+            name=f"test_evaluation_{uuid.uuid4()}",
+            description="A test evaluation without optional messages",
+            content_source=test_project.id,
+        )
+        created_evaluation = await evaluation.store_async(synapse_client=self.syn)
```
BryanFauble: Could add an assert statement here to also assert the behavior from the test_update_evaluation_without_optional_message_fields test, and then we can remove the test_update_evaluation_without_optional_message_fields test so there are fewer that need to run.

BryanFauble (approved):
A few items in the testing structure to take a look at, otherwise LGTM!

synapseclient/models/evaluation.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Implemented — switched to `delete_none_keys(request_body)`.

tests/integration/synapseclient/models/async/test_evaluation_async.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Done — added `None` assertions for both message fields.

tests/integration/synapseclient/models/async/test_evaluation_async.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Done — folded the update scenario into `test_create_evaluation_without_optional_message_fields` and removed `test_update_evaluation_without_optional_message_fields`.

tests/integration/synapseclient/models/async/test_evaluation_async.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Good call — switched `TestEvaluationCreation` to use the session-scoped `project_model` fixture from conftest. Evaluations use unique UUIDs and are cleaned up individually, so no cross-test contamination.

tests/integration/synapseclient/models/async/test_evaluation_async.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Expanded to cover all three required fields using `pytest.mark.parametrize`.

jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Updated the `_build_evaluation_with_all_fields` docstring — it said "all required fields" but `submission_instructions_message` and `submission_receipt_message` are optional. I'm going to leave the submission message inputs in `test_to_synapse_request_create` to test that these optional attributes are handled correctly when building the request body and sending it to the endpoint.

synapseclient/models/evaluation.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Refactored to use per-field checks — this is the dominant pattern across the broader codebase (e.g. `curation.py`, `link.py`, mixins).

jaymedina: good point - [updated!](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1401/commits/89dd01207f2b7c96d4c76504bae599955efcaa0f)

tests/integration/synapseclient/models/async/test_evaluation_async.py
jaymedina: _Note: This comment has been generated with AI assistance and reviewed by me ✍️_

Moved — all 5 tests raise `ValueError` before any API call so they don't need network access. Relocated to `unit_test_evaluation.py` and removed `TestEvaluationValidation` from the integration file.

jaymedina:
Thanks all for the reviews~

@linglp can I get your eyes on those last unresolved threads before your approval so I can merge? Thanks in advance!

linglp (approved):
LGTM! Thanks!

PR #1405, opened 2026-06-11
Add compliance-config.json

allaway:
## Summary

Adds `compliance-config.json` at the repo root to satisfy the **ARPA-H BDF ENHANCE Scorecard** compliance-configuration check. The file is the regulatory-applicability questionnaire (HIPAA / FDA / EHI / children's-data boolean flags), using the canonical 18-key format also used by other BDF program tools (e.g. `helxplatform/Koios`, `alico-cra/bdf-mock-tool`).

### Profile chosen

`synapseclient` is characterized as **low-risk, data-agnostic research data-management infrastructure** — a client library/CLI that transfers whatever files a user provides, with PHI/governance handled at the Synapse platform level rather than inherent to the client. Accordingly, every flag is `false` except `is_low_risk: true`.

| Flag | Value |
|------|-------|
| `collects_health_info` | `false` |
| `has_identifiable_health_info` | `false` |
| `is_health_plan` | `false` |
| `is_healthcare_provider` | `false` |
| `offers_certified_hit` | `false` |
| `enables_ehi_exchange` | `false` |
| `requires_prescription` | `false` |
| `works_for_covered_entity` | `false` |
| `intended_for_medical_use` | `false` |
| `is_administrative_or_lifestyle_only` | `false` |
| `is_low_risk` | **`true`** |
| `has_fda_regulated_function` | `false` |
| `is_consumer_facing` | `false` |
| `interacts_with_phr` | `false` |
| `intended_for_children` | `false` |
| `has_child_oriented_features` | `false` |
| `children_using_app` | `false` |
| `offers_substance_use_treatment` | `false` |

### Validation
- Parses as valid JSON (18 keys).
- Repo `pre-commit` hooks (`check-json`, etc.) pass.

### ⚠️ For maintainers
These are **compliance attestations**, not dummy values. Please have the Sage program lead / compliance confirm the regulatory posture before treating this as authoritative — in particular `collects_health_info`, `has_identifiable_health_info`, `works_for_covered_entity`, and `is_consumer_facing` were judgment calls (both example BDF tools set `is_consumer_facing: true`; this PR sets it `false` since the client is a developer/researcher tool).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


compliance-config.json
```diff
@@ -0,0 +1,20 @@
...
+    "is_healthcare_provider": false,
+    "offers_certified_hit": false,
+    "enables_ehi_exchange": false,
+    "requires_prescription": false,
+    "works_for_covered_entity": false,
+    "intended_for_medical_use": false,
+    "is_administrative_or_lifestyle_only": false,
+    "is_low_risk": true,
+    "has_fda_regulated_function": false,
+    "is_consumer_facing": false,
```
andrewelamb, line 14: With the caveat that I don't know what these fields mean exactly: Is this correct? I would argue that the PythonClient IS consumer facing.

compliance-config.json
```diff
@@ -0,0 +1,20 @@
...
+    "interacts_with_phr": false,
```
andrewelamb, line 15: I'd also argue the PythonClient DOES interact with PHR

thomasyu888: PHR? personal health record?

allaway: I disagree on this one, at least from an intent/design perspective - it's against the synapse TOS to upload these types of data

compliance-config.json
allaway: IT's a good point and I had similar thoughts. After digging a bit, this is specifically talking about contexts where we would have a reporting obligation to the FTC if end-user's PHR data were leaked. in this context, I understand the consumers to be mobile health app users, not just anyone who uses the repo. 


Thanks for the review!

allaway (commented):
I did some deeper dives on the couple of points @andrewelamb raised and I think I still want to stick with the current profile of answers.

compliance-config.json
thomasyu888: @allaway on mere technicality - we do have limited datasets and/or patient/sample metadata that the client does download.

That said, maybe that doesn't classify as PHR

allaway: This is from some claude-ing. I don't think the data that we are positioned to collect qualifies as PHR: 

>For the synapseclient call, the discriminating questions are whether the data it moves is controlled "by or primarily for the individual" data subject, and whether it's PHR identifiable health information provided by or on behalf of that individual. A research-data client where the records are managed by researchers and institutions — not by the individual whose data it might be, and often de-identified — has a solid argument that it does not interact with PHRs in the regulatory sense, even though it can technically read and write health-related files. That's the substance behind the reviewer's "DOES interact with PHR" comment: it's true in a loose everyday sense but not obviously true under §318.2's "controlled by or primarily for the individual" standard. And again, with collects_health_info: false the tool never reaches this branch anyway, so it's an attestation about posture rather than a trigger.

I also asked for some examples to better understand what _does_ qualify, here are some examples:

>Here are platforms that do collect/maintain PHRs, grouped by how they map to the §318.2 definition (a record "controlled by or primarily for the individual," drawing from multiple sources):
>Patient-controlled aggregators (textbook PHRs): Apple Health Records, PicnicHealth, Sync.MD, OneRecord, plus consumer vault apps like Capzule, CareClinic, and Tidy Health — the individual stores and controls their own records.
>Tethered patient portals (PHRs linked to a provider's EHR): Epic MyChart, FollowMyHealth, healow, athenahealth/Oracle Health portals. Note: usually operated for HIPAA-covered entities, so they fall under HIPAA, not the FTC's Health Breach Notification Rule.
>Non-HIPAA consumer apps the FTC pursued under the HBNR (most relevant, since interacts_with_phr feeds that rule): GoodRx, charged as a vendor of personal health records for disclosing consumers' PHR identifiable health info to ad platforms, plus Flo Health, Easy Healthcare, Monument, and Kochava — note the reach extends to period/fertility trackers and substance-use apps.
Discontinued (don't use as live examples): Microsoft HealthVault, Google Health.

allaway: IF it would be helpful, I wonder if we should get a review from Jon on this?

allaway:
@jonlong-sagebionetworks Would you be willing to take a look at the couple of open comment threads and provide your opinion? 

As a bit of context, this is a checklist the BDF IV&V partner wants us to include in our repository. The BDF ENHANCE Scorecard checks for a compliance-config.json at the repo root, so we need one to pass. It's an 18-flag regulatory-applicability questionnaire (HIPAA, FDA, EHI, children's data) in the canonical format the IV&V partner expects.
I characterized synapseclient as low-risk, data-agnostic research infrastructure that just moves user files, with PHI and governance handled at the platform level. So every flag is false except is_low_risk.

Where we're stuck: a couple were judgment calls. @andrewelamb  thinks the client is consumer-facing and interacts with a PHR (both currently false), but I think that it probably is referring to people working with their own personal health data, which is not the intended use of this software (even if it is technically possible, but then that's every mobile app or cli tool that allows you to enter any sort of data). 

Do you have any thoughts that could guide us? thanks!

compliance-config.json
jonlong-sagebionetworks: What is a PHR?  (https://www.hhs.gov/sites/default/files/ocr/privacy/hipaa/understanding/special/healthit/phrs.pdf))

There is currently no universal definition of a PHR, although several relatively similar
definitions exist within the industry. In general, a PHR is an electronic record of an
individual’s health information by which the **individual controls access** to the information and
may have the ability to manage, track, and participate in his or her own health care. A PHR
should not be confused with an electronic health record (EHR). An EHR is held and
maintained by a health care provider and may contain all the information that once existed in a
patient’s paper medical record, but in electronic form.

PHRs universally focus on providing individuals with the ability to manage their health
information and to control, to varying extents, who can access that health information. A PHR
has the potential to provide individuals with a way to create a longitudinal health history and
may include common information such as medical diagnoses, medications, and test results.
Most PHRs also provide individuals with the capability to control who can access the health
information in the PHR, and because PHRs are electronic and generally accessible over the
Internet, individuals have the flexibility to view their health information at any time and from
any computer at any location. The accessibility of health information in a PHR may facilitate
appropriate and improved treatment for conditions or emergencies that occur away from an
individual’s usual health care provider. Additionally, the ability to access one’s own health
information in a PHR may assist individuals in identifying potential errors or mistakes in their
information. 

Does Synapse provide a way for data owners (i.e. Grandma) to access and control their records directly, or is this not possible?  From what I understand the users of Synapse are not patients, but perhaps entities who maintain that relationship with patients directly.  Synapse is upstream and relies on its users to follow the TOS with regard to the HIPAA Privacy Rule as it relates to protected health information.  Christine or Kimberly should also weigh in as they will be more familiar with this space than I am.

milen-sage: A bit late to the discussion here. Thanks @jonlong-sagebionetworks for the input!

@allaway yes, I agree currently Synapse is not a PHR management system (e.g. we are not PicnicHealth).

milen-sage (approved):

thomasyu888 (approved):
🔥 LGTM!

PR #1407, opened 2026-06-12
Reuse pre-signed URL for multi-threaded downloads

BryanFauble:
## Problem

When downloading a file larger than 8 MB (the multi-threaded download path), the client requests **two** pre-signed URLs from the `/fileHandle/batch` endpoint:

1. In `download_by_file_handle`, to retrieve file handle metadata (concrete type, content size, MD5, storage location) needed to choose the download strategy. Its `preSignedURL` is **discarded** on the multi-threaded path.
2. Again inside `_MultithreadedDownloader` (via `PresignedUrlProvider._get_pre_signed_info`), to actually download the chunks.

The first URL is thrown away, so every large-file download generates a redundant pre-signed URL — doubling the pre-signed URL count and polluting download metrics. (Small/single-threaded downloads and cache hits are unaffected — they already use the URL from step 1, or none at all.)

```mermaid
sequenceDiagram
    actor Client
    participant DBFH as download_by_file_handle
    participant API as /fileHandle/batch
    participant MTD as _MultithreadedDownloader
    participant Provider as PresignedUrlProvider

    Client->>DBFH: download (file > 8 MB)
    DBFH->>API: getFileHandleForDownload (#1)
    API-->>DBFH: metadata + preSignedURL
    Note over DBFH: preSignedURL DISCARDED<br/>(only metadata used)
    DBFH->>MTD: download_from_url_multi_threaded()
    MTD->>Provider: get_info()
    Provider->>API: getFileHandleForDownload (#2)
    API-->>Provider: preSignedURL
    Note over API: 2 pre-signed URLs<br/>requested per download
    loop each 8 MB chunk
        MTD->>Provider: get_info() (reuses cached URL)
    end
```

## Solution

Forward the pre-signed URL already retrieved in step 1 into `download_from_url_multi_threaded` (the function already accepts a `presigned_url` parameter, used today by `WikiPage`). The downloader seeds `PresignedUrlProvider._cached_info` with it and skips the second batch call.

Crucially, the file handle identifiers (`file_handle_id` / `object_id` / `object_type`) are now also threaded into the `DownloadRequest` on the seeded-URL path, so the **mid-download expiry/refetch flow is preserved**: if the seeded URL nears expiry during a long transfer, `PresignedUrlProvider.get_info()` still refetches a fresh URL exactly as before. Wiki callers omit these identifiers, so their existing (no-refetch) behavior is unchanged.

```mermaid
sequenceDiagram
    actor Client
    participant DBFH as download_by_file_handle
    participant API as /fileHandle/batch
    participant MTD as _MultithreadedDownloader
    participant Provider as PresignedUrlProvider

    Client->>DBFH: download (file > 8 MB)
    DBFH->>API: getFileHandleForDownload (#1)
    API-->>DBFH: metadata + preSignedURL
    Note over DBFH: build PresignedUrlInfo<br/>from the SAME response
    DBFH->>MTD: download_from_url_multi_threaded(presigned_url=..., file_handle_id=...)
    MTD->>Provider: get_info()
    Note over Provider: returns seeded URL<br/>NO second API call
    loop each 8 MB chunk
        MTD->>Provider: get_info() (reuses cached URL)
        alt URL near expiry mid-download
            Provider->>API: getFileHandleForDownload (refetch)
            API-->>Provider: fresh preSignedURL
        end
    end
    Note over API: 1 pre-signed URL<br/>per download (refetch only if expired)
```

## Testing

- **`unit_test_download_functions.py`** — new tests spy on `get_file_handle_for_download_async` and `download_from_url_multi_threaded` to assert the batch endpoint is called **exactly once** and that the retrieved URL + file handle identifiers are forwarded; plus that the constructed `DownloadRequest` carries the identifiers (and that the identifier-less wiki path leaves them `None`).
- **`unit_test_download_async.py`** — new `TestSeededPresignedUrlProvider`: a seeded unexpired URL is reused with **zero** refetch calls; a seeded near-expiry URL triggers **exactly one** refetch keyed on the correct identifiers (not `None`).
- **`unit_test_download.py`** — updated `test_multithread_true_s3_file_handle` to reflect the new forwarding behavior.
- Full regression run of the download, multithread_download, and wiki unit suites: **139 passed**. ruff / black / isort clean on changed files.

### DIFF
--- a/synapseclient/core/download/download_functions.py
+++ b/synapseclient/core/download/download_functions.py
@@ -602,6 +602,18 @@ async def download_by_file_handle(
             ):
                 span.set_attribute("synapse.storage.provider", "s3")
 
+                # Reuse the pre-signed URL already retrieved above rather than
+                # requesting a second one inside the multi-threaded downloader. The
+                # file_handle_id/object_id/object_type are still passed so the
+                # downloader can refetch the URL if it expires mid-download.
+                presigned_url_info = PresignedUrlInfo(
+                    file_name=file_handle["fileName"],
+                    url=file_handle_result["preSignedURL"],
+                    expiration_utc=_pre_signed_url_expiration_time(
+                        file_handle_result["preSignedURL"]
+                    ),
+                )
+
                 # run the download multi threaded if the file supports it, we're configured to do so,
                 # and the file is large enough that it would be broken into parts to take advantage of
                 # multiple downloading threads. otherwise it's more efficient to run the download as a simple
@@ -612,6 +624,7 @@ async def download_by_file_handle(
                     object_type=entity_type,
                     destination=destination,
                     expected_md5=actual_md5,
+                    presigned_url=presigned_url_info,
                     synapse_client=syn,
                 )
 
@@ -738,7 +751,13 @@ async def download_from_url_multi_threaded(
             # If the destination is a directory, then the file name should be the same as the file name in the presigned url
             # This is added to ensure the temp file can be copied to the desired destination without changing the file name
             destination = os.path.join(destination, presigned_url.file_name)
+        # Pass through the file handle identifiers so the downloader can refetch a
+        # fresh pre-signed URL if the seeded one expires mid-download. Callers that
+        # cannot refetch (e.g. wiki attachments) omit these, leaving them as None.
         request = DownloadRequest(
+            file_handle_id=int(file_handle_id) if file_handle_id is not None else None,
+            object_id=object_id,
+            object_type=object_type,
             path=temp_destination,
             debug=client.debug,
             presigned_url=presigned_url,
--- a/tests/unit/synapseclient/core/download/unit_test_download_async.py
+++ b/tests/unit/synapseclient/core/download/unit_test_download_async.py
@@ -152,6 +152,84 @@ class TestPresignedUrlProvider:
         assert expected == download_async._pre_signed_url_expiration_time(url)
 
 
+class TestSeededPresignedUrlProvider:
+    """Tests that a PresignedUrlProvider seeded with an already-fetched URL reuses
+    it on the happy path but still refetches (using the request's file handle
+    identifiers) once the seeded URL nears expiry mid-download."""
+
+    @pytest.fixture(scope="function", autouse=True)
+    def setup_method(self) -> None:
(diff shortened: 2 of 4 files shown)


thomasyu888 (commented):
🔥 Thanks for taking such a quick look at this. My biggest concern is whether this would work for large file downloads, and I think it will.

I'll defer to @linglp / @andrewelamb for final review

synapseclient/core/download/download_functions.py
```diff
@@ -738,7 +751,13 @@ async def download_from_url_multi_threaded(
             # If the destination is a directory, then the file name should be the same as the file name in the presigned url
             # This is added to ensure the temp file can be copied to the desired destination without changing the file name
             destination = os.path.join(destination, presigned_url.file_name)
+        # Pass through the file handle identifiers so the downloader can refetch a
+        # fresh pre-signed URL if the seeded one expires mid-download. Callers that
+        # cannot refetch (e.g. wiki attachments) omit these, leaving them as None.
         request = DownloadRequest(
```
linglp, line 757: Could you clarify or add a comment about the asymmetry here? We are doing: 
```
        request = DownloadRequest(
            file_handle_id=int(file_handle_id) if file_handle_id is not None else None,
            object_id=object_id,
            object_type=object_type,
            path=temp_destination,
            debug=client.debug,
            presigned_url=presigned_url,
        )
```
But in the else statement, I saw: 
```
        request = DownloadRequest(
            file_handle_id=int(file_handle_id),
            object_id=object_id,
            object_type=object_type,
            path=temp_destination,
            debug=client.debug,
        )
```
I am curious why `None` guard is only needed in the `if presigned_url is not None` branch? Is it because file_handle_id is guaranteed non-None in else?

linglp (approved):
LGTM! Just one question

andrewelamb (approved):
LGTM!

synapseclient/core/download/download_functions.py
BryanFauble: file_handle_id is guaranteed non-None if you are not passing in a presigned_url, because the expectation is that ahead of time the presigned_url was not retrieved - And instead, it is being retrieved from the Rest API based on the file_handle_id

PR #1409, opened 2026-06-19
[SYNPY-1857] When creating RecordSets, the UpsertKeys are the first columns

andrewelamb:
# **Problem:**

When creating a RecordSet via `create_record_based_metadata_task`, the CSV template columns were generated directly from the order in which the schema properties were extracted from the schema. This meant the upsert key columns — which serve as the row identifiers in the Grid curation UI — could appear anywhere in the template rather than as the leftmost columns. Curators expect the columns that identify each row to lead the template for readability and ease of data entry.

Additionally, there was no validation that the supplied `upsert_keys` actually correspond to columns defined in the schema. An upsert key that did not match any schema property would silently produce a template that could not function as intended, with no clear error to the user.

# **Solution:**

- Added a `_reorder_columns_with_upsert_keys_first` helper that moves the upsert key columns to the front of the template DataFrame. The relative order of the upsert keys is preserved as provided by the caller, and the remaining columns keep their original order. Upsert keys not present among the columns are skipped, and empty DataFrames are handled gracefully.
- In `create_record_based_metadata_task`, after the schema properties are extracted, the template is now validated and reordered:
    - If any of the requested `upsert_keys` are not found among the schema properties, a `ValueError` is raised naming the missing keys. This gives the user a clear, actionable error instead of a silently malformed template.
    - The template columns are then reordered so the upsert keys appear first, in the given order.

# **Testing:**

Added unit tests in `tests/unit/synapseclient/extensions/unit_test_curator.py`:

- `test_create_record_based_metadata_task_reorders_upsert_keys_first` — verifies the generated template has the upsert keys as the leftmost columns in the provided order.
- `test_create_record_based_metadata_task_raises_for_missing_upsert_keys` — verifies a `ValueError` naming the missing key is raised when an upsert key is absent from the schema properties.
- `test_reorder_columns_with_upsert_keys_first` — parametrized cases covering moving keys to the front, preserving provided key order, skipping keys missing from columns, and handling an empty DataFrame.

### DIFF
--- a/synapseclient/extensions/curator/record_based_metadata_task.py
+++ b/synapseclient/extensions/curator/record_based_metadata_task.py
@@ -86,26 +86,22 @@ def _reorder_columns_with_upsert_keys_first(
     The upsert keys serve as the row identifiers in the Grid curation UI, so they
     should always be the leftmost columns of the CSV template. The relative order of
     the upsert keys is preserved as given, followed by the remaining columns in their
-    original order. Upsert keys that are not present among the columns are skipped.
+    original order. Callers are expected to validate that every upsert key is present
+    among the columns before calling this function.
 
     Args:
         df: DataFrame whose columns should be reordered.
         upsert_keys: List of column names to move to the front, in the desired order.
 
     Returns:
-        DataFrame with the upsert key columns moved to the front. If the DataFrame has
-        no columns or no upsert keys are present, a DataFrame with the same columns in
-        their original order is returned.
+        DataFrame with the upsert key columns moved to the front.
     """
-    existing_columns = df.columns.tolist()
-
-    ordered_upsert_keys = [key for key in upsert_keys if key in existing_columns]
-    ordered_upsert_key_set = set(ordered_upsert_keys)
+    upsert_key_set = set(upsert_keys)
     remaining_columns = [
-        column for column in existing_columns if column not in ordered_upsert_key_set
+        column for column in df.columns.tolist() if column not in upsert_key_set
     ]
 
-    return df[ordered_upsert_keys + remaining_columns]
+    return df[upsert_keys + remaining_columns]
 
 
 def extract_schema_properties_from_web(
--- a/synapseclient/extensions/curator/record_based_metadata_task.py
+++ b/synapseclient/extensions/curator/record_based_metadata_task.py
@@ -77,6 +77,37 @@ def extract_schema_properties_from_dict(schema_data: Dict[str, Any]) -> DATA_FRA
     return df
 
 
+def reorder_columns_with_upsert_keys_first(
+    df: DATA_FRAME_TYPE, upsert_keys: list[str]
+) -> DATA_FRAME_TYPE:
+    """
+    Reorder a DataFrame's columns so the upsert key columns appear first.
+
+    The upsert keys serve as the row identifiers in the Grid curation UI, so they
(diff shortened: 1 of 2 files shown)


synapseclient/extensions/curator/record_based_metadata_task.py
```diff
@@ -77,6 +77,37 @@ def extract_schema_properties_from_dict(schema_data: Dict[str, Any]) -> DATA_FRA
...
+        upsert_keys: List of column names to move to the front, in the desired order.
+
+    Returns:
+        DataFrame with the upsert key columns moved to the front. If the DataFrame has
+        no columns or no upsert keys are present, a DataFrame with the same columns in
+        their original order is returned.
+    """
+    existing_columns = df.columns.tolist()
+
+    ordered_upsert_keys = [key for key in upsert_keys if key in existing_columns]
```
linglp: Since we now raise a ValueError at the validation step if any upsert_key is missing from the schema, the if key in existing_columns filter in _reorder_columns_with_upsert_keys_first is redundant — ordered_upsert_keys will always equal upsert_keys at this point. Can we simplify to just use upsert_keys directly?

tests/unit/synapseclient/extensions/unit_test_curator.py
```diff
@@ -1485,6 +1596,45 @@ def test_extract_schema_properties_from_web_success(self, mock_schema_cls):
...
+            (
+                "empty DataFrame yields no columns",
+                [],
+                ["specimenID"],
+                [],
+            ),
+        ]
+
+        for name, columns, upsert_keys, expected in cases:
+            with self.subTest(name):
```
linglp, line 1624: why use `subTest` here instead of `@pytest.mark.parametrize`?

linglp (approved):
LGTM! Just two minor comments.

tests/unit/synapseclient/extensions/unit_test_curator.py
andrewelamb: The Class this test is in `class TestRecordBasedHelperFunctions(unittest.TestCase)`  (see line 1514) and inherits from unittest. These are incompatible with pytest, but this is the unit test version of that.

I thought about refactoring these to use pytest, but I figured it's outside the scope for this ticket.

synapseclient/extensions/curator/record_based_metadata_task.py
andrewelamb: Yep, that's correct.

synapseclient/extensions/curator/record_based_metadata_task.py
```diff
@@ -77,6 +77,33 @@ def extract_schema_properties_from_dict(schema_data: Dict[str, Any]) -> DATA_FRA
...
+    among the columns before calling this function.
+
+    Args:
+        df: DataFrame whose columns should be reordered.
+        upsert_keys: List of column names to move to the front, in the desired order.
+
+    Returns:
+        DataFrame with the upsert key columns moved to the front.
+    """
+    upsert_key_set = set(upsert_keys)
```
thomasyu888, line 99: @andrewelamb it's ok this was merged, and just checking if `set` auto changes the ordering?

andrewelamb: It might; sets don't preserve order. But, the set isn't used to determine the final order; the original list is (see line 104 below).

PR #1410, opened 2026-06-19
[SYNPY-1859] When creating entity views for Curator, the name column now goes first

andrewelamb:
# **Problem:**

When Curator created an entity view (via `create_json_schema_entity_view` in `file_based_metadata_task.py`), the columns were reordered so that ["id", "name", "createdBy"] were the first columns. It has been requested that "name" go first.

# **Solution:**

Changed the order of the `reorder_column` calls so that `name` is moved to index 0 last and therefore ends up first. 

# **Testing:**

- Updated the existing unit test (`unit_test_curator.py`) to assert the exact `reorder_column` call sequence (`id`, `createdBy`, `name`) rather than just asserting each call happened in any order, so the column ordering is now actually verified.
- Added an integration test (`tests/integration/.../test_file_based_metadata_task.py`) that creates a real JSON schema, folder, and file-based metadata task, then asserts that the resulting EntityView and CurationTask are created and that the leading columns are ordered `["name", "createdBy", "id"]`.


synapseclient/extensions/curator/file_based_metadata_task.py
```diff
@@ -77,10 +77,12 @@ def create_json_schema_entity_view(
...
     ).store(synapse_client=syn)
-    # This reorder is so that these show up in the front of the EntityView in Synapse
+    # This reorder is so that these show up in the front of the EntityView in Synapse.
+    # Order matters: each call moves its column to index 0, so the LAST call wins the
+    # front slot. The resulting front order is name, then createdBy, then id.
     view.reorder_column(name="createdBy", index=0)
-    view.reorder_column(name="name", index=0)
     view.reorder_column(name="id", index=0)
+    view.reorder_column(name="name", index=0)
     view.store(synapse_client=syn)
```
danlu1, line 84: Is it possible to set the first three columns all at once? As the code, it seems the `name`,`id` and `createdBy` are the first three columns.

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
```diff
@@ -0,0 +1,137 @@
...
+@pytest.fixture(scope="function")
+def folder(
+    syn: Synapse,
+    project: Project,
+    schedule_for_cleanup: Callable[[Any], None],
+) -> Generator[Folder, None, None]:
+    """Returns a Synapse Folder"""
+    folder = Folder(name=_test_name(), parent_id=project.id).store(synapse_client=syn)
+    schedule_for_cleanup(folder.id)
+    yield folder
```
danlu1: why do you need to yield folder?

danlu1 (approved):
LGTM

synapseclient/extensions/curator/file_based_metadata_task.py
```diff
@@ -65,7 +65,7 @@ def create_json_schema_entity_view(
         synapse_client=syn,
     )
     assert isinstance(entity, (Folder, Project))
-    jsb = entity.get_schema()
+    jsb = entity.get_schema(synapse_client=syn)
```
linglp, line 68: Nit: we don't have to do `synapse_client=syn` here?

synapseclient/extensions/curator/file_based_metadata_task.py
linglp: Can we try what Dan suggested? Also, if this pattern comes up more often, I would recommend adding a function so that we could change the column order in bulk rather than calling `reorder_column` and changing the order one at a time

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
```diff
@@ -0,0 +1,137 @@
+"""Integration tests for create_file_based_metadata_task."""
```
linglp: same feedbacks as: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1411/changes#diff-f9d0ff39e8d64faf64be0d788208f1f81a38989cb2ca25b43fe330041020a1aa

synapseclient/extensions/curator/file_based_metadata_task.py
andrewelamb: So, as far as I can tell, currently there is no function that does this in one line for OrderedDicts.

andrewelamb: I rewrote it to be more readable.

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
andrewelamb: This turns this into a generator (rather than a normal fixture). When a fixture is a generator the code that comes after it is always done on fixture cleanup. This means that at the end of any test that uses this fixture any JSON Schemas bound to it are unbound.

synapseclient/extensions/curator/file_based_metadata_task.py
andrewelamb: @linglp see https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1411#discussion_r3453115090

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
andrewelamb: This has been fixed in the other PR.

linglp (approved):
LGTM!

PR #1411, opened 2026-06-19
[SYNPY-1864] Added return_entities param to create_file_based_metadata_task

andrewelamb:
# **Problem:**

- `create_file_based_metadata_task` returned only a tuple of ID strings (`entity_view_id`, `task_id`). Callers who needed to read or modify the newly created `EntityView` or `CurationTask` had to make a second round-trip to Synapse to fetch the full objects.
- This was also inconsistent with `create_record_based_metadata_task`, which returns the entity objects directly.

# **Solution:**

- Added a `return_entities` parameter to `create_file_based_metadata_task`. When `True`, the function returns the created `EntityView` and `CurationTask` objects instead of their ID strings, avoiding the extra round-trip and matching the return shape of `create_record_based_metadata_task`.
- Defaults to `False` for backwards compatibility. When called without `return_entities=True`, a deprecation warning is logged indicating the entity-returning shape will become the default in v5.0.0, at which point the parameter will be removed (tracked by [SYNPY-1865](https://sagebionetworks.jira.com/browse/SYNPY-1865)).
- Refactored the internal helper `create_json_schema_entity_view` → `_create_json_schema_entity_view` (now private) to return the full `EntityView` object rather than just its ID, so the created entity can be reused without re-fetching.
- Updated docstrings, return-type annotations (`Union[Tuple[str, str], Tuple[EntityView, CurationTask]]`), and added a usage example.

# **Testing:**

- Added unit tests in `tests/unit/synapseclient/extensions/unit_test_curator.py` covering both return modes (ID strings vs. entity objects) and the deprecation warning.
- Added integration tests in `tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py` validating the end-to-end behavior of the new `return_entities` parameter against Synapse.

[SYNPY-1865]: https://sagebionetworks.jira.com/browse/SYNPY-1865?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


synapseclient/extensions/curator/file_based_metadata_task.py
```diff
@@ -57,15 +57,15 @@ def create_json_schema_entity_view(
...
+        The created EntityView object
     """
     entity = get(
         file_options=FileOptions(download_file=False),
         synapse_id=synapse_entity_id,
         synapse_client=syn,
     )
     assert isinstance(entity, (Folder, Project))
-    jsb = entity.get_schema()
+    jsb = entity.get_schema(synapse_client=syn)
```
linglp, line 69: Is this intentional? `synapse_client` should just be an optional parameter?

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
```diff
@@ -0,0 +1,187 @@
...
+    CurationTask,
+    EntityView,
+    Folder,
+    JSONSchema,
+    Project,
+    SchemaOrganization,
+)
+
+
+def _test_name() -> str:
```
linglp: Both `_test_name()` and `patient_schema_uri` are duplicates from `test_file_based_metadata_task.py` and `test_record_based_metadata_task.py`. Is there a way to avoid the duplicates?

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
```diff
@@ -0,0 +1,187 @@
...
+        for js in org.get_json_schemas(synapse_client=syn):
+            js.delete(synapse_client=syn)
+        org.delete(synapse_client=syn)
+
+    request.addfinalizer(cleanup)
+
+    return schema.uri
+
+
+@pytest.fixture(scope="function")
```
linglp: this is also very similar to folder fixture inside `synapsePythonClient/tests/integration/synapseclient/extensions/curator/test_record_based_metadata_task.py`
```
@pytest.fixture(scope="function")
def folder(syn: Synapse, project: Project, request: pytest.FixtureRequest) -> Folder:
    """Create a Folder for the test and tear it down on completion.

    The finalizer unbinds any JSON schema from the folder before deletion so
    that the module-scoped schema cleanup in patient_schema_uri can succeed
    (the server refuses to delete a schema that is still bound to an entity).
    """
    folder = Folder(name=_test_name(), parent_id=project.id).store(synapse_client=syn)

    def cleanup():
        folder.unbind_schema(synapse_client=syn)
        folder.delete(synapse_client=syn)

    request.addfinalizer(cleanup)

    return folder
```

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
```diff
@@ -0,0 +1,187 @@
...
+            folder_id=folder.id,
+            curation_task_name=test_name,
+            instructions="Contribute Patient data.",
+            entity_view_name=test_name,
+            schema_uri=patient_schema_uri,
+            synapse_client=syn,
+        )
+        schedule_for_cleanup(entity_view_id)
+
+        # The CurationTask is not a deletable entity via syn.delete, so it keeps a
```
linglp: Instead of having each individual function to cleanup, I think the easiest way is just to modify schedule_for_cleanup function by adding `CurationTask` to the isinstance tuple alongside `Team`, `SubmissionView`, etc. .  Since CurationTask has a delete_async method, it fits the existing `await item.delete_async(synapse_client=syn)`

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
```diff
@@ -0,0 +1,187 @@
...
+            curation_task_name=test_name,
+            instructions="Contribute Patient data.",
+            entity_view_name=test_name,
+            schema_uri=patient_schema_uri,
+            return_entities=True,
+            synapse_client=syn,
+        )
+        schedule_for_cleanup(entity_view.id)
+
+        # The CurationTask is not a deletable entity via syn.delete, so it keeps a
```
linglp: same comment as above.

tests/unit/synapseclient/extensions/unit_test_curator.py
```diff
@@ -484,6 +484,112 @@ def test_create_file_based_metadata_task_with_assignee(
...
+    )
+    @patch(
+        "synapseclient.extensions.curator.file_based_metadata_task.Synapse.get_client"
+    )
+    @patch(
+        "synapseclient.extensions.curator.file_based_metadata_task._create_json_schema_entity_view"
+    )
+    @patch("synapseclient.extensions.curator.file_based_metadata_task.Folder")
+    @patch("synapseclient.extensions.curator.file_based_metadata_task.CurationTask")
+    def test_create_file_based_metadata_task_id_return_logs_deprecation_warning(
```
linglp: Isn't that this test is very similar to: `test_create_file_based_metadata_task_success_with_schema`? Can we assert warnings in that test and avoid adding an additional one?

linglp (commented):
@andrewelamb Thanks for your work! I have some suggestions about better organizing tests and improving `schedule_for_cleanup` so that `Task` can be cleaned up more easily! Let me know if anything doesn't make sense.

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
thomasyu888, line 1: Just checking that integration tests didn't exist before, since this is an entirely new file.

thomasyu888:
🔥 LGTM!

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
andrewelamb: That's correct.

synapseclient/extensions/curator/file_based_metadata_task.py
andrewelamb: Yes. This is only called by `create_file_based_metadata_task` which has the argument `synapse_client: Optional[Synapse] = None`. This gets passed to this function(`_create_json_schema_entity_view`) so it could be `None` or a synapse client. However, the signature for `_create_json_schema_entity_view` doesn't reflect this so I'll fix that.

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
andrewelamb: That's fair!

tests/unit/synapseclient/extensions/unit_test_curator.py
andrewelamb: Agreed! I don't think we even need to test for deprecation warnings TBH.

synapseclient/extensions/curator/file_based_metadata_task.py
linglp: I think my original comment might not be clear. I was curious if `entity.get_schema()` will work fine.

linglp (approved):
LGTM!

synapseclient/extensions/curator/file_based_metadata_task.py
andrewelamb: I still don't think I understand. :) Could you elaborate?

linglp: Hmm I was thinking that we could just do `entity.get_schema()` without doing `entity.get_schema(synapse_client=syn)`

andrewelamb: Why though? If a user supplies a Synapse client via `create_file_based_metadata_task`  shouldn't that get passed on to every function that could use it?

BryanFauble: `synapse_client` should get passed through in all cases for any code that the user is not directly calling to ensure that everything can work when they pass in a constructed Synapse class instance.

@andrewelamb Is correct in this case.

tests/integration/synapseclient/extensions/curator/test_file_based_metadata_task.py
BryanFauble, line 1: Since we are already integrating testing all of the individual building blocks of the client I question if integration tests would be required at this higher level extensions package since it would be duplicating a significant amount of tests that are already running. Could we instead look to move this file over to unit tests to improve CI run times and lower the total number of IT tests that need to run? This would be a good file to start with, and then create a follow-up tech debt ticket to tackle others.

BryanFauble (approved):
Testing structure could be improved, but otherwise changes LGTM!

PR #1414, opened 2026-06-24
[SYNPY-1863] deepcopy annotations in SubmissionStatus._set_last_persistent_instance

jaymedina:
# **Problem:**

When calling `<SubmissionStatusObject>.submission_annotations.update()` on a `SubmissionStatus` object, the dataclass does not register a  change and throws a warning that the object has not been modified, resulting in the annotations never being correctly posted to Synapse.

# **Solution:**

Implement the pattern already established in the repository (example: [Project OOP model](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/fea5caa1226fe4b05882a54b5c8b4e98ad63abe0/synapseclient/models/project.py#L254)) by performing `deepcopy` on mutable objects, such as dicts, while performing `replace` (shallow copy) on immutable objects where it doesn't matter if the 2 instances (last and current) share references.

This should prevent issues with the client not detecting changes to the dataclass instance when a user performs in-place mutations instead of re-assigning the attribute, i.e:

```
# doing this
my_status.annotations.update({"status": "PASSED"})

# instead of this
my_status.annotations = {"status": "PASSED"}
```

# **Testing:**

### ✅ Can no longer reproduce user-error

image

### New tests that cover in-place mutation scenarios are passing:

image


tests/integration/synapseclient/models/async/test_submission_status_async.py
```diff
@@ -390,6 +390,58 @@ async def test_store_submission_status_without_id(self):
         ):
             await submission_status.store_async(synapse_client=self.syn)
 
+    async def test_in_place_mutation_of_submission_annotations_detected_and_stored(
```
BryanFauble: I would rather us not add in new integration tests for this when this is a method that could be verified in a unit test. Adding more integration tests is going to keep bloating our CI pipeline run.

BryanFauble (changes requested):
Can we get away with a unit test only for this change to verify the `has_changed` and this `_set_last_persistent_instance` logic?

tests/integration/synapseclient/models/async/test_submission_status_async.py
```diff
@@ -390,6 +390,58 @@ async def test_store_submission_status_without_id(self):
...
+        Test that in-place mutations (e.g. update()) on submission_annotations are
+        detected as a change and persisted to Synapse.
+        """
+        # GIVEN a submission status with an initial submission annotation stored
+        test_submission_status.submission_annotations = {
+            "initial_key": ["initial_value"]
+        }
+        stored = await test_submission_status.store_async(synapse_client=self.syn)
+        version_after_first_store = stored.status_version
+        assert not stored.has_changed
```
andrewelamb: Make a Jira ticket [Spike] for deciding long-term on has-changed.

andrewelamb (approved):
LGTM!

tests/integration/synapseclient/models/async/test_submission_status_async.py
jaymedina: https://sagebionetworks.jira.com/browse/SYNPY-1868

BryanFauble (approved):

PR #1415, opened 2026-06-25
Fix  incorrect typing on create_file_based_metadata_task

andrewelamb:
# **Problem:**
The type hint `create_file_based_metadata_task`  returned `CurationTask` is `str` when it should be `int`

# **Solution:**
Type hint is corrected


linglp (approved):
Thanks for the quick fix.

PR #1416, opened 2026-06-25
updated cryptography to 49.0.0

andrewelamb:
# **Problem:**
The `cryptography` version of this package was flagged as a security issue.

# **Solution:**
`cryptography` was bumped to 49.0.0 in the lock file.
the setup.cfg file was updated to have 48.0.1 as the minimum version


Pipfile.lock
BryanFauble, line 1: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/develop/setup.cfg would also need to get updated. The Pipfile.lock isn't what is used in all cases so we want to restrict this in the install script itself that 49.0.0 is the minimum version.

BryanFauble (changes requested):

BryanFauble (approved):

PR #1417, opened 2026-06-25
[SYNPY-1708] None values don't become annotations

andrewelamb:
# **Problem:**

Setting an annotation key to `None` (or to a value that contains no real data) did not behave intuitively when storing an entity:

- A key set to `None` was stored as the literal string `"None"` rather than being treated as "no value".
- A list containing only `None` values (e.g. `[None, None]`) was stored as `["None", "None"]`.
- A list mixing `None` with real values (e.g. `[None, "value"]`) was stored as `["None", "value"]`, polluting the data with a bogus `"None"` element.

# **Solution:**

`_convert_to_annotations_list` now treats `None` as the absence of a value:

- A value of `None` is skipped entirely (the key is omitted), rather than being stringified.
- `None` elements are stripped from list values before type inference, so a mixed list like `[None, "value"]` is stored as `["value"]`.
- A value that resolves to an empty list after stripping `None` (including `None`, `[]`, and `[None, None]`) is omitted.

Type inference now runs on the `None`-stripped element list, so the inferred Synapse type (STRING/BOOLEAN/LONG/DOUBLE/TIMESTAMP_MS) is no longer skewed by `None` placeholders. The literal string `"None"` is unaffected and is still stored as a normal string value.

The function also gained a full docstring documenting the conversion behavior, including the heterogeneous-list fallback to STRING.

# **Testing:**

- Added integration test `test_setting_annotation_key_to_none` in `tests/integration/synapseclient/models/async/test_file_async.py` that stores a file with one annotation set to a primitive `None`, one to a list of `None` values, one to an empty list, one to a list mixing `None` with a real value, and one to the literal string `"None"`, then retrieves a fresh copy and asserts:
    - `None`, list-of-`None`, and empty-list keys are not stored at all.
    - The mixed list `[None, "None"]` round-trips as `["None"]` (the `None` stripped).
    - The literal string `"None"` is stored as a normal string value.
- Added unit tests in `tests/unit/synapseclient/unit_test_annotations.py` covering `_convert_to_annotations_list` directly:
    - `test__convert_to_annotations_list__none_values` — asserts a primitive `None`, a list of only `None` (`[None, None]`), and an empty list are all omitted; a mixed list (`[None, "value", None]`) has its `None`s stripped to `["value"]`; and the literal string `"None"` is kept as a normal STRING value.
    - `test__convert_to_annotations_list__mixed_list_preserves_type` — asserts that stripping `None` from typed lists still infers the correct Synapse type (LONG, DOUBLE, BOOLEAN) rather than falling back to STRING.


tests/integration/synapseclient/models/async/test_file_async.py
```diff
@@ -572,6 +572,46 @@ async def test_removing_annotations_to_empty_dict(
         )
         assert not file_copy.annotations and isinstance(file_copy.annotations, dict)
 
+    async def test_setting_annotation_key_to_none(
```
BryanFauble: I would rather we not add another integration test for this. We can accomplish and verify the changed lines via a unit test only.

synapseclient/annotations.py
```diff
@@ -555,10 +555,41 @@ def to_synapse_annotations(annotations: Annotations) -> dict[str, Any]:
...
+        where value is a list of stringified elements.
+    """
     nested_annos = {}
     for key, value in annotations.items():
-        elements = to_list(value)
+        # None values don't become annotations
+        if value is None:
+            continue
+        # Strip None elements so they aren't stored as the string "None"
+        elements = [element for element in to_list(value) if element is not None]
```
thomasyu888: what happens when you store

`file_entity.annotations.key = float('nan')`
`file_entity.annotations.key = ""`
`file_entity.annotations.key = pd.NA`
`file_entity.annotations.key = np.nan`

These should be treated similarly to "None"

thomasyu888: Look out for "np.nan" and "pd.NA" and see if the same catching of these values work with `float('nan')` My rationale for this is that we don't necessarily want to import `pandas` or `numpy` just for this case

synapseclient/annotations.py
```diff
@@ -555,10 +555,41 @@ def to_synapse_annotations(annotations: Annotations) -> dict[str, Any]:
...
+        # Empty lists or lists with only None values don't become annotations
+        if not elements:
+            continue
         element_cls = _annotation_value_list_element_type(elements)
```
thomasyu888, line 644: Look at what this function does

synapseclient/annotations.py
thomasyu888, line 1: This is a bug fix, so lets push forward in 4.14 release

synapseclient/annotations.py
andrewelamb: this comment refers to `_annotation_value_list_element_type(elements)` which is now on line 631. This determines the type the annotaiton is going to be in Synpase based on the Python types.

BryanFauble (approved):

tests/unit/synapseclient/unit_test_annotations.py
```diff
@@ -120,6 +123,115 @@ def test__convert_to_annotations_list():
     assert expected_annos == actual_annos
 
 
+@pytest.mark.parametrize(
+    "value",
+    [
+        None,
```
linglp, line 170: Add `math.nan` so that we know this is covered as well? I also realized that there are other representation of None that we probably want to add? 
* np.float32("nan") / np.float16("nan")
* decimal.Decimal("nan")

linglp: I tested it out really quick, and it seems like: 
```
FAILED tests/unit/synapseclient/unit_test_annotations.py::test__is_missing_annotation_value__missing[numpy_float16_nan] - assert np.False_ is True
FAILED tests/unit/synapseclient/unit_test_annotations.py::test__is_missing_annotation_value__missing[decimal_nan] - AssertionError: assert False is True

```

synapseclient/annotations.py
```diff
@@ -555,10 +556,78 @@ def to_synapse_annotations(annotations: Annotations) -> dict[str, Any]:
...
+        if _is_missing_annotation_value(value):
+            continue
+        # Strip missing elements so they aren't stored as "None"/"nan"/etc.
+        elements = [
+            element
+            for element in to_list(value)
+            if not _is_missing_annotation_value(element)
+        ]
+        # Empty lists or lists with only missing values don't become annotations
+        if not elements:
```
linglp, line 642: the raise `ValueError` in  `annotation_value_list_element_type` is now dead code because here we ensures `_annotation_value_list_element_type` is never called with an empty list.

linglp (approved):
Hi @andrewelamb ! I could approve too. I found some edge cases of null values that we could consider to cover, but I am also fine with how you have written here. There's also a small piece of dead code that can be removed.

PR #1418, opened 2026-07-06
[SYNPY-1745] Prep for v5.0.0

andrewelamb:
# **Problem:**

The Synapse Python Client ships two parallel APIs: the original legacy classes/functions (`Entity`, `Project`, `Folder`, `File`, `Link`, `DockerRepository`, `Activity`, `Wiki`, `SubmissionStatus`, `JsonSchemaService`, and many `Synapse` client methods) and the newer object-oriented (OOP) models under `synapseclient.models`. This split created several problems ahead of the planned 5.0.0 removal of the legacy code:

- Many legacy classes and methods were slated for removal but were only marked with inline `# TODO: Deprecate ...` comments — they emitted no deprecation warning and gave users no migration guidance. Affected code included the entity classes in `entity.py`, the evaluation/submission/wiki methods on the `Synapse` client (`getEvaluation`, `getSubmission`, `getWiki`, `getWikiAttachments`, `downloadTableColumns`, etc.), `Activity`/`is_used_entity`/`is_used_url` in `activity.py`, `Wiki`/`WikiAttachment` in `wiki.py`, `SubmissionStatus` in `evaluation.py`, and `download_file_entity` in the download layer.
- Existing deprecation messages (e.g. `JsonSchemaService`, `create_s3_storage_location`, `get_available_services`/`service`) said only "To be removed in 5.0.0" without pointing users to the OOP replacement.
- The OOP models were still framed as "experimental" in the docs (dedicated experimental async/sync reference pages), and the guides, tutorials, and example scripts still demonstrated the legacy APIs.
- Non-deprecated internal code depended on `escape_column_name`/`join_column_names` living in the deprecated `synapseclient.table` module.
- Unit and integration tests for deprecated classes were interleaved with tests for the current OOP code, making it hard to see what will be removed in 5.0.0.

# **Solution:**

- Added `@deprecated` decorators (version 4.14.0, targeting removal in 5.0.0) with explicit replacement guidance to the legacy classes and functions that previously carried only TODO comments — the entity classes, `Activity`/`is_used_entity`/`is_used_url`, `Wiki`/`WikiAttachment`, `SubmissionStatus`, `download_file_entity`, and the evaluation/submission/wiki/`downloadTableColumns`/`create_s3_storage_location` methods on the `Synapse` client. Each message names the OOP replacement (e.g. use `synapseclient.models.Evaluation`, `Submission`, `WikiPage`, `StorageLocation`) and a matching `WARNING`/`Deprecated:` note was added to the corresponding docstrings.
- Enriched already-deprecated items (`JsonSchemaService`, `create_s3_storage_location`, `get_available_services`/`service`, and `escape_column_name`/`join_column_names` in `table.py`) so their messages point to the concrete OOP equivalents.
- Moved `escape_column_name` and `join_column_names` into `synapseclient.core.utils` so current internal code (migration, describe, etc.) can use them without importing the deprecated `synapseclient.table` module; the versions in `table.py` remain but are deprecated and delegate to the new location. Updated internal callers accordingly.
- Removed the "experimental" framing from the OOP models: dropped the experimental async/sync reference pages from `mkdocs.yml` and added the models to the standard reference docs.
- Updated the guides (`data_storage.md`), tutorials, and example scripts (object-oriented POC, `uploadBenchmark.py`, etc.) to use the OOP models instead of the legacy APIs.
- Deleted an unused table protocol and added TODO markers on legacy transfer-progress calls to be replaced by 5.0.0.

# **Testing:**

- Relocated the unit and integration tests that exercise the deprecated classes into dedicated `deprecated/` folders (with a new `conftest.py` for shared fixtures) so they remain in place — and keep covering the deprecated code until it is removed in 5.0.0 — while being clearly separated from the OOP test suite.
- Updated the remaining (non-deprecated) unit and integration tests to use the OOP models instead of legacy code paths, and added/updated unit tests for the relocated `escape_column_name`/`join_column_names` in `unit_test_utils.py`.
- These changes are behavior-preserving: the deprecation decorators emit warnings but do not alter functionality, and the moved utilities produce identical output. Verified via the existing unit test suite (`pytest -sv tests/unit`) and the integration suite.

# **Tickets:**
The current PR already touches a lot of files, but more needs to be done. Additional work in these tickets:

https://sagebionetworks.jira.com/browse/SYNPY-1870
https://sagebionetworks.jira.com/browse/SYNPY-1871
https://sagebionetworks.jira.com/browse/SYNPY-1872
https://sagebionetworks.jira.com/browse/SYNPY-1873
https://sagebionetworks.jira.com/browse/SYNPY-1874
https://sagebionetworks.jira.com/browse/SYNPY-1875
https://sagebionetworks.jira.com/browse/SYNPY-1877
https://sagebionetworks.jira.com/browse/SYNPY-1878
https://sagebionetworks.jira.com/browse/SYNPY-1879


synapseclient/core/upload/multipart_upload.py
```diff
@@ -320,6 +320,7 @@ def _upload_parts(self, part_count, remaining_part_numbers):
                 file_size,
             )
 
+            # TODO: needs to be replaced by 5.0.0
```
BryanFauble: I don't think that there is a need to touch this code here in `synapseclient/core/upload/multipart_upload.py`. If I recall - `synapseclient/core/upload/upload_functions_async.py` and `synapseclient/core/upload/multipart_upload_async.py` are the replacements for this entire lineage of code.

synapseclient/models/file.py
```diff
@@ -883,22 +884,19 @@ async def store_async(
...
+                "etag": self.etag,
+                "name": self.name,
+                "parentId": parent.id if parent else self.parent_id,
+                "dataFileHandleId": self.data_file_handle_id,
+                "modifiedOn": self.modified_on,
+                "versionLabel": self.version_label,
+                "versionNumber": self.version_number,
+                "versionComment": self.version_comment,
+                "concreteType": FILE_ENTITY,
+            }
```
BryanFauble, line 902: ⚠️ 
!! There are several fields here that I think are going to get dropped unintentionally. For example `contentType`, `contentSize` were being set on the previous `Synapse_File` class, however, now those fields are not being put into this dictionary.

I looked at the code of the `store_entity` function and these fields would only get used if passed in at the second `entity` argument (Which we are no longer passing in).

synapseutils/sync.py
```diff
@@ -177,6 +177,13 @@ def syncFromSynapse(
     )
 
 
+@deprecated(
+    version="4.14.0",
+    reason=(
+        "To be removed in 5.0.0. Use StorableContainer.sync_from_synapse instead, "
```
BryanFauble: `StorableContainer` is not something users would directly be using. Instead let's point them to Project/Folder

mkdocs.yml
```diff
@@ -90,7 +90,7 @@ nav:
       - Permissions: reference/permissions.md
       - Core: reference/core.md
       - REST Apis: reference/rest_apis.md
-      - Experimental:
+      - Object-Oriented Models:
```
BryanFauble: My suggestion is to structure this differently. Put all of these under the top `API Reference:` heading, and then moving most of the things currently under the `API Reference:` heading under a `Deprecated Interfaces:` section.

docs/scripts/object_orientated_programming_poc/synapse_project.py
BryanFauble, line 1: These POC files and directory can be deleted.

docs/tutorials/python/annotation.md
```diff
@@ -121,11 +121,7 @@ files in the synapse web UI. It should look similar to:
...
-- [Annotations][synapseclient.Annotations]
 - [File][file-reference-sync]
+- [Folder][folder-reference-sync]
+- [Project][project-reference-sync]
 - [syn.login][synapseclient.Synapse.login]
-- [syn.findEntityId][synapseclient.Synapse.findEntityId]
-- [syn.getChildren][synapseclient.Synapse.getChildren]
-- [syn.get_annotations][synapseclient.Synapse.get_annotations]
-- [syn.set_annotations][synapseclient.Synapse.set_annotations]
-- [syn.store][synapseclient.Synapse.store]
```
BryanFauble, line 131: Nit: For some of these we could link directly to the header on the Project or Folder page. Not a strict requirement, but it might help folks go right to exactly what they want to read more about.

This comment applies across all the tutorial pages.

synapseclient/services/json_schema.py
```diff
@@ -501,11 +501,24 @@ def create_json_schema(
...
 class JsonSchemaService:
     """Json Schema Service
 
+    Deprecated: To be removed in 5.0.0. Use the OOP JSON Schema models instead:
+    synapseclient.models.SchemaOrganization for organization management,
+    synapseclient.models.JSONSchema for creating and retrieving schemas, and the
+    JSON Schema methods on entity models (e.g. File, Folder, Project) such as
+    bind_schema(), get_schema(), validate_schema(), and unbind_schema() for binding
+    and validating schemas against entities.
+
```
BryanFauble, line 518: I'll make the comment once, although it applies to a bunch of the files. I would prefer that we include in the `@deprecated` annotation the important statement that folks need to use. Example:

```
    @deprecated(
        version="4.13.0",
        reason=(
            "Use `synapseclient.operations.download_list_manifest()` instead. "
            "See synapseclient.operations.download_list_operations for the new interface."
        ),
    )
```

And then in the docstring add an example of what it looked like before, compared to what it looks like after migration:

```
        Example: Migration to new method
            &nbsp;

            ```python
            # Old approach (DEPRECATED)
            # manifest_handle = syn._generate_manifest_from_download_list()

            # New approach (RECOMMENDED)
            from synapseclient.operations import download_list_manifest

            syn = Synapse()
            syn.login()

            # Generate and download the manifest CSV
            manifest_path = download_list_manifest()
            ```
```

BryanFauble (changes requested):
Great work. I think the comment in `file.py` is the most pressing out of these.

synapseclient/core/upload/multipart_upload.py
andrewelamb: Nice! I'll go ahead and deprecate everything in here then!

mkdocs.yml
```diff
@@ -71,94 +71,94 @@ nav:
...
+      - Factory Operations: reference/experimental/sync/factory_operations.md
+      - Agent: reference/experimental/sync/agent.md
+      - Project: reference/experimental/sync/project.md
+      - Folder: reference/experimental/sync/folder.md
+      - File: reference/experimental/sync/file.md
+      - Evaluation: reference/experimental/sync/evaluation.md
+      - Submission: reference/experimental/sync/submission.md
+      - SubmissionStatus: reference/experimental/sync/submission_status.md
+      - SubmissionBundle: reference/experimental/sync/submission_bundle.md
+      - Table: reference/experimental/sync/table.md
```
BryanFauble, line 83: Good idea not the mode the .md files yet, because when we do it will change the URLs that they are served at, currently they're the same:

https://synapsepythonclient--1418.org.readthedocs.build/en/1418/reference/experimental/sync/table/
https://python-docs.synapse.org/en/stable/reference/experimental/sync/table/


When we do move these over we can set up redirects for all of the pages in a pretty generic way using wildcards in ReadtheDocs:

Image

BryanFauble (approved):
This is looks really good. Thank you for this work!

PR #1423, opened 2026-07-14
[SYNPY-1851] Add high level function for creating grids

andrewelamb:
# **Problem:**

A data contributor assigned a `CurationTask` had no simple, idempotent way to open the Grid for that task. The only option was `CurationTask.create_grid_session()`, which *always* creates a brand-new session and re-links it to the task, replacing any session already linked. This made it easy to accidentally start over:

- Calling it a second time (e.g. to fix validation errors) would discard the existing session and create a fresh one.
- There was no supported way to fetch a Grid by its `session_id` alone — the `Grid` model could be created but not retrieved from Synapse.

Affected code: `synapseclient/extensions/curator/`, `synapseclient/models/curation.py` (Grid model), and `synapseclient/api/curation_services.py`.

# **Solution:**

Add a high-level, get-or-create entry point plus the lower-level pieces it needs:

- **`synapseclient.extensions.curator.get_curator_grid(task_id, ...)`** — the primary addition. It:
    1. Fetches the `CurationTask`.
    2. Reads the task status for an attached grid session.
    3. If a session is attached, fetches and returns that existing Grid.
    4. If the attached session no longer exists (404), or none is attached, creates a new session, links it to the task, and returns it.

  The first call starts a session; subsequent calls return the same one, so contributors can pick up where they left off without accidentally starting over. `record_set_id` and authorization mode are taken from the task properties automatically.

- **`Grid.get()` / `Grid.get_async()`** — retrieve a Grid session from Synapse by its `session_id` and populate the model via `fill_from_dict()`.

- **`get_grid_session()`** in the API layer — wraps `GET /grid/session/{sessionId}`.

Non-grid execution details on the task raise a `ValueError`; a 404 on the linked session is handled by recreating, while any other HTTP error propagates. Docs (`metadata_contribution.md` and the async/sync curator reference) are updated to steer contributors to `get_curator_grid`, and the cleanup section now shows how to clear the task's stale session reference after `delete()`.

# **Testing:**

- **Unit tests** (`tests/unit/synapseclient/extensions/unit_test_get_curator_grid.py`) covering every branch of `get_curator_grid`: create-when-no-session, get-existing-by-id, recreate-on-404, non-404 error propagation, general error propagation, create-when-execution-details-None, and raise-on-non-grid-execution-details.
- **Integration test** (`tests/integration/.../test_grid_async.py::test_get_grid_session_async`) creating a Grid from a record set and re-fetching it via a fresh `Grid` instance, asserting the round-tripped fields match.


synapseclient/extensions/curator/get_grid.py
```diff
@@ -0,0 +1,142 @@
...
+from synapseclient import Synapse  # type: ignore
+from synapseclient.core.exceptions import SynapseHTTPError  # type: ignore
+from synapseclient.models import (  # type: ignore
+    CurationTask,
+    Grid,
+    GridExecutionDetails,
+)
+
+
+def get_curator_grid(
```
linglp: I think we might want to rename this function as `get_or_create_curator_grid`? `get_curator_grid` sounds like a pure fetch but this function would create a grid if a grid is not available.

docs/guides/extensions/curator/metadata_contribution.md
cconrad8: I'm getting an error for task.name 

should this be something different? is name data_type?

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -266,10 +270,17 @@ This writes the Grid annotation values back to each file as Synapse annotations.
 ### Cleaning up a Grid session
 
 ```python
+# Delete the grid session
```
linglp, line 273: Nit: I think it's very easy to forget unlinking the grid and the task. What about creating something like `CurationTask.delete_active_grid_session_async()` that deletes the Grid and clears the link in one coordinated call in the future?

linglp (approved):
LGTM! I have small design questions but overall it looks good. Thanks for the fixes.

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: @cconrad8 What do you think?

synapseclient/extensions/curator/get_or_create_grid.py
```diff
@@ -0,0 +1,142 @@
...
+    # Step 1: get the CurationTask.
+    client.logger.info(f"Attempting to get CurationTask {task_id}.")
+    task = CurationTask(task_id=task_id).get(synapse_client=client)
+    client.logger.info(f"Got CurationTask {task_id}.")
+
+    # Step 2: check whether a grid session is already attached to the task.
+    client.logger.info(f"Attempting to get status for CurationTask {task_id}.")
+    status = task.get_status(synapse_client=client)
+    client.logger.info(f"Got status for CurationTask {task_id}.")
+
```
BryanFauble: There is quite a lot of logging on the "happy path" scenarios in this function. Are all of them actually needed for the user?

The failure path scenario should be sufficient here for user logged. For example:

```
import asyncio
from synapseclient.api import get_curation_task

syn = Synapse()

async def main():
    await get_curation_task(9999999, synapse_client=syn)

asyncio.run(main())

```

Raises the correct exception already:

`synapseclient.core.exceptions.SynapseHTTPError: 404 Client Error: Task not found: 9999999`

No exception implies this line was successful.

synapseclient/extensions/curator/get_or_create_grid.py
```diff
@@ -0,0 +1,142 @@
...
+from synapseclient import Synapse  # type: ignore
+from synapseclient.core.exceptions import SynapseHTTPError  # type: ignore
+from synapseclient.models import (  # type: ignore
+    CurationTask,
+    Grid,
+    GridExecutionDetails,
+)
+
+
+def get_or_create_curator_grid(
```
BryanFauble, line 20: Could you also create a `get_or_create_curator_grid_async` function too?

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/0058ba08dc57cf5d6d5e0bd18f725637a3937aa2/synapseclient/extensions/curator/schema_management.py#L20-L78

Is a good example of this. That way `get_or_create_curator_grid` just wraps `get_or_create_curator_grid_async`, and you are free to use any async logic within `get_or_create_curator_grid_async`.

The reason why is detailed here: https://python-docs.synapse.org/en/stable/explanations/asyncio_in_python_3_14/


Without also providing these `_async` functions, then Python 3.14 or above would never be able to call these in Jupyter notebooks (Or an asyncio event loop).

docs/guides/extensions/curator/metadata_contribution.md
BryanFauble: Alternative is to make `get_or_create_curator_grid` a context manager that handled this automatically with a boolean flag to not clean up.

That way in the finally you can handle this for the user (The following is not correct code, but just mocking up what it roughly looks like).

```
try:
    yield latest_grid
finally:
    if cleanup:
        latest_grid.delete()
        
        # Remove the deleted session's reference from the task, so the task no longer
        # points at a session that no longer exists
        status = CurationTask(task_id).get_status()
        status.execution_details = None
        curation_task.update_status(curation_task_status=status)
```

synapseclient/extensions/curator/get_or_create_grid.py
```diff
@@ -0,0 +1,142 @@
...
+                f"Linked grid session {active_session_id} no longer exists; "
+                "creating a new one."
+            )
+
+    # Step 4: no usable session attached, so create one, attach it, and return it.
+    client.logger.info(
+        f"CurationTask {task_id} has no attached grid session; "
+        "attempting to create one."
+    )
+    grid = task.create_grid_session(
```
BryanFauble: nit: This would double log `if active_session_id` is True and you get back an HTTP 404 with basically the same message.

tests/unit/synapseclient/extensions/unit_test_get_or_create_curator_grid.py
```diff
@@ -0,0 +1,267 @@
...
+)
+
+grid_module = sys.modules["synapseclient.extensions.curator.get_or_create_grid"]
+
+
+def _build_mock_client():
+    """A Synapse-spec mock with a usable logger."""
+    client = Mock(spec=Synapse)
+    client.logger = Mock()
+    return client
```
BryanFauble: This is a great candidate to move to `tests/unit/conftest.py`

BryanFauble (approved):
There are some changes to make before merging, approving now though.

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: We can add this as part of another ticket if requested.

PR #1424, opened 2026-07-14
[SYNPY-1871] Moved functions from annotations.py to where they are used/utils

andrewelamb:
# **Problem:**

- `synapseclient/annotations.py` had grown into a catch-all module holding annotation-conversion helpers that were only used by a single caller each, alongside functions that are genuinely shared. This coupling made the module hard to reason about and left several helpers exposed as public API when they are really internal implementation details.
- Affected code:
    - Submission-status-specific converters (`to_submission_status_annotations`, `to_submission_annotations`, `is_submission_status_annotations`) lived in `annotations.py` but were only consumed by `models/submission_status.py`.
    - The generic list-conversion helper `_convert_to_annotations_list` lived in `annotations.py` but is a general utility used by `api/annotations.py`.
    - The `ANNO_TYPE_TO_FUNC` type-dispatch table lived in `annotations.py` but was only used by `models/annotations.py`.
- Part of the effort to slim down the legacy `annotations.py` module ahead of the 5.0.0 release.

# **Solution:**

- Moved each helper to the module that actually uses it, and deprecated the now-relocated public functions in `annotations.py` (targeting removal in 5.0.0):
    - Submission-status converters moved into `models/submission_status.py` as private helpers: `_is_submission_status_annotations`, `_to_submission_status_annotations`, `_to_submission_annotations`. Call sites in `SubmissionStatus` updated accordingly.
    - `_convert_to_annotations_list` moved to `core/utils.py` and promoted to the public `convert_to_annotations_list`, along with two supporting private helpers, `_annotation_value_list_element_type` and `_is_missing_annotation_value`. `api/annotations.py` now imports from `core/utils.py`.
    - `ANNO_TYPE_TO_FUNC` moved into `models/annotations.py` as the private `_ANNO_TYPE_TO_FUNC` (built with a `defaultdict` that raises on unknown types).
- Added `@deprecated` markers to the functions left behind in `annotations.py` (`is_synapse_annotations`, `is_submission_status_annotations`, `to_submission_status_annotations`, `to_submission_annotations`, `set_privacy`, `to_synapse_annotations`, `from_synapse_annotations`, `convert_old_annotation_json`), each pointing to its replacement or noting no replacement is provided.
- `from_submission_status_annotations` and `check_annotations_changed` are still used by `Synapse.store_async`, so they are left in place with a TODO referencing [SYNPY-1874](https://sagebionetworks.jira.com/browse/SYNPY-1874) to deprecate them once that dependency is removed.

# **Testing:**

- Added `tests/unit/synapseclient/models/unit_test_submission_status.py` covering the relocated submission-status helpers: type categorization, boolean/date coercion, the `is_private` flag, idempotency, scalar-to-list wrapping, and the nested submission-annotation format.
- Added `TestConvertToAnnotationsList`, `TestAnnotationValueListElementType`, and `TestIsMissingAnnotationValue` to `tests/unit/synapseclient/core/unit_test_utils.py`, covering type inference (including numpy/bool/int subclass edge cases), mixed-type fallback to STRING, and the full matrix of missing-value handling (scalars, empty lists, all-missing lists, mixed lists, and the numpy/decimal/pandas sentinels).

[SYNPY-1874]: https://sagebionetworks.jira.com/browse/SYNPY-1874?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


synapseclient/core/utils.py
```diff
@@ -1624,3 +1624,123 @@ def coerce_enum_list(enum_class: type[E], values: list[Union[E, str]]) -> list[s
...
+        elif issubclass(element_cls, float):
+            nested_annos[key] = {"type": "DOUBLE", "value": [str(e) for e in elements]}
+        elif issubclass(element_cls, (datetime.date, datetime.datetime)):
+            nested_annos[key] = {
+                "type": "TIMESTAMP_MS",
+                "value": [str(to_unix_epoch_time(e)) for e in elements],
+            }
+        else:
+            nested_annos[key] = {"type": "STRING", "value": [str(e) for e in elements]}
+    return nested_annos
```
andrewelamb, line 1753: These have been copied from `synapseclient/annotations.py` and are only used in `synpaseclient.api/annotations.py`

synapseclient/models/submission_status.py
```diff
@@ -838,3 +840,159 @@ async def main():
...
+                "value": [to_unix_epoch_time(v) for v in value_list],
+            }
+        else:
+            # Default to string representation
+            submission_annos["annotations"][key] = {
+                "type": "STRING",
+                "value": [str(v) for v in value_list],
+            }
+
+    return submission_annos
```
andrewelamb, line 998: These have been copied from `synapseclient/annotations.py` and are only used here.

synapseclient/core/utils.py
```diff
@@ -1633,6 +1633,126 @@ def coerce_enum_list(enum_class: type[E], values: list[Union[E, str]]) -> list[s
...
+    # NaN is the only value not equal to itself. This catches NaN of any numeric type
+    # (float, numpy float16/32/64, Decimal) without calling math.isnan on non-numbers,
+    # which return False here since any normal object equals itself.
+    if value != value:
+        return True
+    # Reached only for normal scalars, so the comparison is unambiguous
+    return value == ""
+
+
+def convert_to_annotations_list(
```
linglp, line 1693: Nit: I am a bit curious why we you moved the original `_convert_to_annotations_list` to `utils.py` instead of `api/annotations.py`. `convert_to_annotations_list` doesn't seem to be used anywhere else?

synapseclient/models/annotations.py
```diff
@@ -1,18 +1,41 @@
...
-from synapseclient.annotations import ANNO_TYPE_TO_FUNC
 from synapseclient.api import set_annotations_async
 from synapseclient.core.async_utils import async_to_sync
+from synapseclient.core.utils import from_unix_epoch_time
 from synapseclient.models.protocols.annotations_protocol import (
     AnnotationsSynchronousProtocol,
 )
 
 
+def _identity(x):
```
linglp, line 17: `def _identity`, `def _raise_anno_type_error`, and `_ANNO_TYPE_TO_FUNC` are all in `synapseclient/annotations.py` before. This move makes sense.

linglp (approved):
LGTM overall! Just a small question

synapseclient/core/utils.py
andrewelamb: You're right; that is true for now. There are some other functions that will use it, but I couldn't deprecate them because they are used by non-deprecated functions.

PR #1425, opened 2026-07-15
[SYNPY-1870] Rewrite test_caching.py to not use deprecated legacy code

andrewelamb:
# **Problem:**

The integration test `tests/integration/synapseclient/core/test_caching.py` was written against the legacy pre-OOP API and emitted a large number of `DeprecationWarning`s (all slated for removal in 5.0.0) on every run. Running the test produced warnings for the legacy `Project`/`File`/`Entity` classes, `Synapse.store`, `Synapse.get`, `Synapse.getChildren`, and — indirectly, through the store path — `Synapse.findEntityId`, along with the internal legacy machinery those calls fanned out to (`_getEntityBundle`, `_check_entity_restrictions`, `_getWithEntityBundle_async`, `Annotations`/`set_annotations`, `download_file_entity`).

While migrating the test off the deprecated public API, the `findEntityId` warning turned out to originate in production code, not the test: `synapseclient/models/services/search.py::get_id()` (used by every model `store`/`get` for name+parent resolution) still called the deprecated `Synapse.findEntityId` via an `asyncio` executor shim, carrying a `# TODO: SYNPY-1623` marker. So the OOP models themselves were still routing through a deprecated method.

# **Solution:**

Migrated the test to the object-oriented models and operations layer, and fixed the underlying production code path so the models no longer depend on a deprecated method.

- `tests/integration/synapseclient/core/test_caching.py`:
    - `Project`/`File` now come from `synapseclient.models` instead of the legacy root classes.
    - `syn.store(...)` / `syn.get(...)` replaced with the model methods `entity.store(synapse_client=syn)` / `entity.get(synapse_client=syn)` (and `File(id=id).get(...)` for fetch-by-id).
    - `syn.getChildren(...)` replaced with `synapseclient.api.get_children` driven through `wrap_async_generator_to_sync_generator` (the same sync-consumption helper the `StorableContainer` mixin uses), since the workers run in sync executor threads. This preserves the lightweight "list child IDs" behavior rather than switching to the heavier `sync_from_synapse` (which would download files and change what the test exercises).
- `synapseclient/models/services/search.py`:
    - `get_id()` now calls `find_entity_id_async` (per the deprecation guidance, `from synapseclient.operations import find_entity_id`), replacing the deprecated `findEntityId` + `run_in_executor` shim. The async variant is used because `get_id` is a coroutine; it is imported lazily to avoid an `operations → models → search` circular import (mirroring how `find_entity_id_async` itself lazily imports `api.get_child`).
    - Removed the now-unused `import asyncio` and the SYNPY-1623 TODO, and updated the docstring reference. Behavior is equivalent: `find_entity_id_async` resolves the parent via `id_of()` and calls `api.get_child`, which supports `parentId=None` for the Project-by-name lookup — matching the old `findEntityId(name, parent)` contract.

This resolves the SYNPY-1623 tech debt as a side effect. `get_id()` has callers across `project`, `folder`, `file`, `link`, `recordset`, and `table_components`; the signature and return contract are unchanged, so all callers are unaffected.

# **Testing:**

- Ran the affected unit tests before the production change to confirm the failure mode: 7 tests in `unit_test_project_async.py` / `unit_test_folder_async.py` failed with `pytest_socket.SocketBlockedError` because they mocked `Synapse.findEntityId`, which `get_id` no longer calls (the real `get_child` attempted a network call).
- Updated those tests to mock the new dependency (`patch("synapseclient.operations.find_entity_id_async", new_callable=AsyncMock, ...)`) and to include `synapse_client=self.syn` in the corresponding `assert_called_once_with` assertions.
- After the fix: **83 passed** across the two model test files, and **1185 passed** across `tests/unit/synapseclient/models/` and `tests/unit/synapseclient/operations/`.
- Verified that importing the model store/get path with `DeprecationWarning` promoted to an error raises nothing, and confirmed no circular import from the lazy `find_entity_id_async` import.

### DIFF
--- a/synapseclient/models/services/search.py
+++ b/synapseclient/models/services/search.py
@@ -7,21 +7,49 @@ from synapseclient.core.exceptions import SynapseNotFoundError
 from synapseclient.models.services.storable_entity_components import FailureStrategy
 
 if TYPE_CHECKING:
-    from synapseclient.models import File, Folder, Project
+    from synapseclient.models import (
+        Dataset,
+        DatasetCollection,
+        EntityView,
+        File,
+        Folder,
+        Link,
+        MaterializedView,
+        Project,
+        RecordSet,
+        SubmissionView,
+        Table,
+        VirtualTable,
+    )
 
 
 async def get_id(
-    entity: Union["Project", "Folder", "File"],
+    entity: Union[
+        "Dataset",
+        "DatasetCollection",
+        "EntityView",
+        "File",
+        "Folder",
+        "Link",
+        "MaterializedView",
+        "Project",
+        "RecordSet",
+        "SubmissionView",
+        "Table",
+        "VirtualTable",
+    ],
     failure_strategy: Optional[FailureStrategy] = FailureStrategy.RAISE_EXCEPTION,
     *,
     synapse_client: Optional[Synapse] = None,
-) -> Union[str, None]:
+) -> Optional[str]:
     """
     Get the ID of the entity from either the ID field or the name/parent of the entity.
     This is a wrapper for the [synapseclient.operations.find_entity_id_async][] function
     that is used in order to search by name/parent.
 
     Arguments:
+        entity: The entity to resolve the ID for. Resolution uses the id field if
+            set, otherwise the name and parent_id fields.
         failure_strategy: Determines how to handle failures when getting the entity
             from Synapse and an exception occurs. Only RAISE_EXCEPTION and None are
             supported.
--- a/synapseclient/models/services/search.py
+++ b/synapseclient/models/services/search.py
@@ -1,6 +1,5 @@
 """Functional interface for searching for entities in Synapse."""
 
-import asyncio
 from typing import TYPE_CHECKING, Optional, Union
 
 from synapseclient import Synapse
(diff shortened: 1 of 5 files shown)


tests/integration/synapseclient/core/test_caching.py
```diff
@@ -265,13 +271,18 @@ def sleep_for_a_bit() -> int:
 
 def get_all_ids_from_Project(syn: Synapse, project: Project):
```
linglp: I think the convention of the code base is to do all lower case? something like: `get_all_ids_from_project`?

tests/integration/synapseclient/core/test_caching.py
```diff
@@ -207,10 +213,10 @@ def thread_keep_storing_one_File(syn: Synapse, project: Project, schedule_for_cl
 
         if stored is not None:
             myPrecious = stored
-        elif "id" in myPrecious:
+        elif myPrecious.id is not None:
             # only attempt to retrieve if the entity was initially saved above without encountering a 412 error
             # and thus has a retrievable synapse id
-            myPrecious = syn.get(myPrecious)
+            myPrecious = myPrecious.get(synapse_client=syn)
```
linglp: Nit: I think maybe it also makes sense to use the snakecase to be consistent with the rest of the code base

synapseclient/models/services/search.py
linglp: Do we know if the type hints here are still accurate? should this match `await find_entity_id_async`?

linglp, line 26: Nit: Also, I don't think there's a dedicated unit test for this function?

linglp (approved):
LGTM! Just some minor comments.

BryanFauble (approved):

PR #1426, opened 2026-07-15
[SYNPY-1486] Route all progress bars through a silent-aware factory

BryanFauble:
# **Problem:**

- tqdm progress bar creation was scattered across the client (upload, download, MD5, multipart, async jobs, tables, SFTP/S3 wrappers). Each call site constructed its own bar with inconsistent options and, critically, did not consistently honor the client's `silent` flag.
- Some long-running operations (notably download-side MD5 calculation) showed no progress at all, leaving users to assume the script was "stuck".

# **Solution:**

- Added a single `create_progress_bar()` factory in `core/transfer_bar.py` that honors the resolved client's `silent` flag via `disable=`. It returns a real (possibly disabled) tqdm so call sites can mutate `desc`/`total` and call `update`/`close` unconditionally.
- Routed all live bar creation through the factory: `client._waitForAsync`, `asynchronous_job` job-wait, `storable_container` bulk upload, four `table_components` table-op bars, SFTP/S3 upload wrappers, and both multipart-upload-async fallbacks.
- Added a download-side "Calculating MD5" bar (real `os.path.getsize` total) so it mirrors the upload side; `md5_for_file_hex` gained a `progress_bar` param.
- Sites that start with an unknown total (`_waitForAsync`, SFTP) upgrade the bar to a real `total` once known, so users see a filling bar rather than a spinner.
- The factory tolerates a missing/uncached client by falling back to a visible bar, preserving the behavior of public static methods (`SFTPWrapper.upload_file`, `S3ClientWrapper.upload_file`) callable standalone.
- Already-`@deprecated` legacy bars are left untouched (slated for removal in 5.0.0).

# **Testing:**

- Added `tests/unit/.../unit_test_transfer_bar.py` covering silent/non-silent behavior, mutation of a disabled bar, and the no-client fallback.
- Added a test asserting `md5_for_file_hex` forwards its `progress_bar`; updated affected download tests to mock `os.path.getsize`.
- I ran the benchmarking uploading/download scripts to verify behavior.

### DIFF
--- a/synapseclient/core/transfer_bar.py
+++ b/synapseclient/core/transfer_bar.py
@@ -127,6 +127,54 @@ def _is_context_managed_download_bar() -> bool:
     return getattr(_thread_local, "progress_bar_download_context_managed", False)
 
 
+def create_progress_bar(
+    total: Optional[int],
+    desc: str,
+    unit: str = "it",
+    postfix: str = None,
+    *,
+    synapse_client: Optional["Synapse"] = None,
+) -> tqdm:
+    """Create a progress bar that honors the client's ``silent`` configuration.
+
+    The returned bar is disabled (a no-op that renders nothing) when the client
+    is running in silent mode, so callers can update ``desc``/``total`` and call
+    ``update``/``refresh``/``close`` unconditionally.
+
+    Arguments:
+        total: The total size/count for the progress bar. May be ``None`` when the
+            total is not known until later.
+        desc: The description shown as the progress bar prefix.
+        unit: The unit of measurement (e.g. ``"B"`` for bytes). Defaults to ``"it"``.
+        postfix: Optional postfix string shown after the bar.
+        synapse_client: If not passed in and caching was not disabled by
+            `Synapse.allow_client_caching(False)` this will use the last created
+            instance from the Synapse class constructor. When no client can be
+            resolved the bar is shown (as if ``silent`` were ``False``).
+
+    Returns:
+        A ``tqdm`` progress bar, disabled when the client is silent.
+    """
+    from synapseclient import Synapse
+    from synapseclient.core.exceptions import SynapseError
+
+    try:
+        silent = Synapse.get_client(synapse_client=synapse_client).silent
+    except SynapseError:
+        silent = False
+
+    return tqdm(
+        total=total,
+        desc=desc,
+        unit=unit,
+        unit_scale=True,
+        smoothing=0,
+        postfix=postfix,
+        leave=None,
+        disable=silent,
+    )
+
+
 def get_or_create_download_progress_bar(
     file_size: int,
     postfix: str = None,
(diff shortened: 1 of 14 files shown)


tests/unit/synapseclient/core/unit_test_utils.py
```diff
@@ -568,6 +568,18 @@ def test_md5_for_file(mock_hashlib: MagicMock) -> None:
...
+    with patch.object(utils, "md5_for_file") as mock_md5_for_file:
+        utils.md5_for_file_hex(file_name, progress_bar=progress_bar)
+
+        mock_md5_for_file.assert_called_once_with(
+            file_name, 2 * utils.MB, None, progress_bar=progress_bar
+        )
+
+
 # TODO: remove test in 5.0.0
 class TestSpinner:
```
andrewelamb, line 614: Should we go ahead and deprecate this?

synapseclient/core/transfer_bar.py
```diff
@@ -127,6 +127,54 @@ def _is_context_managed_download_bar() -> bool:
...
+        desc=desc,
+        unit=unit,
+        unit_scale=True,
+        smoothing=0,
+        postfix=postfix,
+        leave=None,
+        disable=silent,
+    )
+
+
```
andrewelamb, line 177: It seems like this might have a problem with empty files:

```
from pathlib import Path
from tqdm import tqdm
from synapseclient.core.transfer_bar import create_progress_bar

def test_md5_for_file_closes_progress_bar_for_empty_file(
    tmp_path: Path, syn: Synapse
) -> None:
    """md5_for_file should close the progress bar even when the file is empty.

    The download code sizes the MD5 progress bar with os.path.getsize, which is
    0 for an empty file. A tqdm bar with total=0 is falsy, so a truthiness check
    on the bar must not be used to guard the final refresh/close.
    """
    # GIVEN an empty downloaded file
    empty_file: Path = tmp_path / "empty_file.txt"
    empty_file.touch()

    # AND a progress bar created exactly as the download code creates it
    with patch.object(syn, "silent", False):
        progress_bar = create_progress_bar(
            total=os.path.getsize(empty_file),
            desc="Calculating MD5",
            unit="B",
            postfix=empty_file.name,
            synapse_client=syn,
        )

    # WHEN the MD5 is calculated
    utils.md5_for_file(filename=str(empty_file), progress_bar=progress_bar)

    # THEN the progress bar is closed (tqdm marks a closed bar as disabled and
    # deregisters it, so it no longer affects the placement of later bars)
    assert progress_bar.disable is True
```

BryanFauble: Good catch!

tests/unit/synapseclient/core/unit_test_utils.py
BryanFauble: The actual spinner is already deprecated, and this would get removed when we remove the Spinner.

andrewelamb (approved):
LGTM!

PR #1427, opened 2026-07-16
[SYNPY-1872] Deprecate migrate functions

andrewelamb:
# **Problem:**
The migrate functions/class in synapseutils needed to be deprecated.

# **Solution:**
These were deprecated, pointing to the new functions/class they should be using.

### DIFF
--- a/synapseutils/migrate_functions.py
+++ b/synapseutils/migrate_functions.py
@@ -600,7 +600,7 @@ def index_files_for_migration(
         syn = Synapse()
         syn.login()
 
-        project = Project(id="syn123")
+        project = Project(id="syn123").get()
         result = project.index_files_for_migration(
             dest_storage_location_id="12345",
             db_path="/tmp/migration.db",
--- a/synapseutils/migrate_functions.py
+++ b/synapseutils/migrate_functions.py
@@ -791,7 +791,7 @@ def migrate_indexed_files(
         syn = Synapse()
         syn.login()
 
-        project = Project(id="syn123")
+        project = Project(id="syn123").get()
 
         # Index the files first
         index_result = project.index_files_for_migration(
--- a/synapseutils/migrate_functions.py
+++ b/synapseutils/migrate_functions.py
@@ -176,7 +176,17 @@ class MigrationResult:
         )
 
         # Inspect the counts of files by migration status
-        print(result.get_counts_by_status())
+        project = Project(id="syn123").get()
+        index_result = my_migration_folder.index_files_for_migration(
+        dest_storage_location_id = "123456",
+        db_path="/path/to/your/migration.db",
+        include_table_files=False,  # Set True if you also want table-attached files
+        )
+
+        index_result.as_csv("/path/to/your/index_results.csv")
+        print(f"Migration index database: {index_result.db_path}")
+        print(f"Indexed counts by status: {index_result.counts_by_status}")
+
         ```
     """
 
--- a/synapseutils/migrate_functions.py
+++ b/synapseutils/migrate_functions.py
@@ -8,6 +8,8 @@ import traceback
 import typing
 from enum import Enum
 
+from deprecated import deprecated
+
 import synapseclient
 from synapseclient.core import pool_provider, utils
 from synapseclient.core.constants import concrete_types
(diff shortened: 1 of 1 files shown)


linglp (approved):
But I noticed that there's minor inconsistentcy if you check `class StorageLocationConfigurableSynchronousProtocol(Protocol)`:
```
  def index_files_for_migration(
      self,
      dest_storage_location_id: int,   # <-- typed as int
      ...
```
this is marked as integer but here it seems like this is a string? And also, the example in the protocol also uses integer instead of string.

andrewelamb:
> But I noticed that there's minor inconsistentcy if you check `class StorageLocationConfigurableSynchronousProtocol(Protocol)`:
> 
> ```
>   def index_files_for_migration(
>       self,
>       dest_storage_location_id: int,   # <-- typed as int
>       ...
> ```
> 
> this is marked as integer but here it seems like this is a string? And also, the example in the protocol also uses integer instead of string.

Was this comment meant for a different PR?

linglp:
@andrewelamb It was meant for this PR here! I saw that you have the following example: 
```
        # The new MigrationResult is returned by the migration methods on the
        # Project and Folder models
        result = Project(id="syn123").index_files_for_migration(
            dest_storage_location_id="12345", --> string 
        )

```
as the new example. But when I checked `class StorageLocationConfigurableSynchronousProtocol(Protocol)`, I saw: 
```
    def index_files_for_migration(
        self,
        dest_storage_location_id: int, --> int
        db_path: Optional[str] = None,
        *,
        source_storage_location_ids: Optional[List[int]] = None,
        file_version_strategy: str = "new",
        include_table_files: bool = False,
        continue_on_error: bool = False,
    ) -> "MigrationResult":
```
Also here `index_files_for_migration_async` marks `dest_storage_location_id` as integer too : 
```
    async def index_files_for_migration_async(
        self,
        dest_storage_location_id: int, 
        db_path: Optional[str] = None,
        *,
        source_storage_location_ids: Optional[List[int]] = None,
        file_version_strategy: str = "new",
        include_table_files: bool = False,
        continue_on_error: bool = False,
    ) -> MigrationResult:
```
Should we update the example docstring in your PR to use integer?

thomasyu888 (commented):
🔥 LGTM and I defer to @danlu1 and @BryanFauble for a final review here.

Dan contributed the new migration features

synapseutils/migrate_functions.py
```diff
@@ -110,6 +146,38 @@ class MigrationResult:
...
+        syn.login()
+
+        # The new MigrationResult is returned by the migration methods on the
+        # Project and Folder models
+        result = Project(id="syn123").index_files_for_migration(
+            dest_storage_location_id="12345",
+        )
+
+        # Inspect the counts of files by migration status
+        print(result.get_counts_by_status())
```
danlu1:
```suggestion
        project = Project(id="syn123").get()
        index_result = my_migration_folder.index_files_for_migration(
        dest_storage_location_id = "123456",
        db_path="/path/to/your/migration.db",
        include_table_files=False,  # Set True if you also want table-attached files
        )

        index_result.as_csv("/path/to/your/index_results.csv")
        print(f"Migration index database: {index_result.db_path}")
        print(f"Indexed counts by status: {index_result.counts_by_status}")

```

Following the tutorial: https://python-docs.synapse.org/en/stable/tutorials/python/migration/

synapseutils/migrate_functions.py
```diff
@@ -479,6 +557,48 @@ def index_files_for_migration(
 
     Returns:
         A MigrationResult object that can be used to inspect the contents of the index or output the index to a CSV for manual inspection.
+
+    Example: Migration to the new approach
+        &nbsp;
+
+        Use the index_files_for_migration method on the Project or Folder models
+        from synapseclient.models instead. The new method is called on the model
```
danlu1, line 575:
```suggestion
        from synapseclient.models.services instead. The new method is called on the model
```

synapseutils/migrate_functions.py
```diff
@@ -479,6 +557,48 @@ def index_files_for_migration(
...
+        # )
+
+        # New approach (RECOMMENDED)
+        from synapseclient import Synapse
+        from synapseclient.models import Project
+
+        syn = Synapse()
+        syn.login()
+
+        project = Project(id="syn123")
```
danlu1:
```suggestion
        project = Project(id="syn123").get()
```

synapseutils/migrate_functions.py
```diff
@@ -623,6 +752,49 @@ def migrate_indexed_files(
...
```
danlu1:
```suggestion
        project = Project(id="syn123").get()
```

danlu1 (approved):
Thanks for working on this. Just a few comments.

PR #1428, opened 2026-07-17
[SYNPY-1880] feat: added per row validation to Grid data class 

linglp:
# **Problem:**
Add the ability to query rows in a Synapse Grid session and retrieve row level validation results

# **Solution:**
**All dataclasses added:** 
Please note that some data classes are renamed to avoid name collision with the data classes in table (i.e. Query -> GridQuery)
GridQuery	[Query](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/Query.html)
QueryRequest	[QueryRequest](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/QueryRequest.html)
GridQueryJobRequest	[GridQueryJobRequest](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/GridQueryJobRequest.html) / [GridQueryJobResponse](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/GridQueryJobResponse.html)
GridQueryResult	[QueryResult](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/result/QueryResult.html)
GridRow	[Row](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/result/Row.html)
SelectColumn	[SelectColumn](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/result/SelectColumn.html)
GridQueryValidationResult	[ValidationResults](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/result/ValidationResults.html)
SelectItem	[SelectItem](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/SelectItem.html)
SelectByName	[SelectByName](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/SelectByName.html)
SelectAll	[SelectAll](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/SelectAll.html)
CountStar	[CountStar](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/function/CountStar.html)
SelectSelection	[SelectSelection](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/SelectSelection.html)
Filter	[Filter](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/Filter.html)
RowValidationResultFilter	[RowValidationResultFilter](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/RowValidationResultFilter.html)
CellValueFilter	[CellValueFilter](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/CellValueFilter.html)
RowSelectionFilter	[RowSelectionFilter](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/RowSelectionFilter.html)
RowIsValidFilter	[RowIsValidFilter](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/RowIsValidFilter.html)
RowIdFilter	[RowIdFilter](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/RowIdFilter.html)
ValidationOperator	[ValidationOperator](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/ValidationOperator.html)
CellValueOperator	[CellValueOperator](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/CellValueOperator.html)
GridReplica	[GridReplica](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/GridReplica.html)
CreateReplicaRequest	[CreateReplicaRequest](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/CreateReplicaRequest.html)

**Replica support**
* Added the create_grid_replica API call (api/curation_services.py, POST /grid/session/{sessionId}/replica)
* Added GridReplica and CreateReplicaRequest dataclasses
* Added Grid.create_replica_async/create_replica methods

**Grid.validate_rows_async / validate_rows**
New method that submits a QueryRequest against a grid session and returns per-row validation results

**Grid.connect_async / connect**
* Added as an async context manager (@asynccontextmanager) and its true sync counterpart 
* Creates the grid session and one replica up front, binds the replica to the Grid instance for the duration of the with/async with block, and clears it on exit


synapseclient/models/curation.py
```diff
@@ -2777,36 +3934,43 @@ def create(
...
+        may create a replica.
 
         Arguments:
-            timeout: The number of seconds to wait for the job to complete or progress
-                before raising a SynapseTimeoutError. Defaults to 120.
             synapse_client: If not passed in and caching was not disabled by
                 `Synapse.allow_client_caching(False)` this will use the last created
                 instance from the Synapse class constructor.
 
         Returns:
```
linglp, line 3734: All these changes are confusing. I did not delete these and they are still there if you use VS code to see.

synapseclient/models/curation.py
```diff
@@ -2816,60 +3980,104 @@ def export_to_record_set(
             syn = Synapse()
             syn.login()
 
-            # Export modified grid data back to the record set
```
linglp, line 2825: All these changes are confusing. I did not delete these and they are still there if you use VS code to see.

synapseclient/models/curation.py
```diff
@@ -2816,60 +3980,104 @@ def export_to_record_set(
...
+            grid = grid.create()
+
+            replica = grid.create_replica()
+            print(f"Replica created with ID: {replica.replica_id}")
             ```
         """
-        return self
+        return None
 
-    def synchronize(
```
linglp, line 2836: Same above. The function is still there.

synapseclient/models/curation.py
```diff
@@ -3837,3 +5049,401 @@ async def main():
...
+        if not replica_data:
+            raise ValueError(
+                f"Replica could not be created for grid session '{self.session_id}': "
+                f"no replica was returned in the Synapse response."
+            )
+        return GridReplica().fill_from_dict(replica_data)
+
+    @skip_async_to_sync
+    @asynccontextmanager
+    async def connect_async(
```
linglp, line 4907: Based on John Hill's suggestion [here](https://sagebionetworks.slack.com/archives/C09ENGKAJ4B/p1784239407106819?thread_ts=1784223033.076109&cid=C09ENGKAJ4B), we will need to use `connect` function to interact with the replica.

synapseclient/models/curation.py
```diff
@@ -3837,3 +5049,401 @@ async def main():
...
+        request = GridQueryJobRequest(
+            session_id=self.session_id,
+            replica_id=self._replica_id,
+            query_request=query_request,
+        )
+        request = await request.send_job_and_wait_async(
+            timeout=timeout, synapse_client=synapse_client
+        )
+
+        if not request.query_result or not request.query_result.rows:
```
linglp, line 5231: Initially I raised it as an error but eventually decided that here it should be a warning. It could be that there's no row to validate or the Filter or SelectColumn didn't return anything.

synapseclient/models/curation.py
```diff
@@ -3837,3 +5049,401 @@ async def main():
...
+        self._replica_id = replica.replica_id
+        try:
+            yield self
+        finally:
+            self._replica_id = None
+
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: f"Grid_Validate_Rows_Session_ID: {self.session_id}"
+    )
+    async def validate_rows_async(
```
linglp, line 5115: This is also important to review. Examples in the docstring demonstrated how you could use this function with `connect`.

synapseclient/models/curation.py
```diff
@@ -2847,29 +3921,163 @@ def synchronize(
...
+            grid = Grid(initial_query=query)
+            grid = grid.create()
+
+            # Synchronize the grid with the latest state of the file view
+            grid = grid.synchronize()
+            ```
+        """
+        return self
+
+    def create_replica(
```
linglp: Also an important function to review. Creating replica is a pre-requisite of `validate_rows`

synapseclient/models/curation.py
```diff
@@ -2280,6 +2292,990 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
+            columns that will be returned by this query (e.g. SelectAll,
+            CountStar).
+        filters: Each filter must be a complete JSON object with the required
+            'concreteType' property. Multiple filters are combined with AND logic.
+        limit: Limit of the number of rows returned to avoid loading more data
+            than needed into your context window.
+        offset: Specifies where the first returned row begins in the result set.
+        include_validation_messages: Controls whether the
+            'allValidationMessages' array appears in the response. Defaults to
+            false to conserve token usage.
```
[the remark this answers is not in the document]
linglp: This is just to match the synapse documentation

synapseclient/models/__init__.py
```diff
@@ -106,7 +122,23 @@
...
+    "CountStar",
+    "SelectSelection",
+    "Filter",
+    "RowValidationResultFilter",
+    "CellValueFilter",
+    "RowSelectionFilter",
+    "RowIsValidFilter",
+    "RowIdFilter",
+    "ValidationOperator",
+    "CellValueOperator",
```
BryanFauble: This is where we are going to severely pollute the `synapseclient.models` namespace. My opinion is that these should not be exposed as this level. It would be acceptable to have users import this from `synapseclient.models.curation` or, perhaps even better, move the new dozen objects into `synapseclient.models.grid` so they're logically separated.

The reason why I bring this up is that by exposing all of these at this level it is going to be confusing to have the Table schema nouns (Column, ColumnType), and verbs (query), in addition to grid specific nouns (SelectAll, Filter) at the same level when they're not interchangeable.

synapseclient/models/curation.py
```diff
@@ -3837,3 +5058,401 @@ async def main():
...
+    @asynccontextmanager
+    async def connect_async(
+        self,
+        *,
+        attach_to_previous_session: bool = False,
+        timeout: int = 120,
+        synapse_client: Optional[Synapse] = None,
+    ) -> AsyncGenerator["Grid", None]:
+        """
+        Connects to a grid session for a record set and binds a single replica to
```
andrewelamb: What about grids connected to fileviews?

synapseclient/models/__init__.py
linglp: Thanks Bryan. I created this ticket: https://sagebionetworks.jira.com/browse/SYNPY-1888 to track. For this PR, I could have users import from `from synapseclient.models.curation`

synapseclient/models/curation.py
```diff
@@ -3837,3 +5058,401 @@ async def main():
...
+        )
+
+        if not self.session_id:
+            self.create(
+                attach_to_previous_session=attach_to_previous_session,
+                timeout=timeout,
+                synapse_client=synapse_client,
+            )
+
+        replica = self.create_replica(synapse_client=synapse_client)
```
andrewelamb: Shouldn't this be checking if a replica already exists for this session before creating a new one?

synapseclient/models/curation.py
```diff
@@ -3837,3 +5058,401 @@ async def main():
...
+        )
+
+        if not request.query_result or not request.query_result.rows:
+            client = Synapse.get_client(synapse_client=synapse_client)
+            client.logger.warning(
+                f"Validation job for grid session '{self.session_id}' completed but "
+                "did not return any row validation results. This grid may not have "
+                "any rows matching the query."
+            )
+            return None
```
andrewelamb: You may just want to return the query result here or raise an exception. Otherwise, when users do `for row in query_result.rows:` like in the example, they'll get `AttributeError: 'NoneType' object has no attribute 'rows'`.

synapseclient/models/curation.py
andrewelamb: Should these be private? I don't see why users would want to interact with replicas.

synapseclient/models/curation.py
linglp: When I looked at the front-end code [here](https://github.com/nickgros/synapse-web-monorepo/blob/1b79db9d8fec80fd17b10463de8aff72257bd1bd/packages/synapse-react-client/src/components/DataGrid/useInitializeGridConnection.ts#L82-L91), they always create a brand-new replica regardless of whether the session is new or pre-existing, and they also don't check if the replicas exist for this session or not. I think here we should match the front-end behavior.

synapseclient/models/curation.py
linglp: I think grids connected to file view would work too. I will update the docstring.

linglp:
@andrewelamb  Per Nick's comment [here](https://sagebionetworks.slack.com/archives/C09ENGKAJ4B/p1784223988338289?thread_ts=1784223033.076109&cid=C09ENGKAJ4B), I think maybe there's no harsh enforcement of the replica limit? But based on my understanding of John Hill's comment [here](https://sagebionetworks.slack.com/archives/C09ENGKAJ4B/p1784232513996449?thread_ts=1784223033.076109&cid=C09ENGKAJ4B), we shouldn't create replica IDs wastefully (i.e. per function call). There's no way from the client side at least to "delete" a replica, and every replica created just becomes permanent part of the document (see his comment [here](https://sagebionetworks.slack.com/archives/C09ENGKAJ4B/p1784225303209679?thread_ts=1784223033.076109&cid=C09ENGKAJ4B)) 

To answer your second question: the idea here is that we reuse one replica for the lifetime of one with block — every `validate_rows/validate_rows_async` call made inside that block shares the same replica ID. When John said that the replica ID should be re-used in a session [here](https://sagebionetworks.slack.com/archives/C09ENGKAJ4B/p1784232513996449?thread_ts=1784223033.076109&cid=C09ENGKAJ4B), I think he meant within one `with block`, and not separate scripts or the whole Python process.

synapseclient/models/curation.py
linglp: This is a good point. Let me ask John and others in the slack channel.

synapseclient/models/curation.py
```diff
@@ -2280,6 +2292,999 @@ def to_synapse_request(self) -> Dict[str, Any]:
         return request_dict
 
 
+@dataclass
+class SelectColumn:
```
andrewelamb, line 2302: You might want to make most/all of the dataclasses private. Are users exposed to any of them?

synapseclient/models/curation.py
linglp: I confirmed that we should be using a private method here. I just changed it to private.

synapseclient/models/curation.py
linglp: Users will actually need to use them. See example below: 
```
from synapseclient.models import CurationTask, Grid, QueryRequest, GridQuery, SelectAll

syn = Synapse()
syn.login()


async def create_test_grid():

    async with Grid(record_set_id="syn1234").connect_async() as grid: 
        select_item = [SelectAll()]
        grid_query = GridQuery(
            column_selection = select_item,
        )

        query_request = QueryRequest(query=grid_query)
        query_result = await grid.validate_rows_async(query_request=query_request)
        rows = query_result.rows

        for row in rows:
            validation = row.validation_results
            print(f"Row ID: {row.row_id}, Validation Result: {validation}")
```
`SelectItem` is required to construct `GridQuery` (See API documentation [here](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/query/Query.html)), and `Filter` classes are optional

synapseclient/models/curation.py
linglp: I like the idea of returning query_result! But I think we shouldn't raise an exception because it seems possible that there's no row to validate based on the `Filter` and `SelectColumn` provided by the user.

synapseclient/models/curation.py
```diff
@@ -4346,6 +4102,14 @@ class Grid(EnumCoercionMixin, GridSynchronousProtocol):
...
         print(f"Created grid session: {grid.session_id}")
 
+        # Validate rows. SelectAll() selects every column, so each row's full
+        # data is returned alongside its validation results.
+        with Grid(session_id=grid.session_id).connect() as session:
+            query_request = QueryRequest(query=GridQuery(column_selection=[SelectAll()]))
+            result = session.validate_rows(query_request=query_request)
+            for row in result.rows:
+                print(f"Row ID: {row.row_id}, Validation Result: {row.validation_results}")
+
```
BryanFauble, line 4115:
```
        from synapseclient.models import Grid
        from synapseclient.models.curation import GridQuery, QueryRequest, SelectAll

        syn = Synapse()
        syn.login()

        # Create a new grid session from a record set
        grid = Grid(record_set_id="syn1234567")
        grid = grid.create()
        print(f"Created grid session: {grid.session_id}")

        # Validate rows. SelectAll() selects every column, so each row's full
        # data is returned alongside its validation results.
        with grid.connect() as session: # <---- This is the suggested change
            query_request = QueryRequest(query=GridQuery(column_selection=[SelectAll()]))
            result = session.validate_rows(query_request=query_request)
            for row in result.rows:
                print(f"Row ID: {row.row_id}, Validation Result: {row.validation_results}")
```

synapseclient/models/curation.py
```diff
@@ -4370,7 +4135,14 @@ class Grid(EnumCoercionMixin, GridSynchronousProtocol):
         grid = Grid(initial_query=query)
         grid = grid.create()
 
-        # Work with the grid session...
+        # Validate rows. SelectAll() selects every column, so each row's full
+        # data is returned alongside its validation results.
+        with Grid(session_id=grid.session_id).connect() as session:
```
BryanFauble: Same comment https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1428/changes/6449b72eac79e4f7ae37ee55d7c5b192aa98b442..b853a52fb44aac42cc6f0c0f066f9675bd83cfe3#r3624809952

synapseclient/models/curation.py
```diff
@@ -3837,3 +4830,407 @@ async def main():
...
+                self.session_id
+            ).to_synapse_request(),
+            synapse_client=synapse_client,
+        )
+        replica_data = grid_replica.get("replica")
+        if not replica_data:
+            raise ValueError(
+                f"Replica could not be created for grid session '{self.session_id}': "
+                f"no replica was returned in the Synapse response."
+            )
```
BryanFauble, line 4902: This is a great candidate for a unit test, while the overall end-to-end happy path is great for an integration test!

BryanFauble (changes requested):
Let's look to add a few end-to-end integration tests, the rest of the logic looks good to me!

synapseclient/models/curation.py
andrewelamb: The typing isn't quite right then. It says it returns a `GridQueryResult`, but if `GridQueryJobRequest.query_result` is `None`, that is what will be returned. All the examples assume it will not be None as well. I would strongly consider raising an exception here.

synapseclient/models/curation.py
```diff
@@ -2719,6 +3480,84 @@ def fill_from_dict(
...
+class CreateReplicaRequest:
+    """
+    Request to create a new replica. A replica represents an 'in-memory' grid
+    document identified by a unique replicaId.
+
+    This request is modeled from: <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/grid/CreateReplicaRequest.html>
+
+    Attributes:
+        grid_session_id: The ID of the grid session.
+        replica: Information about a replica. Populated after calling the
```
andrewelamb: This doesn't seem to be a real attribute?

synapseclient/models/curation.py
```diff
@@ -2843,33 +3682,166 @@ def synchronize(
...
+    def _create_replica(
+        self, *, synapse_client: Optional[Synapse] = None
+    ) -> "GridReplica":
+        """
+        Creates a new replica for this grid session.
+
+        A grid replica is an in-memory document that represents a 'copy' of the
+        grid. Each replica is identified by a unique replicaId, issued by the
+        'hub'. A user can have more than one replica at a time (i.e. using
+        multiple browser tabs/machines). A user is limited to 10 replicas
```
andrewelamb: I thought we determined this limit didn't exist?

synapseclient/models/curation.py
```diff
@@ -3837,3 +4830,407 @@ async def main():
...
+    async def _create_replica_async(
+        self, *, synapse_client: Optional[Synapse] = None
+    ) -> "GridReplica":
+        """
+        Creates a new replica for this grid session.
+
+        A grid replica is an in-memory document that represents a 'copy' of the
+        grid. Each replica is identified by a unique replicaId, issued by the
+        'hub'. A user can have more than one replica at a time (i.e. using
+        multiple browser tabs/machines). A user is limited to 10 replicas
```
andrewelamb: see above comment

synapseclient/api/curation_services.py
```diff
@@ -284,3 +284,38 @@ async def delete_grid_session(
...
+    synapse_client: Optional["Synapse"] = None,
+) -> dict[str, Any]:
+    """
+    Create a new replica for a grid session. A replica represents an
+    'in-memory' grid document identified by a unique replicaId.
+
+    https://rest-docs.synapse.org/rest/POST/grid/session/sessionId/replica.html
+
+    Note: Only the user that started the grid session may create a replica.
+    A user is limited to 10 replicas per-hour per-grid-session.
```
andrewelamb: I thought we determined this limit doesn't exist?

andrewelamb (changes requested):

BryanFauble (approved):
Thanks for the updates! Nothing additional beyond what has already been flagged.

synapseclient/models/curation.py
linglp: Thanks for the comment, I will update the docstring.

linglp:
For [this](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1428#discussion_r3641216748), I think we should match how UI handles empty grid? Currently, for empty grid, the UI is showing "no validation errors" with green text: 
Screenshot 2026-07-27 at 11 35 20 AM

andrewelamb (approved):
LGTM!

PR #1429, opened 2026-07-21
[SYNPY-1883] convert date to epoch ms

danlu1:
# **Problem:**

Synapse DATE columns store an exact moment in time as an integer, milliseconds, since the Unix epoch. But the client only handled that conversion in one direction:

1. Read side ✅ — _convert_df_date_cols_to_datetime converted epoch integers back into datetime objects when querying.
2. Write side ❌ — nothing converted datetime values into epoch integers before upload. A datetime object (or a formatted date string in a CSV) was passed straight through to the API, which rejected it with a 400 error.

This hit every input type: `pandas DataFrames`, `dict` and `csv` files — and the gap was hidden because the integration test exercising itmanually pre-converted dates with [utils.to_unix_epoch_time before calling store_rows](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/855ef24f92e27a901735614c0b4e863745ddc2da/tests/integration/synapseclient/models/async/test_table_async.py#L1568-L1572), instead of relying on the client to do it.

Also, there is unresolved **timezone ambiguity**. Miss document to explain how different datetime and date, including `tz-aware`, `naive`, `mixed timezone` datetime and plain date,  are converted in the client. Also, csv file columns with mixed UTC offsets had no normalization rule either. 

# **Solution:**
The fix adds the missing write-side conversion, plumbed into every path that sends rows to Synapse.

**New conversion function** [_convert_df_date_cols_to_epoch_time](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L4644-L4679) auto-detects datetime/date columns and converts them via [to_unix_epoch_time](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/core/utils.py#L680-L720) and is called in [DataFrame](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L3846), [csv](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L4775) uploads and [upserts](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L2125). 

**New CSV parameters** [date_columns/date_format](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L3232-L3233) in `store_rows_async` since raw CSVs have no dtype info, the client needs to be told which columns are dates and how to parse them. 
[_parse_df_date_cols_to_datetime](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L4682-L4744) is introduced to parses the strings, [_convert_csv_date_cols_to_epoch_time](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L4746-L4792) converts and writes a temp CSV that's actually [uploaded](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L3821-L3828)in finally.

**Explicit timezone policy**: tz-aware datetimes convert to UTC exactly (recommended); naive datetimes/dates use the upload machine's current local timezone (documented DST caveat in to_unix_epoch_time's docstring; CSV columns with mixed offsets are normalized to UTC.

**New tutorial section** [table.md](https://github.com/Sage-Bionetworks/synapsePythonClient/blob/c6ffc6982ce58b0fecc951ed0976b88630fc2e4d/synapseclient/models/mixins/table_components.py#L3821-L3828) covers tz-aware vs. naive datetimes, plain dates, CSV date columns, and querying back with convert_to_datetime=True.

# **Testing:**

Three new integration tests in `test_table_async.py` upload DataFrames/CSVs with tz-aware, DST-crossing, naive, and date-object columns without any manual pre-conversion, verifying the exact epoch values land correctly .

Unit tests for newly added functions and `store_row_async` have been added.


docs/tutorials/python/table.md
```diff
@@ -216,8 +219,241 @@ Qux2,4,203001,204001,+,False
...
+    )
+
+    # Localize the values so each one carries its own timezone information.
+    # tz_localize looks up the UTC offset in effect on each value's own date:
+    # UTC-8 for the February value, UTC-7 for the October value.
+    df["collected_on"] = df["collected_on"].dt.tz_localize("America/Los_Angeles")
+
+    table.store_rows(values=df)
+
+    ```
```
BryanFauble: All cases like this need to get un-tabbed. It will incorrectly render:

https://synapsepythonclient--1429.org.readthedocs.build/en/1429/tutorials/python/table/#storing-timezone-aware-datetimes-recommended

Image

docs/tutorials/python/table.md
```diff
@@ -216,8 +219,241 @@ Qux2,4,203001,204001,+,False
...
+4        S5  2017-02-14 19:23:00+00:00
+5        S6  2018-10-01 16:30:00+00:00
+```
+</details>
+
+~~~python
+# DATE columns are returned as timezone-aware datetimes in UTC. Use tz_convert to view them on another wall clock:
+results["collected_on"] = results["collected_on"].dt.tz_convert("America/Los_Angeles")
+print(results)
+~~~
```
BryanFauble: nit: Swap to ` character instead of ~

synapseclient/models/mixins/table_components.py
BryanFauble, line 2422: Does this upsert method also need to get a

```
        date_columns: Optional[List[str]] = None,
        date_format: Optional[Union[str, Dict[str, str]]] = None,
```

set of parameters too?

A CSV is valid to be passed into this method.

synapseclient/models/mixins/table_components.py
```diff
@@ -4658,12 +4804,185 @@ def _convert_df_date_cols_to_datetime(
...
+            date_format.get(col) if isinstance(date_format, dict) else date_format
+        )
+        # TODO: add warning for mixed timezones/offsets if the pandas<3.0 pin is ever lifted.
+        #  Mixed offsets will start raising ValueError instead of returning object dtype,
+        # and this spot will need the exception handling back.
+
+        # check if the column holds mixed timezones/offsets by extracting the UTC
+        # offsets; null values have no offset to extract and are dropped so they
+        # don't get counted as a spurious second "offset"
+        offsets = df[col].str.extract(r"([Zz\+\-]\d{2}:?\d{2})$")[0].dropna().unique()
```
BryanFauble:
```suggestion
        offsets = df[col].astype("string").str.extract(r"([Zz\+\-]\d{2}:?\d{2})$")[0].dropna().unique()
```

There is a small edge case here. To recreate this:

```python
import pandas as pd
import io
from synapseclient.models.mixins.table_components import _parse_df_date_cols_to_datetime

syn = Synapse()

# mimic csv_to_pandas_df's read_csv(...).convert_dtypes()
df = pd.read_csv(io.StringIO("col1,date_col\na,\nb,\n")).convert_dtypes()
print(df["date_col"].dtype)  # Int64

res = _parse_df_date_cols_to_datetime(df, date_columns=["date_col"], synapse_client=syn)
print(res["date_col"].dtype)  # datetime64[ns] is what is expected here after patching
# -> AttributeError: Can only use .str accessor with string values!

```

```
    raise AttributeError("Can only use .str accessor with string values!")
AttributeError: Can only use .str accessor with string values!. Did you mean: 'std'?
```

BryanFauble (changes requested):
There are a few small considerations that are required - However, this is looking great.

synapseclient/models/mixins/table_components.py
```diff
@@ -3795,20 +3879,67 @@ async def main():
...
-            and schema_storage_strategy == SchemaStorageStrategy.INFER_FROM_DATA
+        elif isinstance(values, str) and (
+            schema_storage_strategy == SchemaStorageStrategy.INFER_FROM_DATA
+            or date_columns
         ):
-            values = csv_to_pandas_df(filepath=values, **(read_csv_kwargs or {}))
+            # ROW_ID/ROW_VERSION must stay as regular columns so they survive the
+            # round-trip to the temporary upload file when `date_columns` is used —
+            # dropping them would turn a row update into an append.
+            values = csv_to_pandas_df(
```
andrewelamb, line 4026: Shouldn't this use the same parameters as `df.to_csv` at line 4971? For example, this test currently fails:

```
async def test_store_rows_async_csv_date_cols_respects_non_default_separator(self):
        # GIVEN a TAB-separated CSV and a matching csv_table_descriptor telling
        # the client the file is tab-delimited
        table = self.ClassForTest(id="syn123", name="test_table")
        table._last_persistent_instance = self.ClassForTest(
            id="syn123", name="test_table"
        )
        with tempfile.NamedTemporaryFile(
            mode="w", suffix=".csv", delete=False
        ) as csv_file:
            csv_file.write("col1\tdate_col\na\t01/15/2024\nb\t02/20/2024\n")

        uploaded = {}

        def capture_upload(**kwargs):
            # The temp upload file is written with the descriptor's separator,
            # so read it back with the same separator
            uploaded["df"] = pd.read_csv(kwargs["path_to_csv"], sep="\t")

        try:
            with patch.object(
                table, "_chunk_and_upload_csv", new_callable=AsyncMock
            ) as mock_upload:
                mock_upload.side_effect = capture_upload
                # WHEN the tab-delimited file is stored with date_columns
                await table.store_rows_async(
                    values=csv_file.name,
                    date_columns=["date_col"],
                    date_format="%m/%d/%Y",
                    csv_table_descriptor=CsvTableDescriptor(separator="\t"),
                    synapse_client=self.syn,
                )

            # THEN the file is parsed with the tab separator, so date_col is
            # found and converted to epoch ms — rather than the whole header
            # collapsing into a single "col1\tdate_col" column (which raises a
            # "date column(s) not present" ValueError)
            expected_df = pd.DataFrame(
                {
                    "col1": ["a", "b"],
                    "date_col": [
                        1705276800000,
                        1708387200000,
                    ],  # date columns are converted to epoch ms (midnight local timezone, unit tests run with TZ=UTC)
                }
            ).convert_dtypes()
            pd.testing.assert_frame_equal(
                uploaded["df"], expected_df, check_dtype=False
            )
        finally:
            os.remove(csv_file.name)
```

synapseclient/models/mixins/table_components.py
```diff
@@ -4658,12 +4804,185 @@ def _convert_df_date_cols_to_datetime(
...
+
+    df = _convert_df_date_cols_to_epoch_time(df=df)
+
+    fd, temp_path = tempfile.mkstemp(suffix=".csv")
+    os.close(fd)
+    try:
+        df.to_csv(
+            temp_path,
+            index=False,
+            float_format="%.12g",
```
andrewelamb, line 5095: whats the reason for this?

docs/tutorials/python/table.md
```diff
@@ -216,8 +219,241 @@ Qux2,4,203001,204001,+,False
...
+  <summary>The collected_on column now displays timezone-aware datetimes converted to America/Los_Angeles time:</summary>
+```
+  sample_id               collected_on
+0        S1  2017-02-14 19:23:00+00:00
+1        S2  2018-10-01 16:30:00+00:00
+2        S3  2017-02-14 18:23:00+00:00
+3        S4  2017-02-14 07:00:00+00:00
+4        S5  2017-02-14 19:23:00+00:00
+5        S6  2018-10-01 16:30:00+00:00
+```
```
andrewelamb, line 421: Are these correct Los Angeles times?

andrewelamb (changes requested):

docs/tutorials/python/table.md
danlu1: Yes, these are printed results of the above code.

synapseclient/models/mixins/table_components.py
danlu1: This follows the convention of writing csv in the client. Here is the reference: https://github.com/Sage-Bionetworks/synapsePythonClient/blob/8e6954862e455b53656d2959f9c7898c95f3d267/synapseclient/table.py#L2418-L2424

synapseclient/models/mixins/table_components.py
danlu1: The root cause is that the input data isn't being parsed correctly. It only has one column `col1\tdate_col'`

danlu1: This can be solved using      
```               
await table.store_rows_async(
                        values=csv_file.name,
                        date_columns=["date_col"],
                        date_format="%m/%d/%Y",
                        csv_table_descriptor=CsvTableDescriptor(separator="\t"),
                        read_csv_kwargs={"sep": "\t"},
                        synapse_client=self.syn,
                    )
```

tests/unit/synapseclient/mixins/unit_test_table_components.py
```diff
@@ -39,10 +43,15 @@
...
     _construct_single_key_where_statement,
+    _convert_csv_date_cols_to_epoch_time,
+    _convert_df_date_cols_to_epoch_time,
     _format_primary_key_value_for_where,
+    _is_date_list_column,
+    _parse_df_date_cols_to_datetime,
     _query_table_csv,
     _query_table_next_page,
     _query_table_row_set,
+    _upsert_rows_async,
```
linglp: this is not used anywhere?

tests/unit/synapseclient/mixins/unit_test_table_components.py
```diff
@@ -1,10 +1,13 @@
+import hashlib
```
linglp: this import could also be removed?

synapseclient/models/mixins/table_components.py
```diff
@@ -4658,12 +4898,191 @@ def _convert_df_date_cols_to_datetime(
...
+    """
+    Convert date columns with datetime values, and DATE_LIST columns holding a
+    Python list of date/datetime values per cell, to epoch time in milliseconds.
+
+    Arguments:
+        df: The pandas dataframe.
+    Returns:
+        A dataframe with datetime columns converted to epoch time in milliseconds
+    """
+    import pandas as pd
```
linglp, line 4941: I think usually we also add: `test_import_pandas()` here. See function: `_parse_df_date_cols_to_datetime`

synapseclient/models/mixins/table_components.py
```diff
@@ -4658,12 +4898,191 @@ def _convert_df_date_cols_to_datetime(
...
+                .apply(
+                    lambda cell: to_unix_epoch_time(cell) if pd.notna(cell) else None
+                )
+                .astype("Int64")
+            )
+        elif dtype == "mixed" and _is_date_list_column(df[col]):
+            df[col] = df[col].apply(
+                lambda cell: (
+                    [
+                        to_unix_epoch_time(item) if item is not None else None
```
linglp, line 4958: Here's a little edge case that this might fail: 
```
import pandas as pd
from datetime import date
from synapseclient.models.mixins.table_components import _convert_df_date_cols_to_epoch_time

df = pd.DataFrame({'col': [[date(2020,1,1), '', date(2020,1,3)]]})
result = _convert_df_date_cols_to_epoch_time(df=df)
```

`is not None` check only guards against literal Python None — it does not treat an empty string (or any other falsy-but-non-None value) as "missing." Anyone who represents nulls in a DATE_LIST column with '' instead of None will likely hit this edge case. 

I saw error message: 
```
  File "pandas/_libs/lib.pyx", line 2999, in pandas._libs.lib.map_infer
  File "/Users/lpeng/code/synapsePythonClient/synapseclient/models/mixins/table_components.py", line 4957, in <lambda>
    to_unix_epoch_time(item) if item is not None else None
    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lpeng/code/synapsePythonClient/synapseclient/core/utils.py", line 713, in to_unix_epoch_time
    dt = datetime.datetime.fromisoformat(dt.replace("Z", "+00:00"))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: Invalid isoformat string: ''
```
I am curious if this is expected or should we treat empty string as None? 

Similary, this will also fail: 
```
import pandas as pd
from datetime import date
from synapseclient.models.mixins.table_components import _convert_df_date_cols_to_epoch_time

df = pd.DataFrame({'col': ''})
result = _convert_df_date_cols_to_epoch_time(df=df)
```
and the error message is a bit confusing: 
```
  File "/Users/lpeng/.local/share/virtualenvs/synapsePythonClient-xkh67UHE/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 114, in arrays_to_mgr
    index = _extract_index(arrays)
            ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/lpeng/.local/share/virtualenvs/synapsePythonClient-xkh67UHE/lib/python3.12/site-packages/pandas/core/internals/construction.py", line 667, in _extract_index
    raise ValueError("If using all scalar values, you must pass an index")
ValueError: If using all scalar values, you must pass an index

```

synapseclient/models/mixins/table_components.py
```diff
@@ -4658,12 +4898,191 @@ def _convert_df_date_cols_to_datetime(
...
+        raise ValueError(
+            f"The date column(s) {', '.join(missing_cols)} listed in `date_columns` "
+            "are not present in the data. Please ensure that the date columns "
+            "are already in the dataframe."
+        )
+    for col in date_columns:
+        col_format = (
+            date_format.get(col) if isinstance(date_format, dict) else date_format
+        )
+        # TODO: add warning for mixed timezones/offsets if the pandas<3.0 pin is ever lifted.
```
linglp: create a ticket for the `TODO` and add the ticket number here?

synapseclient/models/mixins/table_components.py
```diff
@@ -4658,12 +4898,191 @@ def _convert_df_date_cols_to_datetime(
...
+    Timezone-naive inputs are converted to timezone-naive; Timezone-aware inputs
+    with constant time offset are converted to timezone-aware. When the offsets in a column
+    differ between rows (e.g. a mix of `-0800` and `-0700` values), the values are normalized to UTC.
+
+    Arguments:
+        df: A pandas dataframe
+        date_columns: The names of the columns holding formatted date strings.
+        date_format: The strftime format of the strings — a single format string
+            applied to every column, or a dict mapping column names to their
+            formats. When `None` the format is inferred by pandas.
```
linglp, line 4992: `date_format=None` which is the default is not tested in the unit test

synapseclient/models/mixins/table_components.py
```diff
@@ -4658,12 +4898,191 @@ def _convert_df_date_cols_to_datetime(
...
+            will use to parse the uploaded file.
+
+    Returns:
+        The path to a temporary CSV file with the date columns converted to epoch
+        time in milliseconds. The caller is responsible for deleting this file.
+    """
+    test_import_pandas()
+
+    descriptor = csv_table_descriptor or CsvTableDescriptor()
+    if not descriptor.is_first_line_header:
```
linglp, line 5082: I am curious why here we would need to raise a value error if there's no header?

synapseclient/models/mixins/table_components.py
danlu1: For the first edge case, I don't think Python users would intentionally mix data types within a single column in Python, so I'm not sure we need to handle that case.
For the second error message, this is irrelevant to my change, it errored out when constructing the dataframe. 
Screenshot 2026-08-09 at 9 01 28 PM

synapseclient/models/mixins/table_components.py
danlu1: Added

synapseclient/models/mixins/table_components.py
danlu1: Done

synapseclient/models/mixins/table_components.py
danlu1: I added this to upsert method already.

synapseclient/models/mixins/table_components.py
danlu1: The data is written to a temp csv in the next few steps: `df.to_csv(temp_path, index=False, ...)` with no header= argument, so pandas defaults to header=True.

BryanFauble (approved):

danlu1:
@andrewelamb when you get a chance, could you please review my comments to your questions and give me a approval stamp? Thanks!

andrewelamb (approved):
LGTM!

PR #1431, opened 2026-07-22
 [SYNPY-1882] preserve file name casing when uploading/downloading data 

danlu1:
## Problem ##

`utils.normalize_path()` called `os.path.normcase()` on every path, which lowercases the entire string on Windows but not on other operating systems. This corrupted derived entity/file names ([SYNR-1534](https://sagebionetworks.jira.com/browse/SYNR-1543)) — e.g., an uploaded file like File_Name.TSV would get its name mangled to lowercase when the path round-tripped through normalize_path, and downloaded files could be renamed on disk to the lowercased form.

## Solution ##
1. Replace the `os.path.normcase()` with `os.path.abspath` from `normalize_path` so path casing is preserved on all platforms. Since Cache previously relied on that implicit lowercasing for case-insensitive matching, added a new `_match_cache_map_key()` helper in `cache.py` that tries an exact match first and only falls back to a normcase-based case-insensitive comparison. 
2. Wired this helper into `Cache.contains()`, `Cache.get()`,  `Cache.add()` (replacing a legacy lowercased key instead of duplicating it), and `Cache.remove()`. On POSIX, os.path.normcase is a no-op, so the fallback is inert there and behavior is unchanged.

Note: Cache.get() (download path) still only matches by path — a rename/move with unchanged content still forces a real re-download, since there's no MD5 fallback on the download side (unlike upload, which self-heals via _needs_upload's MD5 comparison against Synapse's record).

## Testing ##

1. Unit tests in `unit_test_utils.py` confirm normalize_path() and guess_file_name() preserve casing.
2. Unit tests in `unit_test_Cache.py` cover `_match_cache_map_key()`directly and `Cache.contains()/get()/add()/remove()` against legacy lowercased keys and casing-only renames, simulating Windows via a mocked `os.path.normcase`.
3. New integration tests in `test_upload.py` confirm `File.store_async` avoids redundant re-uploads after local renames, moves, and (on POSIX) casing changes, since MD5 comparison against Synapse's record catches these even when the cache misses by path.
4. Expanded integration tests in `test_download.py` exercise the download-side cache-hit/miss behavior for renamed, moved, and re-cased files under both simulated case-sensitive and case-insensitive filesystems.

[SYNR-1534]: https://sagebionetworks.jira.com/browse/SYNR-1534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


thomasyu888 (commented):
🔥 Thanks for taking a look at this, and what an interesting bug fix.  I'll defer to @BryanFauble for final review of this!

tests/integration/synapseclient/core/test_download.py
```diff
@@ -234,6 +233,493 @@ async def test_download_cached_file_to_new_directory(
...
+        Contrast with the upload-side equivalent, Cache.get() has no equivalent
+        md5 fallback for downloads -- it only ever matches by recorded path, so
+        renaming a cached file, even within the same directory, invalidates
+        the stale cache entry and forces a real re-download.
+        """
+        spy_file_handle = mocker.spy(download_functions, "download_by_file_handle")
+
+        # GIVEN a file stored in Synapse
+        original_file_path = utils.make_bogus_data_file()
+        file = await File(
```
BryanFauble: Integration tests with file uploads takes quite a lot of time to run. Looking at your changed logic I don't see any changes that must run through Integration testing. We know the file uploads and downloads already work. That means any RestAPI calls to Synapse can be mocked.

Only client side code changed how we handle the cache and file name changing. We didn't however, change how we Integrate with Synapse.

BryanFauble (approved):
I would remove all integration tests on this change and replace it with 100% branch code coverage via unit tests instead.

Approving ahead of that because the core logic looks good to me.

tests/integration/synapseclient/core/test_download.py
danlu1: Thanks for flagging this — good point. I added these as integration tests because caching alone doesn't determine whether a file gets uploaded or downloaded; the client also falls back to MD5 and modified-time comparisons on a cache miss, and there wasn't an existing integration test dedicated to casing changes on files or directories. Also, the newly added function affect cache `contains`,`get` and `remove` that are used in `store_async` and `get_async`. But that part has a comprehensive test already. 
That said, I take your point that we should only test what actually changed. For this ticket, I'll trim the integration tests down and keep only the case(s) if any that directly demonstrate the bug from the ticket is resolved.

PR #1433, opened 2026-07-24
[SYNPY-1869] Review fixes for SearchIndex entity and supporting functionality

BryanFauble:
# **Problem:**

Review pass over the SearchIndex work on `synpy-1869-search-index`. Four classes of issue:

1. **`store_async` crashes when upserting onto an existing index.** `SearchIndex.store_async` moved off `TableStoreMixin` to its own implementation that calls `services/storable_entity.store_entity()`. That function dereferences `resource._last_persistent_instance.version_label` unconditionally for models that have a `version_number`. A freshly-constructed `SearchIndex(name=..., parent_id=..., defining_sql=...)` that resolves to an already-existing entity via `get_id()` reaches `store_entity()` with `_last_persistent_instance` still `None`, because `merge_dataclass_entities()` copies field values but does not set it. Reproduced as `AttributeError: 'NoneType' object has no attribute 'version_label'`. `store_entity()` also read `resource.force_version`, which `SearchIndex` does not define — `AttributeError: 'SearchIndex' object has no attribute 'force_version'`.

2. **`SearchIndex` missing from the `get_id()` type union** in `models/services/search.py`, though `store_async` now calls it.

3. **Dangling docs cross-references.** The docstrings add mkdocstrings refs to `synapseclient.models.SearchQuery`, `SearchIndexQuery`, `search_dsl.Query`, `TextAnalyzer`, and `SynonymSet`, but no reference page renders `SearchIndex` or any of the search-management resources, so every one of those refs failed to resolve at build time.

4. **Point-in-time comments.** Several comments and docstrings narrated the change rather than describing the code: `"existing code that passes raw dicts keeps working unchanged"`, `"Left untyped because..."`, `"is deliberately not re-exported here"`, `"States added by Synapse after this client was released"`, `"the entity body is sent to Synapse without an `entity` wrapper"`.

5. `SearchIndex` no longer needs table/versioning semantics. 
6. `SearchIndexStatus` and `SearchIndexState` are removed along with it
7.  added `ForwardCompatibleStrEnum`, a string enum base that accepts undeclared values instead of raising

# **Solution:**

- **`services/storable_entity.py`** — hoist `_last_persistent_instance` to a local and guard it before the `version_label` comparison; guard `force_version` with `getattr(..., False)`. Fixed in the shared function rather than at the `SearchIndex` call site: every model routed through `store_entity()` without a persistent instance hits the same crash, so one guard covers all callers instead of one per model.
- **`models/services/search.py`** — add `SearchIndex` to the `get_id()` union and `TYPE_CHECKING` imports.
- **`models/search_index.py`** — `to_synapse_request()` returns the entity body directly instead of a single-key `{"entity": ...}` wrapper that its only caller immediately unwrapped as `to_synapse_request()["entity"]`. The wrapper was inherited from the `TableStoreMixin` bundle2 shape, which this model no longer uses.
- **Docs** — add sync + async reference pages for `SearchIndex` and for the search configuration resources (`SearchConfiguration`, `TextAnalyzer`, `SynonymSet`, `ColumnAnalyzerOverride`, `SearchConfigBinding`), wired into `mkdocs.yml` nav alongside the other model pages. The org-scoped resources have no delete endpoint, so their pages list `store` / `get` / `list` only.
- **Comments** — rewritten as durable descriptions of what the code *is*. `"Left untyped because `from` is a Python keyword"` → `"Untyped because `from` is a Python keyword and cannot be a `TypedDict` field"`; `"States added by Synapse after this client was released are preserved"` → `"A state Synapse returns that is not declared here is preserved as-is"`; and so on.

One deliberate narrowing of the reviewed diff: the seven `search_dsl` `TypedDict` shapes (`Aggregation`, `SourceFilter`, `Highlight`, `FieldCollapse`, `Rescore`, `AnalyzerRef`, `ScalarValue`) are **not** re-exported from `synapseclient.models`, leaving `models/__init__.py` unchanged from the base branch. Exporting seven of the eight DSL types while `Query` stays behind — because the name collides with the existing SQL table-query `Query` model — is a half-API that needs a comment in `__all__` to explain itself. They remain importable as `from synapseclient.models.search_dsl import Query, Aggregation, ...`, which is one consistent import path for the whole DSL. Say the word if you'd rather have them on the top-level namespace and I'll put them back.

# **Testing:**

- `uv run pytest tests/unit` — 3003 passed, 2 skipped.
- Added `test_store_async_upserts_onto_existing_index`, covering the crash in (1). It patches `put_entity` rather than `store_entity` so the real `store_entity()` runs and the guard is actually exercised. Confirmed the test fails without the fix (`AttributeError: 'SearchIndex' object has no attribute 'force_version'`) and again with only the `force_version` guard restored (`AttributeError: 'NoneType' object has no attribute 'version_label'`), then passes with both.
- Updated `test_to_synapse_request` and `test_store_async_posts_entity` for the unwrapped request body.
- `mkdocs build` — every `search`-related autoref warning is gone. Remaining warnings in the build are pre-existing and unrelated to this branch.
- `pre-commit run` — all hooks pass.
- Integration tests not run (require Synapse credentials); `test_search_index_async.py` is unchanged by this PR beyond what the base branch already had.


synapseclient/models/search_index.py
```diff
@@ -368,23 +425,102 @@ async def main():
...
+        response_parts: Optional[List["SearchQueryPart"]] = None,
+        *,
+        job_timeout: int = 600,
+        synapse_client: Optional[Synapse] = None,
+    ) -> "SearchIndexQuery":
+        """Asynchronously query this search index. Unlike a SQL-backed Table, a
+        SearchIndex is queried with the
+        [OpenSearch Query DSL](https://docs.opensearch.org/latest/query-dsl/)
+        carried by a [SearchQuery](https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/search/SearchQuery.html) — not with
+        Synapse SQL. See [Query][https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/search/dsl/Query.html] for the
```
andrewelamb: Shouldn't this link be in round brackets?

synapseclient/models/mixins/enum_coercion.py
```diff
@@ -31,3 +32,31 @@ def __setattr__(self, name: str, value: Any) -> None:
         ):
             value = enum_cls(value)
         super().__setattr__(name, value)
+
+
+class ForwardCompatibleStrEnum(str, Enum):
```
andrewelamb, line 37: Is this used anywhere?

linglp: Had a discussion with Bryan and he said that we should keep it for future usage. This is a convenient class to make sure that when the server adds a new Enum value, the code won't break.

synapseclient/models/search_index.py
```diff
@@ -226,24 +218,19 @@ def to_synapse_request(self) -> Dict[str, Any]:
...
-            "concreteType": concrete_types.SEARCH_INDEX_ENTITY,
-            "versionNumber": self.version_number,
-            "versionLabel": self.version_label,
-            "versionComment": self.version_comment,
-            "isLatestVersion": self.is_latest_version,
             "definingSQL": self.defining_sql,
             "searchConfigurationId": self.search_configuration_id,
         }
         delete_none_keys(entity)
-        result = {"entity": entity}
```
linglp, line 238: these two lines are causing problems in the integration tests. I deleted them.

synapseclient/models/services/storable_entity.py
```diff
@@ -44,14 +44,15 @@ async def store_entity(
...
                 resource.version_label
-                and resource.version_label
-                != resource._last_persistent_instance.version_label
+                and last_persistent_instance
+                and resource.version_label != last_persistent_instance.version_label
             ):
                 # a versionLabel implicitly implies incrementing
                 increment_version = True
-            elif resource.force_version and resource.version_number:
+            elif getattr(resource, "force_version", False) and resource.version_number:
```
linglp, line 55: after deleting `version` and `version` related fields in search index, I have to put a fix here

synapseclient/operations/delete_operations.py
```diff
@@ -574,6 +578,7 @@ async def main():
             CurationTask,
             Grid,
             DockerRepository,
+            SearchIndex,
```
linglp, line 581: Move `SearchIndex` in the block here because it doesn't have version support

synapseclient/models/search_dsl.py
```diff
@@ -0,0 +1,1205 @@
+"""Typed shapes for the OpenSearch query DSL accepted by Synapse.
```
linglp, line 1: @BryanFauble Do we need to add these to the doc?

BryanFauble (commented):
LGTM! Thank you @linglp for pushing this forward.

PR #1434, opened 2026-07-27
[SYNPY-1875] Copy and deprecate copyWiki function

andrewelamb:
# **Problem:**

`synapseutils.copyWiki()` was the only way to copy a wiki tree from one entity to
another, and it is legacy code that sits outside the object-oriented model layer:

- It is a free-standing, sync-only function that takes a `syn` client as its first
  positional argument, so it cannot be used from an async context and does not follow
  the `synapse_client` keyword pattern used everywhere else in the new API.
- It uses camelCase arguments (`entitySubPageId`, `destinationSubPageId`,
  `updateLinks`, `entityMap`) that are inconsistent with the rest of the OOP models.
- It has no counterpart on the `WikiPage` model, so users who have adopted
  `synapseclient.models` had to drop back down to `synapseutils` to copy a wiki —
  the same gap that was closed for `DownloadList`, `sync_to_synapse` /
  `sync_from_synapse`, storage locations, and migration in earlier work.

This is the `copyWiki` half of the ongoing migration of `synapseutils` bulk
operations onto the object-oriented models (SYNPY-1875).

# **Solution:**

Add `WikiPage.copy_async()` (and its generated sync `WikiPage.copy()`) as the
object-oriented replacement for `copyWiki()`, and deprecate the legacy function.

**New API function** — `synapseclient/api/file_services.py`

Added `post_file_handles_copy()`, wrapping `POST /filehandles/copy`, used to copy
wiki attachment file handles to the destination. It batches requests by
`MAX_FILE_HANDLE_PER_COPY_REQUEST` internally so callers do not have to, and it is
re-exported from `synapseclient.api`. This puts the file-handle-copy REST call in the
`api/` layer where the model can reuse it, rather than reimplementing the batching
that `synapseutils.copyFileHandles()` does.

**Deprecation** — `synapseutils/copy_functions.py`

`copyWiki()` is marked `@deprecated(version="5.0.0")` for removal in 6.0.0, with a
docstring migration example pointing at `WikiPage.copy` / `WikiPage.copy_async`. The
function itself is unchanged, so existing callers keep working.

**Supporting changes**

- `synapseclient/models/protocols/wikipage_protocol.py` — sync `copy()` signature
  added so IDE type hints resolve.
- `docs/reference/experimental/sync/wiki.md` and
  `docs/reference/experimental/async/wiki.md` — `copy` / `copy_async` added to the
  `WikiPage` members lists so mkdocstrings generates the anchors and
  `[synapseclient.models.WikiPage.copy]` cross-references resolve.

The orchestration is split into small module-level helpers
(`_validate_and_format_copy_inputs`, `_copy_wiki_pages`,
`_collect_wiki_sub_tree_headers`, `_update_internal_links`,
`_update_synapse_id_references`, `_ensure_destination_has_no_root_wiki`,
`_get_existing_destination_wiki_page`) so each piece is unit-testable without
standing up a full wiki tree.

# **Testing:**

**Unit tests** — `tests/unit/synapseclient/models/async/unit_test_wiki_async.py`

Added a `TestWikiPageCopy` class covering `copy_async` end to end with mocks, plus a
test class per helper. Coverage includes:

- Validation failures: missing `owner_id`, invalid `destination_owner_id`, invalid
  `entity_map` keys and values, non-numeric sub-page IDs.
- Source edge cases: source entity with no wiki returns an empty list, unknown
  `entity_sub_page_id` raises `ValueError`, sub-page ID given for a source with no
  wiki raises `ValueError`, sub-page ID type coercion.
- Destination edge cases: destination that already has a root wiki without
  `destination_sub_page_id`, nonexistent `destination_sub_page_id`, and error
  propagation from the destination-page and header-tree lookups.
- The four rewrite combinations: `update_links` only (the default), `entity_map`
  only, both, and neither.
- Helper behavior: sub-tree collection ordering (parents before children) and
  non-mutation of the input headers, link and Synapse-ID rewriting across every
  copied page, `None` markdown becoming an empty string, and hierarchy preservation
  in `_copy_wiki_pages` including the create-under and write-into destination cases.

**Integration tests** — `tests/integration/synapseclient/models/async/test_wiki_async.py`

Added a `TestWikiPageCopy` class that copies against a real Synapse backend:

- `test_copy_entire_wiki_tree` — full tree copy, verifying hierarchy, markdown,
  attachments, and rewritten internal links.
- `test_copy_wiki_sub_tree` — copying only the sub-tree rooted at a given page.
- `test_copy_wiki_into_existing_destination_page` — copying into an existing
  destination wiki page via `destination_sub_page_id`.
- `test_copy_wiki_from_entity_without_wiki` — source with no wiki returns an empty
  list.


thomasyu888 (commented):
🔥 From a product perspective, the usage of this makes sense.  I wonder if we should take this opportunity to update the parameters from "destination_" to "target_" which would mimic the unix copy, and thoughts from the team about this?

I'll defer to @BryanFauble / @danlu1 for final review. PInging Dan especially as this will be new copy functionality that exists for objects.

synapseclient/api/file_services.py
```diff
@@ -544,3 +544,43 @@ def get_file_handle_for_download(
...
+            <https://rest-docs.synapse.org/rest/org/sagebionetworks/repo/model/file/FileHandleCopyResult.html>.
+            Failed copies include a failureCode of UNAUTHORIZED or NOT_FOUND.
+    """
+    from synapseclient import Synapse
+    from synapseclient.core.constants.limits import MAX_FILE_HANDLE_PER_COPY_REQUEST
+
+    client = Synapse.get_client(synapse_client=synapse_client)
+
+    copy_results = []
+    for start in range(0, len(copy_requests), MAX_FILE_HANDLE_PER_COPY_REQUEST):
```
BryanFauble: This is great to batch up to the single rest api limit for 1 request, however, if there is more than 100 the HTTP requests are sent sequentially.

Since we are in an asyncio event loop you can execute the requests in parallel

https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f5fa5236360ac2664c669c60f23ed48529cdf9a4/synapseclient/models/services/migration.py#L1131

as an example usage pattern

synapseclient/models/wiki.py
```diff
@@ -1368,3 +1374,687 @@ async def get_markdown_file_async(
...
+        markdown_url = await get_markdown_url(
+            owner_id=self.owner_id,
+            wiki_id=self.id,
+            wiki_version=self.wiki_version,
+            synapse_client=synapse_client,
+        )
+        cache_dir = os.path.join(synapse_client.cache.cache_root_dir, "wiki_content")
+        if not os.path.exists(cache_dir):
+            os.makedirs(cache_dir)
+        downloaded_file_path = download_from_url(
```
BryanFauble: This is a path that would be great to create an async version of the function
https://github.com/Sage-Bionetworks/synapsePythonClient/blob/f5fa5236360ac2664c669c60f23ed48529cdf9a4/synapseclient/core/download/download_functions.py#L847

The issue is that this synchronous HTTP call would block the asyncio event loop. Although it's small I'd like for us to use the async HTTPX client for all calls instead of the requests library.

synapseclient/models/wiki.py
```diff
@@ -1368,3 +1374,687 @@ async def get_markdown_file_async(
...
+            old_wiki_headers=old_wiki_headers,
+            source_owner_id=self.owner_id,
+            destination_owner_id=destination_owner_id,
+            destination_wiki_page=destination_wiki_page,
+            destination_sub_page_id=destination_sub_page_id,
+            synapse_client=client,
+        )
+
+        if update_links:
+            client.logger.info("Updating internal links")
```
BryanFauble: Could you update all of the log messages to include more information about the ID that generated the messages. We need to uniquely identify one instance of the message for wiki A compared to the same message for wiki B.

synapseclient/models/wiki.py
```diff
@@ -1368,3 +1374,687 @@ async def get_markdown_file_async(
...
+            new_wikis = _update_internal_links(
+                new_wikis=new_wikis,
+                wiki_id_map=wiki_id_map,
+                source_owner_id=self.owner_id,
+                destination_owner_id=destination_owner_id,
+            )
+            client.logger.info("Done updating internal links.")
+
+        if entity_map:
+            client.logger.info("Updating Synapse references")
```
BryanFauble: This is great debug information. Leave info level for higher level operations than this.

synapseclient/models/wiki.py
BryanFauble: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1434/changes#r3679663255 also applied here, demote to debug level

synapseclient/models/wiki.py
```diff
@@ -1368,3 +1374,687 @@ async def get_markdown_file_async(
...
+                    synapse_client=client
+                )
+                wiki_data = await put_wiki_page(
+                    owner_id=destination_owner_id,
+                    wiki_id=new_wiki_id,
+                    request=new_wiki.to_synapse_request(),
+                    synapse_client=client,
+                )
+                new_wikis[new_wiki_id] = new_wiki.fill_from_dict(synapse_wiki=wiki_data)
+                client.logger.info(f"Stored wiki page {new_wiki_id}")
```
BryanFauble: add owner_id to this message

synapseclient/models/wiki.py
```diff
@@ -1368,3 +1374,687 @@ async def get_markdown_file_async(
...
+            f"{argument_description} must be a numeric wiki page ID or None, "
+            f"got {sub_page_id}."
+        )
+    return str(int(sub_page_id))
+
+
+def _validate_and_format_copy_inputs(
+    owner_id: str | None,
+    destination_owner_id: str,
+    entity_sub_page_id: str | int | float | None,
```
BryanFauble: What is a situation where this would be a float?

synapseclient/models/wiki.py
```diff
@@ -1368,3 +1374,687 @@ async def get_markdown_file_async(
...
+                }
+                collected_headers = [root_header]
+            else:
+                collected_headers.append(wiki_header)
+        elif wiki_header.get("parentId") == sub_page_id:
+            collected_headers = _collect_wiki_sub_tree_headers(
+                wiki_headers=wiki_headers,
+                sub_page_id=wiki_header["id"],
+                collected_headers=collected_headers,
+            )
```
BryanFauble, line 2069: Logging at debug level the ignored pages would be useful

tests/integration/synapseclient/models/async/test_wiki_async.py
```diff
@@ -801,6 +802,358 @@ async def test_get_wiki_header_tree(
...
+        # contains an internal wiki link and an entity reference
+        root_wiki.markdown = (
+            "# Root\n\n"
+            f"See the sub page: {project.id}/wiki/{sub_wiki.id}\n\n"
+            f"Data is stored at {self.OLD_ENTITY_ID}."
+        )
+        root_wiki = await root_wiki.store_async(synapse_client=syn)
+
+        # Allow the wiki header tree to become consistent
+        await asyncio.sleep(5)
```
BryanFauble, line 862: https://sagebionetworks.jira.com/browse/SYNPY-1901

If you want to take on the jira, this would be updated to use the new path. At a minimum it would be nice to have a tenacity wait for consistency instead of sleep statements for just your code here, and then applied to existing files as a part of SYNPY-1901

BryanFauble (approved):

andrewelamb:
> 🔥 From a product perspective, the usage of this makes sense. I wonder if we should take this opportunity to update the parameters from "destination_" to "target_" which would mimic the unix copy, and thoughts from the team about this?


I would say no to this, unless we wanted to do it across the entire project.

synapseclient/models/wiki.py
andrewelamb: I'll create a ticket for this.

andrewelamb: https://sagebionetworks.jira.com/browse/SYNPY-1903

PR #1436, opened 2026-07-27
[SYNPY-1908] Fix multipart upload part retries missing connection exceptions

BryanFauble:
# **Problem:**

A user reported an upload failing with `SynapseUploadFailedException: Part upload failed` only 3 seconds into a 10.5GB transfer — far too fast to have exhausted the intended part-level retry window.

Tracing the retry logic in `multipart_upload_async.py` found the root cause: `_put_part_with_retry` (async path) passes `retry_exceptions=[requests.exceptions.ConnectionError]` to `with_retry_time_based`, but the `session` performing the PUT is an `httpx.Client`, not a `requests.Session`. httpx never raises `requests.exceptions.ConnectionError` — it raises `httpx.ConnectError`, `httpx.ReadTimeout`, `httpx.RemoteProtocolError`, etc. The retry matcher (`_is_retryable`) checks the caught exception's class/class-name against the supplied allowlist, so none of httpx's exception types ever match, and the retry helper's default-filling logic only substitutes a default when the caller's list is empty — it doesn't merge in the correct list when a (wrong) list is already supplied.

Net effect: any genuine transient network error during a part PUT (a real socket/connect/read/timeout failure, not an HTTP status code) skipped the entire time-boxed backoff window and failed the part immediately. The only resilience left was the coarser whole-upload-attempt retry (`MAX_RETRIES = 7` in `_multipart_upload_async`), which restarts the *entire* `UploadAttemptAsync` rather than just retrying the one failed PUT.

While auditing every `retry_exceptions=` call site for the same class of bug, the **sync** upload path (`multipart_upload.py`) turned out to have a milder version of the same gap: it correctly uses `requests.exceptions.ConnectionError` (since it genuinely uses `requests.Session`), but only that one exception type — missing `requests.exceptions.Timeout` and `ChunkedEncodingError`, both of which a slow/stalled part PUT can legitimately raise. The download side (`download_async.py`, `download_threads.py`) and `client.py`'s standard retry params already use the full, correct list.

# **Solution:**

- `synapseclient/core/upload/multipart_upload_async.py`: swapped `retry_exceptions=[requests.exceptions.ConnectionError]` for the existing `RETRYABLE_CONNECTION_EXCEPTIONS` list (already defined in `core/retry.py` and used correctly on the download side). Removed the now-unused `import requests`.
- `synapseclient/core/upload/multipart_upload.py`: widened the sync path's `retry_exceptions=[requests.exceptions.ConnectionError]` to the same `RETRYABLE_CONNECTION_EXCEPTIONS` list, restoring parity with the async path and the rest of the codebase.

# **Testing:**

`tests/unit/synapseclient/core/upload/test_multipart_upload_async.py` (new file) covers each distinct path through `UploadAttemptAsync._handle_part` / `_put_part_with_retry`:
- **Regression test**: `httpx.ConnectError` on the first PUT, success on the second — confirmed this test fails without the fix (the exception propagates immediately with zero retries) and passes with it.
- Retryable HTTP status code (503) → retry → success.
- 403 expired presigned URL → refresh URL → success (confirms the existing URL-refresh path is unaffected).
- Non-retryable exception (`ValueError`) → fails immediately with no retry (confirms the fix doesn't over-widen retry behavior).

`tests/unit/synapseclient/core/upload/unit_test_multipart_upload.py`: added `test_handle_part__timeout_error`, a regression test for the sync path — `requests.exceptions.Timeout` on the first PUT, success on the second. Confirmed this fails without the sync-path fix and passes with it.

Full upload unit test suite (`tests/unit/synapseclient/core/upload/`, 29 tests) passes, and `pre-commit run --all-files` (ruff, isort, black, bandit) passes clean.

### DIFF
--- a/tests/unit/synapseclient/core/upload/test_multipart_upload_async.py
+++ b/tests/unit/synapseclient/core/upload/test_multipart_upload_async.py
@@ -1,0 +1,109 @@
+from unittest import mock
+
+import httpx
+import pytest
+
+from synapseclient.core.upload.multipart_upload_async import (
+    HandlePartResult,
+    UploadAttemptAsync,
+)
+from synapseclient.core.utils import md5_fn
+
+PART_SIZE = 256
+PART_NUMBER = 1
+
+
+def _init_upload_attempt(syn):
+    upload_request_payload = {
+        "concreteType": "org.sagebionetworks.repo.model.file.MultipartUploadRequest",
+        "contentMD5Hex": "abc",
+        "contentType": "application/text",
+        "fileName": "target.txt",
+        "fileSizeBytes": 1024,
+        "generatePreview": False,
+        "storageLocationId": "1234",
+        "partSizeBytes": PART_SIZE,
+    }
+
+    def part_request_body_provider_fn(part_number):
+        return (f"{part_number}" * PART_SIZE).encode("utf-8")
+
+    upload = UploadAttemptAsync(
+        syn,
+        "target.txt",
+        upload_request_payload,
+        part_request_body_provider_fn,
+        md5_fn,
+        False,
+    )
+    upload._upload_id = "123"
+    upload._pre_signed_part_urls = {
+        PART_NUMBER: ("https://foo.com/1", {"a": 1}),
+    }
+    return upload
+
+
+class TestPutPartWithRetry:
+    """Regression coverage for the retry paths inside
+    UploadAttemptAsync._put_part_with_retry. multipart_upload_async.py previously
+    passed requests.exceptions.ConnectionError as the retryable exception allowlist,
+    but the session performing the PUT is an httpx.Client, which never raises that
+    exception type -- so a transient httpx connection error skipped the retry
+    window entirely and failed the part immediately.
+    """
+
+    def test_handle_part__httpx_connection_error_then_success(self, syn):
+        upload = _init_upload_attempt(syn)
+        mock_session = mock.Mock()
+        mock_session.put.side_effect = [
+            httpx.ConnectError("connection refused"),
+            mock.Mock(status_code=200),
+        ]
+
+        with mock.patch.object(syn, "_requests_session_storage", mock_session):
+            result = upload._handle_part(PART_NUMBER)
+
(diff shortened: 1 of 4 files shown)


tests/unit/synapseclient/core/upload/test_multipart_upload_async.py
```diff
@@ -0,0 +1,109 @@
...
+class TestPutPartWithRetry:
+    """Regression coverage for the retry paths inside
+    UploadAttemptAsync._put_part_with_retry. multipart_upload_async.py previously
+    passed requests.exceptions.ConnectionError as the retryable exception allowlist,
+    but the session performing the PUT is an httpx.Client, which never raises that
+    exception type -- so a transient httpx connection error skipped the retry
+    window entirely and failed the part immediately.
+    """
+
+    def test_handle_part__httpx_connection_error_then_success(self, syn):
```
linglp: Nit: I noticed that there are more errors for retry that we could potentially be added here from `RETRYABLE_CONNECTION_EXCEPTIONS`. Should we add them?

tests/unit/synapseclient/core/upload/unit_test_multipart_upload.py
```diff
@@ -293,6 +293,41 @@ def test_handle_part__connection_error(self, syn):
             None,
         )
 
+    def test_handle_part__timeout_error(self, syn):
```
linglp: Nit: `test_handle_part__connection_error` and `test_handle_part__timeout_error` are structurally identical except for the exception type raised on the first call. What about combining them and also add tests for other exceptions listed in: RETRYABLE_CONNECTION_EXCEPTIONS

linglp (approved):
LGTM!

tests/unit/synapseclient/core/upload/test_multipart_upload_async.py
BryanFauble: Good call, parametrized this test across ConnectError, ReadError, ReadTimeout, ConnectTimeout, and RemoteProtocolError.

PR #1440, opened 2026-08-04
[SYNPY-1894] feat: synchronize active grid session 

linglp:
# Problem
A grid session captures the schema version of its bound RecordSet at creation time and never picks up a newer version on its own. Once a data manager registers and binds a new JSON schema version to a RecordSet, contributors' already-open grid sessions have no way to see the new columns. 

The old `Grid.synchronize()` only supports synchronizing recordset-based grid sessions. 

# Solution 
* Relaxed Grid.synchronize()/synchronize_async() to support both recordset-based grids and file-view based grid. 
* Added `CurationTask.synchronize_active_grid_session()/_async()` as the task-level "synchronize the tasks" function: it creates an active grid session first if the task doesn't have one yet, otherwise reuses the existing active session, then delegates to Grid.synchronize().

# Design decision
For FileBasedMetadataTaskProperties, there's only one meaningful synchronize behavior — PULL_PUSH — since file-based grids write annotations directly back to files with no intermediate versioned state to preview; there's nothing to choose. 

For RecordBasedMetadataTaskProperties, the choice between PULL and PULL_PUSH is consequential: PULL_PUSH commits the grid's current state as a new RecordSet version, an action with a lasting, visible side effect for every other collaborator on that RecordSet, while PULL does not commit anything.

If it defaulted to `PULL_PUSH` (which is the implicit default of the synapse server), users might be surprised by the JSON Schema version bump. If it defaulted to `PULL`, users might get an unexpected uncommitted preview when they meant to finalize all their changes and push them. Currently, **the method requires the caller to explicitly state their intent** and will raise an error if sync_type is not provided, rather than silently picking a default and letting the wrong choice surface later as a data surprise.


synapseclient/models/curation.py
```diff
@@ -2096,6 +2189,139 @@ async def main():
             task = cls().fill_from_dict(synapse_response=task_dict)
             yield task
 
+    @otel_trace_method(
```
linglp, line 2206: There's integration test for `grid_synchronize` so I skipped adding integration test for this function `synchronize_active_grid_session_async` and only added a unit test.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -108,7 +109,24 @@ Each option in Step 2 leaves you with a single `curation_task`. Start a new Grid
 latest_grid = curation_task.create_grid_session()
 ```
 
-### Step 4: Download record-based metadata as a local CSV
+### Step 4: Synchronize the grid session to pick up schema updates
```
linglp, line 117: I currently place the synchronization step right after session creation (Step 3), framed around picking up a schema change before starting to edit. There's also a valid case for calling synchronize after importing local CSV edits (Step 6). Please let me know if you would want me to move it. @cconrad8

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -244,11 +262,11 @@ curation_task.set_task_state(state="COMPLETED")
...
-File-based tasks follow the same overall flow as record-based tasks (Steps 1–8 above), with three key differences:
+File-based tasks follow the same overall flow as record-based tasks (Steps 1–9 above), with three key differences:
 
 **No CSV import.** `import_csv` is not currently supported for file-based grids. Instead, you can either:
 
-- Download the CSV (Step 4) as a local reference, make your edits locally, then copy-paste the values back into the Grid UI
+- Download the CSV (Step 5) as a local reference, make your edits locally, then copy-paste the values back into the Grid UI
 - Make edits directly in the Synapse Grid UI — Step 3 prints the session URL (`https://www.synapse.org/Grid:default?sessionId=...`) after creating the session
 
 **Use `synchronize()` instead of `export_to_record_set()`.** After editing in the Grid UI, push your changes back to the underlying files:
```
thomasyu888: For discussing, should we deprecate the export to record set function and push people to use this?

thomasyu888 (commented):
🔥 I'll defer to the team on the fine grained technical review after sage week but on a high level, this looks great - just had a tiny comment.

docs/guides/extensions/curator/metadata_contribution.md
linglp: I actually noticed some problems if we remove `export_to_record_set_async`. As an example, the following is working as expected: 
```
grid = Grid(session_id="MTM4Mzc0")
grid = await grid.import_csv_async(path="test_grid_import.csv")
await grid.export_to_record_set_async()
grid = await task.synchronize_active_grid_session_async(sync_type="PULL")
grid.download_csv(destination="/Users/lpeng/Downloads")
```
But if I remove `export_to_record_set_async` and change `sync_type` to `PULL_PUSH`, then I noticed that I can't actually "push" the changes from my CSV to the grid. I will ask other devs in curator-dev channel after Sage week.

synapseclient/models/curation.py
```diff
@@ -2096,6 +2189,139 @@ async def main():
...
+                    f"{self.task_id}: FileBasedMetadataTaskProperties tasks always "
+                    "use SyncType.PULL_PUSH."
+                )
+            sync_type = SyncType.PULL_PUSH
+        elif isinstance(self.task_properties, RecordBasedMetadataTaskProperties):
+            if not sync_type:
+                raise ValueError(
+                    "sync_type must be provided for RecordBasedMetadataTaskProperties"
+                )
+
```
andrewelamb, line 2324: There should be an else statement here that raises an exception, something like:

"This type of CurationTask doesn't support grids"

We are currently adding CurationTasks, which are compute tasks that have other types of task_properties.

synapseclient/models/curation.py
```diff
@@ -2096,6 +2189,139 @@ async def main():
...
+            client.logger.info(
+                f"No active grid session found for task {self.task_id}. "
+                "Creating a new grid session..."
+            )
+            active_grid_session = await self.create_grid_session_async(
+                synapse_client=synapse_client
+            )
+            active_grid_session_id = active_grid_session.session_id
+        else:
+            active_grid_session_id = status.execution_details.active_session_id
```
andrewelamb: Should there be a guard for when there is no active session id?

synapseclient/models/curation.py
```diff
@@ -2096,6 +2189,139 @@ async def main():
...
+                    f"Ignoring sync_type={sync_type} for CurationTask "
+                    f"{self.task_id}: FileBasedMetadataTaskProperties tasks always "
+                    "use SyncType.PULL_PUSH."
+                )
+            sync_type = SyncType.PULL_PUSH
+        elif isinstance(self.task_properties, RecordBasedMetadataTaskProperties):
+            if not sync_type:
+                raise ValueError(
+                    "sync_type must be provided for RecordBasedMetadataTaskProperties"
+                )
```
andrewelamb, line 2318: sync_type should be coerced into the enum here if it's a string. This will ensure it's one of the correct values before doing any API calls.

andrewelamb (changes requested):
Looking good! I found a few issues we should address.

synapseclient/models/curation.py
linglp: hmm if there's no active session id, then the if statement should already create an active grid session.

andrewelamb: The guard covers `execution_details is None`, but not `execution_details` being present with `active_session_id` set to `None`. In that path curation.py:2330 assigns `None` into `Grid(session_id=None)` and calls `synchronize_async` with a null session. `active_session_id` is declared `str | None = None` at curation.py:411 and is populated via `.get("activeSessionId")` at line 426, so a response with execution details but no active session yields exactly that. See this script as an example:

```
import uuid

from synapseclient.models import (
    CurationTask,
    EntityView,
    FileBasedMetadataTaskProperties,
    Folder,
    GridExecutionDetails,
    Project,
    ViewTypeMask,
)

syn = Synapse()
syn.login()

project = Project(name=f"demo_{uuid.uuid4()}").store()
print(f"Created project {project.id}")

try:
    folder = Folder(name=str(uuid.uuid4()), parent_id=project.id).store()
    entity_view = EntityView(
        name=str(uuid.uuid4()),
        parent_id=project.id,
        scope_ids=[folder.id],
        view_type_mask=ViewTypeMask.FILE.value,
    ).store()

    task = CurationTask(
        data_type=f"demo_data_type_{str(uuid.uuid4()).replace('-', '_')}",
        project_id=project.id,
        instructions="Demo of the missing activeSessionId guard.",
        task_properties=FileBasedMetadataTaskProperties(
            upload_folder_id=folder.id,
            file_view_id=entity_view.id,
        ),
    ).store()
    print(f"Created file-based CurationTask {task.task_id}")

    # Put the task into the state under discussion: executionDetails present,
    # activeSessionId absent. GridExecutionDetails.to_synapse_request() omits
    # activeSessionId when it is None, so the request body's executionDetails is
    # just {"concreteType": "...GridExecutionDetails"}.
    status = CurationTask(task_id=task.task_id).get_status()
    status.execution_details = GridExecutionDetails(active_session_id=None)
    print(f"Sending status update: {status.to_synapse_request()}")
    CurationTask(task_id=task.task_id).update_status(curation_task_status=status)

    status = CurationTask(task_id=task.task_id).get_status()
    print(f"execution_details         : {status.execution_details}")
    print(f"execution_details is None : {status.execution_details is None}")
    if status.execution_details is not None:
        print(f"active_session_id         : {status.execution_details.active_session_id}")

    # The guard in synchronize_active_grid_session only checks
    # `status.execution_details is None`, so this call takes the else-branch and
    # builds Grid(session_id=None) instead of creating a session.
    try:
        grid = CurationTask(task_id=task.task_id).synchronize_active_grid_session()
        print(f"synchronize_active_grid_session returned grid {grid.session_id}")
    except Exception as exc:
        print(f"synchronize_active_grid_session raised {type(exc).__name__}: {exc}")

    print()
    print(
        "Expected: either create a grid session for this task (same as the\n"
        "executionDetails-is-None path), or raise an error naming the task, e.g.\n"
        f'"CurationTask {task.task_id} has grid execution details but no active '
        'session id."'
    )
finally:
    Project(id=project.id).delete()
    print(f"Deleted project {project.id}")
```

synapseclient/models/curation.py
```diff
@@ -2097,6 +2190,154 @@ async def main():
...
+                    f"{sync_type!r} is not a valid SyncType. "
+                    f"Expected one of: {[s.value for s in SyncType]}."
+                ) from exc
+
+        if not self.task_properties:
+            await self.get_async(synapse_client=synapse_client)
+
+        if isinstance(self.task_properties, FileBasedMetadataTaskProperties):
+            if sync_type is not None and sync_type != SyncType.PULL_PUSH:
+                client.logger.warning(
```
andrewelamb, line 2308: I'm thinking this should be a ValueError or a TypeError rather than a warning.

docs/guides/extensions/curator/metadata_contribution.md
```diff
@@ -112,7 +113,24 @@ from synapseclient.extensions.curator import get_or_create_curator_grid
 latest_grid = get_or_create_curator_grid(task_id=curation_task.task_id)
 ```
 
-### Step 4: Download record-based metadata as a local CSV
+### Step 4: Synchronize the grid session to pick up schema updates
+
+A Grid session captures the JSON schema in place at the moment it's created — it does not automatically pick up a newer schema version. If the administrator adds or changes a column in the schema *after* you created your session in Step 3, synchronize the session to pull in the latest schema and data from the RecordSet before you continue working.
+
+For record-based grids, `sync_type` is required — you must explicitly choose `"PULL"` or `"PULL_PUSH"`:
```
andrewelamb: This is a bit unclear to users. `sync_type` is actually optional and defaults to `None`. What happens then? What about file-based grids?

synapseclient/models/curation.py
```diff
@@ -3703,15 +3954,27 @@ def export_to_record_set(
         return self
 
     def synchronize(
-        self, *, timeout: int = 120, synapse_client: Optional[Synapse] = None
+        self,
+        *,
+        sync_type: Optional[SyncType] = None,
```
andrewelamb: Should this include strings as well?

synapseclient/models/curation.py
```diff
@@ -2097,6 +2190,154 @@ async def main():
...
+    @otel_trace_method(
+        method_to_trace_name=lambda self, **kwargs: (
+            f"CurationTask_SynchronizeActiveGridSession: ID: {self.task_id}"
+        )
+    )
+    async def synchronize_active_grid_session_async(
+        self,
+        *,
+        sync_type: Optional[Union["SyncType", str]] = None,
+        synapse_client: Optional[Synapse] = None,
```
andrewelamb, line 2215: Should `timeout` be added as a parameter, to be passed to `create_grid_session_async`?

andrewelamb (changes requested):
I found a few thigns that shoudl be adressed.

synapseclient/models/curation.py
```diff
@@ -131,6 +131,24 @@ class AuthorizationMode(str, Enum):
     ownership team. User visibility of rows depends on their individual permissions."""
 
 
+class SyncType(str, Enum):
```
BryanFauble: Could you try out using https://github.com/Sage-Bionetworks/synapsePythonClient/blob/21fde7938cbbbba266100348957163663de76adc/synapseclient/models/mixins/enum_coercion.py#L37 so that later on if more are added we are automatically covered? This would be a good spot to prove it's usage out, and then we can adopt it wider later on.

synapseclient/models/curation.py
```diff
@@ -2097,6 +2190,154 @@ async def main():
...
+
+        # Coerce sync_type to enum early to fail fast on invalid strings
+        if sync_type is not None and not isinstance(sync_type, SyncType):
+            try:
+                sync_type = SyncType(sync_type)
+            except ValueError as exc:
+                raise ValueError(
+                    f"{sync_type!r} is not a valid SyncType. "
+                    f"Expected one of: {[s.value for s in SyncType]}."
+                ) from exc
```
BryanFauble: Using https://github.com/Sage-Bionetworks/synapsePythonClient/blob/21fde7938cbbbba266100348957163663de76adc/synapseclient/models/mixins/enum_coercion.py#L7 means that you do not need to have this logic in place and it will automatically handle it for you.

synapseclient/models/curation.py
```diff
@@ -2097,6 +2190,154 @@ async def main():
...
+        status = await self.get_status_async(synapse_client=synapse_client)
+        if status.execution_details is None:
+            client.logger.info(
+                f"No active grid session found for task {self.task_id}. "
+                "Creating a new grid session..."
+            )
+            active_grid_session = await self.create_grid_session_async(
+                synapse_client=synapse_client
+            )
+            active_grid_session_id = active_grid_session.session_id
```
BryanFauble: If there is no grid session, what does calling `synchronize_active_grid_session_async` actually accomplish? If I am understanding the changes, only an active grid session needs to be synchronized - and if no grid session is active a new one would already get the latest schema applied to it.

It seems to me that this should instead be a no-op operation that logs a warning.

BryanFauble (changes requested):
I have a few questions on behavior

synapseclient/models/curation.py
linglp: Based on Bryan's comment, if there's no active session available, there should be a no op. Since we are no longer calling `create_grid_session_async`, this is no longer relevant.

docs/guides/extensions/curator/metadata_contribution.md
linglp: `sync_type` is optional for file-based grids because `PULL_PUSH` is the only supported value — there's nothing to choose. When omitted (or set to anything else), it defaults to `PULL_PUSH` and a warning is logged if you passed something different. Since this section is only for record-based task, I didn't include the explanation of file-based task. But if you look at the section of "File-Based Curation Tasks", you should be able to see the note!

synapseclient/models/curation.py
linglp: I think for record-based grids, `sync_type = None` is ambiguous. But for file-based grids, `PULL_PUSH` is the only valid value, so I think it is safe to log a warning and override the user's input?

synapseclient/models/curation.py
linglp: Thanks for the example. I am curious why would someone do something like below in practice? 
```
status = CurationTask(task_id=task.task_id).get_status()
status.execution_details = GridExecutionDetails(active_session_id=None)
CurationTask(task_id=task.task_id).update_status(curation_task_status=status)
```
Based on the docstring, updating status should be something like: 
```
status = CurationTask(task_id=task.task_id).get_status()
current.state = TaskState.COMPLETED
updated = task.update_status(curation_task_status=current)
```
That said, it doesn't hurt to add the guard, so I'll add it.

synapseclient/models/curation.py
```diff
@@ -812,6 +833,81 @@ def create_grid_session(
         """
         return Grid()
 
+    def synchronize_active_grid_session(
```
andrewelamb, line 837: The types and docstring need to be updated here to match the async version

docs/guides/extensions/curator/metadata_contribution.md
andrewelamb: "For record-based grids, `sync_type` is required — you must explicitly choose `"PULL"` or `"PULL_PUSH"`:"

```
 async def synchronize_async(
        self,
        *,
        sync_type: Optional[Union[SyncType, str]] = None,
        timeout: int = 120,
    )
```

sync_type is not required

tests/unit/synapseclient/models/async/unit_test_curation_async.py
```diff
@@ -1393,6 +1394,318 @@ async def test_list_async_state_filter_invalid_string_raises(self) -> None:
...
+                record_set_id=RECORD_SET_ID
+            ),
+        )
+
+        with (
+            patch(
+                "synapseclient.models.curation.get_curation_task_status",
+                new_callable=AsyncMock,
+                return_value=_get_curation_task_status_response(),
+            ) as mock_get_status,
```
andrewelamb: nit: mock_get_status doesn't appear to get used anywhere

tests/unit/synapseclient/models/async/unit_test_curation_async.py
```diff
@@ -1393,6 +1394,318 @@ async def test_list_async_state_filter_invalid_string_raises(self) -> None:
...
 
 
+class TestCurationTaskSynchronizeActiveGridSession:
+    """Unit tests for CurationTask.synchronize_active_grid_session_async."""
+
+    @pytest.fixture(autouse=True, scope="function")
+    def init_syn(self, syn: Synapse) -> None:
+        self.syn = syn
+
+    async def test_record_based_creates_session_when_none_active(self) -> None:
```
andrewelamb: This needs to be renamed, see line 1439 below

andrewelamb (approved):
I find a few more things to fix, but overall LGTM

thomasyu888 (commented):
@linglp 

Couple of things:

1. Before merging this, this is from Nick. lets do that in this ticket if possible.

> We deprecated the backend service to export a grid to a RecordSet because we added RecordSet support to the synchronize service, which performs the same task more safely. We don't have a plan to remove this immediately but let us know if there is a reason to use this over the synchronize service

2. The documentation will be reviewed by Christina _after_ all the recent Curator-related features are merged so that she can do a thorough walkthrough and give us feedback.  I did a merge conflict fix, so do update your local branch accordingly!

docs/guides/extensions/curator/metadata_contribution.md
linglp: I changed the documentation to something like: `
For record-based grids, choose `sync_type` explicitly — `"PULL"` or `"PULL_PUSH"` — rather than omitting it:`

linglp:
@andrewelamb based on our conversation on slack, I changed `sync_type` in `synchronize_active_grid_session` under `CurationTask` to "required", but I keep `sync_type` in `Grid.synchronize` as "optional". 

I think `Grid.synchronize` is low-level primitive. It just wraps the REST `SynchronizeGridRequest`, whose own syncType field is optional server-side (the server defaults to PULL_PUSH when omitted) and we should just match that since the `synchronize` function just wraps around the API. 

For `synchronize_active_grid_session`, this is a higher level function that already branches on task type before ever calling into Grid. I make `sync_type` required there so that: 

* keep consistent calling convention for both task types 
* For a `RecordBasedMetadataTaskProperties` task, PULL and PULL_PUSH have different, and non-reversible, effects. I want to make sure that users do not accidentally push their changes to recordset 

Let me know if this makes sense.

andrewelamb:
@BryanFauble I'm going to finish this up for Lingling if needed. Could you re-review when you get a chance? TIA!

BryanFauble (approved):
LGTM! I look on https://synapsepythonclient--1440.org.readthedocs.build/en/1440/guides/extensions/curator/metadata_contribution/ and the formatting of the changes looks good there!

PR #1447, opened 2026-08-17
[SYNPY-1893] Add search index tutorial

thomasyu888:
# **Problem:**

No search index tutorial

# **Solution:**
Added tutorial

# **Testing:**
https://synapsepythonclient--1447.org.readthedocs.build/en/1447/tutorials/python/search/


docs/tutorials/python/tutorial_scripts/search.py
```diff
@@ -0,0 +1,539 @@
...
+                response_parts=[SearchQueryPart.TOTAL_HITS],
+            )
+            if results.total_hits:
+                print(f"Index {index.id} is queryable with {results.total_hits} rows")
+                return
+        except SynapseError:
+            pass  # The index has not been created yet
+        print("Waiting for the search index to build...")
+        time.sleep(10)
+    raise TimeoutError(f"{index.id} did not finish building within {timeout} seconds")
```
BryanFauble: I don't think that this is needed. The asynch job and query pattern should be handling for this automatically. The only thing that might be a race condition is the query returning some, but not all results right away (Which this `size=1` wouldn't catch)

On the Synapse backend side, even we don't know and we don't wait until all of the individual rows are queryable. The Synapse backend only budgets the time to ensure all of the data is accepted, and then turned on user side querying.

thomasyu888: Thanks!  let me refine this tutorial first, as I haven't had a deep look at any of the claude generated content.

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,407 @@
...
+  {'study_name': 'Parkinson Comparative Cohort', 'participant_count': '180'}
+  {'study_name': 'Healthy Aging Single Cell Atlas', 'participant_count': '120'}
+```
+</details>
+
+**Note**: Each page is a separate asynchronous job either way, not a cheap follow-up
+GET, so ask for the largest `size` you can use rather than walking a big index in small
+pages.
+
+## Advanced: Tune matching with synonyms and analyzers
```
thomasyu888: I didn't run the code in this section, so this does need special attention.

docs/tutorials/python/tutorial_scripts/search.py
```diff
@@ -0,0 +1,502 @@
...
+    results = index.query(
+        search_query=SearchQuery(
+            query=Query(match={"abstract": MatchFieldOptions(query="AD")}),
+            source=SourceFilter(includes=["study_name", "diagnosis"]),
+            size=10,
+        ),
+        response_parts=[SearchQueryPart.HITS, SearchQueryPart.TOTAL_HITS],
+    )
+    print("Abstracts matching the abbreviation 'AD':")
+    print_hits(results)
```
andrewelamb: This function doesn't seem to exist anywhere in this file.

docs/tutorials/python/tutorial_scripts/search.py
```diff
@@ -0,0 +1,502 @@
...
+# Create a SearchIndex over a single table and wait for it to build.
+index = SearchIndex(
+    name="Study Summaries Search Index",
+    description="Full text search over the study summary table",
+    parent_id=project_id,
+    # The defining SQL must reference exactly one table-like entity
+    defining_sql=f"SELECT * FROM {table.id}",
+)
+index = index.store()
+print(f"Created SearchIndex with ID: {index.id}")
```
andrewelamb, line 125: Should we put in a `sleep` call here to ensure the index is stored before it's queried on line 133?

docs/tutorials/python/tutorial_scripts/search.py
```diff
@@ -0,0 +1,502 @@
...
+    search_after = results.next_search_after
+    if not search_after or not results.hits:
+        break
+    page += 1
+
+# --8<-- [end:pagination_cursor]
+
+
+# --8<-- [start:search_configuration]
+def create_search_configuration() -> str:
```
andrewelamb: Is this function used anywhere in the tutorial?

docs/tutorials/python/tutorial_scripts/search.py
```diff
@@ -0,0 +1,502 @@
...
+    ).store()
+    print(f"Created SearchConfiguration: {configuration.id}")
+    return configuration.id
+
+
+# --8<-- [end:search_configuration]
+
+
+# --8<-- [start:apply_search_configuration]
+def create_index_with_configuration(search_configuration_id: str) -> SearchIndex:
```
andrewelamb: Is this function used anywhere in the tutorial?

andrewelamb (changes requested):

docs/tutorials/python/tutorial_scripts/search.py
thomasyu888: Good catch, removed

docs/tutorials/python/tutorial_scripts/search.py
thomasyu888: I'm still thinking whether this "sage only" section should live here or if it should actually live within the "explanations" folder as an "advanced topic" since this isn't "open to the world"

docs/tutorials/python/tutorial_scripts/search.py
BryanFauble: I don't think it is needed. The creation of the entity is synchronous on the backend, and the query polling should automatically handle waiting for the index to become queryable

docs/tutorials/python/tutorial_scripts/search.py
```diff
@@ -0,0 +1,505 @@
...
+    # The defining SQL must reference exactly one table-like entity
+    defining_sql=f"SELECT * FROM {table.id}",
+)
+index = index.store()
+print(f"Created SearchIndex with ID: {index.id}")
+
+# --8<-- [end:create_index]
+
+
+# --8<-- [start:full_text_search]
```
BryanFauble, line 281: It would be good to show the fuzzy parameter being used too for typo tolerance.

docs/tutorials/python/tutorial_scripts/search.py
```diff
@@ -0,0 +1,505 @@
...
+                "default": {
+                    "type": "custom",
+                    "tokenizer": "standard",
+                    "filter": ["lowercase", "english_stop", "english_stemmer"],
+                },
+                "default_search": {
+                    "type": "custom",
+                    "tokenizer": "standard",
+                    "filter": [
+                        "lowercase",
```
BryanFauble: If it isn't mentioned in the tutorial a not about casing should be included. In all of the synonyms that I created for our portal I lowercase the synonyms inside the SynonymSet. The reason for this is that when this default_search filter pipeline is used it will lowercase the user's search terms before applying synonyms. If the synonyms themselves are not lowercase then it would not match to a synonym and expand it.

docs/tutorials/python/tutorial_scripts/search.py
BryanFauble: I would elect to have both. Where the tutorial touches on the concept, but then links to a more in depth explanation that talks about it more

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,407 @@
...
+```python
+--8<-- "docs/tutorials/python/tutorial_scripts/search.py:create_index"
+```
+
+<details class="example">
+  <summary>Creating the index should look like:</summary>
+
+```
+Created SearchIndex with ID: syn68123456
+Waiting for the search index to build...
```
BryanFauble: I think you had taken this out from the tutorial script

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,407 @@
...
+  {'study_name': 'MSBB RNA Sequencing', 'participant_count': '300'}
+  {'study_name': 'MCI Plasma Biomarkers', 'participant_count': '220'}
+Page starting at offset 4:
+total_hits=6, returned=2
+  {'study_name': 'Parkinson Comparative Cohort', 'participant_count': '180'}
+  {'study_name': 'Healthy Aging Single Cell Atlas', 'participant_count': '120'}
+```
+</details>
+
+### Cursor paging with `search_after`
```
BryanFauble: search_after also needs to be used when paginating beyond 10k rows. That was what I recall from the opensearch docs that the offset working right.

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,407 @@
...
+</details>
+
+**Note**: Each page is a separate asynchronous job either way, not a cheap follow-up
+GET, so ask for the largest `size` you can use rather than walking a big index in small
+pages.
+
+## Advanced: Tune matching with synonyms and analyzers
+
+!!! warning "Restricted and permanent"
+    Creating and updating the following resources is restricted to Sage Bionetworks employees,
```
BryanFauble: Everything including SearchIndex creation is gated behind the Sage Bio employee restriction

BryanFauble (approved):
There are a few suggested changes to improve this, but I'll approve now because it's shaping up well.

Feel free to ignore the build status on this PR since it isn't affecting anything tested.

docs/tutorials/python/tutorial_scripts/search.py
thomasyu888: It isn't executed by design, because of the nuance described in the markdown file

docs/tutorials/python/tutorial_scripts/search.py
thomasyu888: I'll create a ticket for the "advanced topic" for later, but it would be good for samia and I to coordinate on what that looks like

docs/tutorials/python/search.md
thomasyu888: Do you think that we should just have the `search_after`? or is having both pagination and search after critical?

docs/tutorials/python/search.md
thomasyu888: I see, so there's the setup, and then there's the usage.  Perhaps I'll shuffle things around so that the setup is first, then the usage of the indexes is second?

docs/tutorials/python/tutorial_scripts/search.py
thomasyu888: Done

docs/tutorials/python/search.md
BryanFauble: Sounds fine to me. The querying part is free to use by everyone whereas the setup is sage only.

docs/tutorials/python/search.md
BryanFauble: Using just search_after would be better if the framing is "Enumerate every result", but if the framing is I want the 504th result, then offset would be used.

Both can be kept because they are genuinely different use cases

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,417 @@
...
+
+If you need to search across several tables, build a
+[Materialized View](materializedview.md) first and index that.
+
+Any of these can be the source, whichever one the SQL selects from:
+
+* a [Table][synapseclient.models.Table]
+* an [EntityView][synapseclient.models.EntityView]
+* a [DatasetCollection][synapseclient.models.DatasetCollection]
+* a [MaterializedView][synapseclient.models.MaterializedView]
```
thomasyu888, line 72: @BryanFauble Are `VirtualTable` or `Datasets` supported?

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,417 @@
...
+  {'study_name': 'Parkinson Comparative Cohort', 'participant_count': '180'}
+  {'study_name': 'Healthy Aging Single Cell Atlas', 'participant_count': '120'}
+```
+</details>
+
+**Note**: Each page is a separate asynchronous job either way, not a cheap follow-up
+GET, so ask for the largest `size` you can use rather than walking a big index in small
+pages.
+
+## Advanced: Tune matching with synonyms and analyzers
```
thomasyu888: @BryanFauble what does the index.query functions look like in this tutorial if the following is all set up already?  

would we still have to fine-tune the query parameters like we do in sections 3,4,5,6,7?

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,417 @@
...
+Misspelling 'sequencng' still finds:
+total_hits=3, returned=3
+  {'study_name': 'MSBB RNA Sequencing'}
+  {'study_name': 'Mayo Clinic Whole Genome'}
+  {'study_name': 'Healthy Aging Single Cell Atlas'}
+```
+</details>
+
+Hits come back ranked by relevance, and each one carries its score on
+[`hit.score`][synapseclient.models.SearchHit] along with the `row_id` and `row_version`
```
thomasyu888: Not returning row_id and row_version, but look at hit score

thomasyu888: @BryanFauble how is score determined?

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,417 @@
...
+    abstract: ['Bulk RNA <em>sequencing</em> across four brain regions in a cohort']
+  {'study_name': 'Mayo Clinic Whole Genome', 'assay': 'wholeGenomeSeq'}
+    abstract: ['Whole genome <em>sequencing</em> of temporal cortex samples from']
+  {'study_name': 'Healthy Aging Single Cell Atlas', 'assay': 'snrnaSeq'}
+    abstract: ['Single nucleus RNA <em>sequencing</em> of hippocampus from']
+```
+</details>
+
+**Note**: Highlighting, like relevance scoring, depends on the column being indexed as
+analyzed text. Step 9 covers how to control that with a
```
thomasyu888: no longer step 9

docs/tutorials/python/search.md
```diff
@@ -0,0 +1,417 @@
...
+
+## 5. Combine scored clauses with unscored filters, and sort the results
+
+A [`bool`](https://docs.opensearch.org/latest/query-dsl/compound/bool/) clause is how
+you build a real search request out of several conditions:
+
+* `must` clauses have to match and **do** contribute to the relevance score
+* `filter` and `must_not` clauses have to match (or not match) but **do not** affect
+  the score — use these for hard constraints like a numeric cutoff
+* `should` clauses boost the rows that match them without excluding the rows that don't
```
thomasyu888, line 248: More explanation of how this is actually working would be great.

docs/tutorials/python/search.md
thomasyu888: Created this ticket to track: [SYNPY-1917](https://sagebionetworks.jira.com/browse/SYNPY-1917)

docs/tutorials/python/search.md
thomasyu888: Will cover this here: https://sagebionetworks.jira.com/browse/SYNPY-1917

docs/tutorials/python/search.md
thomasyu888: Captured here: https://sagebionetworks.jira.com/browse/SYNPY-1917

docs/tutorials/python/search.md
thomasyu888: Captured here: https://sagebionetworks.jira.com/browse/SYNPY-1917

docs/tutorials/python/search.md
thomasyu888: Done here: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1447/commits/5680c672fccf14747b689ebc39c4b2d5f8ec3d2a

docs/tutorials/python/search.md
thomasyu888: https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1447/commits/f39a8795e1e102c90b5689398383ad17f4d5b46a

docs/tutorials/python/tutorial_scripts/search.py
thomasyu888: [42e2f9f](https://github.com/Sage-Bionetworks/synapsePythonClient/pull/1447/commits/42e2f9fb56155d1fa49771b440f739f80932fdde)

docs/tutorials/python/tutorial_scripts/search.py
thomasyu888: Captured here: https://sagebionetworks.jira.com/browse/SYNPY-1917

PR #1449, opened 2026-08-20
Fix unit test to use temp path instead of writing files locally

andrewelamb:
# **Problem:**

Running the unit test suite left two artifacts behind in the current working directory that were never cleaned up: a `manifest.csv` file and a `folder_name/` directory (containing its own `manifest.csv`).

- Reproduced by running `pytest tests/unit` from the repo root and checking `git status` afterwards.
- The source is `test_sync_from_synapse_manifest_is_suppress` in `tests/unit/synapseutils/unit_test_synapseutils_sync.py`, which calls `syncFromSynapse` with `path="./"`.

# **Solution:**

Changed the test to use pytest's `tmp_path` fixture instead of `"./"`, so any files written by the sync land in a pytest-managed temp directory instead of the repo root:

- Added a `tmp_path: Path` parameter to the test signature (with the `pathlib.Path` import).
- Passed `path=str(tmp_path)` to `syncFromSynapse` and updated the two `download_location` assertions to match.

# **Testing:**

- Ran the affected test from the repo root: `pytest tests/unit/synapseutils/unit_test_synapseutils_sync.py::test_sync_from_synapse_manifest_is_suppress` — passes.
- Confirmed with `git status` that no `manifest.csv` or `folder_name/` artifacts are created in the working directory after the run.

### DIFF
--- a/tests/unit/synapseutils/unit_test_synapseutils_sync.py
+++ b/tests/unit/synapseutils/unit_test_synapseutils_sync.py
@@ -7,6 +7,7 @@ import os
 import random
 import tempfile
 from io import StringIO
+from pathlib import Path
 from typing import Any, Dict
 from unittest.mock import AsyncMock, MagicMock, Mock, call, create_autospec, patch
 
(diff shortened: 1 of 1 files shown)


thomasyu888 (approved):
🔥 LGTM! I'm going to pre-approve, but I think it would be good to have another team member approve.

I also re-triggered the fails jobs

BryanFauble (approved):

PR #1455, opened 2026-09-18
clarify get_invalid_validation to mention only results from direct ch…

andrewelamb:
# **Problem:**

A user reported that `get_invalid_validation()` on a Folder did not behave as they expected. They called it on a top-level folder and expected results for:
- A child folder
- A file nested two levels down, inside the child folder

# **Solution:**

Added a Note to both docstrings in `synapseclient/models/mixins/json_schema.py`, stating that the method checks direct children only. It does not check the entity itself or grandchildren.

This is a documentation-only change. No code path changed, and no method signature changed.


danlu1 (approved):
LGTM

